PR #1, opened 2022-11-10
Spinning up compiler repo

rsheeter:
Rudimentary rigging for IR w/toml and bincode as text and binary representation respectively. I'm more interested in feedback on the structure than the naming, we can file an issue for the inevitable naming debate :)


.gitignore
```diff
@@ -0,0 +1,5 @@
+/target
+Cargo.lock
+.venv/
+
+.DS_Store
```
dfrg, line 5: ty for this! Need to add to our other .gitignore files too

ir/src/ir.rs
```diff
@@ -0,0 +1,37 @@
+//! Serde types for font IR. See TODO:PublicLink.
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct Axis<'a> {
+    tag: &'a str,
```
dfrg, line 7: I know this is just scaffolding but this should likely be an owned String.

ufo2ir/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+license = "MIT/Apache-2.0"
+description = "Converts UFO or UFO+designspace to font ir for compilation."
+repository = "https://github.com/googlefonts/fontations"
+readme = "README.md"
+categories = ["text-processing", "parsing", "graphics"]
+
+[features]
+
+[dependencies]
+font-types = { git = 'https://github.com/googlefonts/fontations' }
```
dfrg, line 14: Shouldn’t need this explicit dep since it’s now re-exported from read-fonts.

dfrg (approved):
Let’s go!

ir/src/ir.rs
rsheeter: Done

ir/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "ir"
+version = "0.0.1"
+edition = "2021"
+license = "MIT/Apache-2.0"
+description = "IR for font compilation."
+repository = "https://github.com/googlefonts/fontations"
```
anthrotype, line 7: copy/pasta in this repository URL

ir/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "ir"
```
anthrotype, line 2: "font-ir"? let the bikeshedding begin

ir/Cargo.toml
rsheeter: ty, will fix

PR #2, opened 2022-11-11
Design space axis to IR axis

rsheeter:
Starting to rough in some of the patterns for ufo2ir. We are as yet somewhat far from doing anything useful :)

### DIFF
--- a/ir/src/ir.rs
+++ b/ir/src/ir.rs
@@ -4,11 +4,11 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Debug, PartialEq)]
 pub struct Axis {
-    tag: String,
-    min: f32,
-    default: f32,
-    max: f32,
-    hidden: bool,
+    pub tag: String,
+    pub min: f32,
+    pub default: f32,
+    pub max: f32,
+    pub hidden: bool,
 }
 
 #[cfg(test)]
--- a/ufo2ir/src/error.rs
+++ b/ufo2ir/src/error.rs
@@ -1,0 +1,6 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum UfoToIrError {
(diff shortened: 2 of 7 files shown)


ir/src/ir.rs
```diff
@@ -4,11 +4,11 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Debug, PartialEq)]
 pub struct Axis {
-    tag: String,
-    min: f32,
-    default: f32,
-    max: f32,
-    hidden: bool,
+    pub tag: String,
+    pub min: f32,
```
cmyr, line 8: Is `f32` the most appropriate type here? Genuinely a question.

ufo2ir/src/error.rs
```diff
@@ -0,0 +1,6 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum UfoToIrError {
+    // It is in fact inconceivable that this would fail
```
cmyr, line 5: 👍

ufo2ir/src/toir.rs
```diff
@@ -0,0 +1,46 @@
+use ir::ir;
+use norad::designspace::{self, DesignSpaceDocument};
+
+use crate::error::UfoToIrError;
+
+pub fn designspace_to_ir(designspace: DesignSpaceDocument) -> Result<Vec<ir::Axis>, UfoToIrError> {
```
cmyr, line 6: So just an observation, but I could imagine using a trait to encapsulate "convertibility to IR".

cmyr (approved):
some observations inline, looks good!

ir/src/ir.rs
rsheeter: Good question. I don't want to block on or forget this so I've filed https://github.com/googlefonts/fontmake-rs/issues/6

PR #4, opened 2022-11-12
[WIP] Dig upem out of ufo files

rsheeter:
Builds on #2, diff will show those changes until merged.

Note to self: @anthrotype taking this over and looking at glyph:ir


ir/src/ir.rs
```diff
@@ -11,6 +19,28 @@ pub struct Axis {
...
+
+/// A variable definition of a single glyph.
+///
+/// Defined in at least once position in designspace. If defined in
+/// many, presumed to vary continuously between positions and required
+/// to have variation compatible structure.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct Glyph {
+    pub name: String,
+    pub instances: Vec<GlyphInstance>,
```
anthrotype, line 32: didn't we want to call this `sources` and have a map of DesignSpaceLocation to GlyphInstances?
https://github.com/googlefonts/oxidize/pull/37/commits/a3efa92f95b22e8bf1ee6d9a68e0d3b400111783

I prefer `sources` to `instances` here, the former aligns with designspace terminology and conveys that these are the input to the interpolation, not the output. Instances can be used as sources (by associating them with a DS location and passed as inputs to the variation model) but can also be the result of the interpolation, when instancing the designspace at a location that's different the the sources' location.

Also I prefer if the DS location is the key and is not embedded in the GlyphInstance struct; that would ensure sources have unique locations; two sources with the same location make no sense and should be an error.

ufo2ir/src/toir.rs
```diff
@@ -1,19 +1,109 @@
...
+    designspace::{self, DesignSpaceDocument, Dimension},
+    fontinfo::NonNegativeIntegerOrFloat,
+};
 
 use crate::error::UfoToIrError;
 
-pub fn designspace_to_ir(designspace: DesignSpaceDocument) -> Result<Vec<ir::Axis>, UfoToIrError> {
-    // Truly we have done something amazing here today
-    let ir_axes: Vec<ir::Axis> = designspace.axes.into_iter().map(to_ir_axis).collect();
+struct Font {
```
anthrotype, line 11: I think naming this internal struct `Font` could be confusing as elsewhere we use `fontir::Font` already. I think you can call it `Source` (or `FontSource`), the DS calls the corresponding elements `<source>`

ufo2ir/src/toir.rs
```diff
@@ -1,19 +1,109 @@
...
+
+            let glyph_instance = ir::GlyphInstance {
+                location: font.location.clone(),
+                width: None,
+                height: None,                
+            };
+        }
+    }
+
+    Ok((ir::Font { upem, axes }, vec![]))
```
anthrotype, line 101: the assumption that one designspace file maps to only one (variable) font does no longer hold for DesignSpace version 5, where one can have multiple `<variable-fonts>`, with each covering a subset of the whole designspace, see:

https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#variable-fonts-element

so I suppose this method should return a Vec[fontir::Font].

I don't see the purpose of returning the Vec[fontir::Glyph] separately from the Font, since normally a Font owns the Glyph. Unless you are thinking about reusing glyphs in multiple fonts......

ufo2ir/src/toir.rs
rsheeter: +1, will rename

ir/src/ir.rs
rsheeter: sgtm

ir/src/ir.rs
```diff
@@ -11,6 +19,28 @@ pub struct Axis {
...
+}
+
+/// A Glyph at a specific position in designspace.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct GlyphInstance {
+    pub location: DesignSpaceLocation,
+    pub width: Option<f32>,
+    pub height: Option<f32>,
+    // TODO: outlines, a Vec<Shape>
```
rsheeter, line 41: @anthrotype it seems per https://github.com/googlefonts/oxidize/pull/40 we should split contours and components.

ir/src/ir.rs
rsheeter: The mapping seems to reflect what we want quite well, perhaps the thing to do is make DesignSpaceLocation a type that is hashable? - I believe the main thing you'd have to do is quantize the f32

anthrotype: > make DesignSpaceLocation a type that is hashable

you mean making `DesignSpaceLocation` an alias of `Vec[DesignSpaceDimension]` where the dimension is a struct containing an axis name and the associated value? Implementing `Hash` trait for `HashMap` itself is [non trivial](https://internals.rust-lang.org/t/implementing-hash-for-hashset-hashmap/3817) and beyond me at this point.

> quantize the f32

that would be lossy, no?... How about using an [`ordered_float::OrderedFloat`](https://docs.rs/ordered-float/latest/ordered_float/struct.OrderedFloat.html) instead (as suggested [here](https://internals.rust-lang.org/t/f32-f64-should-implement-hash/5436)) which, unlike f32/f64, does implements `Eq` and `Hash` and can thus be used as key in hash maps/sets?

https://docs.rs/ordered-float/latest/ordered_float/struct.OrderedFloat.html

AFAIU, the only difference with built-in floats is that 

> NaN is sorted as greater than all other values and equal to itself, in contradiction with the IEEE standard

but NaNs are explicitly disallowed in UFOs, see https://github.com/unified-font-object/ufo-spec/pull/203

rsheeter: Quantizing only for hashing wouldn't destroy the original value. https://crates.io/crates/ordered-float also sounds fine to me.

HOI support strikes me as something we should think about separately. Maybe we should file an issue?

Even under HOI I don't think it makes sense to have a Glyph with multiple definitions for the same location in design space. To my understanding HOI is about how you compute the values in between instances, not altering the actual instances. In abstract you could imagine an additional field defining how to interpolate, for napkin math purposes `how_to_interpolate: Map<(Location, Location), Curve>`.

ufo2ir/src/toir.rs
rsheeter: I think we should rename Font to something like FontInfo and not return glyphs at all. The compiler will likely require the FontInfo (basic stuff like what axes, upem, etc) be completed before it processes glyphs and will then process glyphs in parallel so we'll actually never want a "do the whole thing at once" method at all.

ir/src/ir.rs
anthrotype: ok I am using ordered-float in https://github.com/googlefonts/fontmake-rs/pull/4/commits/498878e6de2c67be0eb70c4c6d9cf6a80d9e4bd8

I made DesignSpaceLocation a BTreeMap which already implements Hash and has a similar interface as HashMap but is ordered; I don't know how to implement Hash for HashMap ([this thread makes it look tricky](https://internals.rust-lang.org/t/implementing-hash-for-hashset-hashmap/3817)).

And I moved the location field from inside GlyphInstance struct to the key of the Glyph::sources hashmap, and also added a check that it be unique before inserting a new one.

Regarding HOI, 

> Even under HOI I don't think it makes sense to have a Glyph with multiple definitions for the same location in design space.

I didn't mean having multiple glyph sources at the same location; I mean a given location containing multiple axes with the same tag, which would be impossible if the location map was keyed on the axis tag. But since a designspace location is keyed on the axis _name_ (which by definition must be unique as it identifies a given designspace axis element) and not its _tag_, it's in theory possible for a location to refer to multiple axes with distinct names (e.g. "Weight#1", "Weight#2") but that are associated with the same tag ("wght"). So I will keep DesignSpaceLocation as a map instead of making it a vector of dimensions, which is also in line with the designspaceLib doc's about location: "Encodes a dictionary of { axisname: axisvalue }".

ufo2ir/src/toir.rs
anthrotype: renamed Font=>FontInfo in https://github.com/googlefonts/fontmake-rs/pull/4/commits/cd06f47c52d0448ecceed964aecb672a0925e87f

anthrotype:
Initially, @rsheeter was assuming DS sources only use a UFO default layer (the "glyphs" subdirectory) and was listing the directory parsing each individual .glif file with norad::Glyph::load(glif_file), ignoring layercontents.plist as well as contents.plist.

After picking this PR up, at first I changed it so we take into account the fact that DS sources can reference other layers within a given UFO, and used the `plist` crate to directly parse layercontents.plist and the contents.plist in order to get to the path of glif files (rather than listing the layer directories).

Now, after @cmyr changed norad to allow filtering selected layers to be loaded in https://github.com/linebender/norad/pull/280, I modified this PR again to no longer mess with internal UFO .plist files but rely on norad to load only the layers that are used.

However, given the direction Rod is taking in #10 and #12, we may have to either go back to parsing the .plist files ourselves, or change norad to provide a method to only load specific glyphs (i.e. the ones that changed for incremental builds purposes) as opposed to all the glyphs within particular layers.

ir/src/ir.rs
rsheeter: > made DesignSpaceLocation a BTreeMap which already implements Hash

SGTM. That may explain why I saw BTreeMap in a few places where I didn't find it obvious why HashMap wouldn't be used.

rsheeter: > I will keep DesignSpaceLocation as a map

+1, I think that is more aligned with what a glyph instance is: what the glyph looks like at one specific position in design space. HOI defines how to move non-linearly between those points. At a glance it doesn't seem readily expressible in the designspace/ufo format at all...?

rsheeter:
> given the direction Rod is taking in https://github.com/googlefonts/fontmake-rs/pull/10 and https://github.com/googlefonts/fontmake-rs/pull/12

Apologies for the inconvenience but I did indeed self-parse the plists in #15.

PR #5, opened 2022-11-14
Starting on change tracking

rsheeter:
Toys to help figure out when things changed, meant to facilitate incremental compilation.

### DIFF
--- a/ir/src/depgraph.rs
+++ b/ir/src/depgraph.rs
@@ -1,0 +1,424 @@
+use filetime::FileTime;
+use serde::{Deserialize, Serialize};
+
+use std::{
+    collections::HashMap,
+    fs, io,
+    path::{Path, PathBuf},
+    vec,
+};
+
+/// Helps identify changes in a set of files.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(from = "DepGraphSerdeRepr", into = "DepGraphSerdeRepr")]
+pub struct DepGraph {
+    entries: HashMap<PathBuf, Option<DepGraphEntry>>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct DepGraphEntry {
+    mtime: FileTime,
+    size: u64,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct DepGraphSerdeRepr {
+    entries: Vec<DepGraphEntrySerdeRepr>,
+}
+
+impl From<DepGraphSerdeRepr> for DepGraph {
+    fn from(from: DepGraphSerdeRepr) -> Self {
+        DepGraph {
+            entries: from
+                .entries
+                .iter()
+                .map(|d| {
+                    (
+                        PathBuf::from(&d.path),
+                        if d.has_graph_entry() {
+                            Some(DepGraphEntry {
+                                mtime: FileTime::from_unix_time(d.unix_seconds, d.nanos),
+                                size: d.size,
+                            })
+                        } else {
+                            None
+                        },
+                    )
+                })
+                .collect(),
+        }
+    }
+}
+
+impl From<DepGraph> for DepGraphSerdeRepr {
+    fn from(dg: DepGraph) -> Self {
+        DepGraphSerdeRepr {
+            entries: dg
+                .entries
+                .iter()
+                .map(|e| {
+                    let path = e.0.to_str().expect("Only UTF names please").to_string();
+                    if let Some(dep_entry) = e.1 {
+                        DepGraphEntrySerdeRepr {
+                            path,
+                            unix_seconds: dep_entry.mtime.unix_seconds(),
+                            nanos: dep_entry.mtime.nanoseconds(),
+                            size: dep_entry.size,
+                        }
+                    } else {
+                        DepGraphEntrySerdeRepr {
+                            path,
+                            unix_seconds: 0,
+                            nanos: 0,
+                            size: 0,
+                        }
+                    }
+                })
+                .collect(),
+        }
+    }
+}
+
+/// SystemTime lacks a platform independent representation we can
+/// depend on so use FileTime's unix_seconds,nanos.
+/// unix_seconds = nanos = size = 0 is used to represent lack of a DepGraphEntry.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+struct DepGraphEntrySerdeRepr {
+    path: String,
+    unix_seconds: i64,
+    nanos: u32,
+    size: u64,
+}
+
+impl DepGraphEntrySerdeRepr {
+    fn has_graph_entry(&self) -> bool {
+        !(self.unix_seconds == 0 && self.nanos == 0 && self.size == 0)
+    }
+}
+
+pub struct Change {
+    path: PathBuf,
+    updated_entry: DepGraphEntry,
+}
+
+pub enum InitialState {
+    NotChanged,
+    Changed,
+}
+
+impl DepGraphEntry {
+    fn new(path: &Path) -> Result<DepGraphEntry, io::Error> {
+        let metadata = path.metadata()?;
+        Ok(DepGraphEntry {
+            mtime: FileTime::from_system_time(metadata.modified()?),
+            size: metadata.len(),
+        })
+    }
+}
+
+impl DepGraph {
+    pub fn new() -> DepGraph {
+        DepGraph {
+            entries: HashMap::new(),
+        }
+    }
+}
+
+impl Default for DepGraph {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl DepGraph {
+    pub fn is_tracked(&self, path: &PathBuf) -> bool {
+        self.entries.contains_key(path)
+    }
+
+    pub fn track(&mut self, path: &PathBuf, default_state: InitialState) -> Result<(), io::Error> {
+        if !self.entries.contains_key(path) {
+            self.entries.insert(
+                path.clone(),
+                match default_state {
+                    InitialState::NotChanged => Some(DepGraphEntry::new(path)?),
+                    InitialState::Changed => None,
+                },
+            );
+
+            // For a dir to register unchanged we need to add it's current contents
+            if let InitialState::NotChanged = default_state {
+                if path.is_dir() {
+                    if let Some(new_paths) = self.new_paths(path) {
+                        for new_path in new_paths {
+                            self.entries
+                                .insert(new_path.clone(), Some(DepGraphEntry::new(&new_path)?));
+                        }
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
+    pub fn has_changed(&self, change: &Change) -> bool {
+        return self
+            .entries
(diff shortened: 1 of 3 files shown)


ir/src/depgraph.rs
```diff
@@ -0,0 +1,424 @@
...
+
+        // Track the parent dir from it's current state
+        let mut dg = DepGraph::new();
+        dg.track(
+            &temp_dir.path().to_path_buf(),
+            super::InitialState::NotChanged,
+        )
+        .unwrap();
+
+        assert_eq!(
```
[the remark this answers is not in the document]
rsheeter: In particular we should *not* detect the files inside to have changed, that is we should have captured their state as a result of the dir being tracked with initial state unchanged.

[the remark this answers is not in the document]
rsheeter: (will add comments)

ir/src/depgraph.rs
```diff
@@ -0,0 +1,424 @@
...
+        fs::write(&file, "whoa").unwrap();
+        let changes = dg.changed().unwrap();
+        assert_eq!(
+            vec![&file],
+            changes.iter().map(|c| &c.path).collect::<Vec<&PathBuf>>()
+        );
+
+        // We're happy with that change
+        dg.update(&changes);
+        assert_eq!(
```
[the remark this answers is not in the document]
rsheeter: Yes, ty.

ir/src/depgraph.rs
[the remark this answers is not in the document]
rsheeter: Added comments and fixed a bug I found in elaborating the test

ir/src/depgraph.rs
```diff
@@ -0,0 +1,433 @@
...
+
+/// Helps identify changes in a set of files.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(from = "DepGraphSerdeRepr", into = "DepGraphSerdeRepr")]
+pub struct DepGraph {
+    entries: HashMap<PathBuf, Option<DepGraphEntry>>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct DepGraphEntry {
```
cmyr, line 19: can we think of a better name here? `FileState`? `FileInfo`? `FileStat`?

ir/src/depgraph.rs
```diff
@@ -0,0 +1,433 @@
...
+    mtime: FileTime,
+    size: u64,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct DepGraphSerdeRepr {
```
cmyr, line 25: so this is used entirely for serialization? If so, does it need to be public? If we can get away with it I might be tempted to just put all the serialization code in a separate module.

ir/src/depgraph.rs
```diff
@@ -0,0 +1,433 @@
...
+
+impl DepGraph {
+    pub fn new() -> DepGraph {
+        DepGraph {
+            entries: HashMap::new(),
+        }
+    }
+}
+
+impl Default for DepGraph {
```
cmyr, line 127: Is this not derivable?

ir/src/depgraph.rs
```diff
@@ -0,0 +1,433 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl DepGraph {
+    pub fn is_tracked(&self, path: &PathBuf) -> bool {
```
cmyr, line 134: I might prefer `contains`; this would be the method name I would look for if I wanted to find this functionality on something that looked vaguely like a collection.

ir/src/depgraph.rs
```diff
@@ -0,0 +1,433 @@
...
+            .map(|maybe_entry| {
+                maybe_entry
+                    .as_ref()
+                    .map(|entry| change.updated_entry != *entry)
+                    .unwrap_or(true)
+            })
+            .unwrap_or(true);
+    }
+
+    pub fn update(&mut self, changes: &Vec<Change>) {
```
cmyr, line 178: It is a convention to avoid `&Vec` in favour of `&[]`, which is strictly more flexible (unless  you need mutability)

ir/src/depgraph.rs
```diff
@@ -0,0 +1,433 @@
...
+            } else {
+                true
+            };
+
+            if has_changed {
+                changes.push(Change {
+                    path: path.to_owned(),
+                    updated_entry,
+                });
+            }
```
cmyr, line 255: stared at this for a while thinking about how I'd write it, and it would probably look like,

```rust
let has_changed = self
                .entries
                .get(path)
                .map(|maybe_current| maybe_current.as_ref() != Some(&updated_entry))
                .unwrap_or(true);

// as you were
```

ir/src/depgraph.rs
```diff
@@ -0,0 +1,433 @@
...
+    fn assert_no_changes(dg: &DepGraph) {
+        assert_eq!(
+            Vec::<&PathBuf>::new(),
+            dg.changed()
+                .unwrap()
+                .iter()
+                .map(|c| &c.path)
+                .collect::<Vec<&PathBuf>>()
+        )
+    }
```
cmyr, line 280: this logic is confusing, why not just,

```rust
assert_eq!(db.changed().unwrap().iter().count(), 0);
```

is it that you want to see the entries if the assert fails?

ir/src/depgraph.rs
```diff
@@ -0,0 +1,433 @@
...
+                    }
+                })
+                .collect(),
+        }
+    }
+}
+
+/// SystemTime lacks a platform independent representation we can
+/// depend on so use FileTime's unix_seconds,nanos.
+/// unix_seconds = nanos = size = 0 is used to represent lack of a DepGraphEntry.
```
cmyr, line 84: So it's worth keeping in mind how rustdoc renders doc comments. tl;dr, it's kind of like git commits: everything up until the first blank line is considered a 'summary', and is shown on index pages, where anything after the first line break is considered additional detail, and is only shown on the specific page for that item.

So basically: prefer a short summary sentence, and then further explanation below an empty (`/// `) line as needed.

More generally, though, this looks like it is just a comment and not really intended as public documentation?

ir/src/depgraph.rs
```diff
@@ -0,0 +1,433 @@
...
+        Self::new()
+    }
+}
+
+impl DepGraph {
+    pub fn is_tracked(&self, path: &PathBuf) -> bool {
+        self.entries.contains_key(path)
+    }
+
+    pub fn track(&mut self, path: &PathBuf, default_state: InitialState) -> Result<(), io::Error> {
```
cmyr, line 138: curious about the state param. I haven't been following along with the design discussion here too closely, but are these both common cases? Is one more common?

ir/src/depgraph.rs
cmyr: Also (and just thinking out loud, not a suggestion per se) instead of wrapping this in an `Option` it could be its own enum, like,

```rust
pub enum FileStat {
    Seen { mtime: FileTime, byte_size: u64 },
    Unknown,
}
```

which might make some of your other logic easier? Or not 💁

cmyr (approved):
Some notes inline. Without having a lot of context this looks sane?

ir/src/depgraph.rs
rsheeter: I specifically want to see the mismatched entries on failure

ir/src/depgraph.rs
rsheeter: I love how that reads left>right and down, much better.

ir/src/depgraph.rs
rsheeter: > just a comment and not really intended as public documentation?

Are doc comments of internals, aimed at maintainers, not a thing?

ir/src/depgraph.rs
rsheeter: TIL, ty.

ir/src/depgraph.rs
rsheeter: Ty!

ir/src/depgraph.rs
rsheeter: I think the pub is a copy/paste error, fixing

ir/src/depgraph.rs
rsheeter: You can derive default? What else haven't they told me!

ir/src/depgraph.rs
rsheeter: Fixed the short/long pattern

ir/src/depgraph.rs
rsheeter: I quite like the enum idea, will try it out

rsheeter: *much* better

ir/src/depgraph.rs
rsheeter: It's admittedly slightly speculative, worst case I'll delete one of the options as I get deeper into it.

I also haven't taken advantage of basic opts like if the dir mtime hasn't changed don't check the files within.

ir/src/depgraph.rs
```diff
@@ -0,0 +1,421 @@
...
+                .collect(),
+        }
+    }
+}
+
+/// The serde-friendly representation of a DepGraphEntry.
+///
+/// SystemTime lacks a platform independent representation we can
+/// depend on so use FileTime's unix_seconds,nanos.
+/// unix_seconds = nanos = size = 0 is used to represent lack of a DepGraphEntry.
```
anthrotype, line 85:
```diff
- /// unix_seconds = nanos = size = 0 is used to represent lack of a DepGraphEntry.
+ /// unix_seconds = nanos = size = 0 is used to represent lack of a PathState.
```

ir/src/depgraph.rs
```diff
@@ -0,0 +1,421 @@
...
+        }
+        results
+    }
+
+    /// Set of files that have either changed themselves or had
+    /// something they depend on change.
+    ///
+    /// Directories can be tracked but never report as changed.
+    pub fn changed(&self) -> Result<Vec<Change>, io::Error> {
+        let mut paths: Vec<&PathBuf> = self.entries.iter().map(|(k, _)| k).collect();
```
anthrotype, line 216: HashMap has `keys()` method to iterate over the keys
https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.keys

ir/src/depgraph.rs
```diff
@@ -0,0 +1,421 @@
...
+/// Helps identify changes in a set of files.
+#[derive(Serialize, Deserialize, Debug, Default, Clone)]
+#[serde(from = "DepGraphSerdeRepr", into = "DepGraphSerdeRepr")]
+pub struct DepGraph {
+    entries: HashMap<PathBuf, PathState>,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+enum PathState {
+    Seen { mtime: FileTime, size: u64 },
```
anthrotype, line 20: noob question: if we already track the modified time of a file, why do we need also bother checking its size? Can the size change without file modified time changing as well?

PR #8, opened 2022-11-15
Let loose the mighty fontir, fenris lesser known cousin

### LINKED ISSUE #3: Package naming
I think @anthrotype suggested s/ir/font-ir/. To offer an alternative, a fontmake prefix (e.g. fontmake-ir) might make sense?

rsheeter:
Packages updated per #3, fixes #3


dfrg (approved):
Happy to see the name inspiring some fun.

One note: in Rust, a crate can generate both a library and a binary by providing lib.rs and main.rs. This probably makes sense for the fontc driver.

PR #11, opened 2022-11-23
Add a very stripped down sample of a designspace to use in fontc tests

rsheeter:
This sort of sample is helpful to several packages (ex ufo to ir *and* fontc) so I put it in a global resources/testdata.

Originally created for #4.

Ok2merge.


dfrg (approved):
+1

PR #12, opened 2022-11-27
Rough in parallel designspace glyph to IR processing

rsheeter:
Won't *do* anything until #4.

### DIFF
--- a/fontc/src/main.rs
+++ b/fontc/src/main.rs
@@ -54,16 +54,14 @@ impl Config {
     }
 }
 
-fn require_dir(dir: &Path) -> PathBuf {
+fn require_dir(dir: &Path) -> Result<PathBuf, io::Error> {
     if dir.exists() && !dir.is_dir() {
         panic!("{:#?} is taken by something that isn't a directory", dir);
     }
     if !dir.exists() {
-        if let Err(e) = fs::create_dir(dir) {
-            panic!("Unable to create {:#?} {}", dir, e);
-        }
+        fs::create_dir(dir)?
     }
-    dir.to_path_buf()
+    Ok(dir.to_path_buf())
 }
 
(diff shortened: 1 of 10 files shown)


fontc/src/main.rs
```diff
@@ -52,34 +54,32 @@ impl Config {
     }
 }
 
-fn establish_build_dir() -> PathBuf {
-    let build_dir = Path::new("build");
-    if build_dir.exists() && !build_dir.is_dir() {
-        panic!("The name build is taken by something that isn't a directory");
+fn require_dir(dir: &Path) -> PathBuf {
```
cmyr, line 57: this could very reasonably return an error, I'm fine either way

fontc/src/main.rs
```diff
@@ -143,33 +156,93 @@ fn glyphs_deleted<'a>(current_inputs: &Input, prev_inputs: &'a Input) -> HashSet
...
-    let build_dir = establish_build_dir();
-    let (config, prev_inputs) = init(&build_dir, Args::parse()).map_err(Error::IoError)?;
+struct CompileContext {
+    paths: Paths,
+    ir_source: Box<dyn Source>,
+    prev_inputs: Input,
+    current_inputs: Input,
+}
+
+fn establish_context(paths: Paths, args: Args) -> Result<CompileContext, Error> {
```
cmyr, line 209: this feels like it would make sense as a type method/constructor?

cmyr (approved):
My eyes are glazing over a little bit but nothing is jumping out at me 👍

fontc/src/main.rs
rsheeter: Hm. I strongly agree! Will fix.

fontc/src/main.rs
rsheeter: Hm. In re-reading I agree. Good catch.

PR #13, opened 2022-11-28
[testdata] add required layercontents.plist, plus DS source with 'layer' attribute

anthrotype:
the [layercontents.plist](https://unifiedfontobject.org/versions/ufo3/layercontents.plist/) is required in UFO3, and ufoLib chokes with an error if it's missing, so I added one to the UFOs in resources/testdata.

Also, I added a new layer to WghtVar-Regular.ufo and accompanying new source element to wght_var.designspace that references that using the source [`layer` attribute].(https://github.com/fonttools/fonttools/blob/main/Doc/source/designspaceLib/xml.rst). This allows to specify which UFO layer a given DS source contributes to the designspace (if omitted, uses the UFO default layer). It's used e.g. by glyphsLib when converting Glyphs.app so-called "brace" layers to designspace+UFO. Basically it allows to optionally define DS sources from (usually sparse) layers within a given UFO, to add intermediate masters for specific glyphs that may need those.


anthrotype:
(ready to be merged) 
also I need this for #4 but since these testdata are shared with other crates I figured I make a separate PR

cmyr:
looks good!

PR #15, opened 2022-11-29
Pay attention to layer when loading glif files

rsheeter:
OK to merge.


ufo2fontir/src/source.rs
```diff
@@ -35,6 +26,59 @@ impl DesignSpaceIrSource {
...
+    layer_cache: &'a mut HashMap<String, HashMap<String, PathBuf>>,
+    source: &designspace::Source,
+) -> Result<&'a PathBuf, Error> {
+    static DEFAULT_LAYER_NAME: &str = "public.default";
+
+    if !layer_cache.contains_key(&source.filename) {
+        let contents = layer_contents(ufo_dir)?;
+        layer_cache.insert(source.filename.clone(), contents);
+    }
+    let layer_name = source.layer.as_deref().unwrap_or(DEFAULT_LAYER_NAME);
```
anthrotype, line 74: this is incorrect. If source.layer is None, you should not assume that the default layer is called "public.default". The default layer is _default_ because the directory it maps to is called just "glyphs", but a default layer can be given any name, e.g. "foreground" or "rsheeter" if you like. Only when the default layer (the one associated with the "glyphs" directory) is not given a user-defined name, it defaults to "public.default".
See https://unifiedfontobject.org/versions/ufo3/layercontents.plist/#specification

You should do something similar to what I did in https://github.com/googlefonts/fontmake-rs/commit/2062c2f298907806be457716e63be013f8738bf8

```rust
        let mut layer_path = self.path.join("glyphs");
        if let Some(layer_name) = &self.layer {
            layer_path = self.path.join(
                self.layer_contents()?
                    .get(layer_name)
                    .ok_or_else(|| UfoToIrError::LayerNotFoundError(layer_name.to_string()))?,
            );
        }
```

ufo2fontir/src/source.rs
```diff
@@ -35,6 +26,59 @@ impl DesignSpaceIrSource {
     }
 }
 
+fn glif_files<'a>(
+    ufo_dir: &Path,
+    layer_cache: &'a mut HashMap<String, HashMap<String, PathBuf>>,
+    source: &designspace::Source,
+) -> Result<BTreeMap<String, PathBuf>, Error> {
+    let layer_name = layer_name(ufo_dir, layer_cache, source)?;
```
anthrotype, line 34: I would not call this method or local variable `layer_name`, because it's not the _name_ of a layer (that's the _key_ in the layercontents ordered dict), it's the directory path associated with that layer (or the _value_ of layercontents.plist entries).

ufo2fontir/src/source.rs
rsheeter: Ew. Ty, will fix.

rsheeter: Fixed I think?

ufo2fontir/src/source.rs
rsheeter: Fixed

ufo2fontir/src/source.rs
```diff
@@ -35,6 +26,61 @@ impl DesignSpaceIrSource {
     }
 }
 
+fn glif_files<'a>(
+    ufo_dir: &Path,
+    layer_cache: &'a mut HashMap<String, HashMap<String, PathBuf>>,
```
cmyr, line 31: not a blocker, just an observation: depending on how convoluted this gets, I often end up writing a tiny `LayerCache` type here that wraps the hashmap and provides a simple API tailored to my needs.

cmyr (approved):
lgtm

ufo2fontir/src/source.rs
rsheeter: I waffled hard on whether it was justified here. I feel like it's not *quite* but I reserve the right to decide I"m an idiot in future.

cmyr: totally, it's definitely not clear cut either way.

PR #16, opened 2022-11-29
Provide the location when creating glif work

rsheeter:
Provide a set of Path:DesignSpaceLocation(s) to the create IR function. Intended to leave us in a good position to actually build glyph IR.

### DIFF
--- a/ufo2fontir/src/source.rs
+++ b/ufo2fontir/src/source.rs
@@ -98,18 +98,21 @@ impl DesignSpaceIrSource {
     }
 
     // When things like upem may have changed forget incremental and rebuild the whole thing
-    fn global_rebuild_triggers(&self, designspace: &DesignSpaceDocument) -> Result<FileStateSet, Error> {
+    fn global_rebuild_triggers(
+        &self,
+        designspace: &DesignSpaceDocument,
+    ) -> Result<FileStateSet, Error> {
         let mut font_info = FileStateSet::new();
         font_info.insert(&self.designspace_file)?;
         for source in designspace.sources.iter() {
-            let font_info_file = self
-                .designspace_dir
-                .join(&source.filename)
-                .join("fontinfo.plist");
-            if !font_info_file.is_file() {
-                return Err(Error::FileExpected(font_info_file));
+            let ufo_dir = self.designspace_dir.join(&source.filename);
+            for filename in ["fontinfo.plist", "layercontents.plist"] {
+                let font_info_file = ufo_dir.join(filename);
+                if !font_info_file.is_file() {
+                    return Err(Error::FileExpected(font_info_file));
+                }
+                font_info.insert(&font_info_file)?;
             }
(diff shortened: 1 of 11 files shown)


ufo2fontir/src/source.rs
```diff
@@ -1,27 +1,37 @@
...
 };
 use log::debug;
 use norad::designspace::{self, DesignSpaceDocument};
 
 pub struct DesignSpaceIrSource {
     designspace_file: PathBuf,
+    designspace_dir: PathBuf,
     ir_paths: Paths,
+    // A cache of locations, valid provided no font info file changes
+    glif_locations: (FileStateSet, HashMap<PathBuf, Vec<DesignSpaceLocation>>),
```
anthrotype, line 21: why would a change in UFO fontinfo.plist invalidate the cache of glyph locations? 
oh maybe I misunderstood, seems like you used the name `font_info` (the return value of font_info_files) to mean a FileStateSet that includes not just source ufos' fontinfo.plist (which I think are irrelevant to that glif_locations cache) but also the designspace file itself (which defines those locations and thus is relevant).
Perhaps use a different name that can't be misunderstood this way

rsheeter: Naming is the hardest part. Will update. 

My rationale was that font_info.plist includes information that seemed to warrant a global rebuild on change, e.g. upem. I'll add a comment as well as renaming.

rsheeter: Did that help?

anthrotype: yes, thanks

ufo2fontir/src/source.rs
```diff
@@ -110,43 +130,107 @@ impl Source for DesignSpaceIrSource {
...
+    ) -> Result<Vec<Box<dyn Work<()>>>, Error> {
+        // A single glif could be used by many source blocks that use the same layer
+        // *gasp*
+        // So resolve each file to 1..N locations in designspace
+
+        let mut work: Vec<Box<dyn Work<()>>> = Vec::new();
+
+        // Do we have the location of glifs written down?
+        // TODO: consider just recomputing here instead of failing
+        if input.font_info != self.glif_locations.0 {
```
anthrotype, line 176: I find a bit opaque the use of an unnamed tuple for this cache of glif locations, when you access things with self.glif_locations.0 and self.glif_locations.1 I wonder what that means. Why do those two have to go together as a tuple, can't they be distinct field of DesignSpaceIrSource struct?
or maybe it's the name of the self.glif_locations that I find confusing; it is appropriate for its .1 item (the hashmap of glif locations), but perhaps not for the tuple as a whole..

rsheeter: Clearer?

anthrotype: yes thanks

anthrotype (approved):

PR #18, opened 2022-11-30
Start emitting IR for UFO glifs

anthrotype:
reworking of #4 on top of latest changes

### DIFF
--- a/fontir/src/error.rs
+++ b/fontir/src/error.rs
@@ -38,4 +38,10 @@ pub enum Error {
 pub enum WorkError {
     #[error("IO failure")]
     IoError(#[from] io::Error),
+    // I can't use Box(<dyn error::Error>) here because it's not Send, but
+    // if I convert error to string I lose the backtrace... What to do?
+    #[error("Conversion of glyph '{0}' to IR failed: {1}")]
+    GlyphIrWorkError(String, String),
+    #[error("yaml error")]
+    YamlSerError(#[from] serde_yaml::Error),
 }
--- a/fontir/src/error.rs
+++ b/fontir/src/error.rs
@@ -2,6 +2,8 @@ use std::{error, io, path::PathBuf};
 
 use thiserror::Error;
 
+use crate::ir::DesignSpaceLocation;
+
 #[derive(Debug, Error)]
 pub enum Error {
     #[error("Directory expected")]
@@ -24,6 +26,11 @@ pub enum Error {
     UnableToCreateGlyphIrWork,
     #[error("Unexpected state encountered in a state set")]
     UnexpectedState,
+    #[error("Duplicate location for {what}: {loc:?}")]
+    DuplicateLocation {
+        what: String,
+        loc: DesignSpaceLocation,
+    },
 }
 
 /// An async work error, hence one that must be Send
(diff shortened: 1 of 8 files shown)


ufo2fontir/src/source.rs
```diff
@@ -163,10 +165,7 @@ impl Source for DesignSpaceIrSource {
             let ufo_dir = self.designspace_dir.join(&source.filename);
 
             // TODO less dumbed down version of dim => location mapping
```
rsheeter, line 167: remove the TODO?

ufo2fontir/src/source.rs
```diff
@@ -163,10 +165,7 @@ impl Source for DesignSpaceIrSource {
...
-            let mut location = DesignSpaceLocation::new();
-            source.location.iter().for_each(|dim| {
-                location.insert(dim.name.clone(), dim.xvalue.map(|v| v as i32).unwrap());
-            });
+            let location = to_ir_location(&source.location);
```
rsheeter, line 168: I think if you define From this could be .into()?

https://doc.rust-lang.org/std/convert/trait.From.html

ufo2fontir/src/toir.rs
```diff
@@ -1,8 +1,16 @@
 use fontir::ir;
-use norad::designspace::{self, DesignSpaceDocument};
+use norad::designspace::{self, DesignSpaceDocument, Dimension};
+use ordered_float::OrderedFloat;
 
 use crate::error::Error;
 
+// TODO we will need the ability to map coordinates and a test font that does. Then no unwrap.
```
rsheeter, line 7: Maybe we should file an issue for this as it's somewhat important and surprisingly involved given the mapping mechanism?

ufo2fontir/src/source.rs
anthrotype: I can't `impl From<&[Dimension]> for ir::DesignSpaceLocation` because ir::DesignSpaceLocation is just an alias to BTreeMap at the moment, and also it belongs to a different crate (fontir vs ufo2fontir), plus slices are always foreign

ufo2fontir/src/source.rs
anthrotype: done

ufo2fontir/src/toir.rs
anthrotype: https://github.com/googlefonts/fontmake-rs/issues/22

ufo2fontir/src/source.rs
rsheeter: Oh right ... no traits across boundaries

fontir/src/error.rs
```diff
@@ -24,11 +26,22 @@ pub enum Error {
     UnableToCreateGlyphIrWork,
     #[error("Unexpected state encountered in a state set")]
     UnexpectedState,
+    #[error("Duplicate location for {what}: {loc:?}")]
```
rsheeter, line 29: ...what is this wizardry?! TIL.

rsheeter (approved):
LGTM. We'll want to beef up tests at some point but we don't need to block this PR on that.

PR #19, opened 2022-12-01
A .glyphs VF for testing

rsheeter:
Has a hyphen and an explanation mark that vary on wght from 400 to 700. Original "artwork."


anthrotype (approved):
LGTM! Nice exclamation point!!!

PR #20, opened 2022-12-01
Add the ability to capture the state of a slice.

rsheeter:
Intended use is to capture slices from .glyphs files so we can see things like which glyphs have changed since the prior compile (wip of this in #21)

### DIFF
--- a/fontir/src/stateset.rs
+++ b/fontir/src/stateset.rs
@@ -120,11 +120,13 @@ impl StateSet {
 
     /// Pay attention to slice, we'd quite like to know if it changes
     ///
-    /// Path is used only as an identifier, invent any scheme you like.
-    /// my_file/my_subsection say.
-    pub fn track_slice(&mut self, path: &Path, slice: &str) -> Result<(), io::Error> {
-        self.entries
-            .insert(path.to_path_buf(), State::Slice(SliceState::of(slice)?));
+    /// Identifier can be whatever you like. For file fragments a path-like construct
+    /// is suggested, e.g. myfile.glyphs/glyphs/layer/glyph perhaps.
+    pub fn track_slice(&mut self, identifier: &str, slice: &str) -> Result<(), io::Error> {
+        self.entries.insert(
+            PathBuf::from(identifier),
+            State::Slice(SliceState::of(slice)?),
+        );
         Ok(())
     }
 
@@ -247,8 +249,8 @@ mod tests {
 
     #[test]
     fn detect_slice_change() {
-        let p1 = Path::new("Font.glyphs/glyphs/space");
-        let p2 = Path::new("Font.glyphs/glyphs/hyphen");
+        let p1 = "Font.glyphs/glyphs/layer/space";
+        let p2 = "Font.glyphs/glyphs/layer/hyphen";
 
         let mut s1 = StateSet::new();
         s1.track_slice(p1, "this is a glyph").unwrap();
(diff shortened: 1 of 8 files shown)


fontir/src/stateset.rs
```diff
@@ -79,23 +103,31 @@ impl FileStateSet {
...
+                    .insert(new_path.clone(), State::File(FileState::of(&new_path)?));
             }
         }
         Ok(())
     }
 
+    /// Pay attention to slice, we'd quite like to know if it changes
+    ///
+    /// Path is used only as an identifier, invent any scheme you like.
+    /// my_file/my_subsection say.
```
anthrotype, line 124: > Path is used only as an identifier

then why do you use `Path` and not simply a string?

anthrotype: it doesn't seem like you're using any Path-specific methods for the SliceSet. 
Path/PathBuf are wrappers around OsString/OsStr so on windows they will use backslashes and do other win-specific things. Any methods that attempts to access the underlying filesystem is bound to fai..
It feels weird to me. If there was something like python's pathllib.PurePath maybe.. but it doesn't seem to be the case in rust.
I think it'd be clearer if SliceSet used a String for the entries key.

fontir/src/stateset.rs
```diff
@@ -345,21 +400,25 @@ mod tests {
...
         assert_eq!(fs, restored);
     }
 
     #[test]
     fn read_write_bincode() {
         let temp_dir = tempdir().unwrap();
 
-        let (_, _, fs) = one_changed_file_one_not(&temp_dir);
+        let (_, _, mut fs) = one_changed_file_one_not(&temp_dir);
+        fs.track_slice(Path::new("file.glyphs/glyph/glyph_name"), "Hi World!")
```
anthrotype, line 417: what I think would be clearer and more explicit as type of the key in a SliceSet is a tuple/struct where one field is an actual Path or PathBuf (i.e. the path to the file containing the slice) and a second field contains a vector of strings defining the "path" components to identify the given slice

from `Path::new("file.glyphs/glyph/glyph_name"` to `(Path::new("file.glyphs"), vec!["glyph", "glyph_name"])`

anthrotype: or as a struct:

```rust
let glyph_slice_path = { path: Path::new("file.glyphs"), keys: vec!["glyph", "glyph_name"] };
```

rsheeter: My thought was that path captures that the user is intended to use a "pathlike" construct as the key, e.g. even for slices you probably want a structured key. I suppose I could hide that, and track_slice could input a String.

fontir/src/stateset.rs
rsheeter: I view Path/PathBuf as equivalent to Python's Path, is that wrong?

anthrotype: yes. pathlib.Path is same as rust's Path/PathBuf, but it seems like there's no equivalent in Rust for a [PurePath](https://docs.python.org/3/library/pathlib.html#pure-paths), ie. "objects provide path-handling operations which don’t actually access a filesystem."

rsheeter: A Path you never use for file IO is?

But I think you are right, this doesn't need to be a Path. I made it a String.

fontir/src/stateset.rs
rsheeter: track_slice now takes &str identifiers

fontir/src/stateset.rs
anthrotype: > A Path you never use for file IO is?

it's a string that has additional path-like methods to compose, split, etc. like a path, very convenient at times.

rsheeter:
Merging per IM discussion

PR #21, opened 2022-12-01
[WIP] Add the ability to read glyphs files

rsheeter:
In a manner that permits us to tell if bits of them change


glyphs2fontir/src/plist.rs
```diff
@@ -0,0 +1,450 @@
...
+        assert_eq!(0usize..raw_plist.trim_end().len(), range);
+
+        let Plist::Integer(upem, range) = plist.get("unitsPerEm").unwrap() else {
+            panic!("upem isn't an int?!");
+        };
+        assert_eq!(*upem, 1000i64);
+        assert_eq!("1000", raw_plist[range.start..range.end].trim());
+
+        let Plist::Array(glyphs, _) = plist.get("glyphs").unwrap() else {
+            panic!("upem isn't an int?!");
```
anthrotype, line 444: copypasta from the previous panic! statement?
should read something like 

```suggestion
            panic!("'glyphs' isn't an array?!");
```

rsheeter: Oh yes ty!

glyphs2fontir/src/source.rs
```diff
@@ -0,0 +1,44 @@
+use std::{collections::HashSet, fs, path::PathBuf};
+
+use fontir::error::Error;
+use fontir::source::{Input, Paths, Source, Work};
+
+use crate::plist::Plist;
```
anthrotype, line 6: I think this module should be renamed `openstep_plist` to avoid confusions with the `plist` crate which we use to parse xml plists

rsheeter: Good idea.

rsheeter: Done

anthrotype (approved):

PR #24, opened 2022-12-02
Find the glyphs in a .glyphs and sketch out wiring to process them

rsheeter:
For now don't do anything meaningful with them.

NOTE: setup to target something other than main. @cmyr thought that once the prior PR in the chain (#21) merges this might auto-redirect to main so... lets find out.

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -1,13 +1,17 @@
-use std::{collections::HashSet, fs, path::PathBuf};
+use std::collections::HashSet;
+use std::path::Path;
+use std::{collections::HashMap, fs, path::PathBuf};
 
 use fontir::error::Error;
 use fontir::source::{Input, Paths, Source, Work};
+use fontir::stateset::StateSet;
 
 use crate::openstep_plist::Plist;
 
 pub struct GlyphsIrSource {
     glyphs_file: PathBuf,
     ir_paths: Paths,
+    plist_cache: Option<PlistCache>,
 }
 
 impl GlyphsIrSource {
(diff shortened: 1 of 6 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -168,4 +191,40 @@ mod tests {
...
+            "glyphname".to_string(),
+            Plist::String("space".to_string(), 0..1),
+        );
+        assert_eq!(
+            "space",
+            glyph_name(
+                Path::new("f.glyphs"),
+                &glyph,
+                &(1usize..12usize),
+                "I'm a plist"
```
anthrotype, line 276: wait, why is this "I'm a plist"? Shouldn't this be something like `"glyphname = space"`?

anthrotype: ah, it's unused if glyph["glyphname"] is a Plist::String.. still it's confusing

glyphs2fontir/src/source.rs
```diff
@@ -27,15 +27,21 @@ impl GlyphsIrSource {
 }
 
 struct PlistCache {
-    _global_metadata: StateSet,
+    global_metadata: StateSet,
     _root_dict: HashMap<String, Plist>,
```
anthrotype, line 31: underscore prefix is unnecessary for a private-by-default struct field, no?

glyphs2fontir/src/source.rs
```diff
@@ -94,21 +100,32 @@ impl GlyphsIrSource {
...
-        _root_dict: &HashMap<String, Plist>,
+        root_dict: &HashMap<String, Plist>,
+        raw_plist: &str,
     ) -> Result<StateSet, Error> {
-        todo!()
+        // Naive mk1: if anything other than glyphs and date changes do a global rebuild
+        // TODO experiment with actual glyphs saves to see what makes sense
+        let mut state = StateSet::new();
+        for (key, plist) in root_dict {
+            if key == "glyphs" || key == "date" {
```
anthrotype, line 110: this is probably too coarse / too sensitive to changes, but I'm ok to keep like that until we have a better idea of what can change globally that would/would not affect single glyphs IR.

glyphs2fontir/src/source.rs
rsheeter: I will update to something clearer, e.g. "Not used by test"

glyphs2fontir/src/source.rs
rsheeter: I would have sworn it was warning about it, let me double-check

glyphs2fontir/src/source.rs
rsheeter: SGTM

anthrotype (approved):

glyphs2fontir/src/source.rs
rsheeter: global_metadata is now used, I tried with root_dict and it did complain:

```
warning: field `root_dict` is never read
  --> glyphs2fontir/src/source.rs:31:5
   |
29 | struct PlistCache {
   |        ---------- field in this struct
30 |     global_metadata: StateSet,
31 |     root_dict: HashMap<String, Plist>,
   |     ^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default
```

PR #25, opened 2022-12-02
Use hashable types and derived plist conversion instead of raw access

### LINKED ISSUE #23: Implement hasing on Plist?
#20 adds the ability to capture the state of part of a string by hashing it. However, if a plist was written in a different but semantically equivalent order we'd consider it changed. We could improve this by implementing hashing of Plist. Only matters if Glyphs does things like write fields in randomized order every save.

rsheeter:
Grab Raph's plist-derive glyphs-reading, tweak it to support Hash, and use hash on types to track fragments of Glyphs files instead of relying on slices of the raw string.

Fixes #23 .

### DIFF
--- a/fontir/src/stateset.rs
+++ b/fontir/src/stateset.rs
@@ -5,7 +5,7 @@ use serde::{Deserialize, Serialize};
 use std::{
     collections::{hash_map, HashMap, HashSet},
     fs,
-    hash::Hash,
+    hash::{Hash, Hasher},
     io,
     path::{Path, PathBuf},
     vec,
@@ -18,20 +18,18 @@ pub struct StateSet {
     pub(crate) entries: HashMap<StateIdentifier, State>,
 }
 
-// Sometimes state comes from a file, sometimes it comes from a
-// a slice of a string.
+// The key for a stateful thing of some specific type
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
 pub enum StateIdentifier {
     File(PathBuf),
     Memory(String),
 }
 
-// Sometimes state comes from a file, sometimes it comes from a
-// a slice of a string.
+// A stateful thing of some specific type
 #[derive(Debug, Clone, PartialEq)]
 pub(crate) enum State {
     File(FileState),
-    Memory(SliceState),
+    Memory(MemoryState),
 }
 
 #[derive(Debug, Copy, Clone, PartialEq)]
@@ -41,9 +39,9 @@ pub(crate) struct FileState {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct SliceState {
+pub struct MemoryState {
     pub(crate) hash: blake3::Hash,
-    pub(crate) size: usize,
+    pub(crate) size: u64,
 }
 
 impl FileState {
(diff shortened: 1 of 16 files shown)


glyphs-reader/src/font.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+#[derive(Clone, Debug, FromPlist, ToPlist, Hash)]
+pub struct Layer {
+    pub layer_id: String,
+    pub width: OrderedFloat<f64>,
+    pub paths: Option<Vec<Path>>,
+    pub components: Option<Vec<Component>>,
+    pub anchors: Option<Vec<Anchor>>,
+    pub guide_lines: Option<Vec<GuideLine>>,
```
anthrotype, line 39: the two-word camel case seems odd, I'd expect one `guidelines`. Was it auto-snake_cased?

BTW, I think in Glyphs3 source format, these are now just called `guides`

see https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md#differences-between-version-2

> guideLines has been renamed guides.

anthrotype: usually we don't really care aboud guidelines for compilation purposes, you may even put them in the `#[rest]` of the other_stuff

anthrotype (commented):
Does Raph's glyphs parser support Glyphs3 file format? I think we should primarily aim to support that one, or at least initially, given it's the current one. We may also support .glyphs version 2 if need be.

anthrotype:
> Does Raph's glyphs parser support Glyphs3 file format?

looks like it does not, I can tell from the way nodes are serialized, in glyphs3 they are no longer a string like "100 200 LINE" but a tuple of (x, y, type, user_data). More differences b/w 2 vs 3 here: https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md#differences-between-version-2

glyphsLib can parse both v2 and v3 sources, and you can try use that as a reference. I think we should focus on supporting v3 only for simplicity, in theory one can always open up a v2 file in Glyphs.app and save it as v3 (not sure how lossless-ly though).

rsheeter:
> v2 and v3 sources

My v3 test file parses to Plist fine so I think we have a solid foundation with this rig. Worst case we'll have two types, one per glyphs format.

anthrotype (approved):
the core parsing mechanism (Plist + Rust types + macros) LGTM
I suggest you file a separate issue for Glyphs 2 vs 3 support and we can refine with subsequent PRs

glyphs-reader/src/font.rs
rsheeter: Good point, I think I will just drop them

fontir/src/stateset.rs
```diff
@@ -133,14 +156,18 @@ impl StateSet {
...
-    /// Pay attention to slice, we'd quite like to know if it changes
+    /// Pay attention to this hashable thing, we'd quite like to know if it changes
     ///
     /// Identifier can be whatever you like. For file fragments a path-like construct
     /// is suggested, e.g. myfile.glyphs/glyphs/layer/glyph perhaps.
-    pub fn track_slice(&mut self, identifier: String, slice: &str) -> Result<(), io::Error> {
+    pub fn track_memory(
+        &mut self,
+        identifier: String,
+        memory: Box<impl Hash>,
```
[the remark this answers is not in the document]
rsheeter: I thought that was what you "should" do but after further poking I'm no longer sure about that, I think in fact what I want is `memory: &(impl Hash + ?Sized)` which compiles and makes use of the api nices (`track("id", &plist)` vs `track("id", Box::new(plist))`.

glyphs-reader/src/font.rs
rsheeter: Fixed

PR #29, opened 2022-12-05
Add glyphs 2 (as written by glyphs 3) examples

rsheeter:
Save as G2 per https://forum.glyphsapp.com/t/retrocombatibility-glyphs-3-to-glyphs-2/16729.

Ok to merge.


cmyr (approved):
👍

PR #30, opened 2022-12-05
Start to convert Glyphs v2 to Glyphs v3

rsheeter:
Convert a simple G2 to G3 based on https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md#differences-between-version-2.

v2 test files created by opening the v3 files I created in Glyphs and saving in v2 format.

As we inevitably find bugs we can add more test files, hopefully this is a halfway decent start.


glyphs-reader/Cargo.toml
```diff
@@ -11,3 +11,8 @@ edition = "2021"
 kurbo = "0.5.1"
 plist_derive = { path = "plist_derive" }
 ordered-float = "3.4.0"
+
+thiserror = "1.0.37"
+
+[dev-dependencies]
+pretty_assertions = "1.3.0"
```
anthrotype, line 18: didn't know about this one, nice!

glyphs-reader/src/font.rs
```diff
@@ -145,24 +150,48 @@ impl Glyph {
...
+        match &plist {
+            Plist::String(value) => {
+                let mut spl = value.splitn(3, ' ');
+                let x = spl.next().unwrap().parse().unwrap();
+                let y = spl.next().unwrap().parse().unwrap();
+                let pt = Point::new(x, y);
+                let node_type = spl.next().unwrap().parse().unwrap();
+                Node { pt, node_type }
+            }
+            Plist::Array(value) => {
```
anthrotype, line 162: should we not check the .glyphs file's `formatVersion` to conditionally parse nodes as strings or arrays?
Here you're parsing either string or array regardless of format version, which will work but I'd argue would not be "legal"

https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md#element-tree
> .formatVersion int set to 3 for version 3. If that key is missing assume version 2.

glyphs-reader/src/font.rs
```diff
@@ -145,24 +150,48 @@ impl Glyph {
...
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
+            // Glyphs 2 style
             "LINE" => Ok(NodeType::Line),
             "LINE SMOOTH" => Ok(NodeType::LineSmooth),
             "OFFCURVE" => Ok(NodeType::OffCurve),
             "CURVE" => Ok(NodeType::Curve),
             "CURVE SMOOTH" => Ok(NodeType::CurveSmooth),
+            // Glyphs 3 style
+            "l" => Ok(NodeType::Line),
```
anthrotype, line 190: same here, you are ignoring format version and matching  either all-cap node types and Glyphs3's shorthands. Even if it may be simpler to do so, I  think it's more correct to distinguish the two code paths

glyphs-reader/src/font.rs
```diff
@@ -264,3 +294,296 @@ impl Path {
...
+        if tags.len() != 1 || tags[0] != "wght" {
+            return Err(Error::StructuralError("Only wght supported so far".into()));
+        }
+        for master in self.font_master.iter_mut() {
+            let Some(Plist::Integer(wght)) = master.other_stuff.remove("weightValue") else {
+                continue;
+            };
+            master.other_stuff.insert("axesValues".into(), Plist::Array(vec![Plist::Integer(wght)]));
+
+            let name = match master.other_stuff.remove("weight") {
```
anthrotype, line 471: in glyphs2, a master's "weight" attribute is omitted when it's equal to the default value (like most other glyphs plist attributes). In this particular case, the default would be "Regular"

https://github.com/googlefonts/glyphsLib/blob/f6e9c4a29ce764d34c309caef5118c48c156be36/Lib/glyphsLib/classes.py#L1560

so you don't need to do `weight_name(wght).to_string()` when it's not present, just assume that if omitted it's "Regular"

glyphs-reader/src/font.rs
```diff
@@ -264,3 +294,296 @@ impl Path {
...
+                Some(Plist::String(name)) => name,
+                _ => weight_name(wght).to_string(),
+            };
+            master.other_stuff.insert("name".into(), Plist::String(name));
```
anthrotype, line 475: so you want to derive a name for a GSFontMaster, because in Glyphs3 they have a required `name` attribute.
This would be the logic to get a master's name in a way that is compatible with both Glyphs 2 and 3:

https://github.com/googlefonts/glyphsLib/blob/f6e9c4a29ce764d34c309caef5118c48c156be36/Lib/glyphsLib/classes.py#L1637-L1644

basically in Glyphs 2, the master name was implicit in the predefined configuration of weight, width and "custom" axes (see `_joinName` method above); or it could be overridden by a user's defined custom parameter named "Master Name".

glyphs-reader/src/font.rs
```diff
@@ -264,3 +294,296 @@ impl Path {
...
+        let mut is_v2 = true;
+        if let Some(Plist::Integer(version)) = self.other_stuff.get(".formatVersion") {
+            is_v2 = *version < 3; // .formatVersion is only present for v3+
+        }
+        is_v2
+    }
+
+    fn v2_to_v3_axes(&mut self) -> Result<Vec<String>, Error> {
+        let mut tags = Vec::new();
+        if let Some(Plist::Array(custom_params)) = self.other_stuff.get_mut("customParameters") {
```
anthrotype, line 362: you want to get the list of axes defined in the font. As usual, it's a bit more complicated than simply checking if an "Axes" custom parameter is defined :)

That parameter is optional and can be omitted, in which case one has to assume the predefined triplet of default axes definitions, i.e. weight, width and "custom", any of which may be used by a given font master.

see `get_axis_definitions` method from glyphsLib (this is used to define the designspace <axes> element):
 https://github.com/googlefonts/glyphsLib/blob/f6e9c4a29ce764d34c309caef5118c48c156be36/Lib/glyphsLib/builder/axes.py#L530

If you like, since you're only supporting one-axis wght fonts for the time being, you could assume the absence of "Axes" custom param means an implicit {"name": "Axes", "value": [{"Name": "Weight", "Tag": "wght"}]}

glyphs-reader/src/font.rs
```diff
@@ -264,3 +294,296 @@ impl Path {
...
+                self.other_stuff.remove("customParameters");
+            }            
+        }
+        Ok(tags)
+    }
+
+    fn v2_to_v3_metrics(&mut self) -> Result<(), Error> {
+        // metrics are in parallel arrays in v3
+        let v3_metric_names = ["ascender".to_string(), "baseline".to_string(), "descender".to_string(), "cap height".to_string(), "x-height".to_string()];
+        let v2_metric_names = ["ascender".to_string(), "baseline".to_string(), "descender".to_string(), "capHeight".to_string(), "xHeight".to_string()];
```
anthrotype, line 403: maybe these two arrays of strings for the metric names can be constant defined at the top

glyphs-reader/src/font.rs
```diff
@@ -264,3 +294,296 @@ impl Path {
...
+            }
+            master.other_stuff.insert("metricValues".into(), Plist::Array(metric_values));
+        }
+        Ok(())
+    }
+
+    fn v2_to_v3_shapes(&mut self) -> Result<(), Error> {
+        // shapes in v3 encompasses paths and components in v2
+        for glyph in self.glyphs.iter_mut() {
+            // v2 uses single codepoint strings, turn into int to match v3 for now
```
anthrotype, line 434: > v2 uses single codepoint strings

actually, a glyph's `unicode` attribute in v2 can contain a comma-separated list of hexadecimal unicode codepoints, not just one.

And sometimes Georg insisted in not adding quotes to disambiguate these hex strings from equivalent decimal digits so our openstep_plist parser would give as ints but what they really where was hex strings so we had to work around that...

glyphs-reader/src/font.rs
```diff
@@ -264,3 +294,296 @@ impl Path {
...
+                    *v = Plist::Integer(i64::from_str_radix(val, 16).unwrap())
+                }
+            });
+
+            // Paths and components combine in shapes
+            // TODO components
+            for layer in glyph.layers.iter_mut() {
+                let paths = layer.paths.take();
+                if paths.is_some() {
+                    match &mut layer.shapes {
```
anthrotype, line 447: I believe if it has `paths` (v2 only) is doesn't have `shapes` (v3 only) and viceversa

glyphs-reader/src/font.rs
rsheeter: I'd much prefer to keep the logic here "dumb" if possible. I don't think the representations are ambiguous so it's OK.

glyphs-reader/src/font.rs
rsheeter: I think the same reasoning as https://github.com/googlefonts/fontmake-rs/pull/30/files#r1040811887 applies

glyphs-reader/src/font.rs
rsheeter: Added a test and fixed, ty.

glyphs-reader/src/font.rs
rsheeter: I think that's true but this way felt more correct despite arguably being overkill

glyphs-reader/src/font.rs
rsheeter: Good idea, done. Ran afoul of the traditional no global String's but still an improvement I think.

anthrotype (approved):

PR #31, opened 2022-12-06
replace toml-rs with serde_yaml

### LINKED ISSUE #28: replace toml with yaml for IR text format
toml-rs doesn't seem to support arbitrary structures (e.g. glyph contours as a vector of vectors of points), plus the format itself feels unsuitable for deeply nested data structures, so as per internal chat, we shall experiment with serde_yaml instead.

anthrotype:
Fixes #28


cmyr (approved):
yay serde!

PR #37, opened 2022-12-07
Parse unicode from glyphs files

### LINKED ISSUE #34: Parse Glyphs unicode attributes properly
Currently we don't really _use_ the unicode attribute but we will surely need it as we get to things like cmap.

It has more nuances than you'd guess, such as v2 using comma separated unquoted strings.

See https://github.com/googlefonts/fontmake-rs/pull/30#pullrequestreview-1206616519.

rsheeter:
Add examples of all known codepoint formats in glyphs files and prove we can parse them. Fixes #34.

I wound up just bashing it with a regex to make it consistent as that lets me keep the logic simple.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -10,11 +10,11 @@ use std::{fs, path};
 
 use log::warn;
 use ordered_float::OrderedFloat;
+use regex::Regex;
 
 use crate::error::Error;
 use crate::from_plist::FromPlist;
 use crate::plist::Plist;
-use crate::to_plist::ToPlist;
 
 const V3_METRIC_NAMES: [&str; 5] = [
     "ascender",
@@ -25,7 +25,7 @@ const V3_METRIC_NAMES: [&str; 5] = [
 ];
 const V2_METRIC_NAMES: [&str; 5] = ["ascender", "baseline", "descender", "capHeight", "xHeight"];
 
-#[derive(Debug, FromPlist, ToPlist, PartialEq, Eq, Hash)]
+#[derive(Debug, FromPlist, PartialEq, Eq, Hash)]
 pub struct Font {
     pub family_name: String,
     pub axes: Option<Vec<Axis>>,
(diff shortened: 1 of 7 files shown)


glyphs-reader/src/font.rs
```diff
@@ -378,6 +326,19 @@ fn custom_param<'a>(
 }
 
 impl Font {
+    fn get_codepoints(&mut self, radix: u32) {
```
anthrotype, line 329: usually if a method is called get_something I'd expect it to return something, in this case maybe would be better to name it "update_codepoints" or "parse_codepoints"?

glyphs-reader/src/font.rs
```diff
@@ -566,8 +519,17 @@ impl Font {
...
+    pub fn load(glyphs_file: &path::Path) -> Result<Font, Error> {
         let raw_content = fs::read_to_string(glyphs_file).map_err(Error::IoError)?;
+
+        // Glyphs has a wide variety of unicode definitions, not all of them parser friendly
+        // Make unicode always a string, without any wrapping () so we can parse as csv, radix based on format version
+        let re = Regex::new(
+            r#"(?m)^(?P<prefix>\s*unicode\s*=\s*)[(]?(?P<value>[0-9a-zA-Z,]+)[)]?;\s*$"#,
+        )
+        .unwrap();
+        let raw_content = re.replace_all(&raw_content, r#"$prefix"$value";"#);
```
anthrotype, line 531: clever hack.. but it's a pity we have to resort to this. The v3 formatting of unicodes was meant to overcome this ambiguity, so treating all as csv strings feels to me like a push back. But if it's the price to pay to keep supporting v2, I can live with that.

resources/testdata/glyphs2/Unicode-UnquotedHex.glyphs
```diff
@@ -0,0 +1,12 @@
+{
+familyName = WghtVar;
+fontMaster = ();
+glyphs = (
+	{
+		glyphname = name;
+		layers = ();
+		unicode = 03A2;
```
anthrotype, line 8: how about choosing a hex number that could also be interpreted as a decimal one by a naive parser, e.g. one without A-F digits such as "1234"?

anthrotype (approved):

glyphs-reader/src/font.rs
rsheeter: > The v3 formatting of unicodes was meant to overcome this ambiguity

To be fair if we only cared about v3 it would, v3 the only issue is that sometimes it's a scalar and sometimes a sequence. This is the simplest way I could think of to deal with it and I think the simplicity has some value.

anthrotype: well, that's the way it is, in v3 the type is either a scalar or array of scalar, which one can deal with, and a generic parser can work with without extra hacks like this regex find/replace.
Also, now that newlines are no longer escaped with octal escape sequences but added verbatim, I believe (but not 100% certain, need to test) that it's not unthinkable to have multiline strings (e.g. note or FEA features) containing that `unicode = {digits};` pattern may occur and you'd be incorrectly replacing that..

rsheeter: Theoretically true I think. We could reduce the odds by including the exact amount of preceding whitespace (Glyphs app seems to be consistent) or make it fully sound by altering the approach.

Filed https://github.com/googlefonts/fontmake-rs/issues/40 as I think this is sufficient to bootstrap us.

glyphs-reader/src/font.rs
rsheeter: Oh good catch, I think originally it did return them but then I made it not. Renamed.

resources/testdata/glyphs2/Unicode-UnquotedHex.glyphs
rsheeter: Done

PR #39, opened 2022-12-08
Playing with how to move state around

rsheeter:
Exploratory beginning of moving #26 into main fontc. Meant to establish the means to capture data from, and provide data to, work so we can do things like spawn a BE make glyph task that can safely acquire any bits of IR it needs to execute.

### DIFF
--- a/fontc/src/main.rs
+++ b/fontc/src/main.rs
@@ -9,6 +9,7 @@ use clap::Parser;
 use fontc::Error;
 use fontir::{
     error::WorkError,
+    orchestration::Context,
     source::{DeleteWork, Input, Paths, Source, Work},
     stateset::StateSet,
 };
@@ -158,7 +159,7 @@ fn glyphs_deleted<'a>(current_inputs: &Input, prev_inputs: &'a Input) -> HashSet
 
 fn add_glyph_work(
     context: &mut CompileContext,
-    work: &mut Vec<Box<dyn Work<()>>>,
+    work: &mut Vec<Box<dyn Work>>,
 ) -> Result<(), Error> {
     // Destroy IR for deleted glyphs
     work.extend(
@@ -177,12 +178,14 @@ fn add_glyph_work(
     Ok(())
 }
 
-fn do_work(work: Vec<Box<dyn Work<()>>>) -> Result<(), Error> {
+fn do_work(work: Vec<Box<dyn Work>>) -> Result<(), Error> {
     let work_units = work.len();
     debug!("{} work units to execute", work_units);
+    let root_context = Context::new_root();
+    // TODO: https://github.com/googlefonts/fontmake-rs/pull/26 style execution w/task-specific contexts
     let errors: Vec<Result<(), WorkError>> = work
         .into_par_iter()
-        .map(|work| work.exec())
+        .map(|work| work.exec(&root_context))
         .filter(|r| r.is_err())
         .collect();
     for error in errors.iter() {
@@ -238,7 +241,7 @@ fn main() -> Result<(), Error> {
     // perhaps we should do the IO first, e.g. read (but not parse) inputs, etc?
 
     // Accumulate work for the parts of IR that trivially parallelize
-    let mut work: Vec<Box<dyn Work<()>>> = Vec::new();
+    let mut work: Vec<Box<dyn Work>> = Vec::new();
 
     add_glyph_work(&mut context, &mut work)?;
 
@@ -260,7 +263,7 @@ mod tests {
     };
 
     use filetime::FileTime;
-    use fontir::{source::Paths, stateset::StateSet};
+    use fontir::{orchestration::Context, source::Paths, stateset::StateSet};
     use tempfile::tempdir;
 
     use crate::{
@@ -351,8 +354,9 @@ mod tests {
         add_glyph_work(&mut context, &mut work).unwrap();
 
         // Try to do the work
+        let root_context = Context::new_root();
         for work_item in work {
-            work_item.exec().unwrap();
+            work_item.exec(&root_context).unwrap();
         }
 
         finish_successfully(context).unwrap();
--- a/fontc/src/main.rs
+++ b/fontc/src/main.rs
@@ -183,10 +183,9 @@ fn do_work(work: Vec<Box<dyn Work>>) -> Result<(), Error> {
     debug!("{} work units to execute", work_units);
     let root_context = Context::new_root();
     // TODO: https://github.com/googlefonts/fontmake-rs/pull/26 style execution w/task-specific contexts
-    let errors: Vec<Result<(), WorkError>> = work
+    let errors: Vec<WorkError> = work
         .into_par_iter()
-        .map(|work| work.exec(&root_context))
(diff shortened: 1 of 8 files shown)


fontir/src/orchestration.rs
```diff
@@ -0,0 +1,108 @@
+// Helps coordinate the graph execution
+// Eventually has to move outside fontir to let us use for BE work
```
anthrotype, line 2: hm what's "BE work"?

fontir/src/ir.rs
```diff
@@ -4,6 +4,12 @@ use std::collections::BTreeMap;
 
 use serde::{Deserialize, Serialize};
 
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+pub struct GlobalMetadata {}
```
anthrotype, line 8: I wonder if we should call this global font-info stuff that cannot vary (e.g. upem, axes definitions) `StaticMetadata` to distinguish from other `VariableMetadata` properties that can vary across multiple masters like font-wide metrics ascender/descender and such for MVAR, which we haven't added yet as we're focussing on glyph-level variations for now, but eventually we will.

fontir/src/orchestration.rs
anthrotype: oh "back end", right

fontc/src/main.rs
```diff
@@ -158,7 +159,7 @@ fn glyphs_deleted<'a>(current_inputs: &Input, prev_inputs: &'a Input) -> HashSet
 
 fn add_glyph_work(
     context: &mut CompileContext,
-    work: &mut Vec<Box<dyn Work<()>>>,
+    work: &mut Vec<Box<dyn Work>>,
```
cmyr, line 162: not really about this PR, but this fn feels like it could maybe be a method on `CompileContext`?

fontc/src/main.rs
```diff
@@ -177,12 +178,14 @@ fn add_glyph_work(
...
-fn do_work(work: Vec<Box<dyn Work<()>>>) -> Result<(), Error> {
+fn do_work(work: Vec<Box<dyn Work>>) -> Result<(), Error> {
     let work_units = work.len();
     debug!("{} work units to execute", work_units);
+    let root_context = Context::new_root();
+    // TODO: https://github.com/googlefonts/fontmake-rs/pull/26 style execution w/task-specific contexts
     let errors: Vec<Result<(), WorkError>> = work
         .into_par_iter()
-        .map(|work| work.exec())
+        .map(|work| work.exec(&root_context))
```
cmyr, line 188: also not about this pr, but you instead of collecting results, you could just collect the errors, something like:

```rust
let errors = work
    .into_par_iter()
    .filter_map(|work| work.exec(&root_context).err())
    .collect::<Vec<_>>();
```

fontir/src/orchestration.rs
```diff
@@ -0,0 +1,108 @@
+// Helps coordinate the graph execution
```
cmyr, line 1: This line looks like it could pass as documentation for the module, in which case use `//!` :)

fontir/src/orchestration.rs
```diff
@@ -0,0 +1,108 @@
...
+    // Otherwise you totally get to write whatever you like
+    write_mask: Option<WorkIdentifier>,
+
+    // If present, what you can access through this context
+    // Intent is root has None, task-specific Context only allows access to dependencies
+    read_mask: Option<HashSet<WorkIdentifier>>,
+
+    // work results we've completed or restored from disk
+    // We create individual caches so we can return typed results from get fns
+    global_metadata_cache: Arc<RwLock<Option<Arc<GlobalMetadata>>>>,
```
cmyr, line 39: This feels like it might want to be its own type.

fontir/src/orchestration.rs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            panic!("Illegal access");
+        }
+    }
+
+    fn check_write_access(&self, id: &WorkIdentifier) {
+        if !self.write_mask.map(|mask| mask == *id).unwrap_or(true) {
+            panic!("Illegal access");
+        }
+    }
```
cmyr, line 83: So something to consider here: you are describing an invariant ("type must have access in order to perform task") and enforcing it through a convention ("you have to call this method anytime you want to read or write") but you could also consider enforcing it *through the type system itself*.

e.g, imagine the following:

```rust
struct ReadableContext<'a>(&'a Context);

impl Context {
    fn readable(&'a self, id: &WorkIdentifier) -> ReadableContext<'a> {
        if !self.can_read(id) {
            panic!("Illegal access");
        }
        ReadableContext(self)
    }
}
```

(imagine something similar for `WriteableContext`)

And then all the methods for reading/writing exist only on the appropriate contexts, so you actually can't call them at all unless you've gone through this check method.

I'm not sure if it's appropriate here, but in general this is a nice pattern for ensuring invariants are upheld, and it becomes more and more valuable as a project gets more complex.

cmyr (approved):
Overall seems reasonable, a few comments inline. :)

fontir/src/orchestration.rs
rsheeter: The problem I ran into with such a model is that any given context can read some arbitrary subset of readable things and that doesn't express in types very well as far as I can tell. For example, imagine a BE glyph task whose dependencies are {global metadata, IR for the glyph and it's components} ... how do I model that in the type system?

fontir/src/orchestration.rs
rsheeter: Heh +1!

fontir/src/ir.rs
rsheeter: +1, will rename

rsheeter: ...and rip off your comment as type documentation

fontc/src/main.rs
rsheeter: I have to think a little about where the actual task/dep data lives ... that might make sense indeed.

fontc/src/main.rs
rsheeter: ...good change until then though, updating

fontir/src/orchestration.rs
rsheeter: Done

fontc/src/main.rs
rsheeter: Done

fontir/src/orchestration.rs
cmyr: Yea, it may not be the right fit in this instance, but wanted to at least bring it up as a design pattern I would consider. :)

cmyr (approved):
looks good!

fontir/src/orchestration.rs
rsheeter: Hm. OK, only a very shallow type for now as the types cached vary. If I find strong patterns as we add more we can add more types.

fontir/src/orchestration.rs
rsheeter: It's a glorious model, I just couldn't quite see how to make it work.

Issue #27, opened 2022-12-05
Compile Glyphs v2 and 3 => IR


rsheeter:
See https://github.com/googlefonts/fontmake-rs/pull/25#pullrequestreview-1204966830 for context

rsheeter:
Copying @anthrotype's comment about Glyphs 2 vs 3 here:  in glyphs3 they are no longer a string like "100 200 LINE" but a tuple of (x, y, type, user_data). More differences b/w 2 vs 3 here: https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md#differences-between-version-2

rsheeter:
Core strategy is to upgrade v2 to v3 and then convert v3 to IR.

rsheeter:
Ref https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/Lib/glyphsLib/builder/paths.py#L20-L62 for path conversion

rsheeter:
I think we are at the point that individual narrower focus issues are more useful.

PR #42, opened 2022-12-09
Get the axis ranges from .glyphs files

rsheeter:
Step toward #27

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -325,6 +325,24 @@ fn custom_param<'a>(
     None
 }
 
+fn glyphs_v2_field_name_and_default(nth_axis: usize) -> Result<(&'static str, f64), Error> {
+    // Per https://github.com/googlefonts/fontmake-rs/pull/42#pullrequestreview-1211619812
+    // the field to use is based on the order in axes NOT the tag.
+    // That is, whatever the first axis is, it's value is in the weightValue field. Long sigh.
+    // Defaults per https://github.com/googlefonts/fontmake-rs/pull/42#discussion_r1044415236.
+    Ok(match nth_axis {
+        0 => ("weightValue", 400_f64),
+        1 => ("widthValue", 100_f64),
+        2 => ("customValue", 0_f64),
+        _ => {
+            return Err(Error::StructuralError(format!(
+                "We don't know what field to use for axis {}",
+                nth_axis
+            )))
+        }
+    })
+}
+
 impl Font {
     fn parse_codepoints(&mut self, radix: u32) {
         for glyph in self.glyphs.iter_mut() {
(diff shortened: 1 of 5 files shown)


resources/testdata/glyphs2/WghtVar_ImplicitAxes.glyphs
```diff
@@ -20,6 +20,8 @@ capHeight = 0;
 descender = -200;
 id = m01;
 weightValue = 400;
+widthValue = 100;
+customValue = 0;
```
anthrotype, line 24: did you edit these `widthValue` and `customValue` manually? I believe Glyphs doesn't write these fields out when they equal to their default (like many other attributes).

see:

https://github.com/googlefonts/glyphsLib/blob/main/Lib/glyphsLib/classes.py#L1563
https://github.com/googlefonts/glyphsLib/blob/f6e9c4a29ce764d34c309caef5118c48c156be36/Lib/glyphsLib/classes.py#L1517

anthrotype: for v2, the default value of weightValue and widthValue is 100, the default value of customValue is 0

glyphs-reader/src/font.rs
```diff
@@ -391,6 +391,43 @@ impl Font {
...
+                            axis.tag
+                        )))
+                    }
+                };
+                let value = master
+                    .other_stuff
+                    .remove(field_name)
+                    .ok_or_else(|| {
+                        Error::StructuralError(format!(
+                            "Missing '{}' in\n{:#?}",
```
anthrotype, line 415: you shouldn't expect those axis values are present in the master dict as they were omitted in v2 when equal to their default (100 for weightValue and widthValue, 0 for customValue)
see https://github.com/googlefonts/fontmake-rs/pull/42/files#r1044415236

resources/testdata/glyphs2/WghtVar_ImplicitAxes.glyphs
rsheeter: Oh yes I did. OK, I'll take them out and read-or-default.

rsheeter: Removed

glyphs-reader/src/font.rs
rsheeter: Done

glyphs-reader/src/font.rs
```diff
@@ -325,6 +325,24 @@ fn custom_param<'a>(
     None
 }
 
+fn glyphs_v2_field_name_and_default(nth_axis: usize) -> Result<(&'static str, f64), Error> {
+    // Per https://github.com/googlefonts/fontmake-rs/pull/42#pullrequestreview-1211619812
+    // the field to use is based on the order in axes NOT the tag.
+    // That is, whatever the first axis is, it's value is in the weightValue field. Long sigh.
+    // Defaults per https://github.com/googlefonts/fontmake-rs/pull/42#discussion_r1044415236.
+    Ok(match nth_axis {
+        0 => ("weightValue", 400_f64),
```
anthrotype, line 334: the weightValue default is actually 100, in design coordinates, but I see you return 400 in user-space coordinates, deliberately I suppose...

resources/testdata/glyphs2/WghtVar_ImplicitAxes.glyphs
```diff
@@ -20,8 +20,6 @@ capHeight = 0;
 descender = -200;
 id = m01;
 weightValue = 400;
```
anthrotype, line 22: in fact, you'll notice how `weightValue = 400;` here is _not_ omitted by Glyphs.app (or glyphsLib) when serializing .glyphs source file, because != 100 (the actual omit-if-default value)

glyphs-reader/src/font.rs
rsheeter: Fixed

anthrotype (commented):
retroactively LGTM!

Issue #22, opened 2022-12-01
map between design, external, and internal coordinates correctly


anthrotype:
designspace coordinates can be interpreted using so-called user-space (e.g. fvar axis [min,default,max], or OT registered axes predefined ranges) or as internal design-space coordinates (arbitrary numbers convenient to designers for defining source locations).
Designspace axes contain a `<map>` elements to map from the former to the latter, and are the basis for defining the `avar` table that "bends" the default normalization process (the one mapping min to -1, default to 0 and max to +1 and lerp anything in between).

Currently we don't even parse these map elements and assume there's no such user vs design vs normalised space distinction. 

See https://github.com/googlefonts/fontmake-rs/pull/18#discussion_r1036400605

rsheeter:
Copied from @anthrotype's note on https://github.com/googlefonts/fontmake-rs/pull/43#discussion_r1044468469:

> The master axis values in Glyphs are in internal design coordinates. The global axes definitions (the min/default/max that go in fvar, the .designspace axes min/default/max, and hence also our ir:Axis) are in user-space coordinates.
> The mapping between user-space to internal design-space coordinates is defined in Glyphs in multiple ways
> 
> directly via the "Axis Mappings" font custom parameter
> via "Axis Location" custom parameters set on each font master
> implicitly by deriving it for wght and wdth axis only from the user-space location of the instances
> it's one of the most complicated pieces of code in glyphsLib, for your amusement
> 
> https://github.com/googlefonts/glyphsLib/blob/f6e9c4a29ce764d34c309caef5118c48c156be36/Lib/glyphsLib/builder/axes.py#L155

I think we'll want explicit types so you cannot accidentally stick designspace coords in as userspace and vice versa.

rsheeter:
https://github.com/googlefonts/glyphsLib/issues/280#issuecomment-374914714 appears to explain how to interpret "Axis Location": the mapping is 
> design locations = xxxValue and user locations = "Axis Location" parameter

rsheeter:
#83 left out instances: https://github.com/googlefonts/fontmake-rs/pull/83#discussion_r1065795413

rsheeter:
https://github.com/googlefonts/fontc/issues/918 tracks instances Axis Location. As that is not itself a mapping between coordinates issue I think we can close this.

PR #43, opened 2022-12-09
Push Glyphs axes => StaticMetadata using Context

### LINKED ISSUE #44: Process Glyphs Variable Font Origin
https://glyphsapp.com/learn/creating-a-variable-font suggests the first master is default *unless* a variable font origin is specified. Variable Font Origin seems to be a custom parameter on the root that gives the id of the winner:

```
customParameters = (
{
name = "Variable Font Origin";
value = "E09E0C54-128D-4FEA-B209-1B70BEFE300B";
}
);
```

rsheeter:
Begin moving towards graph orchestration akin to that shown in #26.

* Use Context to manage access to data; work reads from and publishes to it.
* Allow writing IR to disk to be turned off.
* Make pushing to Context do the write to disk if it is enabled.
* Teach glyphs and ufo 2ir to write static metadata
   * As a result of this realize you have no idea how to set VF origin in Glyphs so add that; fixes #44.
* Put the hook in for glyphs2ir to write glyphs but don't do anything useful from it

Builds on #42. Step toward #27.

### DIFF
--- a/fontc/src/main.rs
+++ b/fontc/src/main.rs
@@ -37,13 +37,23 @@ struct Config {
     args: Args,
     // The compiler previously used so if the compiler changes config invalidates
     compiler: StateSet,
+
+    build_dir: PathBuf,
 }
 
 impl Config {
-    fn new(args: Args) -> Result<Config, io::Error> {
+    fn new(args: Args, build_dir: PathBuf) -> Result<Config, io::Error> {
         let mut compiler = StateSet::new();
         compiler.track_file(&std::env::current_exe()?)?;
-        Ok(Config { args, compiler })
+        Ok(Config {
+            args,
+            compiler,
+            build_dir,
+        })
+    }
+
+    fn file(&self) -> PathBuf {
+        self.build_dir.join("fontc.yml")
     }
 
     fn has_changed(&self, config_file: &Path) -> bool {
@@ -71,15 +81,10 @@ fn require_dir(dir: &Path) -> Result<PathBuf, io::Error> {
     Ok(dir.to_path_buf())
 }
 
-fn config_file(build_dir: &Path) -> PathBuf {
-    build_dir.join("fontc.yml")
-}
+fn init(config: &Config) -> Result<Input, io::Error> {
+    let config_file = config.file();
 
-fn init(build_dir: &Path, args: Args) -> Result<(Config, Input), io::Error> {
-    let config = Config::new(args)?;
-    let config_file = config_file(build_dir);
-
-    let ir_paths = Paths::new(build_dir);
+    let ir_paths = Paths::new(&config.build_dir);
     let ir_input_file = ir_paths.ir_input_file();
     if config.has_changed(&config_file) {
         info!("Config changed, generating a new one");
@@ -99,7 +104,7 @@ fn init(build_dir: &Path, args: Args) -> Result<(Config, Input), io::Error> {
     } else {
         Input::new()
     };
-    Ok((config, ir_input))
+    Ok(ir_input)
 }
 
 fn ir_source(source: &Path) -> Result<Box<dyn Source>, Error> {
@@ -253,7 +258,8 @@ fn main() -> Result<(), Error> {
     let paths = Paths::new(Path::new("build"));
     let build_dir = require_dir(paths.build_dir())?;
     require_dir(paths.glyph_ir_dir())?;
-    let (config, prev_inputs) = init(&build_dir, Args::parse()).map_err(Error::IoError)?;
+    let config = Config::new(Args::parse(), build_dir)?;
+    let prev_inputs = init(&config).map_err(Error::IoError)?;
 
     let mut change_detector = ChangeDetector::new(config.clone(), paths.clone(), prev_inputs)?;
     let context = Context::new_root(
@@ -299,7 +305,7 @@ mod tests {
     use tempfile::tempdir;
 
     use crate::{
-        add_glyph_work, add_static_metadata_work, config_file, finish_successfully, glyphs_changed,
+        add_glyph_work, add_static_metadata_work, finish_successfully, glyphs_changed,
         glyphs_deleted, init, require_dir, Args, ChangeDetector, Config,
     };
 
@@ -343,9 +349,10 @@ mod tests {
         let temp_dir = tempdir().unwrap();
         let build_dir = temp_dir.path();
         let args = test_args("wght_var.designspace");
+        let config = Config::new(args.clone(), build_dir.to_path_buf()).unwrap();
 
-        init(build_dir, args.clone()).unwrap();
-        let config_file = config_file(build_dir);
+        init(&config).unwrap();
+        let config_file = config.file();
         let paths = Paths::new(build_dir);
         let ir_input_file = paths.ir_input_file();
 
@@ -355,7 +362,9 @@ mod tests {
             "Should not exist: {:#?}",
             ir_input_file
         );
-        assert!(!Config::new(args).unwrap().has_changed(&config_file));
+        assert!(!Config::new(args, build_dir.to_path_buf())
+            .unwrap()
+            .has_changed(&config_file));
     }
 
     #[test]
@@ -363,7 +372,7 @@ mod tests {
         let temp_dir = tempdir().unwrap();
         let build_dir = temp_dir.path();
         let args = test_args("wght_var.designspace");
-        init(build_dir, args.clone()).unwrap();
+        let config = Config::new(args.clone(), build_dir.to_path_buf()).unwrap();
 
         let compiler_location = std::env::current_exe().unwrap();
         let metadata = compiler_location.metadata().unwrap();
@@ -374,15 +383,21 @@ mod tests {
             FileTime::from_system_time(metadata.modified().unwrap()),
             metadata.len() + 1,
         );
-        assert!(Config { args, compiler }.has_changed(&config_file(build_dir)));
+        assert!(Config {
+            args,
+            compiler,
+            build_dir: build_dir.to_path_buf()
+        }
+        .has_changed(&config.file()));
     }
 
     fn compile(build_dir: &Path, source: &str) -> TestCompile {
         let args = test_args(source);
         let paths = Paths::new(build_dir);
         require_dir(paths.glyph_ir_dir()).unwrap();
+        let config = Config::new(args, build_dir.to_path_buf()).unwrap();
 
-        let (config, prev_inputs) = init(build_dir, args).unwrap();
+        let prev_inputs = init(&config).unwrap();
 
         let mut change_detector =
             ChangeDetector::new(config.clone(), paths.clone(), prev_inputs).unwrap();
--- a/fontc/src/main.rs
+++ b/fontc/src/main.rs
@@ -9,7 +9,7 @@ use clap::Parser;
 use fontc::Error;
 use fontir::{
     error::WorkError,
-    orchestration::Context,
+    orchestration::{Context, WorkIdentifier},
     source::{DeleteWork, Input, Paths, Source, Work},
     stateset::StateSet,
 };
@@ -25,6 +25,11 @@ struct Args {
     /// A designspace, ufo, or glyphs file
     #[arg(short, long)]
     source: PathBuf,
+
+    /// Whether to write IR to disk. Must be true if you want incremental compilation.
+    #[arg(short, long)]
+    #[clap(default_value = "true")]
+    emit_ir: bool,
 }
 
 #[derive(Serialize, Deserialize, Debug, PartialEq)]
@@ -97,22 +102,19 @@ fn init(build_dir: &Path, args: Args) -> Result<(Config, Input), io::Error> {
     Ok((config, ir_input))
 }
 
-fn ir_source(source: &Path, paths: Paths) -> Result<Box<dyn Source>, Error> {
+fn ir_source(source: &Path) -> Result<Box<dyn Source>, Error> {
     let ext = source
         .extension()
         .and_then(OsStr::to_str)
         .ok_or_else(|| Error::UnrecognizedSource(source.to_path_buf()))?;
     match ext {
-        "designspace" => Ok(Box::from(DesignSpaceIrSource::new(
-            source.to_path_buf(),
-            paths,
-        ))),
-        "glyphs" => Ok(Box::from(GlyphsIrSource::new(source.to_path_buf(), paths))),
+        "designspace" => Ok(Box::from(DesignSpaceIrSource::new(source.to_path_buf()))),
+        "glyphs" => Ok(Box::from(GlyphsIrSource::new(source.to_path_buf()))),
         _ => Err(Error::UnrecognizedSource(source.to_path_buf())),
     }
 }
 
-fn finish_successfully(context: CompileContext) -> Result<(), Error> {
+fn finish_successfully(context: ChangeDetector) -> Result<(), Error> {
     let current_sources =
         serde_yaml::to_string(&context.current_inputs).map_err(Error::YamlSerError)?;
     fs::write(context.paths.ir_input_file(), current_sources).map_err(Error::IoError)
@@ -122,7 +124,7 @@ fn global_change(current_inputs: &Input, prev_inputs: &Input) -> bool {
     current_inputs.global_metadata != prev_inputs.global_metadata
 }
 
-fn glyphs_changed(context: &CompileContext) -> HashSet<&str> {
+fn glyphs_changed(context: &ChangeDetector) -> HashSet<&str> {
     if global_change(&context.current_inputs, &context.prev_inputs) {
         return context
             .current_inputs
@@ -157,35 +159,55 @@ fn glyphs_deleted<'a>(current_inputs: &Input, prev_inputs: &'a Input) -> HashSet
         .collect()
 }
 
+fn add_static_metadata_work(
+    change_detector: &mut ChangeDetector,
+    context: &Context,
+    work: &mut Vec<Box<dyn Work>>,
+) -> Result<(), Error> {
+    let context = context.copy_for_work(WorkIdentifier::StaticMetadata, None);
+    work.push(
+        change_detector
+            .ir_source
+            .create_static_metadata_work(&context)?,
+    );
+    Ok(())
+}
+
 fn add_glyph_work(
-    context: &mut CompileContext,
+    config: &Config,
+    change_detector: &mut ChangeDetector,
+    context: &Context,
     work: &mut Vec<Box<dyn Work>>,
 ) -> Result<(), Error> {
     // Destroy IR for deleted glyphs
     work.extend(
-        glyphs_deleted(&context.current_inputs, &context.prev_inputs)
-            .iter()
-            .map(|glyph_name| DeleteWork::create(context.paths.glyph_ir_file(glyph_name))),
+        glyphs_deleted(
+            &change_detector.current_inputs,
+            &change_detector.prev_inputs,
+        )
+        .iter()
+        .map(|glyph_name| DeleteWork::create(change_detector.paths.glyph_ir_file(glyph_name))),
     );
 
     // Generate IR for changed glyphs
-    let glyphs_changed = glyphs_changed(context);
-    let glyph_work = context
+    let glyphs_changed = glyphs_changed(change_detector);
+    let glyph_work = change_detector
         .ir_source
-        .create_glyph_ir_work(&glyphs_changed, &context.current_inputs)?;
+        .create_glyph_ir_work(&glyphs_changed, context)?;
+    if config.args.emit_ir {}
     work.extend(glyph_work);
 
     Ok(())
 }
 
-fn do_work(work: Vec<Box<dyn Work>>) -> Result<(), Error> {
+fn do_work(context: &Context, work: Vec<Box<dyn Work>>) -> Result<(), Error> {
     let work_units = work.len();
     debug!("{} work units to execute", work_units);
-    let root_context = Context::new_root();
+
     // TODO: https://github.com/googlefonts/fontmake-rs/pull/26 style execution w/task-specific contexts
     let errors: Vec<WorkError> = work
         .into_par_iter()
-        .filter_map(|work| work.exec(&root_context).err())
+        .filter_map(|work| work.exec(context).err())
         .collect();
     for error in errors.iter() {
         warn!("{:#?}", error);
@@ -203,29 +225,32 @@ fn do_work(work: Vec<Box<dyn Work>>) -> Result<(), Error> {
     Ok(())
 }
 
-struct CompileContext {
+struct ChangeDetector {
     paths: Paths,
     ir_source: Box<dyn Source>,
     prev_inputs: Input,
     current_inputs: Input,
 }
 
-impl CompileContext {
-    fn new(paths: Paths, args: Args) -> Result<CompileContext, Error> {
+impl ChangeDetector {
+    fn new(paths: Paths, args: Args) -> Result<(Config, ChangeDetector), Error> {
         let build_dir = require_dir(paths.build_dir())?;
         require_dir(paths.glyph_ir_dir())?;
         let (config, prev_inputs) = init(&build_dir, args).map_err(Error::IoError)?;
 
         // What sources are we dealing with?
-        let mut ir_source = ir_source(&config.args.source, paths.clone())?;
+        let mut ir_source = ir_source(&config.args.source)?;
         let current_inputs = ir_source.inputs().map_err(Error::FontIrError)?;
 
-        Ok(CompileContext {
-            paths,
-            ir_source,
-            prev_inputs,
-            current_inputs,
-        })
+        Ok((
+            config,
+            ChangeDetector {
+                paths,
+                ir_source,
+                prev_inputs,
+                current_inputs,
+            },
+        ))
     }
 }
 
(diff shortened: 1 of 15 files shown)


glyphs-reader/src/font.rs
```diff
@@ -325,6 +326,40 @@ fn custom_param<'a>(
...
+    key: &str,
+) -> Option<(usize, &'a Plist)> {
+    let Some(custom_params) = custom_params(other_stuff) else {
+        warn!("customParameters isn't an array, omg omg omg");
+        return None;
+    };
+
+    let name_key = "name".to_string();
+    for (idx, custom_param) in custom_params.iter().enumerate() {
+        let Plist::Dictionary(dict) = custom_param else {
```
anthrotype, line 350: sorry but custom parameters can well be strings. They can be anything a plist can be :)

anthrotype: ignore me, I misunderstood the code, of course custom  parameter is a {"name": <string>, "value": Plist} dict

glyphs-reader/src/font.rs
```diff
@@ -588,6 +627,30 @@ impl Font {
...
+                    return 0;
+                };
+                self.font_master
+                    .iter()
+                    .enumerate()
+                    .find(|(_, master)| &master.id == origin)
+                    .map(|(idx, _)| idx)
+                    .unwrap_or_default()
+            })
+            .unwrap_or_default()
```
anthrotype, line 650: took me a bit to understand this, i'm not a big fan of these complicated chains of method calls, but if this is supposed to be idiomatic rust, oh well...

glyphs2fontir/src/source.rs
```diff
@@ -137,6 +144,76 @@ impl GlyphsIrSource {
...
+    fn exec(&self, context: &Context) -> Result<(), WorkError> {
+        let font = self.font.as_ref();
+        debug!("Static metadata for {}", font.family_name);
+
+        let mut axis_values = Vec::new();
+        for master in font.font_master.iter() {
+            master
+                .axes_values
+                .as_ref()
+                .ok_or_else(|| WorkError::InconsistentAxisDefinitions)?
```
anthrotype: what are 'inconsistent' axis definitions exactly?

glyphs2fontir/src/source.rs
```diff
@@ -137,6 +144,76 @@ impl GlyphsIrSource {
...
+                    .max()
+                    .unwrap();
+                let default =
+                    OrderedFloat::<f32>(axis_values[default_master_idx][idx].into_inner() as f32);
+
+                Axis {
+                    name: a.name.clone(),
+                    tag: a.tag.clone(),
+                    hidden: a.hidden.unwrap_or(false),
+                    min,
```
anthrotype, line 203: ok, so this immediately brings https://github.com/googlefonts/fontmake-rs/issues/22 to the front of the queue..
The master axis values in Glyphs are in internal design coordinates. The global axes definitions (the min/default/max that go in fvar, the .designspace axes min/default/max, and hence also our ir:Axis) are in user-space coordinates.
The mapping between user-space to internal design-space coordinates is defined in Glyphs in multiple ways

- directly via the "Axis Mappings" font custom parameter
- via "Axis Location" custom parameters set on each font master
- implicitly by deriving it for wght and wdth axis only from the user-space location of the instances

it's one of the most complicated pieces of code in glyphsLib, for your amusement 

https://github.com/googlefonts/glyphsLib/blob/f6e9c4a29ce764d34c309caef5118c48c156be36/Lib/glyphsLib/builder/axes.py#L155

fontc/src/main.rs
```diff
@@ -203,52 +225,66 @@ fn do_work(work: Vec<Box<dyn Work>>) -> Result<(), Error> {
...
     paths: Paths,
     ir_source: Box<dyn Source>,
     prev_inputs: Input,
     current_inputs: Input,
 }
 
-impl CompileContext {
-    fn new(paths: Paths, args: Args) -> Result<CompileContext, Error> {
+impl ChangeDetector {
+    fn new(paths: Paths, args: Args) -> Result<(Config, ChangeDetector), Error> {
```
anthrotype, line 236: i find it odd that new (supposedly the default constructor of an object?) returns a tuple instead of just the newly initialized object...

anthrotype: i suggest you initialise the Config beforehand from args, and pass that to ChangeDetector::new

fontir/src/ir.rs
```diff
@@ -1,18 +1,27 @@
...
+    pub axes: Vec<Axis>,
+    pub glyph_order: Vec<String>,
+}
+
+impl StaticMetadata {
+    pub fn glyph_id(&self, name: &str) -> Option<u32> {
+        self.glyph_order
+            .iter()
+            .position(|gn| gn == name)
+            .map(|pos| pos as u32)
```
anthrotype, line 23: wonder if we could/should cache this into a HashMap<String, u32> so glyph_id lookup is O(1) instead of O(n)

ufo2fontir/src/source.rs
```diff
@@ -157,34 +179,38 @@ impl Source for DesignSpaceIrSource {
...
                     return Err(Error::FileExpected(glif_file));
                 }
                 let glif_file = glif_file.clone();
-                glyphs
-                    .entry(glyph_name)
-                    .or_default()
-                    .track_file(&glif_file)?;
+                let entry = glyphs.entry(glyph_name.clone());
+                if let hash_map::Entry::Vacant(_) = entry {
+                    glyph_order.push(glyph_name);
```
anthrotype, line 199: the glyph order is not the order in which you find glyphs listed in a UFO layer's contents.plist, the latter is a dictionary and as such unsorted (the fact that you may see it ordered alphabetically is simply a way to serialise the plist dict in a deterministic way).

UFO defines a "public.glyphOrder" key in the UFO lib.plist, which is a list of glyph name strings. That's what fontmake (actually ufo2ft) uses. If it is not defined, ufo2ft sorts the glyphs alphabetically while ensuring that if .notdef is present (which usually is) it is placed in the first (gid 0) position:

https://github.com/googlefonts/ufo2ft/blob/dd738cdcddf61cce2a744d1cafab5c9b33e92dd4/Lib/ufo2ft/util.py#L23-L45

since you want the glyph order of the variable font (right now you're assuming there's 1 VF per designspace file, which isn't the case for v5 but anyway), then you have to get the public.glyphOrder from the default master (the source that is at the default location on all the axes), which is the starting point for the variable font to be built.

anthrotype: for reference: https://unifiedfontobject.org/versions/ufo3/lib.plist/#publicglyphorder

note that public.glyphOrder may contain more or even less glyphs than are in the font.. 😅 
you want to prune glyphs that are not actually in the font, and also append at the end the glyphs that are in the font but are not explicitly listed in the public.glyphOrder

basically what ufo2ft's makeOfficialGlyphOrder does

glyphs2fontir/src/source.rs
```diff
@@ -90,62 +106,129 @@ impl Source for GlyphsIrSource {
...
+                    default,
+                    max,
+                }
+            })
+            .collect();
+
+        let glyph_order = font.glyphs.iter().map(|g| g.glyphname.clone()).collect();
```
anthrotype, line 210: you may be delighted to learn that Glyphs.app has a `glyphOrder` custom parameter as well (ordered list of glyph names), that works similarly to UFO's public.glyphOrder

https://glyphsapp.com/learn/alter-the-glyph-order-in-font-view

if it's not present it's ok to do like you do, take the implicit ordering of the glyphs in the font as the glyph order.

fontir/src/orchestration.rs
```diff
@@ -116,6 +155,7 @@ impl Context {
 
     pub fn set_glyph_ir(&self, glyph_order: u32, ir: ir::Glyph) {
```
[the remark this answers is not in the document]
anthrotype: question: why the WorkIdentifier::GlyphIr variant holds a u32 for the glyph id and not a String for the ir::Glyph.name instead?
If the purpose is that it be unique, the glyph name is by definition as unique as it can be and identifies the given ir::Glyph more than the glyph index.

glyphs-reader/src/font.rs
rsheeter: I think you'll find it grows on you. THe code reads left to right, top to bottom like language in a way thats entirely not true of the normal way of doing it.

glyphs2fontir/src/source.rs
rsheeter: I copied ^ to #22 because I want to deal with it in subsequent PR(s)

fontir/src/ir.rs
rsheeter: We should, I just wanted to start with naively correct.

fontir/src/orchestration.rs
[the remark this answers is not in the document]
rsheeter: Idea was that WorkId is Copy and u32's are small. We could certainly have the name instead.

glyphs2fontir/src/source.rs
rsheeter: and *not* deal with it here, so resolving

fontc/src/main.rs
rsheeter: +1, fixing

rsheeter: Fixed

fontir/src/orchestration.rs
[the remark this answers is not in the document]
rsheeter: Hm. In quickly trying to use name I found I had to copy it quite a lot so my inclination is to start with gid. I renamed the parameter.

glyphs-reader/src/font.rs
rsheeter: To be fair, it's also very possible I still write un-idiomatic Rust

glyphs2fontir/src/source.rs
rsheeter: TIL. Added an example of this and a test.

rsheeter: When I tried this in the actual Glyphs app the glyphs in the list moved around to match my new glyph order. I have manually adjusted a file to make this not true for test purposes.

fontir/src/ir.rs
rsheeter: Added

glyphs-reader/src/font.rs
rsheeter: I think I made it somewhat clearer.

ufo2fontir/src/source.rs
rsheeter: v5 designspace I definitely want to tackle separately :)

rsheeter: I have added mk1 of building the glyph order. Until we do #22 - seems like that needs to be soon - the way it locates the default master will be entirely wrong.

glyphs2fontir/src/source.rs
rsheeter: I don't think this exists anymore.

ufo2fontir/src/source.rs
rsheeter: Also made .glyphs a bit more similar to ufo, though still honoring file order as default.

fontc/src/main.rs
```diff
@@ -233,32 +233,29 @@ struct ChangeDetector {
...
 }
 
 fn main() -> Result<(), Error> {
     env_logger::init();
 
     let paths = Paths::new(Path::new("build"));
-    let (config, mut change_detector) = ChangeDetector::new(paths.clone(), Args::parse())?;
+    let build_dir = require_dir(paths.build_dir())?;
+    require_dir(paths.glyph_ir_dir())?;
+    let (config, prev_inputs) = init(&build_dir, Args::parse()).map_err(Error::IoError)?;
```
anthrotype, line 256: even this `init` function feels a bit double-duty to me. You could make a new config from args and pass it to `init` alongside the build_dir (the latter could also be an attribute if Config btw).

The `init` function could then beme a constructor method for the Input type, that returns an instance of Input given a config

fontir/src/ir.rs
```diff
@@ -9,22 +9,36 @@ use std::collections::{BTreeMap, HashMap};
 ///
 /// For example, upem, axis definitions, etc, as distinct from
 /// metadata that varies across design space such as ascender/descender.
-#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(from = "StaticMetadataSerdeRepr", into = "StaticMetadataSerdeRepr")]
 pub struct StaticMetadata {
     pub axes: Vec<Axis>,
     pub glyph_order: Vec<String>,
+    pub gid_by_name: HashMap<String, u32>,
```
anthrotype, line 17: fwiw fonttools calls this kind of things a glyph_map. `gid_by_name` is also fine

anthrotype: if rust had a map type that remembers the insertion order we could have a single field instead of glyph_order (Vec<String>) and gid_by_name (HashMap<String, u32>); the keys of the ordered map would be the glyph order.

I see there's https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html 
maybe we can consider using that.

In Python 3.6+ dict keeps the insertion order, and you'll find that a lot of python font-related code relies on this inner implicitly ordered dictionary behavior.

glyphs-reader/src/font.rs
```diff
@@ -30,6 +30,7 @@ pub struct Font {
     pub family_name: String,
     pub axes: Option<Vec<Axis>>,
     pub glyphs: Vec<Glyph>,
+    pub glyph_order: Option<Vec<String>>,
```
anthrotype, line 33: technically in the Glyphs.app API, GSFont doesn't have such 'glyph_order' or 'glyphOrder' attribute, but maybe unlike in glyphsLib/classes.py, we don't care about matching that.

glyphs2fontir/src/source.rs
anthrotype: > When I tried this in the actual Glyphs app the glyphs in the list moved around to match my new glyph order

maybe that's a feature, and we can rely on the font.glyphs order matching any optional 'glyphOrder' custom parameter..
In fact I see that glyphsLib itself takes care of sorting the font.glyphs list by the UFO's glyphOrder when converting UFO => Glyphs

https://github.com/googlefonts/glyphsLib/blob/6763648a91ecadee256d0518c731e35eb311c013/Lib/glyphsLib/builder/builders.py#L555-L560

And when going Glyphs => UFO, glyphsLib always sets UFO public.glyphOrder to the order of the font.glyphs list even when no glyphOrder custom parameter is set.

https://github.com/googlefonts/glyphsLib/blob/6763648a91ecadee256d0518c731e35eb311c013/Lib/glyphsLib/builder/font.py#L88-L91

so it looks consistent with your findings

glyphs-reader/src/font.rs
anthrotype: idea: instead of storing an additional `glyph_order` Vec<String> (which also doesn't exist in the Glyphs.app API), you could reorder the font.glyphs Vec<Glyph> in-place such that it matches any glyphOrder custom parameter 

(and that they do _not_ match looks unlikely given https://github.com/googlefonts/fontmake-rs/pull/43#discussion_r1045790918)

ufo2fontir/src/source.rs
```diff
@@ -140,10 +140,20 @@ impl DesignSpaceIrSource {
...
+            for filename in ["fontinfo.plist", "layercontents.plist", "lib.plist"] {
+                // Only track lib.plist for the default master
+                if filename == "lib.plist" && source != default_master {
+                    continue;
+                }
+
                 let font_info_file = ufo_dir.join(filename);
+                // TODO: this is incorrect; several of these files are optional
+                // File tracking curently assumes you only track extant files so keep it for now
                 if !font_info_file.is_file() {
```
anthrotype, line 157: only layercontents.plist is required for UFOv3, lib.plist and fontinfo.plist are optional, though the latter is most likely present as you can't do much, let alone compile a font, if you're not at least given some basic parameters like UPEM and a few others.

ufo2fontir/src/source.rs
```diff
@@ -290,7 +299,73 @@ impl DesignSpaceIrSource {
...
+        .collect();
+    designspace
+        .sources
+        .iter()
+        .find(|source| to_ir_location(&source.location) == default_location)
+}
+
+fn load_lib_plist(ufo_dir: &Path) -> Result<plist::Dictionary, WorkError> {
+    let lib_plist_file = ufo_dir.join("lib.plist");
+    if !lib_plist_file.is_file() {
```
anthrotype, line 322: you should not expecet lib.plist to be present, the whole file is optional. If missing return an  empty plist dict

ufo2fontir/src/source.rs
```diff
@@ -193,21 +203,20 @@ impl Source for DesignSpaceIrSource {
                 if !glif_file.exists() {
                     return Err(Error::FileExpected(glif_file));
                 }
+                glyph_names.insert(glyph_name.clone());
```
anthrotype, line 206: here you are collecting the union of all the glyph names across all the source UFOs, irrespective of them being present or not in the default source, but as I think we discussed only glyphs that are in the default source will exist in the output VF, the non-default sources may contain _less_ glyphs (be sparse), but not _more_ glyphs than the default one.
Since all VF font projects are designed like that it's unlikely that we'll ever find in such a situation where the nomially default source does not already contain the superset of all glyphs defined across all sources.
But I figured I should mention this here. And perhaps we could even add an explicit check for it, skip + warn if there's any such glyph defined elsewhere than default source.

resources/testdata/glyphs3/WghtVar_GlyphOrder.glyphs
```diff
@@ -17,7 +17,7 @@ name = glyphOrder;
 value = (
 hyphen,
 space,
-exclam
+not-a-valid-name
```
anthrotype, line 20: IIRC, Glyphs.app used to crash when a glyph name was present only in the glyphOrder custom parameter but not in the actual glyphs source file... Not sure if this is still the case.

glyphs-reader/src/font.rs
```diff
@@ -394,24 +394,32 @@ impl Font {
...
+                        }
+                    });
+            };
+        };
+
+        // Add anything left over in file order
+        self.glyphs
+            .iter()
+            .filter(|g| valid_names.contains(&g.glyphname))
+            .for_each(|g| glyph_order.push(g.glyphname.clone()));
```
anthrotype, line 420: In python I would use `extend` method on a list, and pass a generator expression or filter.
Here you do iter+filter, followed by for_each and a push. I believe in Rust Vector has a similar extend mode which takes an iterator. Anyway just saying

ufo2fontir/src/source.rs
```diff
@@ -290,7 +299,73 @@ impl DesignSpaceIrSource {
...
+    plist::Value::from_file(&lib_plist_file)
+        .map_err(|e| WorkError::ParseError(lib_plist_file.clone(), format!("{}", e)))?
+        .into_dictionary()
+        .ok_or_else(|| WorkError::ParseError(lib_plist_file, "Not a dictionary".to_string()))
+}
+
+// Per https://github.com/googlefonts/fontmake-rs/pull/43/files#r1044596662
+fn glyph_order(
+    designspace: &DesignSpaceDocument,
+    designspace_dir: &Path,
```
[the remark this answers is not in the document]
rsheeter: I actually think it shouldn't because that would inextricably tie it to the filesystem. It feels right to me to store both independently in places that do care about a filesystem source.

glyphs-reader/src/font.rs
rsheeter: It's probably subjective but I think this reads more cleanly than wrapping extend around it.

glyphs-reader/src/font.rs
rsheeter: I worry a little that maybe I just didn't find the way to make them not match but it does in fact exist.

IMHO the correct typing would be for glyphs to be a name:glyph map and glyph order to be it's own thing - which is what it turns into in IR - but we're limited wrt what the plist reader will readily accomodate.

Re-organizing the exisitng field feels all to easy to miss to me. Maybe the right choice would be to turn the raw Font from plist into another type and compute the derivative information when we transition.

fontir/src/ir.rs
rsheeter: I dislike the python style of not being able to know if order matters. Using an explicit ordered map type selectively does address that but ... is it worth a dependency? - I'm not sure about that.

ufo2fontir/src/source.rs
rsheeter: Tricky bit is if one appears we would want to trigger a recompile. Tracking in https://github.com/googlefonts/fontmake-rs/issues/49.

ufo2fontir/src/source.rs
rsheeter: Tricky bit is if one appears we would want to trigger a recompile. Tracking in https://github.com/googlefonts/fontmake-rs/issues/49.

glyphs-reader/src/font.rs
rsheeter: To my understanding we don't care about matching GSFont, we just want the information to power IR?

I'm going to try the two types approach.

rsheeter: I have now split RawFont, what we parse from plist, and Font which has some post-processing done on it. Font now has a name:glyph map and a glyph order as separate constructs.

rsheeter: I think it's better so resolving but plz reopen if you disagree

ufo2fontir/src/source.rs
rsheeter: Oh good catch. I really hate the idea of throwing away data but I suppose it's correct here.

rsheeter: Added a test to demonstrate the problem and fixed

fontir/src/ir.rs
rsheeter: I propose we not block the PR on this as the map is purely an accelerator (cache). If it proves annoying over time we can always update the data structure.

anthrotype: it's actually very convenient in python to have its default dict type be a *hash* map type that keeps insertion order by default, so that you get constant type lookup _and_ predictable (not random) ordering when iterating. It feels in our rust code sometimes we reach out for BTreeMap just to achive that predictability at expense of O(log2(n)) lookup

fontc/src/main.rs
```diff
@@ -253,7 +258,8 @@ fn main() -> Result<(), Error> {
     let paths = Paths::new(Path::new("build"));
     let build_dir = require_dir(paths.build_dir())?;
     require_dir(paths.glyph_ir_dir())?;
-    let (config, prev_inputs) = init(&build_dir, Args::parse()).map_err(Error::IoError)?;
+    let config = Config::new(Args::parse(), build_dir)?;
```
anthrotype, line 261: It's not ideal that we hardcode our build directory to be "build/" in the current directory, maybe it should become part of Args... then you would only need the args to initialize a Config. 
feel free to do that another time.

glyphs-reader/src/font.rs
```diff
@@ -25,12 +25,24 @@ const V3_METRIC_NAMES: [&str; 5] = [
...
-#[derive(Debug, FromPlist, PartialEq, Eq, Hash)]
+#[derive(Debug, PartialEq, Eq, Hash)]
 pub struct Font {
+    pub family_name: String,
+    pub axes: Vec<Axis>,
+    pub font_master: Vec<FontMaster>,
+    pub default_master_idx: usize,
+    pub glyphs: BTreeMap<String, Glyph>,
+    pub glyph_order: Vec<String>,
+    pub codepoints: BTreeMap<String, Vec<u32>>,
```
anthrotype, line 36: I strongly object to having a top-level `codepoints` mapping glyph names to unicode codepoints (u32), for two reasons:

1) cmap table (which is what this is about) maps codepoints to glyph (indices, or at higher level, glyph names), not the other way around. Thinking like a glyphs has this and that codepoint is the wrong way to think about them, as it may involuntarily lead to multiple glyphs being mapped to the same codepoint, which does not make any sense in a font. Only one glyph can be mapped to a given codepoint inside cmap table (while multiple codepoints may well map to the same glyph). In UFO4 the plan was/is to get rid of uncode elements inside GLIFs and add a cmap.txt csv file at the global level that maps unicodes to glyph names.
2) Moving this (inverted) cmappings to the Font level does not add/improve anything to keeping them inside the Glyph structure and doing the usual font["glyphname"].unicodes, which is what our sources already contain.

So if you like to have this top-level map (i'm not sure we need that here because (2) above), I recommed you invert the keys and values, and make it a map from u32 to String. Maybe we just do that for the IR and here at the source level we stay closer to the sources (where glyph "have" codepoints).

fontc/src/main.rs
anthrotype: https://github.com/googlefonts/fontmake-rs/issues/53

anthrotype (approved):
LGTM, except for the Font.codepoints as {String: u32} map (see [comment](https://github.com/googlefonts/fontmake-rs/pull/43#discussion_r1047127315))

fontc/src/main.rs
rsheeter: Good catch, will fix. +1 to standalone issue.

glyphs-reader/src/font.rs
rsheeter: For context, I want IR to separate the information to build cmap so it can be it's own task and not load glyphs. However, your point about the direction being wrong - even if it does mirror how Glyphs appears to think about it - stands so i will fix the direction.

fontir/src/ir.rs
rsheeter: I agree re convenience, I just dislike the python style of having *all* of them retain order. Having a separate type IndexMap addresses that so maybe we should indeed do that. I really want to land this as building on unmerged PRs is painful so I am going to mvoe that to an issue.

rsheeter: #54

glyphs-reader/src/font.rs
rsheeter: Fixed.

PR #46, opened 2022-12-11
Add a piecewise linear map based on the python version

rsheeter:
Meant for us in addressing #22

### DIFF
--- a/fontir/src/piecewise_linear_map.rs
+++ b/fontir/src/piecewise_linear_map.rs
@@ -1,6 +1,8 @@
 //! Remaps values using a series of linear mappings.
 //!
 //! Useful for things like designspace : userspace mapping.
+//! For example, from a <https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#location>
+//! xvalue to a userspace (fvar) value.
 
 use ordered_float::OrderedFloat;
 
@@ -78,7 +80,7 @@ mod tests {
 
     #[test]
     fn single_segment_map() {
-        // User likes to map wght 0..1000 onto 0..10
+        // User likes to use 0..10 in their sources to mean wght 0..1000 in userspace
         let from_to = vec![
             (OrderedFloat(0_f32), OrderedFloat(0_f32)),
             (OrderedFloat(10_f32), OrderedFloat(1000_f32)),
--- a/fontir/src/piecewise_linear_map.rs
+++ b/fontir/src/piecewise_linear_map.rs
@@ -1,0 +1,104 @@
+//! Remaps values using a series of linear mappings.
+//! 
+//! Useful for things like designspace : userspace mapping.
+
+use ordered_float::OrderedFloat;
+
+pub struct PiecewiseLinearMap {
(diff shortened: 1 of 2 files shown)


fontir/src/piecewise_linear_map.rs
```diff
@@ -0,0 +1,112 @@
+//! Remaps values using a series of linear mappings.
+//!
+//! Useful for things like designspace : userspace mapping.
```
anthrotype, line 3: it's the other way around, `userspace : designspace mapping`

fontir/src/piecewise_linear_map.rs
```diff
@@ -0,0 +1,112 @@
...
+    use ordered_float::OrderedFloat;
+
+    use super::PiecewiseLinearMap;
+
+    #[test]
+    fn single_segment_map() {
+        // User likes to map wght 0..1000 onto 0..10
+        let from_to = vec![
+            (OrderedFloat(0_f32), OrderedFloat(0_f32)),
+            (OrderedFloat(10_f32), OrderedFloat(1000_f32)),
```
anthrotype, line 84: IIUC you want to make wght=1000 (user scale) to 10 in internal design coordinates.. in that case you should swap the from and to in this example, because by definition `from` is user (fvar) coordintes and `to` is design coordinates.

fontir/src/piecewise_linear_map.rs
```diff
@@ -0,0 +1,112 @@
+//! Remaps values using a series of linear mappings.
+//!
+//! Useful for things like designspace : userspace mapping.
+
+use ordered_float::OrderedFloat;
+
+pub struct PiecewiseLinearMap {
+    from: Vec<OrderedFloat<f32>>, // sorted, ||'s to
+    to: Vec<OrderedFloat<f32>>,   // sorted, ||'s from
```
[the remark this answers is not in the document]
anthrotype: I suppose || vectors help with the binary_search you do later, i'm fine you leave it that way.
Also fonttools designspaceLib's AxisDescriptor has it as a list of tuples btw.

fontir/src/piecewise_linear_map.rs
rsheeter: The other way around, 10 in internal design coordinates, e.g. what the source might have in something like a https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#location xvalue, to what fvar might have which IIUC we call user coordinates. I think perhaps the comment is written poorly, will update.

fontir/src/piecewise_linear_map.rs
rsheeter: I updated the comment, to my understanding you would map design : user and not, at least not when going from source to IR, likely the other way around?

rsheeter: (lightbulb) now I see, you meant https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#map-element maps the other way around? 

I would think for building IR we want userspace coords not design coords so the comment stands. That probably means I'm populating it backwards in #47.

PR #47, opened 2022-12-11
Explicitly distinguish design and user coordinates and locations

rsheeter:
Introduce explicit user vs design coord and location types and begin to use them in the right places. Conversion is currently entirely bogus, and explicitly noted so. 

Intent is that in the end we will explicitly move between them using the mapping type added in #46. I current imagine building a design:user map and storing in StaticMetadata.

Builds on #43. Step toward #22.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -1,9 +1,12 @@
 //! Serde types for font IR.
 
-use crate::{error::Error, serde::StaticMetadataSerdeRepr};
-use ordered_float::OrderedFloat;
+use crate::{
+    coords::{UserSpaceCoord, UserSpaceLocation},
+    error::Error,
+    serde::StaticMetadataSerdeRepr,
+};
 use serde::{Deserialize, Serialize};
-use std::collections::{BTreeMap, HashMap};
+use std::collections::HashMap;
 
 /// Global font info that cannot vary.
 ///
@@ -42,15 +45,11 @@ impl StaticMetadata {
 pub struct Axis {
     pub name: String,
     pub tag: String,
-    pub min: OrderedFloat<f32>,
-    pub default: OrderedFloat<f32>,
-    pub max: OrderedFloat<f32>,
+    pub min: UserSpaceCoord,
+    pub default: UserSpaceCoord,
+    pub max: UserSpaceCoord,
     pub hidden: bool,
 }
-
-// Using BTreeMap instead of HashMap and OrderedFloat instead of f32 so that
-// the location is hashable and can be used as a key in Glyph::sources HashMap
-pub type DesignSpaceLocation = BTreeMap<String, OrderedFloat<f32>>;
 
 /// A variable definition of a single glyph.
 ///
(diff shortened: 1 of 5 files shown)


anthrotype:
> I current imagine building a design:user map

As I noted in #46, in designspace axes element, the map goes the other way around: i.e. from user to design

fontir/src/ir.rs
```diff
@@ -60,7 +59,7 @@ pub type DesignSpaceLocation = BTreeMap<String, OrderedFloat<f32>>;
 #[derive(Serialize, Deserialize, Debug, PartialEq)]
 pub struct Glyph {
     pub name: String,
-    pub sources: HashMap<DesignSpaceLocation, GlyphInstance>,
```
anthrotype, line 63: actually I think for these we don't want UserSpaceLocation. Sources in designspace are always in design-space, not user-space. Instances can avail themselves of being optionally defined in user-space, but sources (aka the designer's "masters") are always design locations because that's what the designer works with, and nobody requested to define sources with user-space location so far.

So I don't think it helps anything to have Glyph::sources keyed by user-space location.

ufo2fontir/src/source.rs
```diff
@@ -50,7 +53,7 @@ impl Cache {
     fn new(
         global_metadata_sources: StateSet,
         glyph_names: HashSet<String>,
-        locations: HashMap<PathBuf, Vec<DesignSpaceLocation>>,
+        locations: HashMap<PathBuf, Vec<UserSpaceLocation>>,
```
anthrotype, line 56: so yeah that's incorrect, the locations of sources in designspace (in this particular case, `locations` contains the location of each glif file) are in _design_-space coordinates, not user-space

fontir/src/ir.rs
rsheeter: I had in mind to leave designspace behind when creating IR and work exclusively in user and normalized space. Is that a bad idea?

ufo2fontir/src/source.rs
rsheeter: My rationale was that for IR generation the userspace location would be what you want?

fontir/src/ir.rs
```diff
@@ -42,16 +45,12 @@ impl StaticMetadata {
 pub struct Axis {
     pub name: String,
     pub tag: String,
-    pub min: OrderedFloat<f32>,
-    pub default: OrderedFloat<f32>,
-    pub max: OrderedFloat<f32>,
+    pub min: UserSpaceCoord,
+    pub default: UserSpaceCoord,
+    pub max: UserSpaceCoord,
```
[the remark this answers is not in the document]
rsheeter: I'm a little confused about the from key reference in this context? - IR is not obliged to mirror what UFO does. I wonder if we are conflating what designspace/ufo do with what IR does?

fontir/src/ir.rs
rsheeter: Reminder for self, per f2f I believe the user of internal coordinates was acceptable here.

fontir/src/ir.rs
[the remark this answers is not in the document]
rsheeter: In looking at the python I suspect it's far far easier to become confused than it will be in rust because there are neither types nor comments on coordinates inputs or outputs are in . In rust it'll be very clear that, for example, `f(DesignCoord) -> UserCoord`.

#56 aspires to resolve this in a mutually satisfactory way by making exlicit types with to_other_type methods. Let's resolve here and try to push that through to something we can live with.

rsheeter:
I think I've now changed this so much it makes sense to make a new PR against #56

PR #48, opened 2022-12-12
[WIP] Load basic mapping information from .designspace files

rsheeter:
TODO: reconstruct on top of #56, #57

Build on both #46 and #47 which is hard to express in GH. Please prefer to review the actual linear map in #46.

Step toward #22.


fontir/src/ir.rs
```diff
@@ -49,6 +50,7 @@ pub struct Axis {
     pub default: UserSpaceCoord,
     pub max: UserSpaceCoord,
     pub hidden: bool,
+    pub design_to_user: PiecewiseLinearMap,
```
anthrotype, line 53: designspaceLib calls this just `map`, and -- again -- it's a map from user to design.

each AxisDescriptor has a map_forward and map_backward methods that do the piecewise lerp, where "forward" means from user to design, and "backward" the other way around:

https://github.com/fonttools/fonttools/blob/6aca5be9b772bfb502fcaf29c92a85e3c918ac34/Lib/fontTools/designspaceLib/__init__.py#L929-L945

ufo2fontir/src/toir.rs
```diff
@@ -114,28 +118,84 @@ where
...
+                    tag: "wght".to_string(),
+                    min: user_coord(-1.0),
+                    default: user_coord(-0.1),
+                    max: user_coord(1.125),
+                    hidden: false,
+                    design_to_user: PiecewiseLinearMap::new(vec![
+                        (OrderedFloat(100.0), OrderedFloat(-1.0)),
+                        (OrderedFloat(200.0), OrderedFloat(-0.825)),
+                        (OrderedFloat(300.0), OrderedFloat(-0.55)),
+                        (OrderedFloat(400.0), OrderedFloat(-0.1)),
```
anthrotype, line 171: you can see how the test designspace file has user-space coordinates in the "input" (or `from`), and design-space in the "output" (`to`)

resources/testdata/mapping.designspace
```diff
@@ -0,0 +1,53 @@
...
+            <map input="500" output="0.350000" />
+            <map input="600" output="0.540000" />
+            <map input="700" output="0.730000" />
+            <map input="800" output="0.9275" />
+            <map input="900" output="1.125000" />
+        </axis>
+        <axis default="100" maximum="100" minimum="75" name="width" tag="wdth">
+            <labelname xml:lang="en">Width</labelname>
+        </axis>
+        <axis default="0" maximum="1" minimum="0" name="italic" tag="ital">
```
anthrotype, line 20: a continuous ital axis is relatively uncommon, it's usually a discrete 0|1

ufo2fontir/src/toir.rs
```diff
@@ -24,27 +25,30 @@ pub(crate) fn to_ir_location(loc: &[Dimension]) -> UserSpaceLocation {
...
     ir::Axis {
-        name: axis.name,
-        tag: axis.tag,
-        min: design_to_user(axis.minimum.expect("Discrete axes not supported yet")),
-        default: design_to_user(axis.default),
-        max: design_to_user(axis.maximum.expect("Discrete axes not supported yet")),
+        name: axis.name.clone(),
+        tag: axis.tag.clone(),
         hidden: axis.hidden,
+        min: to_user(&design_to_user, axis.minimum.unwrap()),
```
anthrotype, line 48: in DS, axis minimum/default/maximum are already in user-space coordinates so no mapping is required here

fontir/src/ir.rs
rsheeter: I believe I specifically want design to user because for IR I want user or normalized and *not* design. That probably means I populating it backwards which I will fix.

resources/testdata/mapping.designspace
rsheeter: The example was taken directly from Roboto. Agreed re what's typical.

ufo2fontir/src/toir.rs
rsheeter: Fixed, ty

fontir/src/ir.rs
rsheeter: Fixed, provided I'm right that IR wants user and normalized, not design, coords.

ufo2fontir/src/toir.rs
rsheeter: Done

rsheeter:
I believe everything interesting here has been merged in other PRs; closing

PR #55, opened 2022-12-13
<3 units

rsheeter:
Because I confused myself :)

Hopefully helps with #22.

[Rendered](https://github.com/googlefonts/fontmake-rs/blob/units/resources/text/units.md)

### DIFF
--- a/resources/text/units.md
+++ b/resources/text/units.md
@@ -9,7 +9,7 @@ There are three main classes of coordinates in fonts:
    * Used by the type designer and design tool
    * Not directly present in the font binary at all
    * For example, [<dimension xvalue/>](https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#dimension) in a .designspace file
-1. User, aka External, coordainates
+1. User, aka External, coordinates
    * The values in [fvar](https://learn.microsoft.com/en-us/typography/opentype/spec/fvar), for example Weight 400
    * What you would expose to an end user in a document editor or manipulate via CSS
 1. Internal, or normalized, coordinates
--- a/resources/text/units.md
+++ b/resources/text/units.md
@@ -1,0 +1,96 @@
+# Units
+
+Written because the author confused themselves and wanted to be sure they
+now understand.
+
+There are three main classes of coordinates in fonts:
+
+1. Design coordinates
+   * Used by the type designer and design tool
+   * Not directly present in the font binary at all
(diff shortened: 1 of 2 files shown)


resources/text/units.md
```diff
@@ -0,0 +1,96 @@
...
+Written because the author confused themselves and wanted to be sure they
+now understand.
+
+There are three main classes of coordinates in fonts:
+
+1. Design coordinates
+   * Used by the type designer and design tool
+   * Not directly present in the font binary at all
+   * For example, [<dimension xvalue/>](https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#dimension) in a .designspace file
+1. User, aka External, coordainates
```
cmyr, line 12:
```suggestion
1. User, aka External, coordinates
```

resources/text/units.md
```diff
@@ -0,0 +1,96 @@
...
+   * The values in [fvar](https://learn.microsoft.com/en-us/typography/opentype/spec/fvar), for example Weight 400
+   * What you would expose to an end user in a document editor or manipulate via CSS
```
cmyr, line 14: the use of a numerical value to denote weight feels more like some ancient convention, and I don't really think of this in terms of 'coordinates', per se, but I can see how that makes sense. I'm now very curious to think about where this convention came from...

cmyr (approved):
Reading the introduction here I was getting excited about a deep dive on picos, points, and ems (oh my)

resources/text/units.md
rsheeter: @davelab6 do you happen to know where the weight scale comes from?

resources/text/units.md
```diff
@@ -0,0 +1,96 @@
+# Units
```
anthrotype, line 1: I think this discussion is not about font "units" or "coordinates" in general, but more specifically about _variation axis_ coordinates.

resources/text/units.md
```diff
@@ -0,0 +1,96 @@
...
+
+Written because the author confused themselves and wanted to be sure they
+now understand.
+
+There are three main classes of coordinates in fonts:
```
anthrotype, line 6: ``diff
- There are three main classes of coordinates in fonts:
+ There are three main classes of axis coordinates in variable fonts:
```

rsheeter: Will apply that edit directly as the merge button has gone away

rsheeter: Doen in https://github.com/googlefonts/fontmake-rs/commit/67b44d2b88e0daaac9680c5f40df0d062d0676f0

PR #57, opened 2022-12-14
Use updated converters on .designspace files

rsheeter:
Begin to plumb actual unit types, established in #56, into system. Store a converter in StaticMetadata, serializing it in the standard user:design form (https://github.com/googlefonts/fontmake-rs/blob/main/resources/text/units.md#user--design).

Builds on https://github.com/googlefonts/fontmake-rs/pull/43. Step toward https://github.com/googlefonts/fontmake-rs/issues/22.

Clean copy of #47, this time against the units branch which hopefully addresses the core problems.

### DIFF
--- a/fontir/src/coords.rs
+++ b/fontir/src/coords.rs
@@ -112,8 +112,8 @@ pub fn temporary_design_to_user_conversion(coord: DesignCoord) -> UserCoord {
 
 impl DesignCoord {
     /// We do *not* provide From because we want conversion to be explicit
-    pub fn new(value: OrderedFloat<f32>) -> DesignCoord {
-        DesignCoord(value)
+    pub fn new(value: impl Into<OrderedFloat<f32>>) -> DesignCoord {
+        DesignCoord(value.into())
     }
 
     pub fn to_user(&self, converter: &CoordConverter) -> UserCoord {
@@ -127,8 +127,8 @@ impl DesignCoord {
 
 impl UserCoord {
     /// We do *not* provide From because we want conversion to be explicit
-    pub fn new(value: OrderedFloat<f32>) -> UserCoord {
-        UserCoord(value)
+    pub fn new(value: impl Into<OrderedFloat<f32>>) -> UserCoord {
+        UserCoord(value.into())
     }
 
     pub fn to_design(&self, converter: &CoordConverter) -> DesignCoord {
@@ -142,8 +142,8 @@ impl UserCoord {
 
 impl NormalizedCoord {
     /// We do *not* provide From because we want conversion to be explicit
-    pub fn new(value: OrderedFloat<f32>) -> NormalizedCoord {
-        NormalizedCoord(value)
+    pub fn new(value: impl Into<OrderedFloat<f32>>) -> NormalizedCoord {
+        NormalizedCoord(value.into())
     }
 
     pub fn to_design(&self, converter: &CoordConverter) -> DesignCoord {
--- a/fontir/src/coords.rs
+++ b/fontir/src/coords.rs
@@ -4,11 +4,11 @@
 
 use std::collections::BTreeMap;
 
+use crate::piecewise_linear_map::PiecewiseLinearMap;
+use crate::serde::CoordConverterSerdeRepr;
 use log::warn;
 use ordered_float::OrderedFloat;
 use serde::{Deserialize, Serialize};
-
-use crate::piecewise_linear_map::PiecewiseLinearMap;
 
 /// A coordinate in some arbitrary space the designer dreamed up.
 ///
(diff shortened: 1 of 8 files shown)


rsheeter:
I think I like how individual units convert now, the conversion of sets of units could use some polish but it feels to me like it's moving in the right direction.

fontir/src/coords.rs
```diff
@@ -44,21 +45,27 @@ trait Converter<From, To> {
...
+    pub(crate) default_idx: usize,
+    pub(crate) user_to_design: PiecewiseLinearMap,
     design_to_user: PiecewiseLinearMap,
     design_to_internal: PiecewiseLinearMap,
     internal_to_design: PiecewiseLinearMap,
 }
 
 impl CoordConverter {
+    pub fn nop() -> CoordConverter {
+        CoordConverter::from_user_to_design_examples(Vec::new(), 0)
```
anthrotype, line 60: a no-op CoordConverter (to be used when a designspace axis has no explicit axis mappings) should return behave as if user-space and design-space are the same space, so like an identity mapping.
However internal normalized coordinates must always by definition be mapped to -1(min)..0(default)..+1(max).
So you always need the full user-space triplet of axis minimum, default and maximum (the fvar specifies those) to be able to normalize coordinates accordingly.
The way you made your default CoordConverter simply returns maps any number to itself, IIUC. That may be ok for user<=>design, but not for normalized.

ufo2fontir/src/source.rs
```diff
@@ -428,7 +428,24 @@ impl Work for GlyphIrWork {
...
+        // Migrate glif_files into internal coordinates
+        let axes: HashMap<_, _> = static_metadata.axes.iter().map(|a| (&a.name, a)).collect();
+        let mut glif_files = HashMap::new();
+        for (path, design_locations) in self.glif_files.iter() {
+            let internal_locations: Vec<InternalLocation> = design_locations
+                .iter()
+                .map(|design_location| {
+                    design_location
+                        .iter()
+                        .map(|(an, dl)| (an.clone(), dl.to_user(&axes.get(an).unwrap().converter)))
```
anthrotype, line 441: if this is a Vec<InternalLocation>, why does this do `to_user`?

ufo2fontir/src/toir.rs
```diff
@@ -84,9 +67,44 @@ fn to_ir_glyph_instance(glyph: &norad::Glyph) -> ir::GlyphInstance {
...
+            .iter()
+            .map(|map| {
+                (
+                    UserCoord::new(OrderedFloat(map.input)),
+                    DesignCoord::new(OrderedFloat(map.output)),
+                )
+            })
+            .collect();
+        let default_idx = examples.iter().position(|(u, _)| *u == default).expect(
+            "We currently require that you have a mapping for the default if you have mappings",
```
anthrotype, line 88: to build an avar, fonttools actually requires not only a mapping for the default axis coordinate, but also for the axis minimum and maximum as well. Those are encoded directly in the fvar.
Simply taking the miniumum/maximum of the provided (potentionally incomplete) axis mappings is not sufficient, because what makes a user-space coordinate minimum or maximum (and hence -1 and +1 in normalized coordinates) is the fact that it is globally specificed as such in the fvar table. It's better to enforce this here.

ufo2fontir/src/source.rs
rsheeter: Because I borked the typing of internal locations in #56 :( Fixing.

fontir/src/coords.rs
rsheeter: Hm. That might invalidate the entire concepts of a nop converter as there is no information to make that mapping.

rsheeter: OK, you are right, it doesn't actually make sense. Deleted.

ufo2fontir/src/toir.rs
rsheeter: That's an interesting limitation. I don't actually find it obvious that should be required. I'm going to put in the validation and file an issue.

rsheeter: On a moments further thought the sort of "guessing" at the intended design range that would be implied seems ugly, forget that.

rsheeter:
Per IM from dentist (dedication!), OK to merge if the min/max validation is added.

ufo2fontir/src/toir.rs
```diff
@@ -73,9 +73,16 @@ pub fn to_ir_axis(axis: &designspace::Axis) -> ir::Axis {
             .iter()
             .map(|map| (UserCoord::new(map.input), DesignCoord::new(map.output)))
             .collect();
+        // # mappings is generally small, repeated linear probing is fine
         let default_idx = examples.iter().position(|(u, _)| *u == default).expect(
             "We currently require that you have a mapping for the default if you have mappings",
         );
+        examples.iter().position(|(u, _)| *u == min).expect(
```
anthrotype, line 80: I think you need to check not just that min/max are present but that they are in fact the minimum and the maximum.
Unless CoordConverter sorts.. Or is it the PiecewiseLinearMap that guarantee that the array is sorted?

Both the `from` and the `to` coordinate must be increasing, the fromCoord actually has to be _strictly_ increasing. See discussion in avar spec under 'AxisValueMap record'

PR #58, opened 2022-12-14
Restore smoke test

rsheeter:
And fix a bug it immediately revealed.

Reconstruction of #52 as it was easier to do a new branch than rebase that one.


rsheeter:
Note that as there are no changes beyond the approved #52 I'm inclined to just merge.

PR #59, opened 2022-12-14
[WIP] Load Glyphs v2 axis mappings, example based on Texturina, into converter

rsheeter:


resources/testdata/glyphs2/WghtVar_AxisMappings.glyphs
```diff
@@ -0,0 +1,187 @@
...
+descender = -200;
+id = m01;
+weightValue = 40;
+xHeight = 0;
+},
+{
+ascender = 800;
+capHeight = 500;
+descender = -200;
+id = "medium";
```
anthrotype, line 60: I suspect you wrote this master id by hand? If Glyphs.app doesn't complain, I won't either :)
Usually these are strings in UUID format (`import uuid; uuid.uuid4()` in python)

glyphs2fontir/src/source.rs
```diff
@@ -196,13 +197,49 @@ impl Work for StaticMetadataWork {
...
+                        });
+                    mappings
+                        .iter()
+                        .position(|(_, dc)| *dc == max)
+                        .unwrap_or_else(|| {
+                            panic!("Must have a mapping for max {:?} on {}", max, a.tag)
+                        });
+                    CoordConverter::new(mappings, default_idx)
+                } else {
+                    // There is no mapping; design == user
```
anthrotype, line 233: when there is no Axis Mappings custom parameter, you can't simply assume design == user and go home..

You have to check for Axis Location custom parameters on the masters and optionally on the instances too, and use those to define {user-space: design-space} axis mappings. 

Have a read at "Axis mappings and locations" chapter of https://glyphsapp.com/learn/creating-a-variable-font

Actually, it's interesting that in that doc there's a box on the side with a warning like this:

> Axis Mappings are not always working as expected in 3.0.x, and may create faulty entries in the fvar table. We are considering deprecating this method. If possible, use the Axis Location method.

Well, glypysLib at least gives priority to the more general approach which is Axis Mappings, as the latter allows one to define more mappings than there are masters or instances (if one wished to warp the variation space with greater granularity)...

Finally, if neither Axis Mappings nor Axis Locations are defined.. well, you're in dangerous territory and glyphsLib will do something clever like taking the user-space locations implicitly assigner to instances (which do have weightClass and widthClass attribute corresponding to CSS values) and use those to infer the axis mappings.

resources/testdata/glyphs2/WghtVar_AxisMappings.glyphs
rsheeter: Yes, I manually edited it because reading a UUID was making my head hurt and we really don't care it's a UUID as far as I can tell

glyphs2fontir/src/source.rs
rsheeter: Axis Location's is TODO, I will update the comment to be more explicit about that

If neither is defined and you take the weight/width from attributes I would have thought you assume design == user?

PR #60, opened 2022-12-15
Create blank GlyphInstances at correct NormalizedLocation for .glyphs

rsheeter:
Someday we'll even stuff metrics and outlines into 'em! Baby steps.

It would be nice to be able to call to_whatever on the Location types; filed #61 for that.

### DIFF
--- a/fontir/src/error.rs
+++ b/fontir/src/error.rs
@@ -58,7 +58,7 @@ pub enum WorkError {
     InconsistentAxisDefinitions(String),
     #[error("I am the glyph with gid, {0}")]
     NoGlyphIdForName(String),
-    #[error("I am the glyph with gid, {0}")]
+    #[error("No Glyph for name {0}")]
     NoGlyphForName(String),
     #[error("File expected: {0:?}")]
     FileExpected(PathBuf),
--- a/fontir/src/error.rs
+++ b/fontir/src/error.rs
@@ -58,6 +58,8 @@ pub enum WorkError {
     InconsistentAxisDefinitions(String),
     #[error("I am the glyph with gid, {0}")]
     NoGlyphIdForName(String),
+    #[error("I am the glyph with gid, {0}")]
+    NoGlyphForName(String),
     #[error("File expected: {0:?}")]
     FileExpected(PathBuf),
     #[error("Unable to parse {0:?}: {1}")]
--- a/fontir/src/error.rs
+++ b/fontir/src/error.rs
@@ -66,4 +66,8 @@ pub enum WorkError {
     ParseError(PathBuf, String),
     #[error("No default master in {0:?}")]
     NoDefaultMaster(PathBuf),
+    #[error("No layer {0} exists. Referenced by glyph {1}.")]
+    NoLayerForGlyph(String, String),
+    #[error("Failed to add glyph source: {0}")]
+    AddGlyphSource(String),
 }
--- a/fontir/src/error.rs
+++ b/fontir/src/error.rs
@@ -76,8 +76,8 @@ pub enum WorkError {
     ParseError(PathBuf, String),
     #[error("No default master in {0:?}")]
     NoDefaultMaster(PathBuf),
-    #[error("No layer {0} exists. Referenced by glyph {1}.")]
-    NoLayerForGlyph(String, String),
+    #[error("No master {master} exists. Referenced by glyph {glyph}.")]
+    NoMasterForGlyph { master: String, glyph: String },
     #[error("Failed to add glyph source: {0}")]
     AddGlyphSource(String),
 }
--- a/fontir/src/error.rs
+++ b/fontir/src/error.rs
@@ -2,7 +2,7 @@ use std::{error, io, path::PathBuf};
 
 use thiserror::Error;
 
-use crate::coords::{NormalizedLocation, UserLocation};
+use crate::coords::{DesignCoord, NormalizedLocation, UserLocation};
 
 // TODO: eliminate dyn Error and collapse Error/WorkError
 
@@ -16,6 +16,10 @@ pub enum Error {
     IoError(#[from] io::Error),
     #[error("Unable to parse {0:?}: {1}")]
     ParseError(PathBuf, String),
+    #[error("No axes are defined for master {0}")]
+    NoAxisDefinitions(String),
+    #[error("Axis definitions are inconsistent")]
+    InconsistentAxisDefinitions(String),
     #[error("Illegible source")]
     UnableToLoadSource(Box<dyn error::Error>),
     #[error("Missing layer")]
@@ -39,6 +43,12 @@ pub enum Error {
     InvalidGlobalMetadata,
     #[error("No default master in {0:?}")]
     NoDefaultMaster(PathBuf),
+    #[error("Missing mapping on {axis} for {field} at {value:?}")]
+    MissingMappingForDesignCoord {
+        axis: String,
+        field: String,
+        value: DesignCoord,
+    },
 }
 
 /// An async work error, hence one that must be Send
--- a/fontir/src/error.rs
+++ b/fontir/src/error.rs
@@ -16,7 +16,7 @@ pub enum Error {
     IoError(#[from] io::Error),
     #[error("Unable to parse {0:?}: {1}")]
     ParseError(PathBuf, String),
-    #[error("No axes are defined for master {0}")]
+    #[error("Missing required axis values for {0}")]
     NoAxisDefinitions(String),
     #[error("Axis definitions are inconsistent")]
     InconsistentAxisDefinitions(String),
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -1,11 +1,10 @@
-use fontir::coords::{CoordConverter, DesignCoord, UserCoord};
+use fontir::coords::{CoordConverter, DesignCoord, UserCoord, DesignLocation};
 use fontir::error::{Error, WorkError};
-use fontir::ir;
-use fontir::ir::{Axis, StaticMetadata};
+use fontir::ir::{self, StaticMetadata};
 use fontir::orchestration::Context;
 use fontir::source::{Input, Source, Work};
 use fontir::stateset::StateSet;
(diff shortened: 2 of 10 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -258,20 +276,97 @@ impl Work for StaticMetadataWork {
...
+                dc.to_normalized(&axes.get(tag).unwrap().converter),
+            )
+        })
+        .collect()
+}
+
 struct GlyphIrWork {
     glyph_name: String,
+    font: Arc<Font>,
+    master_indices: Arc<HashMap<String, usize>>,
```
anthrotype, line 302: they are not 'indices' as in 0, 1, 2.. They are (UUID) identifiers, so I would call this `master_ids` for clarity's sake

glyphs2fontir/src/source.rs
```diff
@@ -258,20 +276,97 @@ impl Work for StaticMetadataWork {
...
+        }
+
+        // It's helpful if glyphs are defined at min, default, and max (some of which may be cooincident)
+        for axis in axes.iter() {
+            let min = axis.min.to_normalized(&axis.converter);
+            let max = axis.max.to_normalized(&axis.converter);
+            let default = axis.max.to_normalized(&axis.converter);
+            let positions = axis_positions.get(&axis.tag).unwrap();
+            if !positions.contains(&min) {
+                panic!("{} is undefined at {} min", self.glyph_name, axis.tag);
```
anthrotype, line 357: I think panic! should be used sparingly.. basically a bit like AssertionError in Python, it must never happen in normal nor in exceptional but still predictable circumstances where you would signal with an Exception in Python or Result::Err in Rust. If this can never happen and it's about inner code logic, then it's fine to keep; otherwise, if it can happen sometimes that such condition isn't met then an appropriate WorkError should be returned, I suppose

glyphs2fontir/src/source.rs
```diff
@@ -258,20 +276,97 @@ impl Work for StaticMetadataWork {
...
+        let mut sources = HashMap::new();
+
+        // Glyphs have layers that match up with masters, and masters have locations
+        let mut axis_positions: HashMap<String, HashSet<NormalizedCoord>> = HashMap::new();
+        for layer in glyph.layers.iter() {
+            let Some(master_idx) = self.master_indices.get(&layer.layer_id) else {
+                panic!("glyph {} references layer id {} but no such master exists", self.glyph_name, layer.layer_id);
+            };
+            let master = &font.font_master[*master_idx];
+            let location = to_normalized(axes, &design_location(axes, master));
```
anthrotype, line 330: you are computing this location over and over for each glyph master layer, but it's always going to be the same for a given master. Maybe instead of a `master_indices` map {String: usize} you can store in the GlyphIrWork a `master_locations` hashmap keyed by master.id (string) and the values contain the precomputed normalized locations for each master.

rsheeter: I miss Python's cache annotations

glyphs2fontir/src/source.rs
rsheeter: Fair point. I was actually just thinking we have a bit more unwrap/panic than I'd like.

rsheeter: Made it an error

glyphs2fontir/src/source.rs
```diff
@@ -258,20 +276,97 @@ impl Work for StaticMetadataWork {
...
+
+            // TODO actually populate fields
+            let glyph_instance = GlyphInstance {
+                width: 0_f64,
+                height: None,
+                contours: Vec::new(),
+                components: Vec::new(),
+            };
+
+            sources.insert(location, glyph_instance);
```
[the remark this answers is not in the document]
rsheeter: Didn't notice that existed! Will adopt.

glyphs2fontir/src/source.rs
rsheeter: We use axes from static metadata to compute this so I moved it outside the loop. It could be computed from the Font and shared globally but it makes the computation uglier.

glyphs2fontir/src/source.rs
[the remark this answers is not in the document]
rsheeter: Added a toir.rs and helper functions to better parallel ufo2fontir.

glyphs2fontir/src/source.rs
```diff
@@ -320,16 +325,16 @@ impl Work for GlyphIrWork {
             .get(&self.glyph_name)
             .ok_or_else(|| WorkError::NoGlyphForName(self.glyph_name.clone()))?;
 
-        let mut sources = HashMap::new();
+        let mut ir_glyph = ir::Glyph::new(self.glyph_name.clone());
```
anthrotype, line 194: in ufo2fontir::source::GlyphIrWork::exec() method, we dont create an ir::Glyph inline, but we call ufo2fontir::toir::to_ir_glyph function -- should we do the same for glyphs2fontir?

https://github.com/googlefonts/fontmake-rs/blob/66aa397c76b7d76ba9e9ed8d123dcff2eaacab80/ufo2fontir/src/source.rs#L452-L453

also note how ufo2fontir we map the error returned from to_ir_glyph as a `WorkError::GlyphIrWorkError` variant.

I noticed that glyphs2fontir has no errors.rs of its own, either it's infallible (which I doubt) or it is using the fontir::ir:errors namespace as its own.
I think we should only add to the latter generic errors that would apply to an font IRs in general. E.g. "NoLayerForGlyph" makes sense as a glyphs2fontir::error but not much as, say, ufo2fontir..

glyphs2fontir/src/source.rs
```diff
@@ -330,11 +196,15 @@ impl Work for GlyphIrWork {
         // Glyphs have layers that match up with masters, and masters have locations
         let mut axis_positions: HashMap<String, HashSet<NormalizedCoord>> = HashMap::new();
         for layer in glyph.layers.iter() {
-            let Some(master_idx) = self.master_indices.get(&layer.layer_id) else {
+            let Some(master_idx) = font_info.master_indices.get(layer.layer_id.as_str()) else {
+                // If the glyph has an associatedMasterId this is fine, if not ... not
```
anthrotype, line 200: i don't understand the wording of this inline comment.. "this is fine, if not ... not"
Actually it should be the layer that may/may not have associatedMasterId, not the glyph per se.
Also you may want to mark this with a TODO and link to the associated issue #62

fontir/src/error.rs
```diff
@@ -16,6 +16,10 @@ pub enum Error {
     IoError(#[from] io::Error),
     #[error("Unable to parse {0:?}: {1}")]
     ParseError(PathBuf, String),
+    #[error("No axes are defined for master {0}")]
```
anthrotype, line 19: "No _axis values_ are defined for master {0}", instead of simply "axes".

Also this is looks like a very Glyphs specific error, not even sure if possible to trigger at all with a valid source file.

By the way, in designspace, sources' locations may omit some or even all dimensions and these will be filled in with the default location (in design-space coords) across all globally-defined axes.

https://fonttools.readthedocs.io/en/latest/designspaceLib/python.html#fontTools.designspaceLib.SourceDescriptor.getFullDesignLocation

anthrotype: however because in Glyphs.app there are no global axes definitions with user-space min/default/max (only a list of axes names and tags), we can't infer a default master location in this case. In fact we go the opposite way and determine the global axes min/default/max from the masters locations.
I think in Glyphs we should just assume (assert) masters will have as many axis values as there are axes, like you're already doing with InconsistentAxisDefinitions error.

glyphs2fontir/src/source.rs
anthrotype: I see that in the following commit you're correctly checking for associated_master_id in the layer, not the glyph, but the comment looks stale

glyphs2fontir/src/toir.rs
```diff
@@ -0,0 +1,181 @@
...
+        .map(|(tag, dc)| {
+            (
+                tag.clone(),
+                dc.to_normalized(&axes.get(tag).unwrap().converter),
+            )
+        })
+        .collect()
+}
+
+fn find_by_user_coord(
```
anthrotype, line 31: should this be named `find_by_design_coord`? The `value` which is the key in this lookup is a DesignCoord, so it seems to me you are finding (the index of) the given design coordinate in the mappings array.

glyphs2fontir/src/source.rs
anthrotype: did you? I still see panic! here

rsheeter: Err. I swear I did. Did I not upload it?!

glyphs2fontir/src/source.rs
rsheeter: Wherever errors are reasonable for more than a specific source format or we can make them "generic" +1 to moving them to fontir::Error. I can't think of an immediate example when that wouldn't be true but in abstract I wouldn't want to see fontir::Error::UfoSpecificSituation.

glyphs2fontir/src/toir.rs
rsheeter: Err yes. I think I got halfway to find_user_coord then changed my mind.

glyphs2fontir/src/source.rs
rsheeter: I don't think I want to completely redo errors here, will file an issue.

glyphs2fontir/src/source.rs
rsheeter: fixed comment

fontir/src/error.rs
rsheeter: Fixed message

glyphs2fontir/src/source.rs
rsheeter: #64

anthrotype (approved):
LGTM

glyphs2fontir/src/source.rs
rsheeter: Fixed in #63

PR #63, opened 2022-12-16
Use an enum to distinguish Path from Component, populate for Glyphs 2 and 3

rsheeter:
Manual component generated based on examination of Texturina (Glyphs 2, https://github.com/Omnibus-Type/Texturina, sources/Texturina.glyphs) and Lexend (Glyphs 3, https://github.com/googlefonts/lexend, sources/ofl/lexendgiga/sources/Lexend.glyphs) plus ofc https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md#differences-between-version-2.

Step toward #27. Builds on #60, which is hopefully in good shape now.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -46,11 +46,11 @@ pub struct Font {
 #[derive(Debug, PartialEq, Eq, Hash)]
 pub struct Glyph {
     pub glyphname: String,
-    pub instances: Vec<GlyphInstance>,
+    pub layers: Vec<Layer>,
 }
 
 #[derive(Debug, PartialEq, Eq, Hash)]
-pub struct GlyphInstance {
+pub struct Layer {
     pub layer_id: String,
     pub width: OrderedFloat<f64>,
     pub shapes: Vec<Shape>,
@@ -83,14 +83,14 @@ pub struct Axis {
 
 #[derive(Clone, Debug, FromPlist, PartialEq, Eq, Hash)]
 pub struct RawGlyph {
-    pub layers: Vec<Layer>,
+    pub layers: Vec<RawLayer>,
     pub glyphname: String,
     #[rest]
     pub other_stuff: BTreeMap<String, Plist>,
 }
 
 #[derive(Clone, Debug, FromPlist, PartialEq, Eq, Hash)]
-pub struct Layer {
+pub struct RawLayer {
     pub layer_id: String,
     pub associated_master_id: Option<String>,
     pub width: OrderedFloat<f64>,
@@ -113,9 +113,8 @@ pub struct RawShape {
     pub closed: Option<bool>,
     pub nodes: Option<Vec<Node>>,
 
-    // When I'm a component I have a ref
-    // but it seems to end up other_stuff because r#ref doesn't work
-    pub pos: Option<Vec<OrderedFloat<f64>>>,
+    // When I'm a component I specifically want all my attributes to end up in other_stuff
+    // My Component'ness can be detected by presence of a ref (Glyphs3) or name(Glyphs2) attribute
 
     // Always
     #[rest]
@@ -221,7 +220,7 @@ pub struct FontMaster {
 }
 
 impl RawGlyph {
-    pub fn get_layer(&self, layer_id: &str) -> Option<&Layer> {
+    pub fn get_layer(&self, layer_id: &str) -> Option<&RawLayer> {
         self.layers.iter().find(|l| l.layer_id == layer_id)
     }
 }
@@ -275,26 +274,42 @@ impl std::str::FromStr for NodeType {
     }
 }
 
-impl FromPlist for Component {
-    fn from_plist(plist: Plist) -> Self {
-        let Plist::Dictionary(mut dict) = plist else {
-            panic!("Component must be a dict: {:?}", plist);
-        };
+fn try_f64(plist: &Plist) -> Result<f64, Error> {
+    plist
+        .as_f64()
+        .ok_or_else(|| Error::StructuralError(format!("Bad f64:{:?}", plist)))
+}
+
+impl TryFrom<BTreeMap<String, Plist>> for Component {
+    type Error = Error;
+
+    fn try_from(mut dict: BTreeMap<String, Plist>) -> Result<Self, Self::Error> {
         // Glyphs v3 name has been renamed to ref; look for both
         let glyph_name = if let Some(Plist::String(glyph_name)) = dict.remove("ref") {
             glyph_name
         } else if let Some(Plist::String(glyph_name)) = dict.remove("name") {
             glyph_name
         } else {
-            panic!("Neither ref nor name present: {:?}", dict);
+            return Err(Error::StructuralError(format!(
+                "Neither ref nor name present: {:?}",
+                dict
+            )));
         };
 
-        let transform = if let Some(plist) = dict.remove("transform") {
+        // V3 vs v2: The transform entry has been replaced by angle, pos and scale entries.
+        let mut transform = if let Some(plist) = dict.remove("transform") {
             Affine::from_plist(plist)
         } else {
             Affine::identity()
         };
 
+        if let Some(Plist::Array(pos)) = dict.remove("pos") {
+            assert!(transform == Affine::identity());
+            if pos.len() != 2 {
+                return Err(Error::StructuralError(format!("Bad pos: {:?}", pos)));
+            }
+            transform = Affine::translate(try_f64(&pos[0])?, try_f64(&pos[1])?);
+        }
         // TODO scale, angle, etc
         // See component in <https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md#differences-between-version-2>
         if let Some(..) = dict.remove("angle") {
@@ -304,11 +319,20 @@ impl FromPlist for Component {
             panic!("Scale not supported yet");
         }
 
-        Component {
+        Ok(Component {
             glyph_name,
             transform,
             other_stuff: dict,
-        }
+        })
+    }
+}
+
+impl FromPlist for Component {
+    fn from_plist(plist: Plist) -> Self {
+        let Plist::Dictionary(dict) = plist else {
+            panic!("Component must be a dict: {:?}", plist);
+        };
+        dict.try_into().expect("Unable to parse Component")
     }
 }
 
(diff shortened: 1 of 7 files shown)


glyphs-reader/src/font.rs
```diff
@@ -67,15 +67,33 @@ pub struct Glyph {
...
+///
+/// <https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md#differences-between-version-2>
+#[derive(Clone, Debug, FromPlist, PartialEq, Eq, Hash)]
+pub struct Shape {
+    // When I'm a path
+    pub closed: Option<bool>,
+    pub nodes: Option<Vec<Node>>,
+
+    // When I'm a component I have a ref
+    // but it seems to end up other_stuff because r#ref doesn't work
```
anthrotype, line 80: FWIW, in Glyphs.app's python scripting API (as well as glyphsLib.classes), a GSComponent has a `name` attribute that contains this `ref`

https://docu.glyphsapp.com/#GSComponent.name

glyphs-reader/src/font.rs
```diff
@@ -67,15 +67,33 @@ pub struct Glyph {
     pub other_stuff: BTreeMap<String, Plist>,
 }
 
+/// Represents a path OR a component
```
anthrotype, line 70: why not an enum then? As as struct, it actually represent both a path AND a component which is weird

glyphs-reader/src/font.rs
anthrotype: oh they renamed that component attribute in the plist from `name` (v2) to `ref` (v3), I wonder why..

glyphs-reader/src/font.rs
```diff
@@ -67,15 +67,33 @@ pub struct Glyph {
...
+/// <https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md#differences-between-version-2>
+#[derive(Clone, Debug, FromPlist, PartialEq, Eq, Hash)]
+pub struct Shape {
+    // When I'm a path
+    pub closed: Option<bool>,
+    pub nodes: Option<Vec<Node>>,
+
+    // When I'm a component I have a ref
+    // but it seems to end up other_stuff because r#ref doesn't work
+    pub pos: Option<Vec<OrderedFloat<f64>>>,
```
anthrotype, line 81: there's also scale and angle..

glyphs-reader/src/font.rs
```diff
@@ -550,20 +590,45 @@ impl RawFont {
...
+                        let transform = if let Some(pos) = &shape.pos {
+                            if pos.len() != 2 {
+                                return Err(Error::StructuralError(format!(
+                                    "Strang pos {:?}",
+                                    pos
+                                )));
+                            }
+                            Some(Affine::translate(pos[0].into_inner(), pos[1].into_inner()))
+                        } else {
+                            None
```
anthrotype, line 616: I would make Component::transform just Affine instead of Option<Affine>, and set to identity by default

glyphs-reader/src/font.rs
```diff
@@ -550,20 +590,45 @@ impl RawFont {
...
-                            layer.shapes = Some(paths);
-                        }
+                let Some(mut shapes) = layer.shapes.take() else {
+                    continue;
+                };
+
+                for shape in shapes.iter_mut() {
+                    if let Some(Plist::String(ref_glyph_name)) = shape.other_stuff.remove("ref") {
+                        // only components use ref
+                        let transform = if let Some(pos) = &shape.pos {
```
anthrotype, line 607: maybe add a TODO to later on also handle scale and rotation

glyphs-reader/src/font.rs
```diff
@@ -745,6 +809,8 @@ impl Font {
         let mut raw_font = RawFont::from_plist(raw_content);
         if raw_font.is_v2() {
             raw_font.v2_to_v3()?;
+        } else {
+            raw_font.v3_shapes_to_v2_paths_and_components()?;
```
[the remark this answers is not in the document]
anthrotype: (edited comment above because I need backticks around `<>`)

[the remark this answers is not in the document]
anthrotype: I guess we could have both standalone Path and Component types, and a Shape enum that wrap them as variants e.g.

```rust

struct Path {}

struct Component {}

enum Shape {
    Path(Path),
    Component(Component),
}
```

glyphs-reader/src/font.rs
anthrotype: I guess if we made Shape an enum, we would have to implement a custom FromPlist trait to be able to parse Layer::shapes into this Vec<Shape>, whereas with a simple struct you can simply derive FromPlist?

rsheeter: Yes. Perhaps that woudl be worth doing.

glyphs-reader/src/font.rs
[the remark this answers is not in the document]
rsheeter: We could but it feels particularly odd to migrate v2 to v3 - which merges them - only to promptly split them up again in IR. If IR used a merged vec I'd agree.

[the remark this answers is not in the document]
anthrotype: we are already upgrading v2 to v3 the rest of the font attributes, so this feels like a contradiction.
And the merge followed by split would only be for v2 sources, v3 would be already merged.
And the split is really a loop with a condition, should be pretty fast.

glyphs-reader/src/font.rs
rsheeter: There's *tons* of attributes we don't handle yet, will add a comment.

glyphs-reader/src/font.rs
rsheeter: I believe it has to be Option due to plist parsing

glyphs-reader/src/font.rs
rsheeter: added a general todo for more attributes as there are quite a few. it wasn't immediately obvious which ones we'd need. Certainly scale and rotate :)

glyphs-reader/src/font.rs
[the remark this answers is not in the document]
rsheeter: Per IM, we will keep the v2 => v3 model. Doesn't help with IR but will help our library be a general purpose .glyphs library which is a good outcome.

glyphs-reader/src/font.rs
rsheeter: Per IM discussion will do so

rsheeter: Introduced an enum, it's much nicer now. Good call. Also added use of TryFrom which I think makes the code flow a little better.

glyphs-reader/src/font.rs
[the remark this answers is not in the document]
rsheeter: Upgraded to enums as suggested, much nicer, ty.

glyphs-reader/src/font.rs
rsheeter: Done

glyphs-reader/src/font.rs
rsheeter: Now that it ends up in Component we have a glyph_name.

glyphs-reader/src/font.rs
```diff
@@ -40,13 +43,32 @@ pub struct Font {
     pub axis_mappings: BTreeMap<String, RawUserToDesignMapping>,
 }
 
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct Glyph {
+    pub glyphname: String,
+    pub instances: Vec<GlyphInstance>,
```
anthrotype, line 49: why did you name it `instances: Vec<GlyphInstance>` and not, say, `layers: Vec<Layer>`? I would expect this crate would stick to names that are close to those Glyphs.app itself uses

anthrotype: I see you already have a type called `Layer`, maybe you can rename the latter `RawLayer` and rename `GlyphInstance` -> `Layer`

anthrotype: then instead of `layer_to_glyph_instance` you could `impl TryFrom<RawLayer> for Layer`

glyphs-reader/src/lib.rs
```diff
@@ -6,7 +6,7 @@ mod from_plist;
 mod plist;
 mod to_plist;
 
-pub use font::{Axis, Component, Font, FontMaster, Glyph, Layer, Node, NodeType, Path};
+pub use font::{Axis, Component, Font, FontMaster, Layer, Node, NodeType, Path, RawGlyph};
```
anthrotype, line 9: why pub use font::RawGlyph instead of font::Glyph?

glyphs-reader/src/font.rs
```diff
@@ -211,6 +275,43 @@ impl std::str::FromStr for NodeType {
     }
 }
 
+impl FromPlist for Component {
+    fn from_plist(plist: Plist) -> Self {
+        let Plist::Dictionary(mut dict) = plist else {
+            panic!("Component must be a dict: {:?}", plist);
+        };
+        // Glyphs v3 name has been renamed to ref; look for both
```
anthrotype, line 283: I think Component::from_plist is only being used when deserializing from a v2 plist, the v3 conversion (from RawShape to Shape::Component(Component)) is done inside RawShape::try_from.

anthrotype: maybe you should actually use this Component::from_plist method inside RawShape::try_from, when the latter contains a "ref"... to avoid duplicating the logic in here and over there

glyphs-reader/src/font.rs
```diff
@@ -665,8 +744,116 @@ fn default_master_idx(raw_font: &RawFont) -> usize {
...
+    };
+    src.into_iter().map(map).for_each(|v| dest.push(v));
+}
+
+fn layer_to_glyph_instance(layer: Layer) -> Result<GlyphInstance, Error> {
+    let mut shapes = Vec::new();
+
+    // Glyphs v2 uses paths and components
+    map_and_push_if_present(&mut shapes, layer.paths, Shape::Path);
+    map_and_push_if_present(&mut shapes, layer.components, Shape::Component);
```
anthrotype, line 815: nice use of generics and function pointer 👍

anthrotype (approved):
LGTM, with a few non blocking comments

glyphs-reader/src/lib.rs
rsheeter: Oops, I think that's because I did an automated rename.

glyphs-reader/src/font.rs
rsheeter: Confidence in Rust slowly grows

glyphs-reader/src/font.rs
rsheeter: Oh I like that, updating.

glyphs-reader/src/lib.rs
rsheeter: Yup, ended up with RawLayer there too. Fixed.

glyphs-reader/src/font.rs
rsheeter: Hm. I don't have the plist that made the shape that is a component anymore, but I agree it's duplicative.

rsheeter: OK, if I provide `TryFrom<BTreeMap<String, Plist>> for Component` it can work for both. Sweet.

PR #67, opened 2023-01-05
Use IndexSet<String> for glyph order

### LINKED ISSUE #54: Use IndexMap instead of Vec+HashMap?
StaticMetadata stores an accelerator map plus an ordered vec; coudl just use IndexMap.

See https://github.com/googlefonts/fontmake-rs/pull/43#pullrequestreview-1213452299.

rsheeter:
Fixes #54 

OK to merge.


cmyr (approved):
lgtm!

PR #68, opened 2023-01-06
Make Location a type so it can have impl's

### LINKED ISSUE #61: Make {Design,User,Normalized}Location real types
It would be very nice to be able to say things like DesignLocation.to_normalized(CoordConverter) the same way we can with coords. This appears to be illegal to define when they are just aliases for other types so make them "real" types (e.g. structs)

rsheeter:
Fixes #61

### DIFF
--- a/fontir/src/coords.rs
+++ b/fontir/src/coords.rs
@@ -31,12 +31,11 @@ pub struct NormalizedCoord(OrderedFloat<f32>);
 
 /// A set of per-axis coordinates that define a specific location in a coordinate system.
 ///
-/// E.g. a user location is a `Location<UserCoord>`.
+/// E.g. a user location is a `Location<UserCoord>`. Hashable so it can do things like be
+/// the key for a map of sources by location.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
-pub struct Location<T>(pub BTreeMap<String, T>);
+pub struct Location<T>(BTreeMap<String, T>);
 
-// Using BTreeMap instead of HashMap and OrderedFloat instead of f32 so that
-// the location is hashable and can be used as a key in Glyph::sources HashMap
 pub type DesignLocation = Location<DesignCoord>;
 pub type UserLocation = Location<UserCoord>;
 pub type NormalizedLocation = Location<NormalizedCoord>;
@@ -175,6 +174,10 @@ impl<T> Location<T> {
     pub fn on_axis(name: &str, pos: T) -> Location<T> {
         Location(BTreeMap::from([(name.to_string(), pos)]))
     }
+
+    pub fn iter(&self) -> impl Iterator<Item = (&String, &T)> {
+        self.0.iter()
+    }
 }
 
 impl DesignLocation {
@@ -182,10 +185,10 @@ impl DesignLocation {
         Location::<NormalizedCoord>(
             self.0
                 .iter()
-                .map(|(tag, dc)| {
+                .map(|(name, dc)| {
                     (
-                        tag.clone(),
-                        dc.to_normalized(&axes.get(tag).unwrap().converter),
+                        name.clone(),
+                        dc.to_normalized(&axes.get(name).unwrap().converter),
                     )
                 })
                 .collect(),
--- a/fontir/src/coords.rs
+++ b/fontir/src/coords.rs
@@ -166,11 +166,13 @@ impl NormalizedCoord {
     }
 }
 
-impl<T> Location<T> {
-    pub fn new(positions: BTreeMap<String, T>) -> Location<T> {
-        Location(positions)
+impl<T> FromIterator<(String, T)> for Location<T> {
+    fn from_iter<I: IntoIterator<Item = (String, T)>>(iter: I) -> Self {
+        Location(BTreeMap::from_iter(iter))
     }
+}
 
+impl<T> Location<T> {
     pub fn on_axis(name: &str, pos: T) -> Location<T> {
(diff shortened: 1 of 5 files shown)


fontir/src/coords.rs
```diff
@@ -29,11 +29,17 @@ pub struct UserCoord(OrderedFloat<f32>);
 #[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct NormalizedCoord(OrderedFloat<f32>);
 
+/// A set of per-axis coordinates that define a specific location in a coordinate system.
+///
+/// E.g. a user location is a `Location<UserCoord>`.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Location<T>(pub BTreeMap<String, T>);
```
cmyr, line 36: my personal feeling with these sorts of things is that only very rarely should the inner type be `pub`; it makes it easy to be undisciplined in the future. I would generally prefer to just add little wrapper methods for things like `insert` and `contains`, as you need them.

cmyr: Also if the key here is always a valid opentype tag, I would encourage you to use the `Tag` type from font-types!

fontir/src/coords.rs
```diff
@@ -161,6 +167,32 @@ impl NormalizedCoord {
     }
 }
 
+impl<T> Location<T> {
+    pub fn new(positions: BTreeMap<String, T>) -> Location<T> {
```
cmyr, line 171: instead of requiring this to be a specific type, you could make this constructor take like `impl IntoIterator<Item=(String, T)>` or similar, and we can keep the BTreeMap as an implementation detail.

glyphs2fontir/src/source.rs
```diff
@@ -221,7 +221,7 @@ impl Work for GlyphIrWork {
             let master = &font.font_master[*master_idx];
             let location = &font_info.master_locations[master.id.as_str()];
 
-            for (tag, coord) in location.iter() {
+            for (tag, coord) in location.0.iter() {
```
cmyr, line 224: yea I'd prefer that location just get its own little `iter` method that returns `impl Iterator`.

fontir/src/coords.rs
cmyr: (and you could even get fancy and implement `FromIterator` for location, which would mean you would work with `collect()`, directly)

fontir/src/coords.rs
```diff
@@ -161,6 +167,32 @@ impl NormalizedCoord {
...
+        Location(positions)
+    }
+
+    pub fn on_axis(name: &str, pos: T) -> Location<T> {
+        Location(BTreeMap::from([(name.to_string(), pos)]))
+    }
+}
+
+impl DesignLocation {
+    pub fn to_normalized(&self, axes: &HashMap<&String, &Axis>) -> NormalizedLocation {
```
cmyr, line 181: I might just call this `normalize`? idk, not a blocker or anything, and I'm not totally sure why I have this feeling. Maybe because it takes an argument?

glyphs2fontir/src/source.rs
```diff
@@ -480,9 +480,9 @@ mod tests {
 
         assert_eq!(
             HashSet::from([
-                &BTreeMap::from([("wght".to_string(), NormalizedCoord::new(-1.0))]),
-                &BTreeMap::from([("wght".to_string(), NormalizedCoord::new(0.0))]),
-                &BTreeMap::from([("wght".to_string(), NormalizedCoord::new(1.0))]),
+                &NormalizedLocation::on_axis("wght", NormalizedCoord::new(-1.0)),
```
cmyr, line 483: If the string was a tag, this type could derive `Copy`, and you wouldn't need to be borrowing here.

cmyr (approved):
Nice, this is the sort of helper type I find myself writing over and over again. Some notes inline but nothing blocking, merge when you're happy. :)

fontir/src/coords.rs
rsheeter: It's actually a *name* ... I suppose the non-obviousness of that suggests I should make name a type

fontir/src/coords.rs
rsheeter: I think perhaps to_x makes more sense when you consider each of {design, user, normalized} can reasonable be flipped to any other?

fontir/src/coords.rs
rsheeter: (sigh) that and the fact I called it tag sometimes. Maybe I jet lagged myself into confusion...

glyphs2fontir/src/source.rs
rsheeter: Oh good call. I forgot impl Iterator and didn't think returning the specific impl BTreeMap provided resulted in much hiding of implementation. Fixed.

glyphs2fontir/src/source.rs
rsheeter: Oh dear, this test is horribly misleading, it should say things like Weight not things like wght

rsheeter: https://github.com/googlefonts/fontmake-rs/issues/70 for types to make this a little harder to bungle

rsheeter: In this PR I have updated to make use of name consistent

fontir/src/coords.rs
rsheeter: Oh good call, will update. Definitely not comfortable with some of those nuances yet.

PR #69, opened 2023-01-06
First pass at executing job graph

rsheeter:
Trying to get from the PoC in #26 to a more realistic implementation. Closes #26.

As ever naming is the hardest part. At time of writing "work" is an executable unit and "job" is that plus metadata such as what must happen before this job runs.

### DIFF
--- a/fontc/src/main.rs
+++ b/fontc/src/main.rs
@@ -14,7 +14,7 @@ use fontir::{
     stateset::StateSet,
 };
 use glyphs2fontir::source::GlyphsIrSource;
-use log::{error, info, log_enabled, trace, warn};
+use log::{debug, error, info, log_enabled, trace, warn};
 use serde::{Deserialize, Serialize};
 use std::io::Write;
 use ufo2fontir::source::DesignSpaceIrSource;
@@ -331,7 +331,7 @@ fn main() -> Result<(), Error> {
         while success.len() < job_count && error.is_empty() {
             // Spawn anything that is currently executable (has no unfulfilled dependencies)
             for id in launchable(&pending_jobs, &success) {
-                info!("Start {:?}", id);
+                trace!("Start {:?}", id);
                 let job = pending_jobs.remove(&id).unwrap();
                 let work = job.work;
                 let work_context = root_context.copy_for_work(id.clone(), Some(job.happens_after));
@@ -354,10 +354,10 @@ fn main() -> Result<(), Error> {
                 } {
                     panic!("Repeat signals for completion of {:#?}", completed_id);
                 }
-                info!(
+                debug!(
                     "{}/{} complete, most recently {:?}",
                     error.len() + success.len(),
-                    pending_jobs.len(),
+                    job_count,
                     completed_id
                 );
 
--- a/fontc/src/main.rs
+++ b/fontc/src/main.rs
@@ -1,7 +1,8 @@
 use std::{
     collections::{HashMap, HashSet},
     ffi::OsStr,
-    fs, io,
+    fs,
+    io::{self, Write},
     path::{Path, PathBuf},
 };
 
@@ -15,8 +16,8 @@ use fontir::{
 };
 use glyphs2fontir::source::GlyphsIrSource;
 use log::{debug, error, info, log_enabled, trace, warn};
+use rayon::Scope;
 use serde::{Deserialize, Serialize};
-use std::io::Write;
 use ufo2fontir::source::DesignSpaceIrSource;
 
 /// What font can we build for you today?
@@ -168,10 +169,10 @@ fn glyphs_deleted<'a>(current_inputs: &Input, prev_inputs: &'a Input) -> HashSet
 
 fn add_static_metadata_ir_job(
     change_detector: &mut ChangeDetector,
-    jobs: &mut HashMap<WorkIdentifier, Job>,
+    workload: &mut Workload,
 ) -> Result<(), Error> {
     // TODO don't recreate static metadata if we don't have to
-    jobs.insert(
+    workload.insert(
         WorkIdentifier::StaticMetadata,
         Job {
             work: change_detector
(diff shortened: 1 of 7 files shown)


fontc/src/main.rs
```diff
@@ -14,9 +14,9 @@ use fontir::{
     stateset::StateSet,
 };
 use glyphs2fontir::source::GlyphsIrSource;
-use log::{debug, error, info, warn};
-use rayon::prelude::*;
+use log::{debug, error, info, log_enabled, trace, warn};
 use serde::{Deserialize, Serialize};
+use std::io::Write;
```
cmyr, line 19: nittiest nit nit: keep the std imports together, at the top.

fontc/src/main.rs
```diff
@@ -167,69 +166,61 @@ fn glyphs_deleted<'a>(current_inputs: &Input, prev_inputs: &'a Input) -> HashSet
         .collect()
 }
 
-fn add_static_metadata_work(
+fn add_static_metadata_ir_job(
     change_detector: &mut ChangeDetector,
-    context: &Context,
-    work: &mut Vec<Box<dyn Work>>,
+    jobs: &mut HashMap<WorkIdentifier, Job>,
```
cmyr, line 171: not blocking, but I'd probably prefer some sort sort of `JobSet` type or something, instead of always passing around the hashmap?

fontc/src/main.rs
```diff
@@ -255,8 +246,63 @@ impl ChangeDetector {
...
 
+/// A unit of executable work plus the identifiers of work that it depends on
+struct Job {
+    // The actual task
+    work: Box<dyn Work + Send>,
+    // Things that must happen before us, e.g. our dependencies
+    happens_after: HashSet<WorkIdentifier>,
+}
+
+fn create_jobs(
```
cmyr, line 257: this could be some kind of constructor method on our imagined `JobSet` type? 🤔

fontc/src/main.rs
```diff
@@ -265,31 +311,71 @@ fn main() -> Result<(), Error> {
...
-    add_static_metadata_work(&mut change_detector, &context, &mut work)?;
-    do_work(&context, work)?;
-
-    // Accumulate work for the parts of IR that trivially parallelize
-    let mut work: Vec<Box<dyn Work>> = Vec::new();
+    let mut pending_jobs = create_jobs(&mut change_detector)?;
+    let job_count = pending_jobs.len();
+
+    // Async work will send us it's ID on completion
+    let (send, recv) = crossbeam_channel::unbounded::<(WorkIdentifier, Result<(), WorkError>)>();
```
cmyr, line 318: I feel like for something like this I would probably choose a design where there is some sort of `WorkContext` struct or similar that has a `run()` method, and then the runloop logic would live there? Not a blocker, but something you might consider.

cmyr (approved):
This was a pretty cursory skim from me, but nothing jumped out as Obviously Bad 🤷

fontc/src/main.rs
rsheeter: Good call, will add.

rsheeter: I added a Workload type which is consumed by calling exec. Feels better, good suggestion.

fontc/src/main.rs
rsheeter: Fixed, ty

fontc/src/main.rs
rsheeter: I left this as I like the model of compiler populates workload then runs it. If you have a chance take a look at where I ended and lmk what you think. I'm going to merge but I'm happy to make further changes based on additional feedback.

fontc/src/main.rs
```diff
@@ -255,8 +247,139 @@ impl ChangeDetector {
...
+        Ok(JobResults { success, error })
+    }
+}
+
+/// A unit of executable work plus the identifiers of work that it depends on
+struct Job {
+    // The actual task
+    work: Box<dyn Work + Send>,
+    // Things that must happen before us, e.g. our dependencies
+    happens_after: HashSet<WorkIdentifier>,
```
anthrotype, line 352: I think `dependencies` would be better/clearer name for this field than `happens_after`.. The latter being a verb instead of a noun, it evokes to me an action more than a state, thus more a method name than the name of an attribute..

fontir/src/source.rs
```diff
@@ -56,16 +56,20 @@ pub trait Source {
     /// Create a function that could be called to generate [crate::ir::StaticMetadata].
     ///
     /// When run work should update [Context] with new [crate::ir::StaticMetadata].
-    fn create_static_metadata_work(&self, context: &Context) -> Result<Box<dyn Work>, Error>;
+    fn create_static_metadata_work(&self, input: &Input) -> Result<Box<dyn Work + Send>, Error>;
 
     /// Create a function that could be called to generate IR for glyphs.
     ///
     /// Batched because some formats require IO to figure out the work.
+    /// Expected to return a Vec aligned with the glyph_names input. That is,
```
anthrotype, line 64: > Expected to return a Vec aligned with the glyph_names input...

... but `glyph_names` is typed as a HashSet so I believe its order will be kind of abitrary? If the order matters, should it not also be a Vec (or IndexSet)?

fontc/src/main.rs
rsheeter: For context happens-after is more how you would think about it when reasoning about the concurrency. I think I agree re: renaming though.

fontir/src/source.rs
rsheeter: IndexSet I think, good catch.

rsheeter: https://github.com/googlefonts/fontmake-rs/pull/77

fontc/src/main.rs
rsheeter: https://github.com/googlefonts/fontmake-rs/pull/77

PR #76, opened 2023-01-06
Run fea-rs on features for UFO+designspace when identical

### LINKED ISSUE #45: Restore static metadata from IR if safe
#43 blindly recreates static metadata every cycle, which is cheap enough but not strictly necessary

### LINKED ISSUE #73: Provide rudimentary support for ufo+designspace features
For mk1 just fail if the feature files don't match, ref #72.

rsheeter:
Fixes #73 

Along the way fixed #45.

Has grown large, will split into smaller PRs to submit.


rsheeter:
Split into smaller PRs that are much more plausible to review so closing.

PR #78, opened 2023-01-09
Add a simple test ufo+ds that has features

rsheeter:
Intended for use testing feature compilation (#73). Split out from #76.


cmyr:
So I wonder if it's worth being more principled about test data? For fea-rs, I have a separate repo that contains all my test data, which I include as a submodule, although I appreciate this is not always a popular strategy.

My main thinking is that we are going to want a lot of overlapping test data used in various points in the pipeline. Does it make sense to reuse that data somewhat?

cmyr (approved):

PR #79, opened 2023-01-09
Fontdrasil

rsheeter:
Introduce a crate for things FE and BE both use. Naming based on yggdrasil, the world tree that connects norse realms, to keep the "fontir" style alive.

Intent is a future BE crate will have it's own Context and use the same machinery. Split out of #76.

### DIFF
--- a/fontdrasil/src/lib.rs
+++ b/fontdrasil/src/lib.rs
@@ -1,0 +1,1 @@
+pub mod orchestration;
--- a/fontdrasil/src/lib.rs
+++ b/fontdrasil/src/lib.rs
@@ -1,1 +1,2 @@
 pub mod orchestration;
+pub mod paths;
--- a/fontdrasil/src/orchestration.rs
+++ b/fontdrasil/src/orchestration.rs
@@ -1,11 +1,6 @@
 //! Common parts of work orchestration.
 
-use std::{
-    collections::HashSet,
-    fmt::Debug,
-    hash::Hash,
-    sync::{Arc, RwLock},
-};
+use std::{collections::HashSet, fmt::Debug, hash::Hash};
 
 pub const MISSING_DATA: &str = "Missing data, dependency management failed us?";
 
@@ -19,30 +14,11 @@ pub trait Work<C, E> {
     fn exec(&self, context: &C) -> Result<(), E>;
 }
 
-#[derive(Clone)]
-pub struct Cache<T: Default> {
-    pub item: Arc<RwLock<T>>,
-}
-
-impl<T: Default> Cache<T> {
-    pub fn new() -> Cache<T> {
-        Cache {
-            item: Default::default(),
-        }
-    }
-}
-
-impl<T: Default> Default for Cache<T> {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
-/// Tracks what parts of a context we can read/write.
+/// Access control for context
 ///
 /// Not meant to prevent malicious access, merely to detect mistakes
-/// as the result of mistaken concurrent access can be confusing to track down.
-pub struct Acl<I>
+/// because the result of mistaken concurrent access can be confusing to track down.
+pub struct AccessControlList<I>
 where
     I: Eq + Hash + Debug,
 {
@@ -55,23 +31,23 @@ where
     read_mask: Option<HashSet<I>>,
 }
 
-impl<I: Eq + Hash + Debug> Acl<I> {
-    pub fn read_only() -> Acl<I> {
-        Acl {
+impl<I: Eq + Hash + Debug> AccessControlList<I> {
+    pub fn read_only() -> AccessControlList<I> {
+        AccessControlList {
             write_mask: None,
             read_mask: None,
         }
     }
 
-    pub fn read_write(read: HashSet<I>, write: I) -> Acl<I> {
-        Acl {
+    pub fn read_write(read: HashSet<I>, write: I) -> AccessControlList<I> {
+        AccessControlList {
             write_mask: Some(write),
             read_mask: Some(read),
         }
     }
 }
 
-impl<I: Eq + Hash + Debug> Acl<I> {
+impl<I: Eq + Hash + Debug> AccessControlList<I> {
     pub fn check_read_access(&self, id: &I) {
         if !self
             .read_mask
@@ -84,19 +60,8 @@ impl<I: Eq + Hash + Debug> Acl<I> {
     }
 
     pub fn check_write_access(&self, id: &I) {
-        if !self
-            .write_mask
-            .as_ref()
-            .map(|mask| mask == id)
-            .unwrap_or(false)
-        {
+        if self.write_mask.as_ref() != Some(id) {
             panic!("Illegal access to {:?}", id);
         }
     }
-}
-
-pub fn glyph_file(glyph_name: &str, suffix: &str) -> String {
-    // TODO handle names that are invalid for the filesystem
-    // Ref https://github.com/unified-font-object/ufo-spec/issues/164
-    glyph_name.to_owned() + suffix
 }
--- a/fontdrasil/src/orchestration.rs
+++ b/fontdrasil/src/orchestration.rs
@@ -1,0 +1,102 @@
+//! Common parts of work orchestration.
+
+use std::{
+    collections::HashSet,
+    fmt::Debug,
+    hash::Hash,
+    sync::{Arc, RwLock},
+};
+
+pub const MISSING_DATA: &str = "Missing data, dependency management failed us?";
+
+/// A unit of work safe to run in parallel
+///
+/// Naively you'd think we'd just return FnOnce + Send but that didn't want to compile
+/// See <https://github.com/rust-lang/rust/issues/29625>.
+///
+/// Data produced by work is written into a Context (type parameter C).
+pub trait Work<C, E> {
+    fn exec(&self, context: &C) -> Result<(), E>;
+}
+
+#[derive(Clone)]
+pub struct Cache<T: Default> {
+    pub item: Arc<RwLock<T>>,
+}
+
+impl<T: Default> Cache<T> {
+    pub fn new() -> Cache<T> {
+        Cache {
+            item: Default::default(),
+        }
+    }
+}
+
+impl<T: Default> Default for Cache<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Tracks what parts of a context we can read/write.
+///
+/// Not meant to prevent malicious access, merely to detect mistakes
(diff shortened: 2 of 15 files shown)


fontdrasil/src/orchestration.rs
```diff
@@ -0,0 +1,102 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Tracks what parts of a context we can read/write.
+///
+/// Not meant to prevent malicious access, merely to detect mistakes
+/// as the result of mistaken concurrent access can be confusing to track down.
+pub struct Acl<I>
```
anthrotype, line 45: what does "Acl" mean? "access control" or something?

fontdrasil/src/orchestration.rs
```diff
@@ -0,0 +1,102 @@
...
+            .as_ref()
+            .map(|mask| mask == id)
+            .unwrap_or(false)
+        {
+            panic!("Illegal access to {:?}", id);
+        }
+    }
+}
+
+pub fn glyph_file(glyph_name: &str, suffix: &str) -> String {
```
anthrotype, line 98: this `glyph_file` doesn't feel like it belongs to a module named "orchestration", does it? how about `fontdrasil::paths` or similar

anthrotype (approved):
this mostly moves existing stuff to the new crate so LGTM, but see minor comments

fontdrasil/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod orchestration;
```
cmyr, line 1: root `lib` should have `//! at least some module comment`. This is the text on the root docs page for this crate.

Also, I think it's worth turning on `deny(missing_docs)`; it can be annoying, but it has real benefits, and it's even more annoying to turn on in three months and then have 300 errors to fix. You can always judiciously use `#[allow(missing_docs)]` on specific items where you don't think docs are useful, like large enums.

fontdrasil/src/orchestration.rs
cmyr: +1, would prefer a slightly more descriptive name.

fontdrasil/src/orchestration.rs
```diff
@@ -0,0 +1,102 @@
...
+/// Naively you'd think we'd just return FnOnce + Send but that didn't want to compile
+/// See <https://github.com/rust-lang/rust/issues/29625>.
+///
+/// Data produced by work is written into a Context (type parameter C).
+pub trait Work<C, E> {
+    fn exec(&self, context: &C) -> Result<(), E>;
+}
+
+#[derive(Clone)]
+pub struct Cache<T: Default> {
```
cmyr, line 23: Unless absolutely required, preference is to avoid trait bounds on type declarations, since they're sort of 'contagious'. The preferred approach is to keep the trait bounds to the impl blocks, where you can pick exactly the set of constraints that are necessary for that impl.

I presume that you had some annoyance around `Default`, here?

In any case this is a funny type. Is `T` supposed to be some sort of collection?

fontdrasil/src/orchestration.rs
```diff
@@ -0,0 +1,102 @@
...
+    pub item: Arc<RwLock<T>>,
```
cmyr, line 24: Although I'm not sure exactly how this is used, in general I'd prefer this field not to be `pub`.

fontdrasil/src/orchestration.rs
```diff
@@ -0,0 +1,102 @@
...
+            item: Default::default(),
+        }
+    }
+}
+
+impl<T: Default> Default for Cache<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
cmyr, line 39: Assuming we derive default, we can just write,

```suggestion
impl<T: Default> Cache<T> {
    pub fn new() -> Cache<T> {
        Self::default()
    }
}
```

fontdrasil/src/orchestration.rs
```diff
@@ -0,0 +1,102 @@
...
+    I: Eq + Hash + Debug,
+{
+    // If present, the one and only key you are allowed to write to
+    // None means we're read-only
+    write_mask: Option<I>,
+
+    // If present, what you can access through this context
+    // Intent is root has None, task-specific Context only allows access to dependencies
+    read_mask: Option<HashSet<I>>,
+}
```
cmyr, line 56: Looking at this makes me wonder if we want to be an enum, with `Read` and `ReadWrite` variants or something?

like,

```rust
enum Access {
    Read { read_mask: Option<HashSet<I>> },
    ReadWrite { read_mask: Option<HashSet<I>>, write_mask: I },
}
```

And to keep going, possibly we should have a type, `ReadMask`, like:

```rust
enum ReadMask {
    Root,
    Items(HashSet<I>),
}
```

but maybe not, I'm just spitballing.

fontdrasil/src/orchestration.rs
```diff
@@ -0,0 +1,102 @@
...
+            panic!("Illegal access");
+        }
+    }
+
+    pub fn check_write_access(&self, id: &I) {
+        if !self
+            .write_mask
+            .as_ref()
+            .map(|mask| mask == id)
+            .unwrap_or(false)
```
cmyr, line 91: you can also write this as,

`if self.write_mask.as_ref() != Some(id) { ... }`

fontir/src/orchestration.rs
```diff
@@ -45,89 +40,54 @@ pub struct Context {
...
-struct Cache<T: Default> {
-    item: Arc<RwLock<T>>,
-}
-
-impl<T: Default> Cache<T> {
-    fn new() -> Cache<T> {
-        Cache {
-            item: Default::default(),
+impl Context {
+    fn copy(&self, acl: Acl<WorkIdentifier>) -> Context {
```
cmyr, line 52: I'd maybe call this `with_acl` or something? `copy` is heavily associated with the `Copy` trait.

fontir/src/orchestration.rs
```diff
@@ -137,33 +97,62 @@ impl Context {
             .unwrap();
         let buf_io = BufWriter::new(raw_file);
         serde_yaml::to_writer(buf_io, &content)
-            .map_err(|e| panic!("Unable to serialize to {:?}: {}", file, e))
+            .map_err(|e| panic!("Unable to serialize\n{:#?}\nto {:?}: {}", content, file, e))
```
cmyr, line 100: ultimately we're going to want to actually return errors for things like this. I'm not sure if it's worth doing now, but it's worth keeping in mind...

fontir/src/orchestration.rs
```diff
@@ -137,33 +97,62 @@ impl Context {
...
             .unwrap();
     }
 
+    fn restore<V>(&self, file: &Path) -> V
+    where
+        V: ?Sized + DeserializeOwned,
+    {
+        let raw_file = File::open(file)
+            .map_err(|e| panic!("Unable to read {:?} {}", file, e))
+            .unwrap();
```
cmyr, line 110: if you just `unwrap` this error, you should get approximately the same information as you are manually printing here.

cmyr (approved):
A bunch of nits and observations of unknown value inline. 💁

fontdrasil/src/orchestration.rs
rsheeter: Interesting, I thought that name was very standard for this purpose. Will rename and expand comment.

fontdrasil/src/orchestration.rs
rsheeter: Good catch, I agree

fontdrasil/src/orchestration.rs
rsheeter: In looking again I think I originally had a slightly more interesting idea but ended up with what amounts to a verbose type alias, will delete.

fontdrasil/src/orchestration.rs
rsheeter: Deleted Cache

fontdrasil/src/orchestration.rs
rsheeter: Deleted Cache

rsheeter: But ty, I'm sure that will come in handy for other types

fontdrasil/src/orchestration.rs
rsheeter: I think it works as a single set of enum Access Read(I) Write(I). Let me try that and see if it feels better.

rsheeter: It looked lovely but led me to copy the &I because I wanted to lookup an &I in a HashSet<Access<I>>. Might just be a rust noob mistake, I'll make a gist and send to chat. For the purpose of this PR I will leave as is and then update to Something Better later.

rsheeter: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=07b82a0988efb3869e40e2be855c2354

fontdrasil/src/orchestration.rs
rsheeter: I like it, so updated

fontir/src/orchestration.rs
rsheeter: Hm, I thought it didn't tell me the file but maybe that's a figment of my imagination. Will double-check

fontir/src/orchestration.rs
rsheeter: Good catch, updated

fontir/src/orchestration.rs
rsheeter: https://github.com/googlefonts/fontmake-rs/issues/84 as I think we have a number of places where we panic or unwrap things that could legitimately fail.

fontir/src/orchestration.rs
rsheeter: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=963503e29a7d3b50eee0f1c5e3cd0e6e gives me `Os { code: 2, kind: NotFound, message: "No such file or directory" }` which does not include the path

rsheeter:
> A bunch of nits and observations of unknown value inline. tipping_hand_person

All much appreciated, helps me learn as well as improving the code

fontdrasil/src/lib.rs
rsheeter: Added a string here and filed https://github.com/googlefonts/fontmake-rs/issues/85 to make sure we turn this on globally. It's a great idea but I don't want to fill up this PR with doc additions.

cmyr: oh yes definitely not for this PR :)

PR #82, opened 2023-01-09
Add wiring for backend tasks

rsheeter:
Make fontc use AnyWork, AnyWorkId, etc so it can accomodate FE and BE tasks with non-overlapping id spaces. Subset of changes from #76.

### DIFF
--- a/fontbe/src/orchestration.rs
+++ b/fontbe/src/orchestration.rs
@@ -2,8 +2,9 @@
 
 use std::{collections::HashSet, fs, path::Path, sync::Arc};
 
-use fontdrasil::orchestration::{Acl, Cache, Work, MISSING_DATA};
+use fontdrasil::orchestration::{AccessControlList, Work, MISSING_DATA};
 use fontir::orchestration::{Context as FeContext, WorkIdentifier as FeWorkIdentifier};
+use parking_lot::RwLock;
 use write_fonts::FontBuilder;
 
 use crate::{error::Error, paths::Paths};
@@ -72,11 +73,11 @@ pub struct Context {
     // The final, fully populated, read-only FE context
     pub ir: Arc<FeContext>,
 
-    acl: Acl<AnyWorkId>,
+    acl: AccessControlList<AnyWorkId>,
 
     // work results we've completed or restored from disk
     // We create individual caches so we can return typed results from get fns
-    features: Cache<Option<Arc<Vec<u8>>>>,
+    features: Arc<RwLock<Option<Arc<Vec<u8>>>>>,
 }
 
 impl Context {
@@ -85,8 +86,8 @@ impl Context {
             emit_ir,
             paths: Arc::from(paths),
             ir: Arc::from(ir.read_only()),
-            acl: Acl::read_only(),
-            features: Cache::new(),
+            acl: AccessControlList::read_only(),
+            features: Arc::from(RwLock::new(None)),
         }
     }
 
@@ -99,7 +100,7 @@ impl Context {
             emit_ir: self.emit_ir,
             paths: self.paths.clone(),
             ir: self.ir.clone(),
-            acl: Acl::read_write(dependencies.unwrap_or_default(), work_id.into()),
+            acl: AccessControlList::read_write(dependencies.unwrap_or_default(), work_id.into()),
             features: self.features.clone(),
         }
     }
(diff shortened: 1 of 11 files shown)


fontbe/src/orchestration.rs
```diff
@@ -0,0 +1,150 @@
...
+
+use std::{collections::HashSet, fs, path::Path, sync::Arc};
+
+use fontdrasil::orchestration::{Acl, Cache, Work, MISSING_DATA};
+use write_fonts::FontBuilder;
+
+use crate::{error::Error, paths::Paths};
+
+// Unique identifier of work. If there are no fields work is unique.
+// Meant to be small and cheap to copy around.
```
cmyr, line 11: 1: to be documentation this needs there `///`.
2: treat rust docstrings like git commits: a single short descriptive line, then (if needed) a blank line and then any additional detail, so like:

```suggestion
/// Unique identifier of work.
///
/// If there are no fields work is unique.
/// Meant to be small and cheap to copy around.
```

fontbe/src/orchestration.rs
```diff
@@ -0,0 +1,150 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum WorkIdentifier {
+    Features,
+    Glyph(String),
```
cmyr, line 15: If this is intended to be cheap to copy around, I don't think it should be a `String`, which is expensive to clone.

not necessary for this PR, but I am confident that you will want to use some other type for this eventually, and it's probably worth doing this sooner rather than later. Even if all you do is create a newtype like `GlyphName(String)` for now, then it will be much easier for you to refactor this down the road.

fontbe/src/orchestration.rs
```diff
@@ -0,0 +1,150 @@
...
+    GlyphMerge,
+    FinalMerge,
+}
+
+use fontir::orchestration::{Context as FeContext, WorkIdentifier as FeWorkIdentifier};
```
cmyr, line 20: I'd put this up with the other imports?

fontc/src/work.rs
```diff
@@ -0,0 +1,115 @@
...
+    Fe(FeContext),
+    Be(BeContext),
+}
+
+impl AnyContext {
+    pub fn for_work(
+        fe_root: &FeContext,
+        be_root: &BeContext,
+        id: &AnyWorkId,
+        happens_after: HashSet<AnyWorkId>,
```
cmyr, line 80: I suspect we would find some joy in a `WorkSet` type.

fontc/src/main.rs
```diff
@@ -132,66 +138,75 @@ fn ir_source(source: &Path) -> Result<Box<dyn Source>, Error> {
...
+                        prev_state != *curr_state
+                            || !self
+                                .ir_paths
+                                .target_file(&FeWorkIdentifier::Glyph(glyph_name.to_string()))
+                                .exists()
+                    }
+                    None => true,
+                },
+            )
+            .map(|(glyph_name, _)| glyph_name.as_str())
```
cmyr, line 178: this might be *slightly* easier expressed using `filter_map` 🤷

cmyr (approved):
Some more little notes inline, but lgtm

fontc/src/main.rs
rsheeter: Oh yes indeed. I'm used to having filter, I'm used to having map, apparently I'm not quite yet used to having filter_map

fontbe/src/orchestration.rs
rsheeter: Ha yes indeed. Why doesn't it fix the imports for me!

fontbe/src/orchestration.rs
rsheeter: It's on my todo list to make it a SmolStr. I think that's what fea-rs uses as well?

Filed https://github.com/googlefonts/fontmake-rs/issues/86 to increase the odds of remembering.

fontc/src/main.rs
rsheeter: So updated.

fontc/src/work.rs
rsheeter: An id set? Why? - we only use it for very hashset'ish things currently.

rsheeter: I didn't make that change as I wasn't quite sure why. Happy to make it in a follow-on PR if you wish.

PR #88, opened 2023-01-12
Use SmolStr instead of String in WorkIdentifier::GlyphName

### LINKED ISSUE #86: Make WorkIdentifier String (for both FE and BE) use something cheaper
On my (paper) todo list to look at SmolStr.

As it came up in https://github.com/googlefonts/fontmake-rs/pull/82#discussion_r1066317577, filing here.

Can then make both work identifier types Copy.

rsheeter:
Replaces #87 which made fontc unsuitable for library use.

Fixes #86

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -20,7 +20,7 @@ pub struct FeatureWork {
 impl FeatureWork {
     pub fn create(build_dir: &Path) -> Box<BeWork> {
         let build_dir = build_dir.to_path_buf();
-        Box::from(FeatureWork { build_dir })
+        Box::new(FeatureWork { build_dir })
     }
 }
 
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -2,7 +2,7 @@
 
 use std::path::{Path, PathBuf};
 
-use fea_rs::{Diagnostic, GlyphMap, GlyphName};
+use fea_rs::{Diagnostic, GlyphMap, GlyphName as FeaRsGlyphName};
 use log::{debug, trace, warn};
 use write_fonts::FontBuilder;
 
@@ -102,7 +102,10 @@ impl Work<Context, Error> for FeatureWork {
         if glyph_order.is_empty() {
             warn!("Glyph order is empty; feature compile improbable");
         }
-        let glyph_map = glyph_order.iter().map(Into::<GlyphName>::into).collect();
+        let glyph_map = glyph_order
+            .iter()
+            .map(|n| Into::<FeaRsGlyphName>::into(n.as_str()))
+            .collect();
         let font = self.compile(fea_file, glyph_map)?;
         context.set_features(font);
         Ok(())
--- a/fontbe/src/orchestration.rs
+++ b/fontbe/src/orchestration.rs
@@ -2,7 +2,10 @@
 
 use std::{collections::HashSet, fs, path::Path, sync::Arc};
 
-use fontdrasil::orchestration::{AccessControlList, Work, MISSING_DATA};
+use fontdrasil::{
+    orchestration::{AccessControlList, Work, MISSING_DATA},
+    types::GlyphName,
+};
 use fontir::orchestration::{Context as FeContext, WorkIdentifier as FeWorkIdentifier};
 use parking_lot::RwLock;
 use write_fonts::FontBuilder;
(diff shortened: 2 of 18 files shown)


fontir/src/orchestration.rs
```diff
@@ -20,8 +23,8 @@ use crate::{error::WorkError, ir, paths::Paths, source::Input};
 #[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub enum WorkIdentifier {
     StaticMetadata,
-    Glyph(String),
-    GlyphIrDelete(String),
+    Glyph(GlyphName),
+    GlyphIrDelete(),
```
dfrg, line 27: Should probably remove the parentheses for this variant.

glyphs2fontir/src/source.rs
```diff
@@ -136,9 +137,10 @@ impl Source for GlyphsIrSource {
 
         let mut work: Vec<Box<IrWork>> = Vec::new();
         for glyph_name in glyph_names {
-            work.push(Box::from(
-                self.create_work_for_one_glyph(glyph_name, cache.font_info.clone())?,
-            ));
+            work.push(Box::from(self.create_work_for_one_glyph(
```
dfrg, line 140: `Box::from` is fine but `Box::new` is a bit more idiomatic.

fontbe/src/orchestration.rs
```diff
@@ -16,7 +19,7 @@ use crate::{error::Error, paths::Paths};
 #[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub enum WorkIdentifier {
```
dfrg, line 20: Just below, we have a type called `AnyWorkId`. Preferring shorter names, I'd recommend renaming `WorkIdentifier` and `FeWorkIdentifier` to `WorkId` and `FeWorkId` respectively, but in any case, it would be nice if they were all consistent.

fontbe/src/features.rs
```diff
@@ -2,7 +2,7 @@
 
 use std::path::{Path, PathBuf};
 
-use fea_rs::{Diagnostic, GlyphMap, GlyphName};
+use fea_rs::{Diagnostic, GlyphMap, GlyphName as FeaRsGlyphName};
```
dfrg, line 5: I assume we'll want to pull `fea-rs` into this repo eventually and unify these types somehow.

dfrg (approved):
Simple enough swapping of types. Looks good. I've left comments on some minor nits inline, but nothing blocking.

fontdrasil/Cargo.toml
```diff
@@ -12,6 +12,12 @@ categories = ["text-processing", "parsing", "graphics"]
 
 [dependencies]
 
+parking_lot = "0.12.1"
```
anthrotype, line 15: where's parking_lot being used inside fontdrasil specifically? I see it imported in fontir and fontbe only.

fontbe/src/features.rs
rsheeter: IMHO we should have a common GlyphName both consume from somewhere else. Tracking in https://github.com/googlefonts/fontations/issues/201.

glyphs2fontir/src/source.rs
rsheeter: Huh I'd have sworn I got a red underliine when I guessed Box::new and then thought I was supposed to use from. Will fix, ty.

fontbe/src/orchestration.rs
rsheeter: Oops, will fix.

glyphs2fontir/src/source.rs
rsheeter: Evidently I exclusively use Box::from. Fixed.

fontdrasil/Cargo.toml
rsheeter: That's a leftover from my attempt to make identifiers unnecessarily complicated, removed, ty.

fontir/src/orchestration.rs
rsheeter: I'm mildly surprised that's not a clippy, when would you specifically *want* that?

rsheeter: Fixed.

PR #91, opened 2023-01-13
Compile simple - no token expansion - fea for .glyphs

rsheeter:
Using a workaround for https://github.com/cmyr/fea-rs/issues/115. Will need changes when an official way to compile from memory arrives.

Does not implement the only interesting part of feature compilation: [TokenExpander](https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/Lib/glyphsLib/builder/tokens.py#L12). Tracking token expansion in #92.

Confirmed generated file for Texturina was identical to what I got from Python and used Texturina to build the tests.

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -1,9 +1,14 @@
 //! Feature binary compilation.
 
-use std::path::{Path, PathBuf};
+use std::{
+    fmt::Debug,
+    fs,
+    path::{Path, PathBuf},
+};
 
-use fea_rs::{Diagnostic, GlyphMap, GlyphName as FeaRsGlyphName};
-use log::{debug, trace, warn};
+use fea_rs::{Diagnostic, GlyphMap, GlyphName as FeaRsGlyphName, ParseContext};
+use fontir::ir::Features;
+use log::{debug, error, trace, warn};
 use write_fonts::FontBuilder;
 
 use fontdrasil::orchestration::Work;
@@ -25,7 +30,7 @@ impl FeatureWork {
 }
 
 fn check_diagnostics(
-    fea_file: &Path,
+    feature_source: impl Debug,
     op: &str,
     diagnostics: &Vec<Diagnostic>,
     formatter: impl Fn(&Diagnostic) -> String,
@@ -33,41 +38,51 @@ fn check_diagnostics(
     let mut err = false;
     for diagnostic in diagnostics {
         if diagnostic.is_error() {
-            warn!("{:?} {} error {}", fea_file, op, formatter(diagnostic));
+            warn!(
+                "{:?} {} error {}",
+                feature_source,
+                op,
+                formatter(diagnostic)
+            );
             err = true;
         } else {
-            debug!("{:?} {} {}", fea_file, op, formatter(diagnostic));
+            debug!("{:?} {} {}", feature_source, op, formatter(diagnostic));
         }
     }
     if err {
-        return Err(Error::FeaError(format!("{:?} {} failed", fea_file, op)));
+        return Err(Error::FeaError(format!(
+            "{:?} {} failed",
+            feature_source, op
+        )));
     }
     Ok(())
 }
 
 impl FeatureWork {
-    /// Inspired by (as in shameless copy of) how the fea-rs binary flows.
-    fn compile(&self, fea_file: &Path, glyph_order: GlyphMap) -> Result<FontBuilder, Error> {
-        // Will you not parse?!
-        let parse =
-            fea_rs::parse_root_file(fea_file, Some(&glyph_order), Some(self.build_dir.clone()))
-                .map_err(|e| Error::FeaError(format!("{:?} parsing {:?}", e, fea_file)))?;
+    fn compile_parse(
+        &self,
+        feature_source: &str,
+        parse: ParseContext,
+        glyph_order: GlyphMap,
+    ) -> Result<FontBuilder, Error> {
         let (tree, diagnostics) = parse.generate_parse_tree();
-        check_diagnostics(fea_file, "generate parse tree", &diagnostics, |d| {
+        check_diagnostics(feature_source, "generate parse tree", &diagnostics, |d| {
             format!("{:?}", d)
         })?;
 
         // Maybe even compile?
         let compilation = match fea_rs::compile::compile(&tree, &glyph_order) {
             Ok(compilation) => {
-                check_diagnostics(fea_file, "compile", &compilation.warnings, |d| {
+                check_diagnostics(feature_source, "compile", &compilation.warnings, |d| {
                     tree.format_diagnostic(d)
                 })?;
-                trace!("Compiled {:?} successfully", fea_file);
+                trace!("Compiled {} successfully", feature_source);
                 compilation
             }
             Err(errors) => {
-                check_diagnostics(fea_file, "compile", &errors, |d| tree.format_diagnostic(d))?;
+                check_diagnostics(feature_source, "compile", &errors, |d| {
+                    tree.format_diagnostic(d)
+                })?;
                 unreachable!("errors aren't ... errors?!");
             }
         };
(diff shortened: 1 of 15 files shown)


glyphs-reader/src/font.rs
```diff
@@ -41,6 +41,21 @@ pub struct Font {
...
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub enum FeatureType {
+    Prefix,
+    Class,
+    Feature,
+}
+
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct FeatureSnippet {
+    pub feature_type: FeatureType,
```
anthrotype, line 56: why a separate FeatureType enum contained in the FeatureSnippet struct's feature_type field, and not instead a FeatureSnippet _enum_ where each variant carries with it the respective snippet name/code?

glyphs-reader/src/font.rs
```diff
@@ -862,6 +891,51 @@ impl TryFrom<RawGlyph> for Glyph {
...
+            _ => "???",
+        }
+    }
+}
+
+// https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/Lib/glyphsLib/builder/features.py#L90
+fn prefix_to_feature(prefix: RawFeature) -> Result<FeatureSnippet, Error> {
+    let name = match prefix.name.as_ref().map(|s| s.as_str()) {
+        Some("<anonymous>") => prefix.name.unwrap(),
+        Some(..) => prefix.name.unwrap(),
```
anthrotype, line 907: what's the difference in practive between the two match branches above, namely `Some("<anonymous>")` and `Some(..)`, don't both return the same `prefix.name.unwrap()`?

anthrotype: also, why not simply do

```rust
let name = match prefix.name {
    Some(name) => name,
    None => prefix.generate_name().to_string(),
}
```

glyphs-reader/src/font.rs
```diff
@@ -862,6 +891,51 @@ impl TryFrom<RawGlyph> for Glyph {
     }
 }
 
+impl RawFeature {
+    fn generate_name(&self) -> &'static str {
+        match self.automatic {
+            Some(1i64) => "automatic",
+            _ => "???",
```
anthrotype, line 898: where did you infer this behavior from?
If created from Glyphs.app, feature prefixes (like classes or features proper) will always have a name in the UI and as such also in the .glyphs source file. That special "<anonymous>" named prefix that you saw in glyphsLib source code is simply the special name that glyphsLib resorts to when converting from features from UFO => .glyphs, for anything that doesn't match the `# Prefix: (.*)$` pattern which is used to record a Glyphs' prefix.name when going .glyphs => UFO.
The automatic bool flag doesn't have anything to do with a RawFeature's name. Also that `"???"` seems arbitrarty to me.
I think you can expect that all RawFeatures (be them prefix, class or features) have a name, it's not optional; then you can get rid of generate_name() method altogether.

glyphs-reader/src/font.rs
rsheeter: It seemed the fields are the same which made such an enum feel weird to me. Maybe I can just not have the type field at all.

glyphs-reader/src/font.rs
rsheeter: That's ... either a copy paste error or a series of code manipulations that left nonsensical code I just didn't notice. I think at some point <anonymous> did something else.

glyphs-reader/src/font.rs
rsheeter: It stands in for autostr. I suppose complete consistency with that would be "" but I prefer the indicator something is awry. I don't think we _should_ ever hit ???. Maybe I'll add a warn!.

fontbe/src/features.rs
```diff
@@ -78,26 +93,46 @@ impl FeatureWork {
...
     }
+
+    fn compile_memory(
+        &self,
+        fea_content: &String,
+        glyph_order: GlyphMap,
+    ) -> Result<FontBuilder, Error> {
+        // Will you not parse?!
+
+        // TODO write out the feature content on failure
```
anthrotype, line 110: you already have code to write to debug_file when compile fails, you may as well factor that out to its own function and also call it when parsing fails

fontbe/src/features.rs
```diff
@@ -106,7 +141,32 @@ impl Work<Context, Error> for FeatureWork {
             .iter()
             .map(|n| Into::<FeaRsGlyphName>::into(n.as_str()))
             .collect();
-        let font = self.compile(fea_file, glyph_map)?;
+
+        let font = match &*features {
+            Features::File(fea_file) => self.compile_file(fea_file, glyph_map)?,
+            Features::Memory(fea_content) => {
+                let result = self.compile_memory(fea_content, glyph_map);
+                if result.is_err() || context.emit_ir {
```
anthrotype, line 149: not sure if emit_ir flag should also apply to emitting debug files, they aren't really the same thing and one may want one without the other? maybe you could have a separate --emit-debug or something

fontbe/src/paths.rs
```diff
@@ -10,22 +10,29 @@ use crate::orchestration::WorkId;
 pub struct Paths {
     build_dir: PathBuf,
     glyph_dir: PathBuf,
+    debug_dir: PathBuf,
 }
 
 impl Paths {
     pub fn new(build_dir: &Path) -> Paths {
         let glyph_dir = build_dir.join("glyphs");
+        let debug_dir = build_dir.join("_debug");
```
anthrotype, line 19: what's the meaning of the underscore prefix?

glyphs-reader/src/font.rs
```diff
@@ -5,6 +5,7 @@
 //! where it gets serialized to more Rust-native structures, proc macros, etc.
 
 use std::collections::{BTreeMap, HashSet};
+use std::fmt::Debug;
```
anthrotype, line 8: is this required? isn't Debug already in the preamble?

glyphs-reader/src/font.rs
```diff
@@ -953,6 +978,59 @@ impl TryFrom<RawGlyph> for Glyph {
...
 
+// https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/Lib/glyphsLib/builder/features.py#L43
+fn autostr(feature: &RawFeature) -> String {
+    match feature.automatic {
+        Some(1) => "# automatic\n".to_string(),
+        _ => "".to_string(),
+    }
+}
+
+fn name(feature: &RawFeature) -> Result<String, Error> {
```
anthrotype, line 989: "name" is quite a short and common _name_ to name a function, may inadvertently be confused with some other local variable named like that, maybe rename "feature_name"?

resources/testdata/glyphs2/Fea_Order.glyphs
```diff
@@ -0,0 +1,69 @@
+{
+.appVersion = "3148";
+classes = (
```
anthrotype, line 3: I think you have `classes` entry twice (there's another one probably overwriting this one further below), keys in plist dicts are unique

anthrotype (commented):
LGTM, plus minor non blocking comments

fontbe/src/paths.rs
rsheeter: I just felt it made it more obvious that it was different. I don't feel strongly about it, would just "debug" be better?

anthrotype: i actually like the underscore prefix, it's like when you have an unused local variable to stop clippy from complaining, in this case _debug directory isn't actually "used" for compilation, hence the underscore

anthrotype (approved):

fontbe/src/features.rs
rsheeter: Oh good catch, fixed

glyphs-reader/src/font.rs
rsheeter: Err. Yes. I wonder how I managed to do that!

glyphs-reader/src/font.rs
rsheeter: I feel like name(RawFeature) is arguably clear though I buy the consumption of a common name. Let me just add it directly to RawFeature so it's feature.name()

resources/testdata/glyphs2/Fea_Order.glyphs
rsheeter: Oops, indeed.

glyphs-reader/src/font.rs
rsheeter: Fixed

fontbe/src/features.rs
rsheeter: Oh good call. Will add.

PR #97, opened 2023-01-20
Use upcoming compile api for fea-rs

rsheeter:
Build against https://github.com/cmyr/fea-rs/pull/119.

Ran into some issues:

1. Had to copy the feature content for in memory use
1. Using a closure for the resolve sounded lovely until I tried to actually do it, then I angered the compiler and ultimately made a struct
1. `cargo test -p fontc fea` now fails, apparently absolute paths aren't welcome?
   * `/blah-blah-blah/fontmake-rs/resources/testdata/Static-Regular.ufo/features.fea` triggers `thread 'tests::compile_fea' panicked at 'assertion failed: path.is_relative()', ...blah blah.../fea-rs/src/util/paths.rs:15:9

Update: ^ fixed by latest updates to the fea-rs branch.

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -2,7 +2,9 @@
 
 use std::{
     ffi::{OsStr, OsString},
+    fmt::Display,
     fs,
+    rc::Rc,
 };
 
 use fea_rs::{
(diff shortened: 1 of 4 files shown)


fontbe/src/features.rs
```diff
@@ -18,114 +20,59 @@ use crate::{
...
+    content_path: OsString,
+    content: String,
+}
+
+impl SourceResolver for CursedClosures {
+    fn get_contents(&self, path: &OsStr) -> Result<String, SourceLoadError> {
+        if path == &*self.content_path {
+            return Ok(self.content.clone());
         }
+        // I don't think I can actually construct this given private fields and pub(crate) new
```
cmyr, line 45: mm good point, will look into this.

fontbe/src/features.rs
```diff
@@ -18,114 +20,59 @@ use crate::{
...
+            let resolver = CursedClosures {
+                content_path: root_path.clone(),
+                content: fea_content.clone(),
+            };
+            compiler = compiler.with_resolver(resolver);
+        }
+        match compiler.compile() {
+            Ok(compilation) => compilation
+                .assemble(&glyph_order, Default::default())
+                .map_err(|e| Error::FeaError(format!("{:?} assembling {:?}", e, root_path))),
```
cmyr, line 70: I would make your `FeaError` variant just wrap the error produced by `fea-rs`.

cmyr (commented):
Some notes inline, I agree this doesn't feel like it's quite there yet on  the `fea-rs` side, although I'm not exactly sure what the answer is.

fontbe/src/features.rs
rsheeter: Fixed in fea-rs update

fontbe/src/features.rs
```diff
@@ -18,114 +20,59 @@ use crate::{
...
-        if parse.is_err() {
-            write_debug_fea(context, parse.is_err(), "fea parse failed", fea_content);
+        eprintln!("ROOT: {:?}", root_path);
+        let mut compiler = Compiler::new(root_path.clone(), &glyph_order);
+        if let Features::Memory(fea_content) = features {
+            let resolver = CursedClosures {
+                content_path: root_path.clone(),
+                content: fea_content.clone(),
+            };
+            compiler = compiler.with_resolver(resolver);
```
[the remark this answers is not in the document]
rsheeter: Features::File uses your default impl, Features::Memory provides it's own so it's not quite true the whole Features type is the resolver.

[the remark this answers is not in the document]
rsheeter: For now we fail resolution of anything other than the root item. https://github.com/googlefonts/fontmake-rs/issues/109 covers actual resolution.

PR #98, opened 2023-01-20
glyphs contour and component conversion

### LINKED ISSUE #93: Generate IR for .glyphs paths
References:

* glyphsLib
   * [to_ufo_glyph](to_ufo_layers) triggers path and component translation [here](https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/Lib/glyphsLib/builder/glyph.py#L163)
   * [to_ufo_paths](https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/Lib/glyphsLib/builder/paths.py#L20) translates the actual path.
   * https://github.com/googlefonts/fontmake-rs/blob/main/ufo2fontir/src/toir.rs does the UFO:IR - borderline nop - translation

rsheeter:
Fixes #93. Note that I chose to keep a 2x3 - as Glyphs 2 did - rather than Glyphs 3 style individual fields because it's not obvious what order those fields apply in.

### DIFF
--- a/fontir/Cargo.toml
+++ b/fontir/Cargo.toml
@@ -19,7 +19,10 @@ serde_yaml = "0.9.14"
 bincode = "1.3.3"
 filetime = "0.2.18"
 thiserror = "1.0.37"
+
+kurbo = { version = "0.9.0", features = ["serde"] }
 ordered-float = { version = "3.4.0", features = ["serde"] }
+
 indexmap = "1.9.2"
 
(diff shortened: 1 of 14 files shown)


dfrg (approved):
My concerns were all addressed in #101 so this one looks good to me!

PR #104, opened 2023-01-23
Use kurbo::BezPath instead of our own in-house contour

### LINKED ISSUE #33: Representing path
Per f2f w/@anthrotype, perhaps we use a `Vec<Vec<Command>>` where `Vec<Command>` is an svg subpath, as from https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d

### LINKED ISSUE #103: Use Kurbo's BezPath for IR
Use https://docs.rs/kurbo/0.1.0/kurbo/struct.BezPath.html. This allows us to define shape reuse in terms of kurbo which I believe is desirable. Further, kurbo already has to/from bezier path which will be handy when compiling from svgs.

rsheeter:
Instead of the UFO-style point sequence use `kurbo::BezPath`

Fixes #103. Unexpected perk, it makes the IR easier to read:

```
    contours:
    - - !MoveTo
        x: 120.0
        y: 580.0
      - !CurveTo
        - x: 197.0
          y: 580.0
        - x: 231.0
          y: 626.0
        - x: 240.0
          y: 697.0
...etc...
      - ClosePath
```

Fixes #33.


fontir/src/ir.rs
```diff
@@ -115,41 +116,122 @@ pub struct GlyphInstance {
...
+}
+
+impl<'a> GlyphPathBuilder<'a> {
+    pub fn new(glyph_name: &'a str) -> GlyphPathBuilder {
+        GlyphPathBuilder {
+            glyph_name,
+            offcurve: Vec::new(),
+            path: BezPath::new(),
+            // kurbo *requires* subpaths start with move so this isn't going to get used
+            last_move_to: Point::ZERO,
```
cmyr, line 152: Just a general rust observation: unless it is significantly less ergonomic, I do think that modelling this as `Option<Point>` better captures the intent/invariants of the program, and makes it more likely that we will correctly identify a possible bug here.

fontir/src/ir.rs
```diff
@@ -115,41 +116,122 @@ pub struct GlyphInstance {
...
+                _ => self.check_num_offcurve(|v| v < 3)?,
+            }
+            self.offcurve.clear();
+        } else {
+            self.move_to(p)?;
+        }
+        Ok(())
+    }
+
+    pub fn offcurve(&mut self, p: impl Into<Point>) -> Result<(), WorkError> {
```
cmyr, line 216: Is there a condition where we would expect this to error in the future, or is it about matching the signature of other methods?

fontir/src/ir.rs
```diff
@@ -115,41 +116,122 @@ pub struct GlyphInstance {
...
     pub transform: Affine,
 }
 
+/// Helps convert points-of-type to a bezier path.
+///
+/// Source formats tend to use streams of point-of-type. Curve manipulation is
+/// often easier on bezier path, so provide a mechanism to convert.
+#[derive(Debug)]
+pub struct GlyphPathBuilder<'a> {
+    glyph_name: &'a str,
```
cmyr, line 139: we have a generally cheap `GlyphName` now, would it be easier to use use that and then not worry about the lifetime?

fontir/src/ir.rs
```diff
@@ -115,41 +116,122 @@ pub struct GlyphInstance {
...
+
+    fn check_num_offcurve(&self, expected: impl Fn(usize) -> bool) -> Result<(), WorkError> {
+        if !expected(self.offcurve.len()) {
+            return Err(WorkError::InvalidSourceGlyph {
+                glyph_name: self.glyph_name.into(),
+                message: format!(
+                    "{} unexpected consecutive offcurve points {:?}",
+                    self.offcurve.len(),
+                    self.offcurve
+                ),
```
cmyr, line 164: An observation: seeing this makes me think it might make sense to have a distinct `GlyphConversionError` type, which can have variants representing various possible error conditions?

The thing that stands out here is that you're sort of using the `message` field as an error type, and the format string that you provide looks like it _really_ just wants to be the `Display` impl for your error. Instead, you could have a `TooManyOffCurves` error variant that cloned `self.offcurve`, and then you would produce this string in the `Display` impl.

fontir/src/ir.rs
```diff
@@ -115,41 +116,122 @@ pub struct GlyphInstance {
...
+        if self.is_empty() {
+            self.move_to(p)?;
+        } else {
+            self.path.line_to(p);
+        }
+        Ok(())
+    }
+
+    pub fn qcurve_to(&mut self, p: impl Into<Point>) -> Result<(), WorkError> {
+        todo!("Support qcurve to {}", p.into());
```
cmyr, line 193: Is there a good reason not to implement this now?

cmyr (approved):
Looks good, some notes & nits inline.

fontir/src/ir.rs
rsheeter: For symmetry.

fontir/src/ir.rs
rsheeter: Hm. Yeah, probably right. Just being lazy. Will fix.

fontir/src/ir.rs
rsheeter: Yes, it would indeed. Ty, will fix.

fontir/src/ir.rs
rsheeter: I was worried it would be noisy but having tried I think it's fine. Fixed.

fontir/src/ir.rs
rsheeter: Off the top of my head I wasn't sure what a qcurve was so I figured I'd leave it for a follow-on

rsheeter: https://github.com/googlefonts/fontmake-rs/issues/110

PR #107, opened 2023-01-23
Support tag on feature

rsheeter:


glyphs-reader/src/font.rs
```diff
@@ -92,6 +92,7 @@ struct RawFont {
 pub struct RawFeature {
     pub automatic: Option<i64>,
     pub name: Option<String>,
+    pub tag: Option<String>,
```
cmyr, line 95: Is *this* always an OpenType tag?

cmyr (approved):
One qq inline :)

glyphs-reader/src/font.rs
rsheeter: I can't think what else it would be but I don't have strong documentation to point to.

PR #108, opened 2023-01-23
Update path handling based on smoke test

rsheeter:
Try to avoid setting off debug_assert's in BezPath :)

I mistakenly assumed a BezPath with a move would not be empty but it (correctly) is. Correct our interrogation of the BezPath.

Smoke test UFOs love to open with a curve and close with dangling offcurve; support that. 

https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#point-types suggests more variants than I've supported thus far should work. qcurve is also unimplemented. Baby steps.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -11,6 +11,7 @@ use kurbo::{Affine, BezPath, Point};
 use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
+    fmt::Debug,
     path::{Path, PathBuf},
 };
 
@@ -133,10 +134,12 @@ pub struct Component {
 ///
 /// Source formats tend to use streams of point-of-type. Curve manipulation is
 /// often easier on bezier path, so provide a mechanism to convert.
+#[derive(Debug)]
 pub struct GlyphPathBuilder<'a> {
     glyph_name: &'a str,
     offcurve: Vec<Point>,
     path: BezPath,
+    last_move_to: Point,
 }
 
 impl<'a> GlyphPathBuilder<'a> {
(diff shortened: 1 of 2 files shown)


anthrotype:
> UFOs love to open with a curve and close with dangling offcurve

I'd say that's the 'canonical' way whereby UFO GLIF points are serialized, always starting with an oncurve

anthrotype:
why is this a separate PR and not folded into #104 instead? seems realtively small to be a commit, rather than standalone PR

rsheeter:
> I'd say that's the 'canonical' way whereby UFO GLIF points are serialized, always starting with an oncurve

Good to know. The spec offers no such guidance but reality agrees.

> why is this a separate PR

Purely because I sent the former for review first. I'll fold them by merging this into that (which is how it's setup anyway).

anthrotype:
> Purely because I sent the former for review first

I think you can just add commit to a PR that's already been reviewed, and notify the reviewers that you added something you want to be reviewed. I find confusing/distracting when when there are several PRs piling up on top of one another each one touching the same code; the reviewer has to figure out in which order they are related to one another and to the main branch, then may spend time reviewing some part of the code that will be rendered obsolete by some other follow-up PR that's waiting in the queue. I think it's just better to consolidate all PRs that touch upon the same files into one, but that's just my opinion.

rsheeter:
I agree re merging when it's related or the it's the same bit of code, as was the case here. If it wasn't so related then maybe not even if it was the same files?

The order is apparent from the branches so I'd have hoped that wasn't too time consuming.

anthrotype:
sounds good to me

PR #111, opened 2023-01-24
Fill in qcurve

### LINKED ISSUE #110: Implement GlyphPathBuilder::qcurve_to
...after finding out what exactly a qcurve is :)

https://github.com/googlefonts/fontmake-rs/pull/104/files#r1084564359

https://github.com/TypeNetwork/Josefinslab has lots of qcurve's.

https://stackoverflow.com/questions/20733790/truetype-fonts-glyph-are-made-of-quadratic-bezier-why-do-more-than-one-consecu clarifies, citing https://developer.apple.com/fonts/TrueType-Reference-Manual/RM01/Chap1.html#necessary.

rsheeter:
Fixes #110

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -199,12 +199,12 @@ impl GlyphPathBuilder {
         }
 
         // Insert an implied oncurve between every pair of offcurve points
-        for i in 0..(self.offcurve.len() - 1) {
+        for window in self.offcurve.windows(2) {
+            let curr = window[0];
+            let next = window[1];
             // current offcurve to halfway to the next one
-            let lhs = self.offcurve[i];
-            let rhs = self.offcurve[i + 1];
-            let on = Point::new((lhs.x + rhs.x) / 2.0, (lhs.y + rhs.y) / 2.0);
-            self.path.quad_to(lhs, on);
+            let implied = Point::new((curr.x + next.x) / 2.0, (curr.y + next.y) / 2.0);
+            self.path.quad_to(curr, implied);
         }
         // last but not least, the last offcurve to the provided point
         self.path.quad_to(*self.offcurve.last().unwrap(), p.into());
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -189,7 +189,28 @@ impl GlyphPathBuilder {
     }
 
     pub fn qcurve_to(&mut self, p: impl Into<Point>) -> Result<(), PathConversionError> {
-        todo!("Support qcurve to {}", p.into());
+        // https://github.com/googlefonts/fontmake-rs/issues/110
+        // Guard clauses: degenerate cases
+        if self.is_empty() {
+            return self.move_to(p);
+        }
+        if self.offcurve.is_empty() {
+            return self.line_to(p);
+        }
+
+        // Insert an implied oncurve between every pair of offcurve points
+        for i in 0..(self.offcurve.len() - 1) {
+            // current offcurve to halfway to the next one
+            let lhs = self.offcurve[i];
+            let rhs = self.offcurve[i + 1];
+            let on = Point::new((lhs.x + rhs.x) / 2.0, (lhs.y + rhs.y) / 2.0);
+            self.path.quad_to(lhs, on);
+        }
+        // last but not least, the last offcurve to the provided point
+        self.path.quad_to(*self.offcurve.last().unwrap(), p.into());
+        self.offcurve.clear();
+
+        Ok(())
     }
 
     /// Type of curve depends on accumulated off-curves
@@ -241,6 +262,8 @@ mod tests {
         ir::Axis,
     };
 
+    use super::GlyphPathBuilder;
+
     fn test_axis() -> Axis {
         let min = UserCoord::new(100.0);
         let default = UserCoord::new(400.0);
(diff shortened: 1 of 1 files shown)


anthrotype:
remember we discussed about defining an `ImpliedQuadTo` additional PathEl variant to describe TrueType-style quadratic bezier splines with implied on-curve points?
https://github.com/googlefonts/fontmake-rs/issues/33#issuecomment-1346877319

if we store them in the IR as kurbo BezPaths, where they become simply sequences of quadratic bezier curves in which the previously implicit on-curves are now explicit, then later on when we compile them into TrueType glyphs we will no longer be able to tell apart the cases whereby the source paths contained implicit vs explicit on-curve points.
Since for VF, the source paths for each glyph must be interpolation-compatible (same number of segments, same type of segments, same starting point, etc.), messing with the paths in this way could lead to incompatibilities when we need to build the gvar deltas.
I think we should try retain the information as to original truetype-style qcurves with their implied oncurves; when compiling VFs they would already be in the destination format (truetype glyf) and we would not need to guess whether an on-curve must be omitted because it can be implied.

cmyr:
I don't believe that kurbo should support implied offcurve points, which is a very font-specific concept. If we need to maintain this information we should use a custom representation.

However, thinking out loud: assuming we are *always* making implied offcurves explicit, and assuming that conversion is deterministic, shouldn't we always end up with interpolation-compatible outlines, assuming we *started* with them? As in, if there were two outlines that previously had implied off curves, and were interpolatable, then converting them both to have explicit offcurves should mean they are still interpolatable.

And unless I'm mistaken, the only reason to have implied offcurves at all is that it allows a more efficient representation of the of the glyph data, because you can elide some points. during compilation, I imagine we would want to be analyzing the curve and wherever possible converting segments to use implied off curves?

anthrotype:
> is a very font-specific concept

that's true, but I wasn't impling Kurbo should support, rather than we don't use kurbo BezPaths in our IR but keep paths as list of points instead and convert on-demand

> assuming we are always making implied offcurves explicit, and assuming that conversion is deterministic, shouldn't we always end up with interpolation-compatible outlines, assuming we started with them?

yes

> the only reason to have implied offcurves at all is that it allows a more efficient representation of the of the glyph data

correct

> because you can elide some points during compilation, I imagine we would want to be analyzing the curve and wherever possible converting segments to use implied off curves?

you should always think about not just a single path, but multiples interpolatable paths at different locations in a variable design space; eliding the points would have to be made across all glyph masters for them to continue to be interpolatable, which means they can be implied if they can be made so for all the masters; and being floats that midpoint-ness is always going to be an approximation within some tolerance.
Some sources already contains qcurves with implied oncurves; most of them only contain postscript cubic curves that are cu2qu'd in a way that produces such qcurves with implied oncurves; we already have this information (either from the input or after cu2qu), however we lose that when converting to bezpath and then we'd have to reconstruct that once again in order to build glyf and the gvar deltas.

rsheeter:
Let's pursue whether BezPath should be temporary or transient in #33. I think we will want a BezPath at some point, so we will need code to convert, so I'd like to merge qcurve if it looks sane?

fontir/src/ir.rs
```diff
@@ -189,7 +189,28 @@ impl GlyphPathBuilder {
...
+        }
+        if self.offcurve.is_empty() {
+            return self.line_to(p);
+        }
+
+        // Insert an implied oncurve between every pair of offcurve points
+        for i in 0..(self.offcurve.len() - 1) {
+            // current offcurve to halfway to the next one
+            let lhs = self.offcurve[i];
+            let rhs = self.offcurve[i + 1];
```
dfrg, line 205: The [`windows`](https://doc.rust-lang.org/std/primitive.slice.html#method.windows) method may be appropriate here. 

Patiently waiting for [`array_windows`](https://doc.rust-lang.org/std/primitive.slice.html#method.array_windows) to land in stable which would allow
````rust
for [lhs, rhs] in self.offcurve.array_windows() {
````

**Review by dfrg — APPROVED** [review]
+1 with small suggestion.

File: fontir/src/ir.rs [hunk 2]
```diff
@@ -189,7 +189,28 @@ impl GlyphPathBuilder {
...
+        // Guard clauses: degenerate cases
+        if self.is_empty() {
+            return self.move_to(p);
+        }
+        if self.offcurve.is_empty() {
+            return self.line_to(p);
+        }
+
+        // Insert an implied oncurve between every pair of offcurve points
+        for i in 0..(self.offcurve.len() - 1) {
```
cmyr, line 202: would `self.offcurve.windows(2)` be useful here?

fontir/src/ir.rs
```diff
@@ -189,7 +189,28 @@ impl GlyphPathBuilder {
...
+            // current offcurve to halfway to the next one
+            let lhs = self.offcurve[i];
+            let rhs = self.offcurve[i + 1];
+            let on = Point::new((lhs.x + rhs.x) / 2.0, (lhs.y + rhs.y) / 2.0);
+            self.path.quad_to(lhs, on);
+        }
+        // last but not least, the last offcurve to the provided point
+        self.path.quad_to(*self.offcurve.last().unwrap(), p.into());
+        self.offcurve.clear();
```
cmyr, line 211: I wonder if it might be helpful to write this whole thing as,


```rust
match &self.offcurve {
    [] => self.line_to(p),
    [one] => self.path.quad_to(one, p),
    multiple => { // your implied oncurve logic goes here
    },
}
self.offcurve.clear();
```

cmyr (approved):
modulo the outstanding question this looks okay to me!

fontir/src/ir.rs
rsheeter: I tried that - although I think I wrote it as match on .len, your way is nicer - and I felt it was less clear than having guards then the normal/expected case. Maybe that's just my pre-Rust habbits emerging though.

fontir/src/ir.rs
rsheeter: TIL, tha t seems handy indeed!

fontir/src/ir.rs
rsheeter: It would! So updated.

fontir/src/ir.rs
rsheeter: Per IM this is a matter of taste more than this being the one true Rust way so I will leave it for now.

PR #113, opened 2023-01-25
[glyphs-reader] Fix order of combined affine transform

anthrotype:
I was doing some experiments while fixing some unrelated bug in glyphsLib and noticed that the order of the combined affine transform is _not_ (assuming LTR order) rotate, then scale and finally translate as we first thought, but rather first scale, then rotate and finally translate...

Open the resources/testdata/glyphs3/Components.glyphs file in Glyphs.app and run this script in the Macro Window to confirm:

```python
from Cocoa import NSPoint
from math import isclose

def assert_transforms_almost_equal(t1, t2):
	assert all(isclose(v1, v2, rel_tol=1e-05) for v1, v2 in zip(t1, t2))

l = Glyphs.font.selectedLayers[0]
c = l.components[0]

c.scale = NSPoint(1.0303,2.0303)
c.rotation = 34.88155
c.position = NSPoint(-233,-129)

print(c.transform)

t1 = NSAffineTransform()
t1.shift((-233,-129))
t1.rotate(34.88155)
t1.scale((1.0303,2.0303))
print(t1.matrix)
assert_transforms_almost_equal(c.transform, t1.matrix)

t2 = NSAffineTransform()
t2.shift((-233,-129))
t2.scale((1.0303,2.0303))
t2.rotate(34.88155)
print(t2.matrix)
# this will raise AssertionError
assert_transforms_almost_equal(c.transform, t2.matrix)
```


rsheeter:
This is alarming because I created Component.glyphs *using* Glyphs.app and saving as Glyphs 2 and Glyphs 3. If that doesn't work ... what's up with that?

anthrotype:
picosvg also agrees

```python
>>> import math
>>> from picosvg.svg_transform import Affine2D
>>> s = Affine2D.identity().scale(1.0303,2.0303)
>>> r = Affine2D.identity().rotate(math.radians(34.88155))
>>> t = Affine2D.identity().translate(-233,-129)
>>> a = Affine2D.compose_ltr([s, r, t])
>>> a
Affine2D(a=0.8451922548777853, b=0.5892097608616178, c=-1.161091504879494, d=1.6655283267770236, e=-233.0, f=-129.0)
```

anthrotype:
> what's up with that?

your example has scale set to (2.0303,2.0303); i deliberately changed it to (1.0303,2.0303) so that scaleX and scaleY are different to trigger the issue

anthrotype (commented):
LGTM

rsheeter (approved):

PR #117, opened 2023-02-06
VariationModel in Rust, part 1

rsheeter:
Partial (as far as delta weights) https://github.com/fonttools/fonttools/blob/2f1f5e5e7be331d960a0e30d537c2b4c70d89285/Lib/fontTools/varLib/models.py in Rust. Meant to be used when compiling an IR Glyph (set of outlines at positions in designspace) to binary glyf + gvar.

Comments based on reading the code plus pestering @behdad. Anything helpful is probably his, the errors are mine :)

The three independent parallel vec's don't feel great to me, would like to think more about that. I'd like to land this very close to Python then refine from there.

Built on #118 to avoid noise from clippy; merge that first.

### DIFF
--- a/fontir/src/coords.rs
+++ b/fontir/src/coords.rs
@@ -40,7 +40,7 @@ pub struct NormalizedCoord(OrderedFloat<f32>);
 #[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
 pub struct Location<T>(BTreeMap<String, T>)
 where
-    T: Clone;
+    T: Copy;
 
 pub type DesignLocation = Location<DesignCoord>;
 pub type UserLocation = Location<UserCoord>;
@@ -184,13 +184,13 @@ impl Sub<NormalizedCoord> for NormalizedCoord {
     }
 }
 
-impl<T: Clone> FromIterator<(String, T)> for Location<T> {
+impl<T: Copy> FromIterator<(String, T)> for Location<T> {
     fn from_iter<I: IntoIterator<Item = (String, T)>>(iter: I) -> Self {
         Location(BTreeMap::from_iter(iter))
     }
 }
 
-impl<T: Clone> Location<T> {
+impl<T: Copy> Location<T> {
     pub fn new() -> Location<T> {
         Location(BTreeMap::new())
     }
@@ -213,11 +213,11 @@ impl<T: Clone> Location<T> {
     }
 
     pub fn get(&self, axis_name: &String) -> Option<T> {
-        self.0.get(axis_name).cloned()
+        self.0.get(axis_name).copied()
     }
 }
 
-impl<T: Clone> Default for Location<T> {
+impl<T: Copy> Default for Location<T> {
     fn default() -> Self {
         Self::new()
     }
--- a/fontir/src/coords.rs
+++ b/fontir/src/coords.rs
@@ -3,6 +3,8 @@
 //! See <https://github.com/googlefonts/fontmake-rs/blob/main/resources/text/units.md>
 
 use std::collections::{BTreeMap, HashMap};
+use std::fmt::{Debug, Write};
+use std::ops::Sub;
 
 use crate::serde::CoordConverterSerdeRepr;
 use crate::{ir::Axis, piecewise_linear_map::PiecewiseLinearMap};
@@ -26,15 +28,19 @@ pub struct UserCoord(OrderedFloat<f32>);
 /// Always in [-1, 1].
 ///
 /// Not typically used directly in sources.
-#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(
+    Serialize, Deserialize, Default, Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash,
+)]
 pub struct NormalizedCoord(OrderedFloat<f32>);
 
 /// A set of per-axis coordinates that define a specific location in a coordinate system.
 ///
 /// E.g. a user location is a `Location<UserCoord>`. Hashable so it can do things like be
 /// the key for a map of sources by location.
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
-pub struct Location<T>(BTreeMap<String, T>);
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
+pub struct Location<T>(BTreeMap<String, T>)
+where
+    T: Clone;
 
 pub type DesignLocation = Location<DesignCoord>;
 pub type UserLocation = Location<UserCoord>;
@@ -170,13 +176,21 @@ impl NormalizedCoord {
     }
 }
 
-impl<T> FromIterator<(String, T)> for Location<T> {
+impl Sub<NormalizedCoord> for NormalizedCoord {
+    type Output = NormalizedCoord;
+
+    fn sub(self, rhs: NormalizedCoord) -> Self::Output {
+        NormalizedCoord::new(self.0 - rhs.0)
+    }
+}
+
+impl<T: Clone> FromIterator<(String, T)> for Location<T> {
     fn from_iter<I: IntoIterator<Item = (String, T)>>(iter: I) -> Self {
         Location(BTreeMap::from_iter(iter))
     }
 }
 
-impl<T> Location<T> {
+impl<T: Clone> Location<T> {
     pub fn new() -> Location<T> {
         Location(BTreeMap::new())
     }
@@ -189,9 +203,21 @@ impl<T> Location<T> {
     pub fn iter(&self) -> impl Iterator<Item = (&String, &T)> {
         self.0.iter()
     }
+
+    pub fn axis_names(&self) -> impl Iterator<Item = &String> {
+        self.0.keys()
+    }
+
+    pub fn has(&self, axis_name: &String) -> bool {
+        self.0.contains_key(axis_name)
+    }
+
+    pub fn get(&self, axis_name: &String) -> Option<T> {
+        self.0.get(axis_name).cloned()
+    }
 }
 
-impl<T> Default for Location<T> {
+impl<T: Clone> Default for Location<T> {
     fn default() -> Self {
         Self::new()
     }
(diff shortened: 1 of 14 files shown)


fontir/src/variations.rs
```diff
@@ -0,0 +1,632 @@
...
+impl VariationRegion {
+    fn new() -> Self {
+        VariationRegion(HashMap::new())
+    }
+}
+
+impl VariationRegion {
+    /// The scalar multiplier for the provided location for this region
+    ///
+    /// Returns None for no influence, Some(non-zero value) for influence.
```
anthrotype, line 264: why return an Option? I think supportScalar always returns a float between 0.0 (when location falls outside the tent) and +1.0 (at least with extrapolate=False); plus assumes per-axis scalar of 1.0 when peak is at 0.0 (ot=True), as in "axis does not participate".
I don't see what the None buys us

fontir/src/variations.rs
```diff
@@ -0,0 +1,632 @@
...
+    }
+
+    /// Python has several examples that are really just checking the width peak 0 does nothing
+    /// >>> supportScalar({'wght':2.5, 'wdth':0}, {'wght':(0,2,4), 'wdth':(-1,0,+1)})
+    /// 0.75
+    /// >>> supportScalar({'wght':2.5, 'wdth':.5}, {'wght':(0,2,4), 'wdth':(-1,0,+1)})
+    /// 0.75
+    #[test]
+    fn zero_peak_means_no_impact() {
+        assert_eq!(false, tent(-1.0, 0.0, 1.0).validate());
```
anthrotype, line 581: i'm not sure if "invalid" is the right word for this.. a peak at 0.0 simply means that the axis associated with the given tent does not participate in the delta interpolation in that particular region, i.e. per-axis support scalar is 1.0 so it has no effect when multiplied with other scalars or deltas.

fontir/src/variations.rs
```diff
@@ -0,0 +1,632 @@
...
+}
+
+impl VariationRegion {
+    /// The scalar multiplier for the provided location for this region
+    ///
+    /// Returns None for no influence, Some(non-zero value) for influence.
+    ///
+    /// In Python, supportScalar. We only implement the ot=True, extrapolate=False paths.
+    /// <https://github.com/fonttools/fonttools/blob/2f1f5e5e7be331d960a0e30d537c2b4c70d89285/Lib/fontTools/varLib/models.py#L123>.
+    fn multiplier_at(&self, location: &NormalizedLocation) -> Option<OrderedFloat<f32>> {
```
anthrotype, line 268: everybody else (including OT spec) calls this a 'scalar', but I presume that 'multiplier' reads clearer to you.

fontir/src/variations.rs
anthrotype: ok, I see, you use None to mean scalar==0.0 so you can use Option methods like map() and similar instead of explicitly testing if scalar == 0.0

anthrotype: on the other end, when you are actually _scaling_ the deltas, if you have something that's unconditionally a float then you can simply multiply that, and if scalar is 0, the delta is zeroed accordingly.. whereas with this Option you'd have to branch.

fontir/src/variations.rs
rsheeter: In FontTools non-OT tents are supported and tents like this can make sense. To my understanding in OT this form of tent is correctly described as invalid, it doesn't make sense and the code filters them out when they are encountered in OT mode. In Rust I only support OT mode so they are always invalid.

Or ... did I simply misunderstand the python?

fontir/src/variations.rs
rsheeter: I thought it would make the code simpler but I don't feel tremendously strongly about it, we could just return 0.

fontir/src/variations.rs
rsheeter: I think that's just an error, I should have called it scalar_at. The key change I wanted to make was to move the method onto the region.

fontir/src/variations.rs
rsheeter: It now returns 0

fontir/src/variations.rs
anthrotype: honestly I have never seen supportScalar being used anywhere with anything other than ot=True default...

> To my understanding in OT this form of tent is correctly described as invalid

hm I'd argue that a peak-0 axis tent is the _default_.. In fonttools code for gvar's TupleVariations, a tuple's region is represented by a dict keyed by axis tag, with values being (min,peak,max) tent triplets; when an axis doesn't participate within that region (i.e. peak == 0) then it is omitted from the dict; when compiling, the omitted axis tent is assumed to default to (0.0, 0.0, 0.0).

actually, one can potentially define a region where _all_ the axis tents peak at 0, and that basically means "please always apply my deltas at full scale (scalar=1.0) no matter what instance location"... And to my understanding that is not invalid. It's just the way things works in OTvar.

Maybe rename `Tent::validate` method to something like `does_participate` or something

rsheeter: The argument for invalid is that it appears to me it will *never* participate, it will just *always* be filtered out. If it had influence on some regions and not others then does_participate makes more sense to me, but perhaps I misunderstand.

rsheeter: See in particular the `if ot:` series of continue's at https://github.com/fonttools/fonttools/blob/2f1f5e5e7be331d960a0e30d537c2b4c70d89285/Lib/fontTools/varLib/models.py#L160. Tents that are so formed will *never* have influence. My argument is that this makes them illformed or invalid, whatever the user thought their sources would achieve probably isn't being achieved.

Is that fair or not so much?

rsheeter: Hm. In debugging I think it may be that taking (0,0,0) to mean 1.0 might actually simplify things. Let me play with that more.

fontir/src/variations.rs
```diff
@@ -0,0 +1,795 @@
...
+                }
+
+                non_zero_axes.push(axis_name);
+            }
+            // score -1 for every axis position that matches a location that was directly on an axis
+            on_axis_points += self
+                .on_axis_points
+                .get(axis_name)
+                .map(|on_axis| on_axis.get(pos).map_or(0, |_| -1))
+                .unwrap_or_default();
```
anthrotype, line 185: honestly this chain of method calls looks a bit dense/convoluted.. you want to count the number of on-axis points for the given location, and negate that number so the sort is decreasing (`-len(onPointAxes)`). Here you decrease a counter -1 for each value that is an on-axis point, am I right?

EDIT: Ok nevermind, I read the docs, over time I will get more familiar with all these methods. All good, nothing to do here.

fontir/src/variations.rs
```diff
@@ -0,0 +1,795 @@
...
+    fn norm_loc(positions: &[(&str, f32)]) -> NormalizedLocation {
+        let mut loc = NormalizedLocation::new();
+        for (axis, value) in positions {
+            loc.set_pos(axis.to_owned(), NormalizedCoord::new(*value));
+        }
+        loc
+    }
+
+    fn tent(lower: f32, peak: f32, upper: f32) -> Tent {
+        Tent {
```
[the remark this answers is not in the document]
anthrotype: oh there's one already, maybe you can change it to take `Into<NormalizedCoord>` params

or maybe not

fontir/src/variations.rs
```diff
@@ -0,0 +1,795 @@
...
+        if peak > zero {
+            lower = zero;
+        } else {
+            upper = zero;
+        }
+        Tent { lower, peak, upper }
+    }
+}
+
+impl Tent {
```
anthrotype, line 341: nit: this `impl Tent {` block can be joined with the one immediately preceding, if I'm not mistaken (which is why the first time I somehow got confused and missed the `Tent::new` implementation)

fontir/src/variations.rs
```diff
@@ -0,0 +1,795 @@
...
+    let mut influence = Vec::new();
+    for (i, region) in regions.iter().enumerate() {
+        let axes: HashSet<&String> = region.0.keys().collect();
+        let mut region = region.clone();
+        for prev_region in regions[..i].iter() {
+            if axes != prev_region.0.keys().collect() {
+                continue;
+            }
+            // If prev doesn't overlap current we aren't interested
+            let overlap = region.0.iter().all(|(axis, axis_region)| {
```
anthrotype, line 438:
```diff
- let overlap = region.0.iter().all(|(axis, axis_region)| {
+ let overlap = region.0.iter().all(|(axis, axis_tent)| {
```

I would reserve the term "region" for the multi-axis `Dict[AxisTag, Tuple[float, float, float]]` only, and prefer "tent" for the latter (lower, peak, upper) triplet that's associated with a given axis.
On the other hand, one 'tent' could still be seen as a 'region' with only one axis..

anthrotype (approved):
Good job! I left some minor comments below

rsheeter:
Note to self: https://github.com/fonttools/fonttools/blob/main/Tests/varLib/models_test.py has tests, should implement some or all

fontir/src/coords.rs
```diff
@@ -26,15 +28,19 @@ pub struct UserCoord(OrderedFloat<f32>);
...
 /// A set of per-axis coordinates that define a specific location in a coordinate system.
 ///
 /// E.g. a user location is a `Location<UserCoord>`. Hashable so it can do things like be
 /// the key for a map of sources by location.
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
-pub struct Location<T>(BTreeMap<String, T>);
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
+pub struct Location<T>(BTreeMap<String, T>)
+where
+    T: Clone;
```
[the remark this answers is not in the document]
rsheeter: It was indeed because I wanted cloned. I agree copied is preferable.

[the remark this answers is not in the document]
rsheeter: Fixed, ty

fontir/src/variations.rs
```diff
@@ -0,0 +1,795 @@
...
+            influence: Vec::new(),
+            delta_weights: Vec::new(),
+        }
+    }
+}
+
+/// Gryffindor!
+///
+/// Sorts locations, and thus the resulting regions, based on the intuition that from most
+/// to least influential are the default master, then corner masters, and finally
```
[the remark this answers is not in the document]
rsheeter: In my head an on-axis master often *is* a corner, e.g. in normalized terms the masters at wght -1 and wght 1 are probably on-axis corners.

fontir/src/variations.rs
```diff
@@ -0,0 +1,795 @@
...
+                ordered_axes.push(axis_name.clone());
+            }
+        }
+
+        for (axis_name, pos) in location.iter() {
+            if pos.into_inner() != ZERO {
+                // Rank is the # of axes with non-zero values
+                rank += 1;
+
+                // Score more than the max allowable axes for any axis with non-zero value that has no user-specified order
```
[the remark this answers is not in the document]
rsheeter: Hm, indeed, that's only valid in Python. I don't want an error type for key_for so I just added a comment.

fontir/src/variations.rs
rsheeter: I have a habbit of keeping methods that take self separate of methods that don't but there is no real reason for that. Updated.

fontir/src/variations.rs
rsheeter: Ew good catch, that's a terrible terrible naming choice. I updated it to axis_name, tent.

fontir/src/variations.rs
```diff
@@ -0,0 +1,795 @@
...
+    use ordered_float::OrderedFloat;
+
+    use crate::{
+        coords::{NormalizedCoord, NormalizedLocation},
+        variations::{Tent, ONE},
+    };
+
+    use super::{VariationModel, VariationRegion};
+
+    fn norm_loc(positions: &[(&str, f32)]) -> NormalizedLocation {
```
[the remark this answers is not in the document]
rsheeter: I don't think the macro would save anything over the actual code in this case?

Updated to iter/map/collect, that's much nicer.

fontir/src/variations.rs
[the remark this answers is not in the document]
rsheeter: I deliberately didn't implement Into on the coord types because I wanted the conversion to be deliberate, not just an accidental .into that made say a float that was one type into some other.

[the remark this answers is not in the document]
rsheeter: Hm. `From<(f32,f32,f32)>` for Tent might make sense though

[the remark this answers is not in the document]
rsheeter: That feels more idiomatic, good catch. Fixed.

rsheeter:
Per @behdad advice, ported more of models_test.py. I learned I'm bad at transcribing #s.

PR #119, opened 2023-02-09
Store a VariationModel in StaticMetadata

rsheeter:
Building on #117, construct and retain a variation model along with static metadata. For example, Lexend produces

```yml
glyph_locations:
- Hyper Expansion: 0.0
  Weight: 0.0
- Hyper Expansion: 0.0
  Weight: 0.14432989
- Hyper Expansion: 0.0
  Weight: 0.44329897
- Hyper Expansion: 0.0
  Weight: 1.0
- Hyper Expansion: 1.0
  Weight: 0.0
- Hyper Expansion: 1.0
  Weight: 0.44329897
- Hyper Expansion: 1.0
  Weight: 1.0
```

### DIFF
--- a/fontir/src/error.rs
+++ b/fontir/src/error.rs
@@ -99,6 +99,8 @@ pub enum WorkError {
     },
     #[error("Path conversion error")]
     PathConversionError(#[from] PathConversionError),
+    #[error("Variation model error")]
+    VariationModelError(#[from] VariationModelError),
 }
 
 /// An async work error, hence one that must be Send
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -23,8 +23,13 @@ use std::{
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 #[serde(from = "StaticMetadataSerdeRepr", into = "StaticMetadataSerdeRepr")]
 pub struct StaticMetadata {
+    /// Every axis used by the font being compiled
     pub axes: Vec<Axis>,
+    /// The name of every glyph, in the order it will be emitted
+    ///
+    /// <https://rsheeter.github.io/font101/#glyph-ids-and-the-cmap-table>
     pub glyph_order: IndexSet<GlyphName>,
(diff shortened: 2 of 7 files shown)


anthrotype:
hm.. you are assuming that only one variation model can be applied for the entire glyph set, but that's not always necessarily the case. Well, it is in simpler "multiple-masters" setup where each "master" source contains the _entire_ set of glyphs (or if you prefer, all glyphs contain the same number of masters); but in OTVar each glyph can define its own variation sub-model; and the glyph sources can be "sparse", e.g. some glyphs may define additional intermediate masters (the so-called "brace" layers in Glyphs.app); or in a master-less, variable-first design, each glyphs may define sources at various locations that don't necessarily overlap across the entire glyph set (e.g. glyph "foo" may not have a Bold, or "bar" not vary at all, etc.).

The FontTools VariationModel class has a `getSubModel` method that will create a new VariationModel from a list of master values that may contain `None` (indicating the absence of a master at that index), and it also caches it so that it gets reused if a similar master configuration is 

https://github.com/fonttools/fonttools/blob/bf265ce49e0cae6f032420a4c80c31d8e16285b8/Lib/fontTools/varLib/models.py#L273-L281

here's it used in `_add_gvar`:

https://github.com/fonttools/fonttools/blob/bf265ce49e0cae6f032420a4c80c31d8e16285b8/Lib/fontTools/varLib/__init__.py#L264-L272
 
I think we can do something similar, maybe store a "master" variation model that is global and contains the union of all the normalized locations and then each glyph can cut its own sub-model.

rsheeter:
I had assumed, though evidently not stated, that if I had the VariationModel for all locations that can have at least one glyph instance that I could then derive the submodel for any given glyph when compiling it, and therefore would fill in submodel later. Is that incorrect?

anthrotype:
yes, that's correct! In fact that's what varLib itself does.

anthrotype (approved):

rsheeter:
I updated the comment on StaticMetadata to better explain the intent to compute narrowings from the global copy.

PR #120, opened 2023-02-10
Composite glyphs  (static glyf, 3 of 3)

rsheeter:
Build composite glyphs, mk1. Merge binary glyphs into things that look like `glyf` and `loca` and confirm fontations can read 'em back.


rsheeter:
Closing to reopen as a PR that numbers nicely

PR #126, opened 2023-03-03
Update kurbo

rsheeter:
We need https://github.com/linebender/kurbo/pull/236 for glyph compilation. Split out because the svg format change makes the diff noisy.


dfrg (approved):
kurbo 0.9.1 is now published!

PR #127, opened 2023-03-03
Add the ability to filter glyphs to be processed

rsheeter:
Very handy when trying to track down issues with specific glyphs.

Skip features when a glyph filter is active.


fontc/src/main.rs
```diff
@@ -390,10 +402,28 @@ impl ChangeDetector {
...
 
+        let glyph_name_filter = config
+            .args
+            .glyph_name_filter
+            .map(|raw_filter| Regex::new(&raw_filter))
+            .transpose()
+            .map_err(Error::BadRegex)?;
+
+        if let Some(regex) = &glyph_name_filter {
+            current_inputs.glyphs.retain(|glyph_name, _| {
```
dfrg, line 416: Per im, feature compilating requires the full glyph list. Assuming we want to eventually support filtering in combination with features (compile time subsetting?) it might be worth thinking about how we retain the original glyph order and map between that and the filtered set.

dfrg (approved):
Lgtm!

fontc/src/main.rs
rsheeter: We can store it on Context if it turns out we need it.

PR #131, opened 2023-03-10
Build simple glyphs

rsheeter:
Split out from #120 to try to make review more tractable.


fontbe/src/error.rs
```diff
@@ -11,4 +13,45 @@ pub enum Error {
     FeaAssembleError(#[from] BinaryCompilationError),
     #[error("Fea compilation failure")]
     FeaCompileError(#[from] CompilerError),
+    #[error("'{0}' {1}")]
+    GlyphError(GlyphName, GlyphProblem),
+    #[error("'{0}' {1:?} {2}")]
+    KurboError(GlyphName, BadKurbo, String),
```
cmyr, line 19: My personal rule of thumb is I only use tuple-structs if there is a single member, and I would definitely switch to named fields when there are 3+ members.

fontc/Cargo.toml
```diff
@@ -37,3 +37,5 @@ indexmap = "1.9.2"
 diff = "0.1.12"
 ansi_term = "0.12.1"
 tempfile = "3.3.0"
+
```
cmyr, line 40: intentional?

cmyr (approved):
I feel my brain has largely departed me for the day, but nothing here jumps out as egregious 👍

fontbe/src/error.rs
rsheeter: Oh good catch, will fix. My rule of thumb is if it's not obvious from the types and that String is not obvious. If it was GlyphName BadKurbo SomeInformativeTYpe I'd feel better.

fontc/Cargo.toml
rsheeter: I don't _think_ so; fixed

rsheeter:
@anthrotype although now merged I would still love your feedback on this

fontbe/src/glyphs.rs
```diff
@@ -0,0 +1,280 @@
...
+                        p1: *p1,
+                        p2: *p2,
+                        p3: *p3,
+                    },
+                    _ => unreachable!("'{name}': *must* all be cubic"),
+                })
+                .collect();
+
+            // At long last, actually convert something to quadratic
+            // TODO what should we pass for accuracy
```
anthrotype, line 117: python cu2qu defaults to a _relative_ max_error of 1/1000th of the font's UPEM; e.g. for a usual font with head.unitsPerEm=1000, the _absolute_ max_error is 1 font unit.

anthrotype (commented):
LGTM!

fontc/src/main.rs
```diff
@@ -527,7 +579,10 @@ impl Workload {
 
                 // Block for things to phone home to say they are done
                 // Then complete everything that has reported since our last check
-                self.read_completions(&recv, RecvType::Blocking)?;
+                let successes = self.read_completions(&recv, RecvType::Blocking)?;
+                successes
+                    .into_iter()
+                    .for_each(|s| self.handle_success(&fe_root, s));
```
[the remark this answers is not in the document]
cmyr: I actually agree with you, I think this is confusing, and maybe doing too much.

PR #132, opened 2023-03-10
Composite glyphs (static glyf, 3 of 3)

rsheeter:
Build composite glyphs, mk1. Merge binary glyphs into things that look like glyf and loca and confirm fontations can read 'em back.

Last part of splitting up #120

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -448,6 +448,7 @@ fn compute_composite_bboxes(context: &Context) -> Result<(), Error> {
             let mut missing_boxes = false;
             let boxes: Vec<Bbox> = composite
                 .components()
+                .iter()
                 .filter_map(|c| {
                     if missing_boxes {
                         return None; // can't succeed
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -1,17 +1,24 @@
 //! 'glyf' Glyph binary compilation
 
-use std::collections::{BTreeSet, HashMap, HashSet};
+use std::{
+    cmp,
+    collections::{BTreeSet, HashMap, HashSet},
+};
 
 use fontdrasil::{orchestration::Work, types::GlyphName};
 use fontir::{coords::NormalizedLocation, ir};
-use kurbo::{cubics_to_quadratic_splines, BezPath, CubicBez, PathEl};
-use log::{error, trace, warn};
+use kurbo::{cubics_to_quadratic_splines, Affine, BezPath, CubicBez, PathEl, Rect};
+use log::{trace, warn};
 
-use write_fonts::tables::glyf::SimpleGlyph;
+use read_fonts::{
+    tables::glyf::{self, Anchor, Transform},
+    types::{F2Dot14, GlyphId},
+};
+use write_fonts::tables::glyf::{Bbox, Component, ComponentFlags, CompositeGlyph, SimpleGlyph};
 
 use crate::{
     error::{Error, GlyphProblem},
-    orchestration::{BeWork, Context},
+    orchestration::{BeWork, Context, Glyph},
 };
 
 struct GlyphWork {
@@ -20,6 +27,86 @@ struct GlyphWork {
 
 pub fn create_glyph_work(glyph_name: GlyphName) -> Box<BeWork> {
     Box::new(GlyphWork { glyph_name })
+}
+
+fn create_component(
+    context: &Context,
+    ref_glyph_name: &GlyphName,
+    transform: &Affine,
+) -> Result<(Component, Bbox), GlyphProblem> {
+    // Obtain glyph id from static metadata
+    let gid = context
+        .ir
+        .get_static_metadata()
+        .glyph_id(ref_glyph_name)
+        .ok_or(GlyphProblem::NotInGlyphOrder)?;
+    let gid = GlyphId::new(gid as u16);
+
+    // No known source does point anchoring so we just our transform into a 2x2 + offset
+    let [a, b, c, d, e, f] = transform.as_coeffs();
+    let component = Component::new(
+        gid,
+        Anchor::Offset {
+            x: e as i16,
+            y: f as i16,
+        },
+        Transform {
+            xx: F2Dot14::from_f32(a as f32),
+            yx: F2Dot14::from_f32(b as f32),
+            xy: F2Dot14::from_f32(c as f32),
+            yy: F2Dot14::from_f32(d as f32),
+        },
+        ComponentFlags::default(),
+    );
+
+    // Bbox computation is postponed to glyph merge to ensure all glyphs are available to query
+    Ok((component, Bbox::default()))
+}
+
+fn create_composite(
+    context: &Context,
+    glyph_name: &GlyphName,
+    default_location: &NormalizedLocation,
+    components: &HashMap<(GlyphName, NormalizedLocation), Affine>,
+) -> Result<CompositeGlyph, Error> {
+    let mut errors = vec![];
+    let components_at_default = components
+        .iter()
+        .filter_map(|((ref_glyph_name, loc), transform)| {
+            if default_location == loc {
+                Some((ref_glyph_name, transform))
+            } else {
+                None
+            }
+        })
+        .filter_map(|(ref_glyph_name, transform)| {
+            create_component(context, ref_glyph_name, transform)
+                .map_err(|problem| {
+                    errors.push(Error::ComponentError {
+                        glyph: glyph_name.clone(),
+                        referenced_glyph: ref_glyph_name.clone(),
+                        problem,
+                    })
+                })
+                .ok()
+        });
+
+    let composite = CompositeGlyph::try_from_iter(components_at_default)
+        .map_err(|_| {
+            errors.push(Error::GlyphError(
+                glyph_name.clone(),
+                GlyphProblem::NoComponents,
+            ))
+        })
+        .ok();
+
+    if !errors.is_empty() {
+        return Err(Error::ComponentErrors {
+            glyph: glyph_name.clone(),
+            errors,
+        });
+    }
+    Ok(composite.unwrap())
 }
 
 impl Work<Context, Error> for GlyphWork {
@@ -38,8 +125,9 @@ impl Work<Context, Error> for GlyphWork {
         // TODO refine (submodel) var model if glyph locations is a subset of var model locations
 
         match glyph {
-            CheckedGlyph::Composite { name, .. } => {
-                error!("setting no glyph for {name}; composites not implemented yet");
+            CheckedGlyph::Composite { name, components } => {
+                let composite = create_composite(context, &name, default_location, &components)?;
+                context.set_glyph(name, composite.into());
             }
             CheckedGlyph::Contour { name, contours } => {
                 // Draw the default outline of our simple glyph
@@ -176,6 +264,7 @@ fn cubics_to_quadratics(glyph: CheckedGlyph) -> CheckedGlyph {
 enum CheckedGlyph {
     Composite {
         name: GlyphName,
+        components: HashMap<(GlyphName, NormalizedLocation), Affine>,
     },
     Contour {
         name: GlyphName,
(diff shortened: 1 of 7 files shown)


fontbe/src/paths.rs
```diff
@@ -45,7 +45,8 @@ impl Paths {
         match id {
             WorkId::Features => self.build_dir.join("features.ttf"),
             WorkId::Glyph(name) => self.glyph_file(name.as_str()),
-            WorkId::GlyphMerge => self.build_dir.join("glyf.ttf"),
+            WorkId::GlyphMerge(GlyphMerge::Glyf) => self.build_dir.join("glyf.ttf"),
```
anthrotype, line 48: I would not call the built file for glyf table "glyf.ttf" because that suggests it is a semi-complete font with sfnt structure, whereas in fact this only contains the blob for one (glyf) table, without sfnt header nor the other tables which would be required to parse glyf (i.e. maxp and loca at the very least).
On the other hand, the "features.ttf" file that fea-rs produces _can_ be decompiled with fontTools/ttx, because in addition to GPOS/GSUB it has a maxp).

Maybe since glyf and loca always go together (they are really _one_ table), you could dump both in the same glyf.ttf file, add an sfnt header to locate them, and also add a minimal maxp table just to know how many glyphs are in loca and what loca format it is. This would be enough for fontTools to able to parse, and can be handy if one wants to quickly inspect the build byproducts (`ttx build/glyf.ttf` currently gives "Unknown file type: 'build/glyf.ttf'")

Alternatively if you think it's preferable to serialize the two to different files, give a different file extension to glyf so that it's clear that it's not a font, not even a partial one, but just one table (glyf.table or glyf.sfnt?)

fontbe/src/glyphs.rs
```diff
@@ -278,3 +377,170 @@ fn path_el_type(el: &PathEl) -> &'static str {
...
+        // This isn't overly memory efficient but ... fonts aren't *that* big (yet?)
+        let mut loca = vec![0];
+        let mut glyf: Vec<u8> = Vec::new();
+        glyf.reserve(1024 * 1024);
+        glyph_order
+            .iter()
+            .map(|gn| context.get_glyph(gn))
+            .for_each(|g| {
+                let bytes = g.to_bytes();
+                loca.push(loca.last().unwrap() + bytes.len() as u32);
```
anthrotype, line 538: as you know, `loca` has two version a "short" and a "long" one (with head.indexToLocFormat storing the loca format), you seem to unconditionally opt for the latter; at least add a TODO so we can later on implement the "short" variant

fontbe/src/glyphs.rs
```diff
@@ -22,6 +29,86 @@ pub fn create_glyph_work(glyph_name: GlyphName) -> Box<BeWork> {
...
+
+    // Bbox computation is postponed to glyph merge to ensure all glyphs are available to query
+    Ok((component, Bbox::default()))
+}
+
+fn create_composite(
+    context: &Context,
+    glyph_name: &GlyphName,
+    default_location: &NormalizedLocation,
+    components: &HashMap<(GlyphName, NormalizedLocation), Affine>,
```
anthrotype, line 70: why is this `components` an HashMap and not just a Vec of tuples? It is entirely possible that for a given location, more than one component with the same `ref_glyph_name` exist (they'd likely have a different affine, otherwise they'd exactly overlap one another, which is absurd though still valid)

fontbe/src/glyphs.rs
```diff
@@ -22,6 +29,86 @@ pub fn create_glyph_work(glyph_name: GlyphName) -> Box<BeWork> {
...
+            x: e as i16,
+            y: f as i16,
+        },
+        Transform {
+            xx: F2Dot14::from_f32(a as f32),
+            yx: F2Dot14::from_f32(b as f32),
+            xy: F2Dot14::from_f32(c as f32),
+            yy: F2Dot14::from_f32(d as f32),
+        },
+        ComponentFlags::default(),
```
anthrotype, line 59: FWIW, ufo2ft defaults to 0x4 ROUND_XY_TO_GRID. It also sets USE_MY_METRICS to the first component whose base glyph's advance width is equal to the composite glyph's advance width.

anthrotype: also note, support for additional glyf flags set from UFO public.truetype.* lib keys will be added with https://github.com/googlefonts/ufo2ft/pull/654

fontbe/src/glyphs.rs
```diff
@@ -176,6 +264,7 @@ fn cubics_to_quadratics(glyph: CheckedGlyph) -> CheckedGlyph {
 enum CheckedGlyph {
     Composite {
         name: GlyphName,
+        components: HashMap<(GlyphName, NormalizedLocation), Affine>,
```
anthrotype, line 267: should make this a Vec -- otherwise you're only allowing one component with unique glyph name per composite glyph at given location.

E.g. think about glyph for ELLIPSIS, it may well have three components named "period" ...

fontbe/src/glyphs.rs
```diff
@@ -278,3 +377,170 @@ fn path_el_type(el: &PathEl) -> &'static str {
...
+                            })
+                    });
+                    if bbox.is_none() {
+                        trace!("Can't compute bbox for {glyph_name} because bbox for {ref_glyph_name} isn't ready yet");
+                        missing_boxes = true;
+                        return None; // maybe next time?
+                    };
+
+                    // The transform we get here has changed because it got turned into F2Dot14 and i16 parts
+                    // We could go get the "real" transform from IR...?
```
anthrotype, line 473: I think we should...

fontc/src/main.rs
```diff
@@ -1092,36 +1195,129 @@ mod tests {
...
+    #[test]
+    fn compile_composite_glyphs_to_glyf_loca() {
+        let temp_dir = tempdir().unwrap();
+        let build_dir = temp_dir.path();
+        let result = compile(test_args(build_dir, "glyphs2/Component.glyphs"));
+        let (raw_glyf, raw_loca) = result.raw_glyf_loca();
+
+        // non-uniform scaling of period
+        let period_idx = result.get_glyph_index("period");
+        let comma_idx = result.get_glyph_index("non_uniform_scale");
```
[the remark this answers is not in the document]
rsheeter: Err ... it sure doesn't look like it does it!

fontbe/src/glyphs.rs
```diff
@@ -278,3 +377,170 @@ fn path_el_type(el: &PathEl) -> &'static str {
...
+        compute_composite_bboxes(context)?;
+
+        let static_metadata = context.ir.get_static_metadata();
+        let glyph_order = &static_metadata.glyph_order;
+
+        // Glue together glyf and loca
+        // This isn't overly memory efficient but ... fonts aren't *that* big (yet?)
+        let mut loca = vec![0];
+        let mut glyf: Vec<u8> = Vec::new();
+        glyf.reserve(1024 * 1024);
```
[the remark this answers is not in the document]
rsheeter: It doesn't matter because the vec will resize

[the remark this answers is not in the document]
rsheeter: I'll add a comment

fontc/src/main.rs
```diff
@@ -396,6 +396,52 @@ fn add_glyph_ir_jobs(
...
+        let mut dependencies: HashSet<_> = glyphs_changed
+            .iter()
+            .map(|gn| BeWorkIdentifier::Glyph(gn.clone()).into())
+            .collect();
+        dependencies.insert(FeWorkIdentifier::FinalizeStaticMetadata.into());
+
+        let id: AnyWorkId = BeWorkIdentifier::GlyphMerge(GlyphMerge::Glyf).into();
+        // Write the merged glyphs and write individual glyphs that are updated, such as composites with bboxes
+        let write_access: AccessFn<_> = Arc::new(|id| {
+            matches!(
```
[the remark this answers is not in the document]
rsheeter: oh lol that is slightly unfortunate. It's not the same id but it does look an awful lot like it. I think perhaps if I move the let id down that would be more obvious.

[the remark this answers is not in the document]
rsheeter: Fixed.

fontbe/src/glyphs.rs
rsheeter: Oh that's an interesting point. Maybe I should have a set of (name, loc, affine).

fontbe/src/glyphs.rs
rsheeter: I know :) I should have put a todo, adding now.

rsheeter: Oh sorry, that's entirely intentional, the possible shrinking of loca is supposed to occur during the final merge. Clearly need a comment.

fontbe/src/glyphs.rs
rsheeter: I think I'd like to make that a followup PR, filed as https://github.com/googlefonts/fontmake-rs/issues/135

fontbe/src/glyphs.rs
rsheeter: https://github.com/googlefonts/fontmake-rs/issues/136 for the new flags

fontbe/src/glyphs.rs
rsheeter: I will add a test case that shows off the bug here

rsheeter: Got a lovely test case where a component goes missing now :)

anthrotype: A set would dedup but would lose the order. I think we also want to keep the arbitrary user-defined ordering components within a given composite glyph, hence why I suggested Vec. The order of the components defines the the control points indices in a composite glyph, that can be referenced from truetype hinting instructions (we don't support them now, but you never know).
Also I mean, if one wants to shoot themselves in the feet by defining the exact same component twice (never ever seen that in the wild by the way) why should we not let them, garbage in, garbage out.

fontc/src/main.rs
```diff
@@ -1092,36 +1195,129 @@ mod tests {
...
+                yx: F2Dot14::from_f32(0.0),
+                xy: F2Dot14::from_f32(0.0),
+                yy: F2Dot14::from_f32(1.5)
+            },
+            component.transform
+        );
+
+        // Have ye bbox?
+        // Original period bbox is (250,50) to (375,100), then transform above is applied
+        // Result *should* be [550, 125, 800, 200] but F2Dot14/i16 representation changes it
```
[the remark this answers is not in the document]
rsheeter: Hm, I think you are right but I want to do it later :) https://github.com/googlefonts/fontmake-rs/issues/135

fontbe/src/glyphs.rs
anthrotype: I think you should not prune duplicate components, that's not currently happening in the python pipeline and may be unexpected; also it is conceivable to devise a use case for that; e.g. two components that are on top of one another in some but not all masters; or they are shifted only by truetype instructions. In general, let's try to keep faithful to the sources whenever we can.

rsheeter: In coding it up I concluded the same, plus I believe it might in future - boring expansion - be valid to have overlapping components at one normalized location but not another.

anthrotype: if by irrelevant you mean they would display the same because fill rules don't care about the relative ordering of various subpaths, then yes -- at least, provided the composite glyph is unhinted or contain no TT glyph instruction directly referencing point indices, in which case shuffling components around would break those TT instructions

rsheeter: Fixed, is now a vec

anthrotype: plus, both our source formats define glyph components as lists with a user-defined ordering, it makes more sense to just record that order in a Vec and keep that in glyf composites, as expected by user.

fontbe/src/glyphs.rs
rsheeter: Added a test. I even used multiple period. Now a vec. Fixed.

fontbe/src/glyphs.rs
rsheeter: I set round xy to grid.  I haven't done anything with metrics yet so I've pushed that to https://github.com/googlefonts/fontmake-rs/issues/137. I should probably generate an hmtx at some point I suppose.

anthrotype (approved):
LGTM

PR #134, opened 2023-03-12
Build a simple cmap

rsheeter:
Add a job to generate cmap and wire it through to work in simple cases. Demonstrate it produces the expected glyph order and unicode assignments.

Contributes to #138.

Requires https://github.com/googlefonts/fontations/pull/275.


rsheeter:
Per f2f discussion make a high level API on write-fonts and pass something akin to a map codepoint:gid in. Ref https://github.com/googlefonts/ufo2ft/blob/5b22f201b4650ab9ef1169357702a732e3127a5e/Lib/ufo2ft/outlineCompiler.py#L460-L542 .

fontir/src/ir.rs
```diff
@@ -102,13 +102,15 @@ impl Features {
 #[serde(from = "GlyphSerdeRepr", into = "GlyphSerdeRepr")]
 pub struct Glyph {
     pub name: GlyphName,
+    pub accessors: HashSet<Vec<u32>>, // sequence(s) of unicodes to reach this glyph
```
anthrotype, line 105: why don't you call these `unicodes` to match what the sources (e.g. UFO, Glyphs) also have? I personally don't find the name "accessors" any clearer, perhaps it's my non-native English or my familiarity with font sources.

anthrotype: if you don't like "unicodes", then "codepoints" would also be fine

anthrotype: actually -- I propose that four our IR we completely do away with the flawed approach of storing a list of unicode codepoints at the Glyph level, and instead have a global CharacterMap (which you can store in the static metadata) that is simply a `HashMap<Codepoint, GlyphName>`.
Associating codepoints to a glyph is wrong, not just because it obfuscates the fact that a cmap -- which this piece of data is designed to build -- actually maps from codepoints to glyphs instead of the other way around; but most importantly, it opens the door for ambiguous/invalid/impossible to encode mappings whereby more than one glyph ends up being mapped to the same unicode codepoint (e.g. two glyphs "foo" and "bar" both having unicodes list = [0x0061]; if one types "a", which one will one get, foo or bar?!)

See discussion on the proposal for storing a global cmap.txt and removing GLIF-level unicode elements in UFO4: https://github.com/unified-font-object/ufo-spec/issues/77

fontbe/src/cmap.rs
```diff
@@ -0,0 +1,51 @@
...
+
+pub fn create_cmap_work() -> Box<BeWork> {
+    Box::new(CmapWork {})
+}
+
+impl Work<Context, Error> for CmapWork {
+    /// Generate [cmap](https://learn.microsoft.com/en-us/typography/opentype/spec/cmap)
+    fn exec(&self, context: &Context) -> Result<(), Error> {
+        // cmap only accomodates single codepoint : glyph mappings; collect all of those
+        let static_metadata = context.ir.get_static_metadata();
```
anthrotype, line 22: if you had the global {codepoint : glyph} mappings in the static metadata IR, you'd be done here

fontc/src/main.rs
```diff
@@ -414,30 +415,65 @@ fn add_glyph_merge_be_job(
...
+            .collect();
+        dependencies.insert(FeWorkIdentifier::FinalizeStaticMetadata.into());
+
+        let id: AnyWorkId = BeWorkIdentifier::Cmap.into();
+        workload.insert(
+            id.clone(),
+            Job {
+                work: create_cmap_work().into(),
+                dependencies,
+                // We need to read all glyph IR, even unchanged ones, plus static metadata
```
anthrotype, line 468: > We need to read all glyph IR

if you had the global {character: glyph} mappings, you wouldn't need to read all glyph IRs, just the static metadata

fontir/src/ir.rs
rsheeter: Since we will support legacy versions of both glyphs and ufo for at least a few years I believe the IR should model it in this way and then the BE should convert it.

I do agree that accessors is a stunningly poor choice of name. I like codepoints and use it in other code.

rsheeter: Maybe I will split codepoints from ligatures so it's a bit simpler what goes into cmap vs needs to become ligatures.

anthrotype: you mean, nanoemoji-style GSUB ligatures from RGI sequences? Note that the list of codepoints in a UFO/Glyphs glyph is _not_ understood to be a sequence in GSUB ligature sense, but as the set of codepoints that map to that glyph in the character mapping table.

anthrotype: it is only a _list_ (ordered) and not a _set_ because of the even more flawed notion of a primary unicode codepoint which happens to be the first in that list (and which makes no sense from a cmap point of view).
The whole glyph.unicodes is a legacy that haunts font tooling since the Robofab days and should really be superseded by a global {codepoint: gid} mapping.

rsheeter: Done, it's slightly more verbose but IMHO much clearer

rsheeter: I don't want to lose the disucssion of representing cmap some other way, nor do I want to block this on it. Filed as https://github.com/googlefonts/fontmake-rs/issues/145

fontbe/src/cmap.rs
rsheeter: yup! I have forkecd that discussion to https://github.com/googlefonts/fontmake-rs/issues/145.

fontc/src/main.rs
rsheeter: Agreed, I have forked that discussion to https://github.com/googlefonts/fontmake-rs/issues/145

glyphs-reader/src/font.rs
```diff
@@ -747,18 +749,30 @@ fn parse_glyph_order(raw_font: &RawFont) -> Vec<String> {
     glyph_order
 }
 
-fn parse_codepoints(raw_font: &mut RawFont, radix: u32) -> BTreeMap<Vec<u32>, String> {
+/// Returns two maps, codepoints => glyph name and the reverse
```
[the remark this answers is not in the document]
rsheeter: Err. That does sound dumb. I should just return the cmap version. Note however that I don't think it's quite so clear we can just skip glyph level codepoints, see #145.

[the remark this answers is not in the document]
rsheeter: Hm. I think I misinterpreted .gyphs multiple unicodes to mean a ligature where per https://glyphsapp.com/learn/unicode it should mean multiple cmap entries. Will fix.

fontir/src/ir.rs
rsheeter: > the list of codepoints in a UFO/Glyphs glyph is not understood to be a sequence in GSUB ligature sense, but as the set of codepoints that map to that glyph in the character mapping table.

I managed to get that right for UFO and wrong for glyphs because _clearly_ a single field unicodes with a list means a ligature. Sigh. Fixed.

I did destroy the ordering, which might be a problem if we ever wanted source => IR => source with retention of primary codepoint. I don't currently understand what a primary codepoint *is*

glyphs-reader/src/font.rs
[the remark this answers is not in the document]
rsheeter: Fixed. Made things simpler which is nice.

fontir/src/ir.rs
rsheeter: OK, I think thi sis fixed for this PR.

anthrotype (approved):

Issue #138, opened 2023-03-13
Compile an MVP font that renders


rsheeter:
Where "works" is defined as renders in Chrome on Mac and/or Linux *not* as produced the same ttx as fontmake; that will come later. Per f2f with @anthrotype I believe this encompasses support for:

- [ ] head
- [ ] maxp
- [ ] cmap _PRs outstanding against fontations & fontmake-rs_
- [x] glyf/loca
- [ ] hmtx
- [ ] hhea
- [ ] OS/2
- [ ] name _do browsers accept fonts w/o name?_
- [ ] post _format 3 is simple_
- [ ] Merging individual tables into a valid sfnt

rsheeter:
Oh nice. And I suppose implicit in renders is "passes ots"

PR #139, opened 2023-03-13
Emit mvp post

rsheeter:
No tmeant to be correct, merely to tick the box for #138

### DIFF
--- a/fontbe/src/post.rs
+++ b/fontbe/src/post.rs
@@ -1,7 +1,7 @@
 //! Generates a [post](https://learn.microsoft.com/en-us/typography/opentype/spec/post) table.
 
 use fontdrasil::orchestration::Work;
-use read_fonts::types::{FWord, Fixed};
+use read_fonts::types::Version16Dot16;
 use write_fonts::tables::post::Post;
 
 use crate::{
@@ -19,16 +19,7 @@ impl Work<Context, Error> for PostWork {
     /// Generate [post](https://learn.microsoft.com/en-us/typography/opentype/spec/post)
     fn exec(&self, context: &Context) -> Result<(), Error> {
         // TODO a more serious post
-        let post = Post::new(
-            Fixed::from_i32(0),
-            FWord::new(0),
-            FWord::new(16),
-            0,
-            0,
-            0,
-            0,
-            0,
-        );
+        let post = Post { version: Version16Dot16::VERSION_3_0, ..Default::default() };
         context.set_post(post);
         Ok(())
     }
--- a/fontbe/src/post.rs
+++ b/fontbe/src/post.rs
@@ -1,0 +1,35 @@
+//! Generates a [post](https://learn.microsoft.com/en-us/typography/opentype/spec/post) table.
+
+use fontdrasil::orchestration::Work;
+use read_fonts::types::{FWord, Fixed};
+use write_fonts::tables::post::Post;
+
+use crate::{
+    error::Error,
+    orchestration::{BeWork, Context},
+};
+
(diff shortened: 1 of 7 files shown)


fontbe/src/post.rs
```diff
@@ -0,0 +1,35 @@
...
+
+pub fn create_post_work() -> Box<BeWork> {
+    Box::new(PostWork {})
+}
+
+impl Work<Context, Error> for PostWork {
+    /// Generate [post](https://learn.microsoft.com/en-us/typography/opentype/spec/post)
+    fn exec(&self, context: &Context) -> Result<(), Error> {
+        // TODO a more serious post
+        let post = Post::new(
```
anthrotype, line 22: where is the formatType = 3.0 defined in here? Or does write-fonts take care of setting that automatically?

fontbe/src/post.rs
```diff
@@ -0,0 +1,35 @@
...
+            Fixed::from_i32(0),
+            FWord::new(0),
+            FWord::new(16),
```
anthrotype, line 25: what is 16 here? I think for a viable MVP post format 3 you just need to set format to 3.0 and all the rest of the fields to 0

fontbe/src/post.rs
anthrotype: hm I inspected the `post.table` file you compiled, and the format is 1.0, not 3.0: the first four bytes are `b'\x00\x01\x00\x00'`

anthrotype: post 1.0 is for fonts that contains "exactly the 258 glyphs in the standard Macintosh TrueType font file", which is not the case for you. You want formatType=3.0

fontbe/src/post.rs
anthrotype: looks like the third new() parameter is the post.underlineThickness

https://github.com/googlefonts/fontations/blob/5fc746797c643c04b1dee2091ceb8f9642c84e37/write-fonts/generated/generated_post.rs#L79

I think 16 is too arbitrary. If you don't want to set to 0, then you can do what ufo2ft does (uses a ratio of the upem):

https://github.com/googlefonts/ufo2ft/blob/5b22f201b4650ab9ef1169357702a732e3127a5e/Lib/ufo2ft/fontInfoData.py#L313-L322

fontbe/src/post.rs
rsheeter: Erk good catch. For some reason I thought that would get me 3 :(

fontbe/src/post.rs
rsheeter: I switched to your earlier suggestion of defaults and filed https://github.com/googlefonts/fontmake-rs/issues/146 so I don't forget where to get better underline values from.

fontc/src/main.rs
```diff
@@ -478,6 +484,30 @@ fn add_cmap_be_job(
     Ok(())
 }
 
+fn add_post_be_job(
+    change_detector: &mut ChangeDetector,
+    workload: &mut Workload,
+) -> Result<(), Error> {
+    if change_detector.post_be_change() {
+        let mut dependencies = HashSet::new();
+        dependencies.insert(FeWorkIdentifier::FinalizeStaticMetadata.into());
```
anthrotype, line 493: I think your minimalist post v3 does not have any dependencies, does it? It's a constant string, basically

anthrotype (approved):

fontc/src/main.rs
rsheeter: That's right, will remove.

PR #143, opened 2023-03-14
Rough in hmtx

rsheeter:
First pass at hmtx. Contributes to #138.


fontc/src/main.rs
```diff
@@ -655,11 +701,15 @@ impl Workload {
 
                     debug!("Updating graph for new glyph {glyph_name}");
 
-                    self.jobs_pending
-                        .get_mut(&AnyWorkId::Be(BeWorkIdentifier::Glyf))
-                        .unwrap()
-                        .dependencies
-                        .insert(id.clone());
+                    // It would be lovely if our new glyph was in glyf and hmtx
+                    // loca hides with hmtx
```
anthrotype, line 705: I'm confused by the sentence "loca hides with hmtx"... did you mean "glyf"?

resources/testdata/glyphs2/OneGlyph.glyphs
```diff
@@ -0,0 +1,32 @@
+{
+familyName = "New Font";
+fontMaster = (
+    {
+        id = m01;
+    }
+);
+glyphs = (
+{
+    glyphname = period;
```
anthrotype, line 10: might as well call this '.notdef', which is a required glyph anyway

fontc/src/main.rs
```diff
@@ -1459,4 +1514,46 @@ mod tests {
...
     }
+
+    #[test]
+    fn hmtx_of_one() {
+        let temp_dir = tempdir().unwrap();
+        let build_dir = temp_dir.path();
+        let result = compile(test_args(build_dir, "glyphs2/OneGlyph.glyphs"));
+
+        let raw_hmtx = result.be_context.get_hmtx();
+        let hmtx = Hmtx::read_with_args(FontData::new(&raw_hmtx.buf), &(1, 1)).unwrap();
```
anthrotype, line 1525: what's `&(1, 1)`, I suppose something from hhea needed to parse hmtx.. I wish rust had keyword args sometimes

fontbe/src/hmtx.rs
```diff
@@ -0,0 +1,80 @@
...
+        let static_metadata = context.ir.get_final_static_metadata();
+        let var_model = &static_metadata.variation_model;
+        let default_location = var_model.default_location();
+
+        let mut long_metrics: Vec<LongMetric> = static_metadata.glyph_order.iter()
+            .map(|gn| {
+                let bbox = context.get_glyph(gn).bbox();
+                let ir_glyph = context.ir.get_glyph_ir(gn);
+                let Some(ir_instance) = ir_glyph.sources.get(default_location) else {
+                    panic!("{gn} is not defined at the default location. Unnecessary limitation if it covers the default.");
```
anthrotype, line 36: I would remove that from the error message, as too opinionated: "Unnecessary limitation if it covers the default"

right now, we require the the default location is complete and only allow non-default location to be sparse. and hmtx makes no exception. Changing this assumption is not as trivial as you might think, as it would require to make a temporary VariationModel where another location is assumed to be the default, then the desired default be interpolated from the available ones and re-inserted as the actual default in a new model; but this might also change the effect which the rest of the preexisting masters have, since the default/neutral master by definition applies throughout the variation space, being the base for the other masters..
Anyway that's how OT variation model work nowadays, and font designers know this and don't need to be reminded of this 'limitation' here. File an issue perhas if you like to discuss this further.

fontbe/src/hmtx.rs
```diff
@@ -0,0 +1,80 @@
...
+                };
+                let advance: u16 = ir_instance.width as u16;
+
+                LongMetric {
+                    advance,
+                    side_bearing: bbox.x_min
```
anthrotype, line 42: Source can contain float advance widths as well as float coordinates. Where does the rounding from floats to int happen? in the IR or here in the BE?
To match ufo2ft we want to use [otRound](https://github.com/fonttools/fonttools/blob/444a349ef4dd1bae2d43f6ea0e06d73fb6466b9a/Lib/fontTools/misc/roundTools.py#L23) for this.

fontc/src/main.rs
rsheeter: ...yes

fontc/src/main.rs
rsheeter: I agree, and the docs for read with args for hmtx are pretty obtuse too. @cmyr thoughts?

fontbe/src/hmtx.rs
rsheeter: Fair point. I do think sources that *don't* cover the default shoudl work but it's probably not helpful to note it everywhere, it just needs to be it's own project.

fontc/src/main.rs
cmyr: we have https://github.com/googlefonts/fontations/issues/108 open for this; I did not expect this method to really ever be called directly from user code. I'll hack on this now, haven't written code in a while and this feels like a good warmup.

fontbe/src/hmtx.rs
rsheeter: The IR GlyphInstance has `width: f64,` and I'm using as u16 here to convert it.

fontc/src/main.rs
rsheeter: Fixed

rsheeter:
Apologies, had to force push to tidy up merge errors.

resources/testdata/glyphs2/OneGlyph.glyphs
rsheeter: I like it, so updated

fontc/src/main.rs
rsheeter: I will happily update once we have something in fontations but I would prefer not to block this PR on this.

fontbe/src/hmtx.rs
rsheeter: I think that will make a very nice discrete followon; tracking in https://github.com/googlefonts/fontmake-rs/issues/147.

resources/testdata/glyphs2/OneGlyph.glyphs
anthrotype: thanks. This reminds me of https://github.com/googlefonts/fontmake-rs/issues/149

anthrotype (approved):

rsheeter:
> you can always us git merge instead of git rebase

I know, but I also know I might forget the cleanup step. Gambling that feedback was largely complete I suppose.

PR #148, opened 2023-03-15
Assemble final font

rsheeter:
If you compile a font and try to ttx it sadly it blows up because we don't have maxp yet. Soon!

I would like to merge this and then grow the set of tables over the next couple of PRs as we drive towards #138.

### DIFF
--- a/fontbe/src/font.rs
+++ b/fontbe/src/font.rs
@@ -1,0 +1,41 @@
+//! Merge tables into a font
+
+use fontdrasil::orchestration::Work;
+use read_fonts::types::Tag;
+use write_fonts::FontBuilder;
+
+use crate::{
+    error::Error,
+    orchestration::{BeWork, Bytes, Context, WorkId},
+};
+
+struct FontWork {}
+
+pub fn create_font_work() -> Box<BeWork> {
+    Box::new(FontWork {})
+}
+
+// TODO the tables have the tag, but I'm not sure how to access it
+const TABLES_TO_MERGE: &[(WorkId, Tag)] = &[
+    (WorkId::Cmap, Tag::new(b"cmap")),
+    (WorkId::Hmtx, Tag::new(b"hmtx")),
+    (WorkId::Glyf, Tag::new(b"glyf")),
+    (WorkId::Loca, Tag::new(b"loca")),
+];
+
+impl Work<Context, Error> for FontWork {
+    /// Glue binary tables into a font
+    fn exec(&self, context: &Context) -> Result<(), Error> {
+        // Lets go right ahead and believe those bytes are a font
+        let mut builder = FontBuilder::default();
+
+        // A fancier implementation would mmap the files. We basic.
+        for (work_id, tag) in TABLES_TO_MERGE.iter().cloned() {
+            let bytes = context.read_raw(work_id).map_err(Error::IoError)?;
+            builder.add_table(tag, bytes);
+        }
+
+        context.set_font(Bytes::new(builder.build()));
+        Ok(())
+    }
+}
--- a/fontbe/src/font.rs
+++ b/fontbe/src/font.rs
@@ -31,7 +31,7 @@ impl Work<Context, Error> for FontWork {
 
         // A fancier implementation would mmap the files. We basic.
         for (work_id, tag) in TABLES_TO_MERGE {
-            let bytes = context.read_raw(*work_id).map_err(Error::IoError)?;
+            let bytes = context.read_raw(work_id.clone()).map_err(Error::IoError)?;
             builder.add_table(*tag, bytes);
         }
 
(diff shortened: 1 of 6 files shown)


fontbe/src/font.rs
```diff
@@ -0,0 +1,41 @@
...
+
+struct FontWork {}
+
+pub fn create_font_work() -> Box<BeWork> {
+    Box::new(FontWork {})
+}
+
+// TODO the tables have the tag, but I'm not sure how to access it
+const TABLES_TO_MERGE: &[(WorkId, Tag)] = &[
+    (WorkId::Cmap, Tag::new(b"cmap")),
```
anthrotype, line 20: You're missing perhaps the most important tables of all, maxp, which defines _how many glyphs_ are in the font :)

fontbe/src/font.rs
```diff
@@ -0,0 +1,41 @@
...
+    (WorkId::Hmtx, Tag::new(b"hmtx")),
+    (WorkId::Glyf, Tag::new(b"glyf")),
+    (WorkId::Loca, Tag::new(b"loca")),
```
anthrotype, line 23: needless to say, hmtx won't work without hhea, and loca (thus glyf) won't work without head (likely other tables too depend on head too), but those are easy if you just want something that works and don't care about the particular values in it. OS/2 might be needed for windows I think (just use the same vertical metrics values as hhea).

anthrotype: oh, name table seems to be required by OTS (I just tried to strip it from a font and it complained)

rsheeter: Yup. I think I want to merge this then add more tables as we progress through #138.

rsheeter: To paraphrase me, this puts the pattern to create a working font in place. It doesn't yet fully do it due to missing tables. My current expectation is if we hit the set identified in #138 we're good.

fontbe/src/orchestration.rs
```diff
@@ -210,12 +213,26 @@ where
     T::read(FontData::new(&buf)).unwrap()
 }
 
-fn raw_hmtx_from_file(file: &Path) -> RawHmtx {
+pub struct Bytes {
```
anthrotype, line 216: you have an identical `Raw` struct in fontbe::hmtx module...

fontbe/src/hmtx.rs
```diff
@@ -8,10 +8,10 @@ use write_fonts::{
 
 use crate::{
     error::Error,
-    orchestration::{BeWork, Context},
+    orchestration::{BeWork, Bytes, Context},
 };
 
-pub struct RawHmtx {
+pub struct Raw {
```
anthrotype, line 14: this is no longer used, is it?

anthrotype (approved):

fontbe/src/font.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// Glue binary tables into a font
+    fn exec(&self, context: &Context) -> Result<(), Error> {
+        // Lets go right ahead and believe those bytes are a font
+        let mut builder = FontBuilder::default();
+
+        // A fancier implementation would mmap the files. We basic.
+        for (work_id, tag) in TABLES_TO_MERGE.iter().cloned() {
+            let bytes = context.read_raw(work_id).map_err(Error::IoError)?;
+            builder.add_table(tag, bytes);
+        }
```
cmyr, line 36: Merely for the sake of conversation, I would probably write this as,

```suggestion
        for (work_id, tag) in TABLES_TO_MERGE {
            let bytes = context.read_raw(*work_id).map_err(Error::IoError)?;
            builder.add_table(*tag, bytes);
        }
```

🤷

fontbe/src/orchestration.rs
```diff
@@ -210,12 +213,26 @@ where
...
+    pub(crate) fn new(buf: Vec<u8>) -> Bytes {
+        Bytes { buf }
+    }
+
+    pub fn get(&self) -> &[u8] {
+        &self.buf
+    }
+}
+
+fn raw_from_file(file: &Path) -> Bytes {
```
cmyr, line 230: does this want to be `Bytes::read_from_file` or something?

cmyr (approved):
Some very ignorable comments inline, lgtm

fontbe/src/font.rs
rsheeter: I had that and thought it less readable. Do you find the opposite?

fontbe/src/orchestration.rs
rsheeter: I have 2, need 1, and would prefer ot have 0 rofl. Will fix.

fontbe/src/hmtx.rs
rsheeter: Nope! Deleted.

fontbe/src/orchestration.rs
rsheeter: Keeping this one.

fontbe/src/orchestration.rs
rsheeter: That seems to complicate my macro. Will add a comment.

PR #151, opened 2023-03-15
Properly round advances when building hmtx

### LINKED ISSUE #147: Round hmtx properly
In https://github.com/googlefonts/fontmake-rs/pull/143#pullrequestreview-1339233906 @anthrotype noted

> Source can contain float advance widths as well as float coordinates. Where does the rounding from floats to int happen? in the IR or here in the BE?
> To match ufo2ft we want to use [otRound](https://github.com/fonttools/fonttools/blob/444a349ef4dd1bae2d43f6ea0e06d73fb6466b9a/Lib/fontTools/misc/roundTools.py#L23) for this.

Define - or locate - otRound in fontations and use it in `fontbe/src/hmtx.rs`. Currently hmtx rounds the float advance using `as u16` in #143.

rsheeter:
Fixes #147

### DIFF
--- a/fontbe/Cargo.toml
+++ b/fontbe/Cargo.toml
@@ -27,7 +27,7 @@ env_logger = "0.9.0"
 parking_lot = "0.12.1"
 
 read-fonts = "0.1.2"
-write-fonts = "0.1.2"
+write-fonts = "0.1.3"
 
 kurbo = { version = "0.9.1", features = ["serde"] }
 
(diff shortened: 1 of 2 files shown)


anthrotype (approved):
LGTM (assuming the left sidebearing is already ot-rounded as it comes from glyf)

rsheeter:
> assuming the left sidebearing is already ot-rounded as it comes from glyf

It probably isn't since I dind't have ot round until today. Will file a bug for that.

Issue #152, opened 2023-03-15
Use ot_round when building glyf


rsheeter:
https://github.com/googlefonts/fontmake-rs/pull/151#pullrequestreview-1342084790 notes that we should be using ot round. Since it didn't exist until https://github.com/googlefonts/fontations/pull/287 we probably aren't :)

anthrotype:
I assumed that they were becaue the original ot_round was inside write-fonts glyf module, but it's worth confirming that.

rsheeter:
I think there are some values, e.g. bbox, that fontmake-rs is setting directly in some cases. Perhaps that means the only place it's missing is the bbox of composites.

rsheeter:
We have two types, write-fonts Bbox which has i16 min/max and kurbo Rect which uses floats.

We create simple glyphs from bez path via https://github.com/googlefonts/fontations/blob/main/write-fonts/src/tables/glyf.rs#L151. That computes a Rect which is turned into a Bbox using `From<Rect> for Bbox`, which does ot rounding. So far so good I think.

For composite glyphs we lookup the Bbox for the constituant simple glyphs, transform it (which makes a Rect, a box with floats), then turn it into a Bbox using `rect2bbox(rect: Rect) -> Bbox`, which does *not* do ot rounding. We then reduce accumulated Bbox to produce a final Bbox. I think that means we have two obvious problems:

1. We don't ot round in fontmake-rs when computing composite bboxes
1. We convert Rect > Bbox too early, we should keep Rect's until the reduce phase

I'm not sure if it's OK that we use the Bbox from the simple glyph - rounded - as input or if we should compute the Rect for the path and do only one conversion from Rect => Bbox at the end of the reduce. Wdyt?

anthrotype:
No, it's the correct thing to do, and exactly what ufo2ft also does, compute the composite glyphs bounding boxes from the compiled simple TTGlyphs which have ot-rounded bboxes

anthrotype:
fonttools recalcBBoxes option when saving a TTFont also computes the bboxes from the already rounded coordinates that are in glyf, so ufo2ft marches that and computes bboxes from post-rounded coordinates

anthrotype:
> For composite glyphs we lookup the Bbox for the constituant simple glyphs, transform it (which makes a Rect, a box with floats), then turn it into a Bbox

[...]

PR #153, opened 2023-03-16
Improve composite glyf bbox computation

rsheeter:
I'm confident this improves the situation. I'm not sure it can be said to finish the job. Please read #152 and tell me your thoughts.

Helps with #152.

Just merge.

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -1,9 +1,6 @@
 //! 'glyf' Glyph binary compilation
 
-use std::{
-    cmp,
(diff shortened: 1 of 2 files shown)


anthrotype (approved):
this definitely helps in the sense that delays the rounding the end, avoiding rounding errors to accummulate. But it's incorrect because bbox(transform(bbox(shape))) != bbox(transform(shape)), see my comments in #152

rsheeter:
Baby steps lol. In retrospect I don't know how I convinced myself I could just transform the box.

PR #156, opened 2023-03-16
Add a minimal head table. Our fonts now ttx.

rsheeter:
Contributes to #138.

Put upem into it because hey, we might need that. Turns out we weren't capturing it from source so... do so.

ttx now runs to completion on fontc produced fonts. I'm mildly amazed. Try it yourself!

```shell
cargo run -p fontc -- --source resources/testdata/static.designspace
ttx build/font.ttf
```

### DIFF
--- a/fontbe/src/head.rs
+++ b/fontbe/src/head.rs
@@ -1,4 +1,4 @@
-//! Generates a [maxp](https://learn.microsoft.com/en-us/typography/opentype/spec/maxp) table.
+//! Generates a [head](https://learn.microsoft.com/en-us/typography/opentype/spec/head) table.
 
 use fontdrasil::orchestration::Work;
 use write_fonts::tables::head::Head;
--- a/fontbe/src/head.rs
+++ b/fontbe/src/head.rs
@@ -15,7 +15,7 @@ pub fn create_head_work() -> Box<BeWork> {
 }
 
 impl Work<Context, Error> for HeadWork {
-    /// Generate [maxp](https://learn.microsoft.com/en-us/typography/opentype/spec/maxp)
+    /// Generate [head](https://learn.microsoft.com/en-us/typography/opentype/spec/head)
     fn exec(&self, context: &Context) -> Result<(), Error> {
         let static_metadata = context.ir.get_final_static_metadata();
         let head = Head {
--- a/fontbe/src/head.rs
+++ b/fontbe/src/head.rs
@@ -1,0 +1,29 @@
+//! Generates a [maxp](https://learn.microsoft.com/en-us/typography/opentype/spec/maxp) table.
+
+use fontdrasil::orchestration::Work;
+use write_fonts::tables::head::Head;
+
+use crate::{
+    error::Error,
+    orchestration::{BeWork, Context},
+};
+
+struct HeadWork {}
+
+pub fn create_head_work() -> Box<BeWork> {
+    Box::new(HeadWork {})
+}
+
+impl Work<Context, Error> for HeadWork {
+    /// Generate [maxp](https://learn.microsoft.com/en-us/typography/opentype/spec/maxp)
+    fn exec(&self, context: &Context) -> Result<(), Error> {
+        let static_metadata = context.ir.get_final_static_metadata();
+        let head = Head {
+            units_per_em: static_metadata.units_per_em,
+            index_to_loc_format: 1, // TODO: set based on num glyphs
+            ..Default::default()
+        };
+        context.set_head(head);
+        Ok(())
+    }
+}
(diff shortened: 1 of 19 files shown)


fontbe/src/head.rs
```diff
@@ -0,0 +1,29 @@
...
+};
+
+struct HeadWork {}
+
+pub fn create_head_work() -> Box<BeWork> {
+    Box::new(HeadWork {})
+}
+
+impl Work<Context, Error> for HeadWork {
+    /// Generate [maxp](https://learn.microsoft.com/en-us/typography/opentype/spec/maxp)
```
anthrotype, line 18:
```suggestion
    /// Generate [head](https://learn.microsoft.com/en-us/typography/opentype/spec/head)
```

fontbe/src/head.rs
```diff
@@ -0,0 +1,29 @@
+//! Generates a [maxp](https://learn.microsoft.com/en-us/typography/opentype/spec/maxp) table.
```
anthrotype, line 1: copy pasta 🍝 

```suggestion
//! Generates a [head](https://learn.microsoft.com/en-us/typography/opentype/spec/head) table.
```

ufo2fontir/src/source.rs
```diff
@@ -435,6 +435,37 @@ fn glyph_order(
...
+    designspace_dir: &Path,
+) -> Result<u16, WorkError> {
+    let mut upems: HashSet<u16> = HashSet::new();
+    for source in designspace.sources.iter() {
+        let ufo_dir = designspace_dir.join(&source.filename);
+        if !ufo_dir.join("fontinfo.plist").exists() {
+            continue;
+        }
+        let fontinfo = load_plist(&ufo_dir, "fontinfo.plist")?;
+        let Some(plist::Value::Integer(upem)) = fontinfo.get("unitsPerEm") else {
```
anthrotype, line 449: for whatever reason, UFO allows `unitsPerEm` to be "non-negative integer _or float_" (emphasis mine)... I have never seen a float upem in the wild, and can't even imagine one (a global _grid_ to be used as reference for the rest of font unit coordinates, is the least float-y thing in the world, IMHO).
Anyway, [ufo2ft applies an otRound to it before setting in the head table](https://github.com/googlefonts/ufo2ft/blob/5b22f201b4650ab9ef1169357702a732e3127a5e/Lib/ufo2ft/outlineCompiler.py#L257), we might as well do that here

https://unifiedfontobject.org/versions/ufo3/fontinfo.plist/#generic-dimension-information

(At least Glyphs.app types it as a integer so we are fine in glyphs2fontir).

ufo2fontir/src/source.rs
```diff
@@ -435,6 +435,37 @@ fn glyph_order(
...
+    designspace: &DesignSpaceDocument,
+    designspace_dir: &Path,
+) -> Result<u16, WorkError> {
+    let mut upems: HashSet<u16> = HashSet::new();
+    for source in designspace.sources.iter() {
+        let ufo_dir = designspace_dir.join(&source.filename);
+        if !ufo_dir.join("fontinfo.plist").exists() {
+            continue;
+        }
+        let fontinfo = load_plist(&ufo_dir, "fontinfo.plist")?;
```
anthrotype, line 448: for now this is ok, since we are only reading the unitsPerEm from the fontinfo.plist.. but we will sooner or later read more than just that, so it'd make sense to parse it only once instead of for each individual getter such as in this case for `units_per_em(designspace, designspace_dir)`

anthrotype (approved):
wow we have a font! this is a great milestone 💯

anthrotype:
> ttx now runs to completion on fontc produced fonts

just wanted to point out we got kind of lucky with static.designspace test file, because it produces an hmtx with 2 longHorMetric records (1 per glyph, there's 2 glyphs in total) and when fonttools tries to decompile this hmtx but does not find an hhea table, it assumes that there are as many longHorMetrics as there are maxp.numGlyphs, so decompiling hmtx works.
But as soon as len(longHorMetrics) != numGlyphs (when the "tail" of hmtx glyphs have all same advance but only different sidebearings), then decompiling will fail (not enough bytes to read) in absence of hhea table (which is where the number of longHorMetric records is stored for hmtx).
We'll make an hhea pretty soon anyway. Overall, impressive work Rod!

fontbe/src/head.rs
rsheeter: Important detail :D

ufo2fontir/src/source.rs
rsheeter: Really?! That's amazing. Good catch, will fix.

ufo2fontir/src/source.rs
rsheeter: Good catch. Changed to parse all the font info files then pass a vec of them to the upem fn.

rsheeter:
> we got kind of lucky

I did deliberatley pick a very simple example :D

ufo2fontir/src/source.rs
rsheeter: Fixed.

rsheeter: And added a test to prove it.

PR #159, opened 2023-03-17
[fontc] move Args to args module

cmyr:
No functional change, just doing some cleanup and trying to get up to speed on these crates.


dfrg (approved):
🧹

PR #160, opened 2023-03-17
[fontc] Replace test_args with Args::for_test

cmyr:
A little more cleanup.

(just merge me)

### DIFF
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -1,6 +1,6 @@
 //! Command line arguments
 
-use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


fontc/src/args.rs
```diff
@@ -53,4 +53,24 @@ impl Args {
 
         flags
     }
+
+    /// Manually create args for testing
+    #[doc(hidden)]
```
rsheeter, line 58: Doesn't this mean non-test binaries now carry for_test?

rsheeter (approved):
LGTM in context of #162

PR #161, opened 2023-03-17
[fontc] Move Config to new 'config' module

cmyr:
This also makes the 'init' function into a method on this type, but otherwise avoids any substantive changes.

There are a few FIXMEs in the code here where I have questions about some possible minor changes.

This is based on #160, and shouldn't be merged until that is.

### DIFF
--- a/fontc/src/config.rs
+++ b/fontc/src/config.rs
@@ -1,0 +1,133 @@
+//! State for a (possibly incremental) compiler job
+
+use std::{
+    fs, io,
+    path::{Path, PathBuf},
+};
+
+use fontir::{paths::Paths as IrPaths, source::Input, stateset::StateSet};
+use serde::{Deserialize, Serialize};
+
+use crate::Args;
+
+/// The settings and compiler definition of a single compilation run.
+///
+/// This tracks the input arguments for a compilation, as well as the compiler
+/// version used (so that we do not reuse incremental state between different
+/// compiler versions)
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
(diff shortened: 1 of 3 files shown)


fontc/src/config.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// The settings and compiler definition of a single compilation run.
+///
+/// This tracks the input arguments for a compilation, as well as the compiler
+/// version used (so that we do not reuse incremental state between different
+/// compiler versions)
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct Config {
+    pub args: Args,
+    // The compiler previously used so if the compiler changes config invalidates
```
rsheeter, line 21: config and any IR lurking in the build dir

fontc/src/config.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// Returns the path to the config file for this compilation
+    fn file(&self) -> PathBuf {
+        self.args.build_dir.join("fontc.yml")
+    }
+
+    /// Attempt to initialize the compiler for this run.
+    ///
+    /// This serializes the configuration if necessary, and returns any existing
+    /// incremental state for this run if it exists.
+    //FIXME: this returns a Result but it looks like we never actually return
```
rsheeter, line 42: I think that's an error and we should change the expect's to map_err or ?. Lets do that in it's own PR as it's hard to see diffs when mixed with a move.

rsheeter (approved):
LGTM, the fixme's look worth fixing but lets do them in their own pr plz

PR #165, opened 2023-03-17
[fontc] Make Config::init return errors

cmyr:


fontc/src/config.rs
```diff
@@ -39,32 +39,25 @@ impl Config {
...
+    pub fn init(&self) -> Result<Input, Error> {
         let config_file = self.file();
 
         let ir_paths = IrPaths::new(&self.args.build_dir);
         let ir_input_file = ir_paths.ir_input_file();
         if self.has_changed(&config_file) {
             log::info!("Config changed, generating a new one");
             if ir_input_file.exists() {
-                fs::remove_file(ir_input_file).expect("Unable to delete old ir input file");
+                fs::remove_file(ir_input_file)?;
```
rsheeter, line 50: IIRC this doesn't actually report which file so perhaps use `Error::FileExpected` which does?

fontc/src/config.rs
```diff
@@ -39,32 +39,25 @@ impl Config {
...
-            .expect("Unable to write updated config");
+            fs::write(config_file, serde_yaml::to_string(self)?)?;
         };
 
-        let ir_input = if ir_input_file.exists() {
-            let yml = fs::read_to_string(ir_input_file).expect("Unable to load ir input file");
-            serde_yaml::from_str(&yml).expect("Unable to parse ir input file")
+        if ir_input_file.exists() {
+            let yml = fs::read_to_string(ir_input_file)?;
+            serde_yaml::from_str(&yml).map_err(Into::into)
```
rsheeter, line 57: Is this a sneaky way to say `map_err(Error::YamlSerError)`?

fontc/src/config.rs
```diff
@@ -39,32 +39,25 @@ impl Config {
...
         } else {
-            Input::new()
-        };
-        Ok(ir_input)
+            Ok(Input::new())
+        }
```
rsheeter, line 60: Optional: reading this again I dislike the nesting in both before and after. Guessing this could be converted to a chain, e.g. input_file.exists().then(...).or_else.

rsheeter (approved):

fontc/src/config.rs
cmyr: yes, this is what `?` uses under the covers.

PR #166, opened 2023-03-17
[fontc] Recompute config path as needed

cmyr:
This costs us an extra allocation per init, in exchange for a slightly simpler API.

This also modifies one of the tests in a way that I think leaves it functionally unchanged.

### DIFF
--- a/fontc/src/config.rs
+++ b/fontc/src/config.rs
@@ -1,9 +1,6 @@
 //! State for a (possibly incremental) compiler job
 
-use std::{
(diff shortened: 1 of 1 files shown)


rsheeter (approved):
LGTM, but beware for clippy is angry

cmyr:
yes, and it feels like a clippy bug; `Box<dyn T>` does not itself impl `T`, for annoying reasons I have had to look up several times, and so I can't even act on the suggestion.

PR #167, opened 2023-03-17
Require Glyph to have a default location, make that instance easy to get

### LINKED ISSUE #164: Make a safe way to get the IR instance at the default location
Fail Glyph IR production if no default instance is available and add a method to get it easily.

Currently it is tiresome, you have to ask for sources at default location and you have to get the default location. See https://github.com/googlefonts/fontmake-rs/pull/158/files/e5f58d495c7bf43ab5b3e6ce65f054c69a7aacc8#diff-b0c3b1b1945e81b7eb357d6cbb5c6209560d69467365affd0433811acbfc9f91 for an example of this being annoying.

rsheeter:
`Glyph::default_location -> GlyphInstance` is added and used by hmtx. Various other files are modified for the sheer joy of it.

Fixes #164. Feel free to merge if it seems OK.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -99,18 +99,73 @@ impl Features {
 
 /// A variable definition of a single glyph.
 ///
-/// Defined in at least one position. If defined in
-/// many, presumed to vary continuously between positions and required
-/// to have variation compatible structure.
+/// Guarrantees at least one definition. Currently that must be at
+/// the default location. In theory that limitation is unnecessary,
+/// a variable glyph that covers the default location could compile.
+///
+/// If defined in many locations, presumed to vary continuously
+/// between positions and required to have variation compatible structure.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[serde(from = "GlyphSerdeRepr", into = "GlyphSerdeRepr")]
 pub struct Glyph {
     pub name: GlyphName,
     pub codepoints: HashSet<u32>, // single unicodes that each point to this glyph. Typically 0 or 1.
-    pub sources: HashMap<NormalizedLocation, GlyphInstance>,
+    default_location: NormalizedLocation,
+    sources: HashMap<NormalizedLocation, GlyphInstance>,
 }
 
 impl Glyph {
+    pub fn new(
+        name: GlyphName,
+        codepoints: HashSet<u32>,
+        sources: HashMap<NormalizedLocation, GlyphInstance>,
+    ) -> Result<Self, WorkError> {
+        if sources.is_empty() {
+            return Err(WorkError::InvalidSourceGlyph {
+                glyph_name: name,
+                message: "No sources".into(),
+            });
+        }
+        let defaults: Vec<_> = sources
+            .keys()
+            .filter(|loc| !loc.iter().any(|(_, c)| c.into_inner() != 0.0))
+            .collect();
+        if defaults.len() != 1 {
+            return Err(WorkError::InvalidSourceGlyph {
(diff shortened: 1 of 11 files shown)


fontir/src/ir.rs
```diff
@@ -99,18 +99,73 @@ impl Features {
...
+    ) -> Result<Self, WorkError> {
+        if sources.is_empty() {
+            return Err(WorkError::InvalidSourceGlyph {
+                glyph_name: name,
+                message: "No sources".into(),
+            });
+        }
+        let defaults: Vec<_> = sources
+            .keys()
+            .filter(|loc| !loc.iter().any(|(_, c)| c.into_inner() != 0.0))
```
cmyr, line 131: some thoughts:

- we could impl `PartialEq<f32>` for the various 'coord' types
- we could add `ZERO` consts on these types
- we could ignore me and move on

fontir/src/ir.rs
```diff
@@ -135,6 +190,34 @@ impl Glyph {
...
+
+impl From<&Glyph> for GlyphBuilder {
+    fn from(value: &Glyph) -> Self {
+        Self {
+            name: value.name.clone(),
+            codepoints: value.codepoints.clone(),
+            sources: value.sources.clone(),
+        }
+    }
+}
```
cmyr, line 209: I think we should delete this, and make the caller clone on their side, otherwise this can hide the fact that a clone is happening.

(and if we keep it, the impl can just be `value.clone().into()`)

cmyr (approved):
lg, some notes inline.

fontir/src/ir.rs
rsheeter: > this can hide the fact that a clone is happening.

Interesting. This doesn't reflect what I feel I see in Rust thus far, though ofc I would defer to your greater experience. To me it's no different than `From<&str> for String`: you definitely aren't getting an owned thing from a ref without some allocation, so it's clear from the signature there is some copying going on.

fontir/src/ir.rs
rsheeter: I think PartialEq and ZERO are good ideas but I would like to do it in a separate PR. Hm. Zero arguably makes most sense for normalized coords, it doesn't have special significance for the others.

rsheeter: https://github.com/googlefonts/fontmake-rs/issues/170

fontir/src/ir.rs
rsheeter: Merging this but happy to send a follow-on PR to adjust if it becomes clear what I did was a Bad Idea.

fontir/src/ir.rs
```diff
@@ -99,18 +99,73 @@ impl Features {
...
     pub codepoints: HashSet<u32>, // single unicodes that each point to this glyph. Typically 0 or 1.
-    pub sources: HashMap<NormalizedLocation, GlyphInstance>,
+    default_location: NormalizedLocation,
+    sources: HashMap<NormalizedLocation, GlyphInstance>,
 }
 
 impl Glyph {
+    pub fn new(
+        name: GlyphName,
+        codepoints: HashSet<u32>,
```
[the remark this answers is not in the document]
rsheeter: IMHO use of u32 matches both how sources discuss unicodes and how HB handles them (https://harfbuzz.github.io/harfbuzz-hb-common.html#hb-codepoint-t) so conversion to char should be a BE step, currently handled in https://github.com/googlefonts/fontmake-rs/blob/main/fontbe/src/cmap.rs by way of `char::from_u32`.

PR #171, opened 2023-03-19
Take 'name' entries from the default master for UFO+designspace

rsheeter:
.glyphs names coming soon to a fontc near you.

Contributes to #138. I suspect further adjustment will come once we get to trying to match ttx with fontmake.


ufo2fontir/src/source.rs
```diff
@@ -478,35 +474,66 @@ fn files_identical(f1: &Path, f2: &Path) -> Result<bool, WorkError> {
...
+fn names(font_info: &norad::FontInfo) -> HashMap<NameKey, String> {
+    let mut result = HashMap::new();
+
+    let mut insert_if_present = |maybe_value: &Option<String>, name_id: NameId| {
+        if let Some(value) = maybe_value.as_ref() {
+            // For now let's very optimistically emit unicode names only
+            // <https://learn.microsoft.com/en-us/typography/opentype/spec/name#platform-encoding-and-language-ids>
+            result.insert(
+                NameKey {
+                    platform_id: 0, // Unicode
```
anthrotype, line 504: note that fontmake (ufo2ft) emits "Windows" platform (3) name records only, which also support the full unicode repertoire, and AFAIK they should be more widely supported than strict "Unicode" platform name records.

> When building a Unicode font for Windows, the platform ID should be 3 and the encoding ID should be 1... All string data for platform 3 must be encoded in UTF-16BE.

https://learn.microsoft.com/en-us/typography/opentype/spec/name#windows-encoding-ids

ufo2ft defaults to platform_id=3, encoding_id=1 if name string is BMP-only else 10, langID=0x409 (for American English):
https://github.com/googlefonts/ufo2ft/blob/fca66fe3ea1ea88ffb36f8264b21ce042d3afd05/Lib/ufo2ft/outlineCompiler.py#L430-L432

We should match that.

re: english only, note that multilingual name strings can also be set using `openTypeNameRecords` entries in fontinfo.plist. You can add support for those later on if you like.

ufo2fontir/src/source.rs
```diff
@@ -478,35 +474,66 @@ fn files_identical(f1: &Path, f2: &Path) -> Result<bool, WorkError> {
...
+                    encoding_id: 5, // Unicode full repertoire
```
anthrotype, line 505: you sure you actually meant '5' here?
I believe that's for unicode variation sequences cmap 14:
https://learn.microsoft.com/en-us/typography/opentype/spec/cmap#unicode-platform-platform-id--0

I don't even see that defined in the name table's tables: https://learn.microsoft.com/en-us/typography/opentype/spec/name#platform-specific-encoding-and-language-ids-unicode-platform-platform-id--0

ufo2fontir/src/source.rs
rsheeter: I'm not spotting how to access l10n values of a name from https://docs.rs/norad/latest/norad/fontinfo/struct.FontInfo.html. @cmyr ?

ufo2fontir/src/source.rs
rsheeter: That's a typo, it should say 4. Also I'm deleting this to match the fontmake behavior :)

ufo2fontir/src/source.rs
rsheeter: Oh nevermind, apparently UFO has individual fields that don't specify platform, encoding, lang *and* a list of names with platform, encoding, lang. I was thrown by the individual field with no apparent way to set those fields.

fontir/src/ir.rs
```diff
@@ -63,6 +69,161 @@ impl StaticMetadata {
     }
 }
 
+/// See <https://learn.microsoft.com/en-us/typography/opentype/spec/name#name-ids>
+#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum NameId {
```
cmyr, line 74: this should maybe be part of `write-fonts`?

cmyr (approved):
One note, otherwise lgtm modulo cosimo's approval

fontir/src/ir.rs
rsheeter: Does write-fonts do serde?

rsheeter: (that's the only reason why not that comes to mind and even that seems minor)

rsheeter: > lgtm modulo cosimo's approval

I'm going to merge and then fix all the problems @anthrotype points out in a follow-on so I can establish a clean base for .glyphs support.

ufo2fontir/src/source.rs
```diff
@@ -478,35 +474,196 @@ fn files_identical(f1: &Path, f2: &Path) -> Result<bool, WorkError> {
...
+                builder
+                    .get(NameId::TypographicSubfamilyName)
+                    .unwrap_or_default(),
+            ),
+        );
+
+        // Name's that don't get individual fields
+        if let Some(name_records) = font_info.open_type_name_records.as_ref() {
+            for nr in name_records.iter() {
+                let name_id: u16 = nr.name_id.try_into().unwrap();
```
anthrotype, line 639: UFO's openTypeNameRecords not only specify the nameID, but also the specific platformID, platEncID and langID; you seem to be only using the nameID here and ignoring the rest.

See https://unifiedfontobject.org/versions/ufo3/fontinfo.plist/#name-record-format

ufo2fontir/src/source.rs
```diff
@@ -478,35 +474,196 @@ fn files_identical(f1: &Path, f2: &Path) -> Result<bool, WorkError> {
...
+
+        // Version has a weird fallback
+        // https://github.com/googlefonts/ufo2ft/blob/fca66fe3ea1ea88ffb36f8264b21ce042d3afd05/Lib/ufo2ft/fontInfoData.py#L169
+        if !builder.contains_key(NameId::Version) {
+            let major = font_info.version_major.unwrap_or_default();
+            let minor = font_info.version_minor.unwrap_or_default();
+            builder.add(NameId::Version, format!("Version {major}.{minor:0>3}"));
+        }
+
+        // Full name is always based on typographi family
```
anthrotype, line 622: typo

```diff
- // Full name is always based on typographi family
+ // Full name is always based on typographic family

File: ufo2fontir/src/source.rs (hunk 4 above)
Comment (anthrotype) [reply ufo2fontir/src/source.rs:639]: I suppose you are intentionally ignoring the additional attributes because of the complications of encoding the string, if that's the case I suggest you add a TODO as reminder

File: fontbe/src/font.rs [hunk 6]
```diff
@@ -17,13 +24,14 @@ pub fn create_font_work() -> Box<BeWork> {
 
 // TODO the tables have the tag, but I'm not sure how to access it
```
anthrotype, line 25: TODO comment is now stale

ufo2fontir/src/source.rs
```diff
@@ -478,35 +474,196 @@ fn files_identical(f1: &Path, f2: &Path) -> Result<bool, WorkError> {
...
+
+        // After our first pass at getting values, apply fallbacks
+
+        // https://github.com/googlefonts/ufo2ft/blob/fca66fe3ea1ea88ffb36f8264b21ce042d3afd05/Lib/ufo2ft/fontInfoData.py#L188
+        builder.apply_fallback(NameId::TypographicFamilyName, &[NameId::FamilyName]);
+
+        // https://github.com/googlefonts/ufo2ft/blob/fca66fe3ea1ea88ffb36f8264b21ce042d3afd05/Lib/ufo2ft/fontInfoData.py#L195
+        builder.apply_fallback(NameId::TypographicSubfamilyName, &[NameId::SubfamilyName]);
+
+        // Version has a weird fallback
```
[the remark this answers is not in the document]
anthrotype: you are also missing fallbacks for two other core nameIDs, the absence of which may make your font not work in some environments: namely, nameID=3 (TrueType Unique ID) and nameID=6 (PostScript Name).
If I remember correctly from the time I worked on "obfuscating" names in the subsetter (some foundry like to hide the real name of a webfont and replace with minimal garbage string)-- you need all the nameIDs between 1 and 6 included for a font to work reliably everywhere.
Ufo2ft should give you the right fallback for them.

https://github.com/googlefonts/ufo2ft/blob/fca66fe3ea1ea88ffb36f8264b21ce042d3afd05/Lib/ufo2ft/fontInfoData.py#L178-L185

https://github.com/googlefonts/ufo2ft/blob/fca66fe3ea1ea88ffb36f8264b21ce042d3afd05/Lib/ufo2ft/fontInfoData.py#L283-L293

ufo2fontir/src/source.rs
rsheeter: ...actually it's already fixed

ufo2fontir/src/source.rs
rsheeter: https://github.com/googlefonts/fontmake-rs/issues/188

fontbe/src/font.rs
rsheeter: Removed in #184

ufo2fontir/src/source.rs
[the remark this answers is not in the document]
rsheeter: Oh good catch, will move the fallback mechanism into FontBuilder and add in #184

PR #184, opened 2023-03-21
Support default names from .glyphs 2 and 3

rsheeter:


glyphs-reader/src/font.rs
```diff
@@ -77,20 +77,42 @@ pub enum Shape {
 // The font you get directly from a plist, minimally modified
 // Types chosen specifically to accomodate plist translation.
 #[derive(Debug, FromPlist, PartialEq, Eq)]
+#[allow(non_snake_case)]
 struct RawFont {
     pub units_per_em: Option<i64>,
     pub family_name: String,
+    pub copyright: Option<String>,
+    pub designer: Option<String>,
+    pub designerURL: Option<String>,
```
anthrotype, line 86: do you actually need to `#[allow(non_snake_case)]`?
Can you not do this at the point where you use the attributes:

```diff
- v2_to_v3_name(properties, &self.designerURL, "designerURL");
+ v2_to_v3_name(properties, &self.designer_url, "designerURL");
```

glyphs-reader/src/font.rs
```diff
@@ -527,6 +549,29 @@ fn custom_param<'a>(
     None
 }
 
+fn v2_to_v3_name(properties: &mut Vec<RawName>, v2_prop: &Option<String>, v3_name: &str) {
+    // https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md#properties
+    // Keys ending with "s" are localizable that means the second key is values
```
anthrotype, line 554: > Keys ending with "s" are localizable

🤦‍♂️

glyphs-reader/src/font.rs
```diff
@@ -710,11 +755,46 @@ impl RawFont {
...
+        let properties = self.properties.get_or_insert_with(Default::default);
+
+        v2_to_v3_name(properties, &self.copyright, "copyrights");
+        v2_to_v3_name(properties, &self.designer, "designers");
+        v2_to_v3_name(properties, &self.designerURL, "designerURL");
+        v2_to_v3_name(properties, &self.manufacturer, "manufacturers");
+        v2_to_v3_name(properties, &self.manufacturerURL, "manufacturerURL");
+
+        let mut v2_to_v3_param = |v2_name: &str, v3_name: &str| {
+            eprintln!("{:?}", custom_param(&self.other_stuff, v2_name));
```
anthrotype, line 770: should this be a logging message instead?

glyphs-reader/src/font.rs
```diff
@@ -1068,9 +1148,33 @@ impl TryFrom<RawFont> for Font {
...
+        }
+        names.insert("familyNames".into(), from.family_name);
+        let version_string = names.remove("versionString");
+        let empty_string = String::from("");
+        let version = format!(
+            "{}.{:0>3}{}{}",
+            from.versionMajor.unwrap_or_default(),
+            from.versionMinor.unwrap_or_default(),
+            version_string.as_ref().map(|_| "; ").unwrap_or(""),
+            version_string.as_ref().unwrap_or(&empty_string),
```
anthrotype, line 1172: where did you infer this treatment for versionString? glyphsLib doesn't seem to be doing this, but simply setting whatever is the versionString to the corresponding UFO's openTypeNameVersion (which will become nameID=5); whereas here you are appending it after a semicolor to the version string generated from major/minor.

glyphs-reader/src/font.rs
rsheeter: No, because the plist parser will fail to populate the field at all.

glyphs-reader/src/font.rs
```diff
@@ -1068,9 +1148,33 @@ impl TryFrom<RawFont> for Font {
...
+                .or_else(|| {
+                    name.values
+                        .unwrap_or_default()
+                        .iter()
+                        .find(|v| v.language == "dflt")
+                        .map(|v| v.value.clone())
+                })
+                .and_then(|value| names.insert(name.key, value));
+        }
+        names.insert("familyNames".into(), from.family_name);
```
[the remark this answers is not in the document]
rsheeter: If/when we support name l10n it will stop being single valued.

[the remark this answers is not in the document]
rsheeter: https://github.com/googlefonts/fontmake-rs/issues/187

glyphs-reader/src/font.rs
rsheeter: I made it up based on the ability to set a version - which fits the OT "Should begin with the syntax “Version <number>.<number>” (upper case, lower case, or mixed, with a space between “Version” and the number)." perfectly - and a version string in the UI. However, upon test exporting from a font with both version and version string set name 5 is just the version string and the values I punched for version are just gone.

anthrotype: > name 5 is just the version string and the values I punched for version are just gone.

yep, versionString defines the whole nameID=5

anthrotype (approved):

glyphs-reader/src/font.rs
rsheeter: Testing w/Glyphs App suggests if version string is there, just use it. If not, build the version string from minor/major. Will add a test for this.

dfrg:
We’re close to building a usable font so I don’t suggest we block on this but representing localized strings with an OpenType name record in our IR smells like a layering violation.

In the future, I’d like to see something like:
```rust
pub struct LocalizedString {
    pub language: String,
    pub value: String,
}
// …
HashMap<StringId, Vec<LocalizedString>>
```

And map to the appropriate platform/encoding/language ids in the backend. 

I recognize that UFO complicates this but that seems par for the course.

anthrotype:
what is that 'language' String? A BCP47 sort of language?

> map to the appropriate platform/encoding/language ids in the backend.

if you use that for the IR, then you'd need to map twice: once when translating sources (which may already contain OT-style platID, encID, langID triplets) to IR, and then again to OT name records.

But yeah, I agree we should not block on this one.

rsheeter:
> We’re close to building a usable font so I don’t suggest we block on this but representing localized strings with an OpenType name record in our IR smells like a layering violation.

Since both source formats do it too and we reserve the right to alte the IR format I think this is OK for now.

PR #193, opened 2023-03-21
Move to write-fonts 0.1.5; it fixes cmap

rsheeter:
CI will fail until https://github.com/googlefonts/fontations/pull/299 is submitted and released.


anthrotype (approved):
LGTM, though CI still fails even after write-fonts 0.1.5 was released. It's probably because of https://github.com/cmyr/fea-rs/pull/123?

PR #190, opened 2023-03-21
MVP OS2: the default

rsheeter:

PR #194, opened 2023-03-22
Produce a font that renders

### LINKED ISSUE #138: Compile an MVP font that renders

rsheeter:
Closes #138.

Builds on #190. 

OK to review, but DO NOT SUBMIT until a fea-rs release that compiles against NameId is available; https://github.com/cmyr/fea-rs/pull/123.

A plus compiled with fontc and for comparison the systems default plus:
image


fontir/src/ir.rs
```diff
@@ -54,8 +58,13 @@ impl StaticMetadata {
     ) -> Result<StaticMetadata, VariationModelError> {
         let axis_names = axes.iter().map(|a| a.name.clone()).collect();
         let variation_model = VariationModel::new(glyph_locations, axis_names)?;
+        // https://github.com/googlefonts/ufo2ft/blob/fca66fe3ea1ea88ffb36f8264b21ce042d3afd05/Lib/ufo2ft/fontInfoData.py#L38-L45
+        let ascender = (0.8 * units_per_em as f32).into();
+        let descender = (0.2 * units_per_em as f32).into();
```
anthrotype, line 63: note that those ufo2ft's fallback ascender/descender are ot-rounded, not left as floats. It doesn't look like you are rounding them here.

anthrotype: well you _are_ rounding them before writing them to hhea. I wonder if you should round them already in the IR as well.

anthrotype: I think for IR, we could type those fields as i16, I doubt anybody wants these to ever be floats (like upem)

rsheeter: I made it negative and rounded in hhea. In the light of a new day I think the negative part is in the wrong place but the rounding makes sense to me to defer.

fontir/src/ir.rs
```diff
@@ -299,6 +308,12 @@ impl From<NameId> for u16 {
     }
 }
 
+impl From<font_types::NameId> for NameId {
+    fn from(value: font_types::NameId) -> Self {
+        NameId::from(value.to_u16())
+    }
+}
```
cmyr, line 315: Let's open an issue for migrating to just use `font_types::NameId`?

cmyr (approved):
🥳

PR #198, opened 2023-03-22
[ufo2fontir] Codegolf computing UPM

cmyr:
No functional change, but in reviewing an earlier PR I noticed that this was calling `collect` three different times and I wondered if there was a reasonable alternative.

Conclusion: There is a semi-reasonable alternative?

### DIFF
--- a/ufo2fontir/src/source.rs
+++ b/ufo2fontir/src/source.rs
@@ -435,27 +435,23 @@ fn units_per_em<'a>(
     const MIN_UPEM: f64 = 16.0;
     const MAX_UPEM: f64 = 16384.0;
 
-    let upems: Vec<_> = font_infos
-        .filter_map(|fi| fi.units_per_em)
-        .map(|v| v.as_f64())
-        .collect();
-    for upem in upems.iter() {
-        // Per <https://learn.microsoft.com/en-us/typography/opentype/spec/head>, 16..16384
-        if *upem < MIN_UPEM || *upem > MAX_UPEM {
-            return Err(WorkError::InvalidUpem(format!("{upem}")));
-        }
-    }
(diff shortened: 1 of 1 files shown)


ufo2fontir/src/source.rs
```diff
@@ -435,27 +435,23 @@ fn units_per_em<'a>(
...
-    }
-    let upems: HashSet<u16> = upems.into_iter().map(|v| v.ot_round()).collect();
+    let mut upems: Vec<_> = font_infos
+        .filter_map(|fi| match fi.units_per_em.map(|x| x.as_f64()) {
+            None => None,
+            // Per <https://learn.microsoft.com/en-us/typography/opentype/spec/head>, 16..16384
+            Some(val) if (MIN_UPEM..=MAX_UPEM).contains(&val) => Some(Ok(val.ot_round())),
+            Some(bad_val) => Some(Err(WorkError::InvalidUpem(format!("{bad_val}")))),
+        })
+        .collect::<Result<_, _>>()?;
```
rsheeter, line 445: If you move to the model of collecting result can't you just collect to a set here and skip sort_unstable and dedup?

as_slice doesn't seem to make sense for hashset but you could match on it's length so you get 0 => 1000, 1 => iter.next.unwrap, _ => inconsistent.

cmyr: Two reasons for this choice:

- The error we return takes a `Vec<u16>`, so this lets me reuse a single result in that case
- pattern matching on slices gives us a very nice single expression, whereas using a set would require an unwrap to access the actual items
- all other things being equal I prefer using `Vec` since it's in the prelude, whereas you have to import `HashSet`.

PR #201, opened 2023-03-22
Capture glyphs v3 ascender/descender and plumb through to IR

rsheeter:


glyphs-reader/src/font.rs
```diff
@@ -1110,6 +1140,52 @@ fn raw_feature_to_feature(feature: RawFeature) -> Result<FeatureSnippet, Error>
...
+        master.baseline = Some(v);
+    });
+    copy_metric_to_masters(font, "cap height", |master, v| {
+        master.cap_height = Some(v);
+    });
+    copy_metric_to_masters(font, "x-height", |master, v| {
+        master.x_height = Some(v);
+    });
+    for master in font.font_master.iter_mut() {
+        master.metric_values = None;
```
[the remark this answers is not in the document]
anthrotype: I understand the convenience of doing `font.default_master().ascender` etc. But maybe instead of copying the values, you could add some accessors (e.g. `ascender()`) that would look up the respective metric into the metrics_values?

glyphs-reader/src/font.rs
```diff
@@ -245,11 +260,23 @@ pub struct Anchor {
 #[derive(Debug, Clone, FromPlist, PartialEq, Eq, Hash)]
 pub struct FontMaster {
     pub id: String,
+    pub ascender: Option<OrderedFloat<f64>>,
+    pub descender: Option<OrderedFloat<f64>>,
+    pub baseline: Option<OrderedFloat<f64>>,
+    pub cap_height: Option<OrderedFloat<f64>>,
+    pub x_height: Option<OrderedFloat<f64>>,
     pub axes_values: Option<Vec<OrderedFloat<f64>>>,
+    pub metric_values: Option<Vec<RawMetricValue>>,
```
anthrotype, line 269: this again looks a bit weird to me, because metrics_values technically only exists in v3 sources, whereas ascender/descender/etc. only exist in v2 sources, but you're using the same FontMaster struct for either v2 or v3...

glyphs-reader/src/font.rs
```diff
@@ -685,23 +712,32 @@ impl RawFont {
 
         // in each font master setup the parallel array
         for master in self.font_master.iter_mut() {
-            let mut metric_dicts = Vec::new();
+            if Some(OrderedFloat(0.0)) == master.cap_height {
+                master.cap_height = None;
+            }
+            if Some(OrderedFloat(0.0)) == master.x_height {
```
anthrotype, line 718: why special case these two cap_height and x_height? I agree 0.0 is arguably a good value for them, but why assume they are None?

glyphs2fontir/src/source.rs
```diff
@@ -239,16 +239,23 @@ impl Work<Context, WorkError> for StaticMetadataWork {
...
+            font.units_per_em,
+            names(font),
+            axes,
+            glyph_order,
+            glyph_locations,
+        )
+        .map_err(WorkError::VariationModelError)?;
+
+        if let Some(ascender) = font.default_master().ascender {
+            static_metadata.ascender = (ascender.into_inner() as f32).into();
```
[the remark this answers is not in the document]
rsheeter: I imagine in time we'll make it more like Glyph where we have a map location:these metrics but I dont' think we need to block on that.

glyphs-reader/src/font.rs
anthrotype: glyphsLib defaults to these values when they are omitted

https://github.com/googlefonts/glyphsLib/blob/da1aae7ac1b4c9f3d9e740be74ae2cc59a52dc6f/Lib/glyphsLib/classes.py#L1612-L1613
But to be sure, it might be good to see what Glyphs.app python scripting API returns when loading a font that does not have those set.

glyphs-reader/src/font.rs
rsheeter: I was trying to avoid having RawFontMaster and FontMaster, probably I should just bite the bullet and do it.

glyphs-reader/src/font.rs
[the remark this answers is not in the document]
rsheeter: I need the attribute to parse the objects nicely but if I use RawFontMaster and turn it into FontMaster there's a clear place to shuffle the attributes into metric_values so ... probably I should just not be lazy and actually do that.

glyphs-reader/src/font.rs
rsheeter: We parse both to the same structures so I think that's only avoidable by some really nasty hand-parsing of plists. Per prior comments, I'll make that just on RawFontMaster so you don't "see" it if you interact with only Font.

glyphs-reader/src/font.rs
anthrotype: If I create a new font from Glyphs 3 and immediately save it, I can see that the x-height and cap height metricValues are set respectively to 500 and 700 by default.
If I deliberately remove those two metrics properties from the master tab and save, I can see they disappear from the .glyphs file as expected. However if I open the Macro window and do

```python
print(Glyphs.font.masters[0].xHeight)
```
... I just get garbage that looks like random memory?! (`9223372036854775808`).

Oh... I tried to set the two values to 0 and then saved the file, and now I see what you mean, there's empty {} dict in their places in the metricValues array. It does look like the metrics values are omitted when both `pos` and `over` are zero. It's not just those two that you singlet out here, but all of them. I tried to set to (0,0) the ascender or the descender and they also get written out as empty {} in the serialized .glyphs file.

anthrotype: > v3 appears to omit them where v2 sets 0.

so yeah, in v3 metricValues, if either pos or over or both are not present in each metricValue, they are assumed to be 0.0.

in GSMetricValue constructor, in fact, they default to 0:
https://github.com/googlefonts/glyphsLib/blob/da1aae7ac1b4c9f3d9e740be74ae2cc59a52dc6f/Lib/glyphsLib/classes.py#L1442

glyphs-reader/src/font.rs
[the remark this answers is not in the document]
rsheeter: Done. I did take the liberty of storing a map<name, value> instead of having a parent with names and a child with a parallel value structure because that's just annoying and especially annoying in Rust, you end up with an Rc and a messy lookup.

glyphs-reader/src/font.rs
rsheeter: I modified it to not special case those.

glyphs2fontir/src/source.rs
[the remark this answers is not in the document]
rsheeter: Tracking in https://github.com/googlefonts/fontmake-rs/issues/205.

anthrotype (approved):

PR #204, opened 2023-03-23
add pre-push git hook to check for println! or eprintln! statements in Rust files

anthrotype:
I did this just to brush up on shell scripting and git hooks, so feel free to reject if you think this is not useful.

Also, note this deliberately only checks the specific lines that are added (cached) to the index, not all the lines in an .rs file. And it strips all `//` inline comments so that a commented-out `println!()` or `eprintln!()` would be allowed. If we wanted to support multi-line /**/ comments, I found a crate/CLI tool called [comment-strip](https://crates.io/crates/comment-strip/) but I don't think it's worth it.

### DIFF
--- a/fontc/src/workload.rs
+++ b/fontc/src/workload.rs
@@ -242,13 +242,13 @@ impl Workload {
         while !self.jobs_pending.is_empty() {
             let launchable = self.launchable();
             if launchable.is_empty() {
-                eprintln!("Completed:");
+                log::error!("Completed:");
                 for id in self.success.iter() {
-                    eprintln!("  {id:?}");
+                    log::error!("  {id:?}");
                 }
-                eprintln!("Unable to proceed with:");
+                log::error!("Unable to proceed with:");
                 for (id, job) in self.jobs_pending.iter() {
-                    eprintln!("  {:?}, happens-after {:?}", id, job.dependencies);
+                    log::error!("  {:?}, happens-after {:?}", id, job.dependencies);
                 }
                 assert!(
                     !launchable.is_empty(),
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -31,6 +31,9 @@ jobs:
       - name: cargo doc
         run: cargo doc --all-features --document-private-items --no-deps
 
+      - name: check no println! or eprintln! statements
+        run: resources/scripts/check_no_println.sh
+
   clippy-lint:
     name: Clippy lints
     runs-on: ubuntu-latest
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -362,7 +362,7 @@ impl TryFrom<&ir::Glyph> for CheckedGlyph {
                 .sources()
                 .iter()
                 .flat_map(|(location, instance)| {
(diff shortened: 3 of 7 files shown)


anthrotype:
I added this only to the `pre-commit` hook, but we may want to also add it to the `pre-push` in case one only uses the latter and not the former; or maybe only to `pre-push` so that temporarily one could commit print statements while developing but the pre-push hook would block them from polluting main.

Btw, this could well be a completely bad idea, as there might be valid use cases for println! (over logging) that I am not seeing.

rsheeter:
I like it but I'd hook to CI. I don't use the githooks as githooks - I often want to commit or push w/o the checks - so I run them manually and thus can forget :)

rsheeter:
Also definitely want it in pre-push instead or as well. Instead imho, I don't think you'd want to block commit with print for someone using the hooks as hooks.

rsheeter (approved):

cmyr:
This is interesting! There definitely are situations where `println` makes sense (e.g. when you're actually presenting output to the user on the command line that is not a diagnostic) but I think it's really a project-to-project thing?

I think this would make sense as a pre-push check, at least in terms of how I think about these things; in general I want all commits to compile, but I don't mind if they're a bit messy.

anthrotype:
ok, I can move it to pre-push and also add to CI (maybe in the same ci build job as rustfmt).

anthrotype:
> I don't use the githooks as githooks

hm well.. for this pre-push to work correctly it has to be run as a git hook, because passes the local and remote refs to the hook script's standard input, and I need those get a diff between what we have locally and what's on the remote in order to only check the lines that we are adding, as opposed to every single *.rs file...
I need to think about how I can accommodate both yours and the git-hook use cases. Maybe if it's run as a hook I only check the added lines in the diff as I currently do, whereas if run as a simple script I check everything (which may take longer but you can always add it as a hook if you like).

rsheeter:
You could also add this *only* to CI, similar to the OTS check.

anthrotype:
I moved the logic to a separate `resources/scripts/check_no_println.sh` script, and call that from the `pre-push` as well as from the CI. I didn't add a new job like with the ots-test, because I thought that would be wasting GHA precious time. I simply added a new step to the existing `Rustfmt` job. Let me know if you think it should not live there, but in its own distinct job.

By the way, it is currently failing because of these files containing println!

Screenshot 2023-03-28 at 12 53 32

fontc/src/workload.rs
```diff
@@ -242,13 +242,13 @@ impl Workload {
...
+                log::error!("Completed:");
                 for id in self.success.iter() {
-                    eprintln!("  {id:?}");
+                    log::error!("  {id:?}");
                 }
-                eprintln!("Unable to proceed with:");
+                log::error!("Unable to proceed with:");
                 for (id, job) in self.jobs_pending.iter() {
-                    eprintln!("  {:?}, happens-after {:?}", id, job.dependencies);
+                    log::error!("  {:?}, happens-after {:?}", id, job.dependencies);
```
anthrotype, line 251: I went for `log::error!` in this case because if `launchable` is indeed empty the following assert will fail, and also a panic would occur immediately afterwards where `&launchable[0]`

PR #206, opened 2023-03-24
Use a loc:metrics map

### LINKED ISSUE #205: Shift basic metrics in IR to a location:metrics map
As @anthrotype notes on https://github.com/googlefonts/fontmake-rs/pull/201#discussion_r1146207037 the current model is overly basic.

rsheeter:
Fixes #205. So far only the metrics at the default location are used. Baby steps. JMM if seems sane.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -3,7 +3,7 @@
 use crate::{
     coords::{CoordConverter, NormalizedLocation, UserCoord},
     error::{PathConversionError, VariationModelError, WorkError},
-    serde::{GlyphSerdeRepr, StaticMetadataSerdeRepr},
+    serde::{GlobalMetricsSerdeRepr, GlyphSerdeRepr, StaticMetadataSerdeRepr},
     variations::VariationModel,
 };
 use fontdrasil::types::GlyphName;
@@ -45,11 +45,6 @@ pub struct StaticMetadata {
     /// to narrow (submodel in FontTools terms) to the set of locations they actually use. Use of a
     /// location not in the global model is an error.
     pub variation_model: VariationModel,
-
-    /// The values of metrics at various positions in design space.
-    ///
-    /// At a minimum should be defined at the default location.
-    pub metrics: HashMap<NormalizedLocation, Metrics>,
 }
 
 impl StaticMetadata {
@@ -62,15 +57,9 @@ impl StaticMetadata {
     ) -> Result<StaticMetadata, VariationModelError> {
         let axis_names = axes.iter().map(|a| a.name.clone()).collect();
         let variation_model = VariationModel::new(glyph_locations, axis_names)?;
-        let mut metrics = HashMap::new();
-        metrics.insert(
-            variation_model.default_location().clone(),
-            Metrics::default_for_upem(units_per_em),
-        );
 
         Ok(StaticMetadata {
             units_per_em,
-            metrics,
             names,
             axes,
             glyph_order,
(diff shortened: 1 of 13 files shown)


fontir/src/ir.rs
```diff
@@ -28,26 +28,28 @@ pub struct StaticMetadata {
...
     ///
     /// This copy includes all locations used in the entire font. Users, such as glyph BE, may wish
     /// to narrow (submodel in FontTools terms) to the set of locations they actually use. Use of a
     /// location not in the global model is an error.
     pub variation_model: VariationModel,
+
+    /// The values of metrics at various positions in design space.
+    ///
+    /// At a minimum should be defined at the default location.
+    pub metrics: HashMap<NormalizedLocation, Metrics>,
```
anthrotype, line 52: I think the reason we called this metadata of IR "_Static_ Metadata" is because it is not variable, but static. At least that's how I understood it. If we now introduce a map of metrics at various locations, it is no longer static, in my view. 
Maybe it should just be called FontMetadata (as opposed to Glyph data) or simply Metadata.

Alternatively we could also think about splitting this off and create a separate object/file called Metrics (which would roughly correspond to Glyphs.app's [GSFontMaster](https://docu.glyphsapp.com/#gsfontmaster)) and is going to be used for things like MVAR (font-wide metrics variations table).

anthrotype: I can see already that my previous argument about StaticMetadata being about global font-wide non-variable stuff doesn't hold very well, because we already have a VariationModel in it, which is the cornerstone of variations...

anthrotype: actually, the VariationModel itself does not vary in the same sense of the other pieces of data, it is the thing that link them together, so in that sense it makes sense to be in StaticMetadata.

On the other hand, the very docstring at the top of StaticMetadata says:

> /// Global font info that cannot vary.
> ///
> /// For example, upem, axis definitions, etc, as distinct from
> /// metadata that varies across design space such as ascender/descender.

... which precisely contradicts this PR where "metadata that varies across design space such as ascender/descender" are being added to this StaticMetadata struct.

fontir/src/ir.rs
```diff
@@ -85,6 +81,43 @@ impl StaticMetadata {
...
+
+/// Global metrics. Ascender/descender, cap height, etc.
+///
+/// Represents the values of these metrics at a specific position in design space.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct Metrics {
+    pub ascender: OrderedFloat<f32>,
+    pub descender: OrderedFloat<f32>,
+    pub cap_height: OrderedFloat<f32>,
+    pub x_height: OrderedFloat<f32>,
```
anthrotype, line 106: technically MVAR allows each one of these (and similar other related ones) to variate independently of the others. Maybe we should support that somehow.
Either make these struct fields all optional such that when None we take it as "does not participate" when making the MVAR deltas, or we have map of {location: value} for each one of them, instead of a map {Location: Metric}

anthrotype: This would also be in line with our statede intention to move away from a "multiple-master" approach to font variations

fontir/src/ir.rs
rsheeter: I confess I entirely forgot we put that restriction on static metadata. I think I'd be up for metrics being a separate job and IR entry.

If I made that change would this seem OK?

fontir/src/ir.rs
rsheeter: Good point, I'll update to make them more independent.

fontir/src/ir.rs
anthrotype: > metrics being a separate job and IR entry.

yeah, sounds good

rsheeter: Done

fontir/src/ir.rs
rsheeter: Done

PR #211, opened 2023-03-26
Basic fvar

### LINKED ISSUE #209: Emit basic fvar
Just enough to expose the axes, in particular we could choose to skip named instances for now.

rsheeter:
Fixes #209.


fontbe/src/fvar.rs
```diff
@@ -0,0 +1,72 @@
...
+                .iter()
+                .map(|ir_axis| {
+                    let axis_name_id: u16 = (*reverse_names.get(&ir_axis.name).unwrap()).into();
+                    let mut var = VariationAxisRecord {
+                        axis_tag: Tag::new(ir_axis.tag.as_bytes()),
+                        min_value: Fixed::from_f64(ir_axis.min.into_inner().into_inner().into()),
+                        default_value: Fixed::from_f64(
+                            ir_axis.default.into_inner().into_inner().into(),
+                        ),
+                        max_value: Fixed::from_f64(ir_axis.max.into_inner().into_inner().into()),
```
dfrg, line 49: These conversions are a bit of a mess. I think we should make any use of `OrderedFloat` an implementation detail and not expose it in public APIs unless necessary. In other words, maybe have the `*Coord` types return f32 directly from `into_inner()` or just add a `to_f32()` method.

dfrg (approved):
Looks good with one nit.

fontbe/src/fvar.rs
cmyr: Honestly I'd say these types should just have an `Into<Fixed>` method something?

rsheeter: We sometimes want the ordered float so I think into_inner is right but I should add to_f32 and Into<Fixed>, good calls.

fontbe/src/fvar.rs
```diff
@@ -0,0 +1,72 @@
...
+            .collect();
+
+        let axes_and_instances = AxisInstanceArrays::new(
+            static_metadata
+                .axes
+                .iter()
+                .map(|ir_axis| {
+                    let axis_name_id: u16 = (*reverse_names.get(&ir_axis.name).unwrap()).into();
+                    let mut var = VariationAxisRecord {
+                        axis_tag: Tag::new(ir_axis.tag.as_bytes()),
```
cmyr, line 44: This will panic if the tag is not valid per the spec. Is this verified anywhere? I think it makes sense to just use `Tag` instead of `String` at the source, but otherwise you could use `Tag::new_checked`, which will return an error instead of panicking.

fontbe/src/fvar.rs
```diff
@@ -0,0 +1,72 @@
...
+                        min_value: Fixed::from_f64(ir_axis.min.into_inner().into_inner().into()),
+                        default_value: Fixed::from_f64(
+                            ir_axis.default.into_inner().into_inner().into(),
+                        ),
+                        max_value: Fixed::from_f64(ir_axis.max.into_inner().into_inner().into()),
+                        axis_name_id: axis_name_id.into(),
+                        ..Default::default()
+                    };
+                    if ir_axis.hidden {
+                        var.flags |= 0x0001;
```
cmyr, line 52: It would be nice if this was either a real flags struct or else we at least had constants defined for the various flags?

cmyr (approved):
Two small notes from me, otherwise looks good!

fontbe/src/fvar.rs
rsheeter: Hm. No, probably not. I think I would like to make IR capture Tag actually, but as a separate PR.

rsheeter: https://github.com/googlefonts/fontmake-rs/issues/214

fontbe/src/fvar.rs
rsheeter: Now a const.

fontbe/src/fvar.rs
```diff
@@ -0,0 +1,72 @@
...
+pub fn create_fvar_work() -> Box<BeWork> {
+    Box::new(FvarWork {})
+}
+
+impl Work<Context, Error> for FvarWork {
+    /// Generate [fvar](https://learn.microsoft.com/en-us/typography/opentype/spec/fvar)
+    fn exec(&self, context: &Context) -> Result<(), Error> {
+        let static_metadata = context.ir.get_init_static_metadata();
+        // Guard clause: don't produce fvar for a static font
+        if static_metadata.axes.is_empty() {
```
anthrotype, line 28: note that static_metadata.axes currently is currently never empty when building from a designspace, so we always emit an fvar even when the input contain no variations.
A designspace without any axes makes no sense of course, there's always going to be at least one.
It's not merely the presence of an axis that makes a font variable, but whether those axes contain any variations. E.g. take our static.designspace test file, it has one axis, one source (master) and one instance.. it's effectively a single UFO with a no-op accompanying designspace file whose only purpose is acting as an input for fontc (which currently only supports either .glyphs or .designspace inputs, no single .ufo).
When running fontc on such a static.designspace file, I would expect that no fvar be built, even if a Weight axis is declared in the designspace. That's because there are actually no variations for Weight in this designspace. You can detect this situation by either checking that an axis min==default==max (as in this particular case), or also by seeing if there are more than 1 source (or master) alongside the given axis; if there's only one, there can't be any variation.
I'm not sure about .glyphs file whether it defaults to no axes or implies at least weight, will need to test.

rsheeter: Quite right, I need to filter out point axes. Will send a followon.

PR #215, opened 2023-03-28
Use font_types definition of Tag for IR

### LINKED ISSUE #214: Make IR axis tags of type Tag and fail if they are invalid
See https://github.com/googlefonts/fontmake-rs/pull/211#discussion_r1149833640

rsheeter:
Fixes #214


dfrg (approved):
Lgtm. It might be reasonable to take an optional dependency on serde in font-types.

PR #216, opened 2023-03-28
Use font_types NameId

### LINKED ISSUE #196: Use font_types::NameId
It SHOUTS_AT_YOU but having two definitions is worse.

rsheeter:
Fixes #196

For some reason serde kept puking when I try to handle name id as a u16 so I made it a string. Ex `'Unable to deserialize "/tmp/.tmpRPdlUn/static_metadata.preliminary.yml" names.name_id: invalid type: integer `2`, expected A valid NameId at line 3 column 14'`

### DIFF
--- a/fontbe/src/fvar.rs
+++ b/fontbe/src/fvar.rs
@@ -41,13 +41,12 @@ impl Work<Context, Error> for FvarWork {
                 .axes
                 .iter()
                 .map(|ir_axis| {
-                    let axis_name_id: u16 = (*reverse_names.get(&ir_axis.name).unwrap()).into();
                     let mut var = VariationAxisRecord {
                         axis_tag: Tag::new(ir_axis.tag.as_bytes()),
                         min_value: ir_axis.min.into(),
                         default_value: ir_axis.default.into(),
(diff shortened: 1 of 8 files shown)


dfrg (approved):
Looks good. While reading, I realized that POSTSCRIPT should probably be POST_SCRIPT for consistency. I’ll fix both projects in another set of PRs.

khaledhosny:
Isn't PostScript is one word?

dfrg:
To me, the pascal casing as you’ve written it suggests it should include an underscore when converted to (screaming) snake case. We use `post_script` elsewhere in field/variable names.

rsheeter:
This PR is definiltely CHAD_YELLING_AT_ME

khaledhosny:
HarfBuzz uses `HB_OT_NAME_ID_POSTSCRIPT_NAME` and that feels more natural to me, `post_script` feels odd as well as that reads like the opposite of `pre_script`.

dfrg:
Fair enough. I’m happy to go the other way as long as we’re consistent.

PR #225, opened 2023-03-30
allow to set fontc --emit-ir=false

### LINKED ISSUE #224: [fontc] --emit-ir=false leads to TasksFailed([(Be(Font), "IO failure")])
fontc has an --emit-ir flag which is true by default, however it doesn't seem to be possible to disable it.
I tried --no-emit-r or --emit-ir=false but didn't work. Passing `--emit-ir` does nothing different than not passing one at all, since the default is already true.

Should the default be false maybe? Or maybe it can be true but use clap's ArgAction::SetFalse?

anthrotype:
Fixes #224 

credits to @rsheeter for the hint to using ArgAction::Set

> https://github.com/clap-rs/clap/issues/1649#issuecomment-1156638220 might offer a path

### DIFF
--- a/fontbe/Cargo.toml
+++ b/fontbe/Cargo.toml
@@ -32,7 +32,7 @@ write-fonts = "0.1.6"
 kurbo = { version = "0.9.2", features = ["serde"] }
 
 fea-rs = "0.3.1"
-smol_str = "0.1.18"
+smol_str = { version = "0.1.25", features = ["serde"] }
 
 [dev-dependencies]
 diff = "0.1.12"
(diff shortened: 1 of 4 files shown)


anthrotype:
hm why am I getting:

> the trait bound `SmolStr: Serialize` is not satisfied

when I run my pre-push git hook before pushing, it did _not_ catch this?!
Similarly if I `cargo clean` just to be safe, and then I run `cargo test --all-targets --all-features` like the CI also does, it PASSES on my local machine... 😕

rsheeter (approved):
Approving to see if I can reproduce a GH bug. Also I'm fine if you merge this when ready :)

Issue #146, opened 2023-03-14
Better defaults in post


rsheeter:
Set underline attrs per https://github.com/googlefonts/ufo2ft/blob/5b22f201b4650ab9ef1169357702a732e3127a5e/Lib/ufo2ft/fontInfoData.py#L313-L322

PR #229, opened 2023-03-31
build post format 2.0 containing glyph names by default

anthrotype:
Part of #146 

Only the last commit 413a1b15ff32796c937b0d333e60ba155540c51e is relevant for review; the previous two commits are here because the PR is based off `write-fonts-0-1-7` branch from PR #228 in order to include the commit d592106f5e852846e4435cc85fc8ff3502945cf0 that downgrades smol_str. Running tests off `main` is currently impossible since they v1.1.25 was yanked and no longer downloadable from crates.io. Therefore, #228 should be merged _before_ this PR.


dfrg (approved):
lgtm!

PR #230, opened 2023-04-01
Add get deltas method

rsheeter:
Intended use is to produce data to feed into a future `Gvar::from_glyph_variations` method that will be added in https://github.com/googlefonts/fontations/pull/319 or perhaps a follow-on.

Does not yet implement rounding properly; see https://github.com/googlefonts/fontmake-rs/issues/235.

### DIFF
--- a/fontir/src/variations.rs
+++ b/fontir/src/variations.rs
@@ -3,9 +3,9 @@ use std::{
     cmp::Ordering,
     collections::{HashMap, HashSet},
     fmt::{Debug, Display},
+    ops::{Mul, Sub},
 };
 
-use kurbo::{Point, Vec2};
 use log::{log_enabled, trace};
 use ordered_float::OrderedFloat;
 use serde::{Deserialize, Serialize};
@@ -90,13 +90,6 @@ impl VariationModel {
         let influence = master_influence(regions);
         let delta_weights = delta_weights(&locations, &influence);
 
-        let idx_of_location = locations
-            .iter()
-            .cloned()
-            .enumerate()
-            .map(|(idx, loc)| (loc, idx))
-            .collect();
-
         Ok(VariationModel {
             default,
             locations,
@@ -133,14 +126,21 @@ impl VariationModel {
     /// Note that it is NOT required to provide a point sequence for every location known
     /// to the variation model.
     ///
-    /// Returns a delta, in the form of a [Vec2], for every input point. Intended use is to support
+    /// P is the point type, meant to be absolute position in 1 or 2 dimensional space.
+    /// V is the vector type, such as [Vec2].
+    ///
+    /// Returns a delta, as the vector type, for every input point. Intended use is to support
     /// construction of a variation store.
     ///
     /// Rust version of <https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/varLib/models.py#L449-L461>
-    pub fn deltas(
+    pub fn deltas<P, V>(
         &self,
-        point_seqs: HashMap<NormalizedLocation, Vec<Point>>,
-    ) -> Result<HashMap<NormalizedLocation, Vec<Vec2>>, DeltaError> {
+        point_seqs: &HashMap<NormalizedLocation, Vec<P>>,
+    ) -> Result<HashMap<NormalizedLocation, Vec<V>>, DeltaError>
+    where
+        P: Copy + Default + Sub<P, Output = V>,
+        V: Copy + Mul<f64, Output = V> + Sub<V, Output = V>,
+    {
         let Some(defaults) = point_seqs.get(&self.default) else {
             return Err(DeltaError::DefaultUndefined);
         };
@@ -148,7 +148,7 @@ impl VariationModel {
             return Err(DeltaError::InconsistentNumbersOfPoints);
         }
 
-        let mut result: HashMap<NormalizedLocation, Vec<Vec2>> = HashMap::new();
+        let mut result: HashMap<NormalizedLocation, Vec<V>> = HashMap::new();
         // self.locations is sorted such that[i] is only influenced by[i+1..N]
         // so we know subsequent spins won't invalidate our delta if we go in the same order
         for (loc_idx, points) in self
@@ -161,6 +161,7 @@ impl VariationModel {
 
             let mut deltas = Vec::new();
             for (idx, point) in points.iter().enumerate() {
+                let initial_vector: V = *point - Default::default();
                 deltas.push(
                     // Find other masters that are active (have influence)
                     master_influences
@@ -169,7 +170,7 @@ impl VariationModel {
                         .filter_map(|(loc, w)| result.get(loc).map(|deltas| (deltas, w)))
                         .filter_map(|(deltas, w)| deltas.get(idx).map(|delta| (delta, w)))
                         // Start with a vector to our destination, then subtract away influence from other active masters
-                        .fold(point.to_vec2(), |acc, (other, other_weight)| {
+                        .fold(initial_vector, |acc, (other, other_weight)| {
                             acc - if other_weight != 1.0 {
                                 *other * other_weight.into()
                             } else {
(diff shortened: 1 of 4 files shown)


fontir/src/variations.rs
```diff
@@ -113,6 +121,75 @@ impl VariationModel {
...
+    /// Note that it is NOT required to provide a point sequence for every location known
+    /// to the variation model.
+    ///
+    /// Returns a delta, in the form of a [Vec2], for every input point. Intended use is to support
+    /// construction of a variation store.
+    ///
+    /// Rust version of <https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/varLib/models.py#L449-L461>
+    pub fn deltas(
+        &self,
+        point_seqs: HashMap<NormalizedLocation, Vec<Point>>,
```
anthrotype, line 142: maybe we could make this method generic. Right now this would work for Points only (hence, gvar), but the same method should be used for other single values that are not points or (x,y) tuples. The fonttools original doesn't mention "points" at all, only talks about "master values". When building a gvar, that method is fed GlyphCordinates which is equivalent to Vec<Point> and supports math operations like subtraction or multiplication by a scalar.

fontir/src/variations.rs
```diff
@@ -113,6 +121,75 @@ impl VariationModel {
...
+            for (idx, point) in points.iter().enumerate() {
+                deltas.push(
+                    // Find other masters that are active (have influence)
+                    master_influences
+                        .iter()
+                        .map(|(loc_idx, w)| (&self.locations[*loc_idx], w.into_inner()))
+                        .filter_map(|(loc, w)| result.get(loc).map(|deltas| (deltas, w)))
+                        .filter_map(|(deltas, w)| deltas.get(idx).map(|delta| (delta, w)))
+                        // Start with a vector to our destination, then subtract away influence from other active masters
+                        .fold(point.to_vec2(), |acc, (other, other_weight)| {
```
anthrotype, line 172: wow, I'm lost in the chain of maps...

fontir/src/variations.rs
justvanrossum: Yes. In practice, fonttools' VariationModel works for any object that supports `__sub__`, `__add__` and `__mul__`. This is enormously useful and IMO a key point in Behdad's excellent design.

rsheeter: I _think_ I can do the same but I didn't want to start there. Worst case I can give multiple methods, one for 2d and one for 1d ... hopefully fonts never use 3+.

fontir/src/variations.rs
rsheeter: Will add comments or go back to a for-loop.

fontir/src/variations.rs
rsheeter: Context: am Rust noob

rsheeter: std::ops sure looks like it'll work!

rsheeter: Now with test.

fontir/src/variations.rs
rsheeter: Better?

anthrotype: yes, thanks! :)

anthrotype (approved):
looks like a faithful port of the python original, so LGTM

fontir/src/variations.rs
```diff
@@ -113,6 +114,98 @@ impl VariationModel {
...
+                .enumerate()
+                .filter_map(|(loc_idx, loc)| {
+                    point_seqs
+                        .get(loc)
+                        .map(|points| (loc, points, &self.delta_weights[loc_idx]))
+                })
+        {
+            let mut deltas = Vec::new();
+            for (idx, point) in points.iter().enumerate() {
+                let initial_vector: V = *point - Default::default();
```
cmyr, line 169: I'm staring at this like an idiot: it feels like a convoluted way to convert from the input type to the vector type? Am I missing anything?

fontir/src/variations.rs
```diff
@@ -113,6 +114,98 @@ impl VariationModel {
...
+                                return None;
+                            };
+                            Some((delta, master_weight.into_inner()))
+                        })
+                        .fold(initial_vector, |acc, (other, other_weight)| {
+                            acc - if other_weight != 1.0 {
+                                *other * other_weight.into()
+                            } else {
+                                *other
+                            }
```
cmyr, line 193: does this branch do anything? if `other_weight` is 1.0, wouldn't the mul be a noop anyway?

fontir/src/variations.rs
```diff
@@ -113,6 +114,98 @@ impl VariationModel {
...
+    /// construction of a variation store.
+    ///
+    /// Rust version of <https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/varLib/models.py#L449-L461>
+    pub fn deltas<P, V>(
+        &self,
+        point_seqs: &HashMap<NormalizedLocation, Vec<P>>,
+    ) -> Result<HashMap<NormalizedLocation, Vec<V>>, DeltaError>
+    where
+        P: Copy + Default + Sub<P, Output = V>,
+        V: Copy + Mul<f64, Output = V> + Sub<V, Output = V>,
```
cmyr, line 145: I've been staring at this for twenty minutes, trying to come up with some compelling alternative, and I'm not having a lot of luck.


Just for the exercise of it, one possible solution I could imagine would be to actually use less fancy types here; we could use const generics to make the input and output of this method always be arrays, where we are generic over the length of the array (it would always be either 1 or 2).

(I went and implemented this and it wasn't really an improvement, so... idk)

fontir/src/variations.rs
```diff
@@ -113,6 +114,98 @@ impl VariationModel {
...
+    /// point primitive should suffice.
+    ///
+    /// Returns a delta, as the vector type, for every input point. Intended use is to support
+    /// construction of a variation store.
+    ///
+    /// Rust version of <https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/varLib/models.py#L449-L461>
+    pub fn deltas<P, V>(
+        &self,
+        point_seqs: &HashMap<NormalizedLocation, Vec<P>>,
+    ) -> Result<HashMap<NormalizedLocation, Vec<V>>, DeltaError>
```
cmyr, line 142: Would it make sense to have some sort of `LocationMap` abstraction here?

fontir/src/variations.rs
```diff
@@ -113,6 +114,98 @@ impl VariationModel {
...
+            }
+            result.insert(loc.clone(), deltas);
+        }
+
+        Ok(result)
+    }
+}
+
+#[derive(Debug)]
+pub enum DeltaError {
```
cmyr, line 205: If we're going to use this as an actual public error type we should impl `Error`. I'm open to using some third-party crate like `thiserror` to make this easier?

cmyr (approved):
Some notes inline: I find the signature of the `deltas` method confusing, but I have struggled to come up with a better approach, so 🤷

fontir/src/variations.rs
rsheeter: That's correct. There isn't a nice trait that works for the types - I was hoping for V: From<P> or some such - we want to use this for so the other option I came up with was to pass in a function that converts P => V. I think this is cleaner than that.

@dfrg wdyt about Vec2 having From<Point>, possibly also vice versa?

fontir/src/variations.rs
rsheeter: I dutifully copied that from FontTools. I will ask Behdad as I had a similar reaction but wasn't _quite_ sure it would truly nop.

fontir/src/variations.rs
rsheeter: Oh good catch, will add.

rsheeter: Done

fontir/src/variations.rs
rsheeter: I followed a similar arc. Ultimately I concluded this did express *exactly* the characteristics I want so ... perhaps this is the intrinsic complexity.

fontir/src/variations.rs
rsheeter: I waffled, it didn't seem like it would do anything more than a map would? I suppose it would afford a place to put comments if nothing else?

dfrg: A type alias might be a reasonable compromise?

fontir/src/variations.rs
rsheeter: The answer is to avoid allocation. I might remove this branch in #236 .

PR #231, opened 2023-04-02
Build a simple font w/o OTS warnings

### LINKED ISSUE #220: debug and fix a bunch of OTS warnings
```
$ cargo run -p fontc -- --source resources/testdata/static.designspace
$ python -m ots build/font.ttf
WARNING: bad table directory searchRange
WARNING: bad table directory entrySelector
WARNING: bad table directory rangeShift
WARNING: maxp: Bad maxZones: 0
WARNING: glyf: Number of contour points exceeds maxp maxPoints, adjusting limit.
WARNING: glyf: Number of contour points exceeds maxp maxPoints, adjusting limit.
File sanitized successfully!
```

I think maxp.maxZones should just be 1 by default, that's also [what ufo2ft does](https://github.com/googlefonts/ufo2ft/blob/0f123a791c62ccd09ad41992f6cd52ed21858bc2/Lib/ufo2ft/outlineCompiler.py#L1477).

As for the maxPoints, I think we currently don't set that so it's 0 but the glyf glyphs do contain some points so OTS complains.

The sftn table directory has the searchRange, entrySelector and rangeShift, similar to the ones we have for cmap.
https://learn.microsoft.com/en-us/typography/opentype/spec/otff#table-directory
I'm not sure where these are supposed to be set, I suppose the responsibility is whomever creates the table directory. I guess write-fonts should automatically compute those?

rsheeter:
Fixes #220 

```shell
$ cargo run -p fontc -- --source resources/testdata/static.designspace
$ python -m ots build/font.ttf
File sanitized successfully!
```

Credit @madig for the nudge (https://github.com/googlefonts/fontations/issues/314) to get this done.

`fontbe/src/metrics_and_limits.rs` is not entirely new as the diff suggests, only the maxp part plus some shufffling to make that easier is changed.


fontbe/src/metrics_and_limits.rs
```diff
@@ -0,0 +1,204 @@
...
+            .map(|v| max(v, advance))
+            .or(Some(advance));
+
+        match glyph {
+            Glyph::Simple(simple) => {
+                let max_points = simple
+                    .contours()
+                    .iter()
+                    // TODO: this is stupid, just let me see the damn points
+                    .map(|c| c.clone().into_iter().count())
```
cmyr, line 71: I agree this is stupid, maybe just add a `len` or `num_points` method to `Contour`?

fontbe/src/metrics_and_limits.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+#[derive(Debug, Default)]
+struct GlyphLimits {
+    min_left_side_bearing: Option<i16>,
+    min_right_side_bearing: Option<u16>,
+    x_max_extent: Option<i16>,
+    advance_width_max: Option<u16>,
+    max_points: Option<u16>,
+    max_contours: Option<u16>,
```
cmyr, line 32: maybe a lot of these fields don't need to be `Option`? would clear up a bunch of the other code...

fontbe/src/metrics_and_limits.rs
```diff
@@ -0,0 +1,204 @@
...
+                    .contours()
+                    .iter()
+                    // TODO: this is stupid, just let me see the damn points
+                    .map(|c| c.clone().into_iter().count())
+                    .max()
+                    .unwrap_or_default() as u16;
+                self.max_points = self
+                    .max_points
+                    .map(|v| max(v, max_points))
+                    .or(Some(max_points));
```
cmyr, line 77: I might write this as,
```suggestion
                self.max_points = Some(self.max_points.unwrap_or_default().max(max_points));
```

cmyr (approved):
It feels like a lot of the logic here is more fiddly and annoying than it should be, but overall seems reasonable.

fontbe/src/metrics_and_limits.rs
rsheeter: Definitely, I just didn't want to block on that

fontbe/src/metrics_and_limits.rs
rsheeter: I think you are right. I made them option to match maxp but we would branch at a higher level if we were trying to emit a maxp v0.5 and they are mandatory for 1.0.

fontbe/src/metrics_and_limits.rs
rsheeter: Hm. Maybe that already exists lol.

fontbe/src/metrics_and_limits.rs
rsheeter: Fixed.

fontbe/src/metrics_and_limits.rs
rsheeter: This became nicer by not using Option per your other suggestion

fontc/src/lib.rs
```diff
@@ -502,7 +502,7 @@ fn add_os2_be_job(
     Ok(())
 }
 
-fn add_hmetrics_job(
+fn add_metric_and_limits_job(
```
anthrotype, line 505: sometimes you use plural for both metrics and limits (as in the module name), other times singlular for both as in 
"create_metric_and_limit_work" function, othertimes singular for one and plural for the other; I know it doesn't matter much from a meaning point of view but it is a bit distracting. If you remove one item and only name "metrics", without the "_and_limits" there's one less thing to worry about. You can document that updating the metrics has side effect of also updating other relaated maximum limits.

fontbe/src/metrics_and_limits.rs
```diff
@@ -0,0 +1,186 @@
...
+
+impl GlyphLimits {
+    fn update(&mut self, advance: u16, glyph: &Glyph) {
+        // min side bearings are only for non-empty glyphs
+        // we will presume only simple glyphs with no contours are empty
+        if glyph.is_empty() {
+            return;
+        }
+
+        let bbox = glyph.bbox();
```
anthrotype, line 44: since you are getting the glyph bboxes to do these limit computations, you may as well also collect maximum bounds (union of all non-empty glyph's bboxes) and use that to set the [head](https://learn.microsoft.com/en-us/typography/opentype/spec/head) table font-wide bounds (head.xMin, yMin, xMax, yMax).

PR #232, opened 2023-04-03
[fontbe/head] set created/modified to current timestamp or SOURCE_DATE_EPOCH

anthrotype:
when opening a font compiled with fontc with fontTools.ttLib.TTFont, we get a warning because the created/modified fields are set to 0 and the font looks suspiciously too old.

Here I changed it such that we match fontmake's default behavior: i.e. set those two head fields to the current timestamp (as seconds since the TrueType epoch), or to `SOURCE_DATE_EPOCH` if the latter environment variable is present.
This is common way to ensure reproducible builds and fontmake also supports that: see https://reproducible-builds.org/specs/source-date-epoch/

Note that the head.created in particular (not the head.modified) can also be controlled via a 'date' font attribute in .glyphs source, which is equivalent to 'openTypeHeadCreated' from UFO fontinfo.plist. However for now I added a TODO for that, we can implement that later (I filed a separate issue lest we forget https://github.com/googlefonts/fontmake-rs/issues/233).

### DIFF
--- a/fontbe/Cargo.toml
+++ b/fontbe/Cargo.toml
@@ -26,6 +26,7 @@ env_logger = "0.10.0"
 
 parking_lot = "0.12.1"
 
+font-types = "0.1.3"
 read-fonts = "0.1.3"
 write-fonts = "0.1.8"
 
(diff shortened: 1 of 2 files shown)


rsheeter (approved):
LGTM!

anthrotype:
for the record, this is the warning we get from fontTools when we don't set these fields:

https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/ttLib/tables/_h_e_a_d.py#L61-L65

```python
            if value < 0x7C259DC0:  # January 1, 1970 00:00:00
                log.warning(
                    "'%s' timestamp seems very low; regarding as unix timestamp", stamp
                )
                value += 0x7C259DC0
```

PR #234, opened 2023-04-04
Filename does not align with a single location

rsheeter:
Test to expose bug + fix


anthrotype:
yes, source.name is supposed to be unique and can be used as the key in the map. multiple sources may point to the same UFO filename but have different layer name or locations.

anthrotype (approved):

rsheeter:
Per f2f, in https://github.com/googlefonts/fontmake-rs/blob/f5d631ec66560d2551114b0c6f610ce1839cc0bc/ufo2fontir/src/source.rs#L622 we should *not* establish a metrics record (which would in time feed mvar, etc) if the source has a layer attribute.

PR #236, opened 2023-04-04
Compile simple VFs w/gvar that works in a browser

### LINKED ISSUE #95: Implement BE work for a simple variable glyph
https://github.com/googlefonts/fontations/issues/207 requests an example of doing this w/fontations.

See:

* https://github.com/googlefonts/oxidize/blob/main/text/2022-07-25-PROPOSAL-build-glyphs-in-rust.md
* https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/varLib/__init__.py#L238
* https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/varLib/models.py#L204

### LINKED ISSUE #207: Emit gvar
Produce a gvar table. See https://github.com/googlefonts/oxidize/blob/main/text/2022-07-25-PROPOSAL-build-glyphs-in-rust.md for notes.

### LINKED ISSUE #245: Build a valid gvar for Texturina
Compile Texturina with https://github.com/googlefonts/fontmake-rs/pull/236 and https://github.com/googlefonts/fontations/pull/351 in place, e.g. `cargo run -p fontc -- --source smoke_test/build/texturina/sources/Texturina.glyphs`

Ttx gvar:

```shell
$ ttx -o - -t gvar build/font.ttf
# prints some tuple's then boom
ERROR: Unhandled exception has occurred
Traceback (most recent call last):
  File "venv/lib/python3.10/site-packages/fontTools/ttx.py", line 405, in main
    process(jobs, options)
  File "venv/lib/python3.10/site-packages/fontTools/ttx.py", line 387, in process
    action(input, output, options)
  File "venv/lib/python3.10/site-packages/fontTools/misc/loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "venv/lib/python3.10/site-packages/fontTools/ttx.py", line 278, in ttDump
    ttf.saveXML(output,
  File "venv/lib/python3.10/site-packages/fontTools/ttLib/ttFont.py", line 241, in saveXML
    self._saveXML(writer, **kwargs)
  File "venv/lib/python3.10/site-packages/fontTools/ttLib/ttFont.py", line 296, in _saveXML
    self._tableToXML(tableWriter, tag, splitGlyphs=splitGlyphs)
  File "venv/lib/python3.10/site-packages/fontTools/ttLib/ttFont.py", line 327, in _tableToXML
    table.toXML(writer, self)
  File "venv/lib/python3.10/site-packages/fontTools/ttLib/tables/_g_v_a_r.py", line 190, in toXML
    variations = self.variations.get(glyphName)
  File "/usr/lib/python3.10/_collections_abc.py", line 824, in get
    return self[key]
  File "venv/lib/python3.10/site-packages/fontTools/ttLib/tables/_g_v_a_r.py", line 49, in __getitem__
    v = v()
  File "venv/lib/python3.10/site-packages/fontTools/ttLib/tables/_g_v_a_r.py", line 122, in decompileVarGlyph
    return decompileGlyph_(numPointsInGlyph, sharedCoords, axisTags, gvarData)
  File "venv/lib/python3.10/site-packages/fontTools/ttLib/tables/_g_v_a_r.py", line 254, in decompileGlyph_
    return tv.decompileTupleVariationStore(
  File "venv/lib/python3.10/site-packages/fontTools/ttLib/tables/TupleVariation.py", line 707, in decompileTupleVariationStore
    result.append(decompileTupleVariation_(
  File "venv/lib/python3.10/site-packages/fontTools/ttLib/tables/TupleVariation.py", line 751, in decompileTupleVariation_
    deltas_x, pos = TupleVariation.decompileDeltas_(
  File "venv/lib/python3.10/site-packages/fontTools/ttLib/tables/TupleVariation.py", line 474, in decompileDeltas_
    assert len(result) == numDeltas
AssertionError
```

Make it happy.

rsheeter:
Builds on #230, #239. Fixes #207. Fixes #95. 

Update fontations to pickup gvar fixes and thereby fixes #245.

`cargo run -p fontc -- --source resources/testdata/mov_xy.designspace` now produces a VF that works in a browser.

I have punted some feedback to issues as I believe just compiling any vf that works is worth capturing, specifically:

1. https://github.com/googlefonts/fontmake-rs/issues/240
1. https://github.com/googlefonts/fontmake-rs/issues/241

Sufficient to compile an at least somewhat functional Oswald:

```shell
$ git clone https://github.com/googlefonts/OswaldFont
$ cd fontmake-rs
$ cargo run -p fontc -- --source ../OswaldFont/sources/Oswald.glyphs
```

In browser:

image

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -216,7 +216,7 @@ impl Work<Context, Error> for GlyphWork {
             }
         };
 
-        // Everybody loves gvar!
+        // Contour (aka Simple) and Composite both need gvar
         let deltas = var_model
             .deltas(&point_seqs)
             .map_err(|e| Error::GlyphDeltaError(self.glyph_name.clone(), e))?;
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -157,10 +157,13 @@ fn point_seqs_for_composite_glyph(ir_glyph: &ir::Glyph) -> HashMap<NormalizedLoc
         .sources()
         .iter()
         .map(|(loc, inst)| {
+            // We need 1 point per component for it's X/Y, plus phantoms
+            // See https://github.com/fonttools/fonttools/blob/1c283756a5e39d69459eea80ed12792adc4922dd/Lib/fontTools/ttLib/tables/_g_v_a_r.py#L243
             let mut points = Vec::new();
-
-            // TODO: composites arguably need more than just phantoms
-
+            for component in inst.components.iter() {
+                let [.., dx, dy] = component.transform.as_coeffs();
+                points.push((dx, dy).into());
+            }
             add_phantom_points(inst.width.ot_round(), &mut points);
 
             (loc.clone(), points)
@@ -174,7 +177,7 @@ impl Work<Context, Error> for GlyphWork {
 
         let static_metadata = context.ir.get_final_static_metadata();
         let var_model = &static_metadata.variation_model;
-        let default_location = var_model.default_location();
+        let default_location = static_metadata.default_location();
         let ir_glyph = &*context.ir.get_glyph_ir(&self.glyph_name);
         let glyph: CheckedGlyph = ir_glyph.try_into()?;
 
@@ -217,6 +220,7 @@ impl Work<Context, Error> for GlyphWork {
         let deltas = var_model
             .deltas(&point_seqs)
             .map_err(|e| Error::GlyphDeltaError(self.glyph_name.clone(), e))?;
+
         context.set_gvar_fragment(name, GvarFragment { deltas });
 
         Ok(())
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -1,28 +1,34 @@
-//! 'glyf' Glyph binary compilation
+//! 'glyf' and 'gvar' compilation
+//!
+//! Each glyph is built in isolation and then the fragments are collected
+//! and glued together to form a final table.
 
 use std::collections::{BTreeSet, HashMap, HashSet};
 
 use fontdrasil::{orchestration::Work, types::GlyphName};
 use fontir::{coords::NormalizedLocation, ir};
-use kurbo::{cubics_to_quadratic_splines, Affine, BezPath, CubicBez, PathEl, Rect};
+use kurbo::{cubics_to_quadratic_splines, Affine, BezPath, CubicBez, PathEl, Point, Rect};
 use log::{trace, warn};
 
 use read_fonts::{
     tables::glyf::{self, Anchor, Transform},
     types::{F2Dot14, GlyphId},
 };
-use write_fonts::tables::glyf::{Bbox, Component, ComponentFlags, CompositeGlyph, SimpleGlyph};
+use write_fonts::{
+    tables::glyf::{Bbox, Component, ComponentFlags, CompositeGlyph, SimpleGlyph},
+    OtRound,
+};
 
 use crate::{
     error::{Error, GlyphProblem},
-    orchestration::{BeWork, Context, Glyph},
+    orchestration::{BeWork, Context, Glyph, GvarFragment},
 };
 
 struct GlyphWork {
     glyph_name: GlyphName,
 }
 
-pub fn create_glyph_work(glyph_name: GlyphName) -> Box<BeWork> {
+pub fn create_glyf_work(glyph_name: GlyphName) -> Box<BeWork> {
     Box::new(GlyphWork { glyph_name })
 }
 
@@ -110,6 +116,58 @@ fn create_composite(
     Ok(composite.unwrap())
 }
 
+/// * <https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L335-L367>
+/// * <https://docs.microsoft.com/en-us/typography/opentype/spec/tt_instructing_glyphs#phantoms>
+fn add_phantom_points(advance: u16, points: &mut Vec<Point>) {
+    // FontTools says
+    //      leftSideX = glyph.xMin - leftSideBearing
+    //      rightSideX = leftSideX + horizontalAdvanceWidth
+    // We currently always set lsb to xMin so leftSideX = 0, rightSideX = advance.
+    points.push(Point::new(0.0, 0.0)); // leftSideX, 0
+    points.push(Point::new(advance as f64, 0.0)); // rightSideX, 0
+
+    // TODO: vertical phantom points
+    points.push(Point::new(0.0, 0.0));
+    points.push(Point::new(0.0, 0.0));
+}
+
+fn point_seqs_for_simple_glyph(
+    ir_glyph: &ir::Glyph,
+    instances: HashMap<NormalizedLocation, SimpleGlyph>,
+) -> HashMap<NormalizedLocation, Vec<Point>> {
+    instances
+        .into_iter()
+        .map(|(loc, glyph)| {
+            let mut points = glyph
+                .contours()
+                .iter()
+                .flat_map(|c| c.iter())
+                .map(|cp| Point::new(cp.x as f64, cp.y as f64))
+                .collect();
+
+            add_phantom_points(ir_glyph.sources()[&loc].width.ot_round(), &mut points);
+
+            (loc, points)
+        })
+        .collect()
+}
+
+fn point_seqs_for_composite_glyph(ir_glyph: &ir::Glyph) -> HashMap<NormalizedLocation, Vec<Point>> {
+    ir_glyph
+        .sources()
+        .iter()
+        .map(|(loc, inst)| {
+            let mut points = Vec::new();
+
+            // TODO: composites arguably need more than just phantoms
+
+            add_phantom_points(inst.width.ot_round(), &mut points);
+
+            (loc.clone(), points)
+        })
+        .collect()
+}
+
 impl Work<Context, Error> for GlyphWork {
     fn exec(&self, context: &Context) -> Result<(), Error> {
         trace!("BE glyph work for {}", self.glyph_name);
@@ -123,33 +181,50 @@ impl Work<Context, Error> for GlyphWork {
         // Hopefully in time https://github.com/harfbuzz/boring-expansion-spec means we can drop this
         let glyph = cubics_to_quadratics(glyph);
 
-        // TODO refine (submodel) var model if glyph locations is a subset of var model locations
-
-        match glyph {
+        let (name, point_seqs) = match glyph {
             CheckedGlyph::Composite { name, components } => {
                 let composite = create_composite(context, &name, default_location, &components)?;
-                context.set_glyph(name, composite.into());
+                context.set_glyph(name.clone(), composite.into());
+                (name, point_seqs_for_composite_glyph(ir_glyph))
             }
-            CheckedGlyph::Contour { name, contours } => {
-                // Draw the default outline of our simple glyph
-                let Some(path) = contours.get(default_location) else {
+            CheckedGlyph::Contour { name, paths } => {
+                // Convert paths to SimpleGlyph so we can get consistent point streams
+                let instances = paths
+                    .into_iter()
+                    .map(|(loc, path)| {
+                        match SimpleGlyph::from_kurbo(&path).map_err(|e| Error::KurboError {
+                            glyph_name: self.glyph_name.clone(),
+                            kurbo_problem: e,
+                            context: path.to_svg(),
+                        }) {
+                            Ok(glyph) => Ok((loc, glyph)),
+                            Err(e) => Err(e),
+                        }
+                    })
+                    .collect::<Result<HashMap<_, _>, Error>>()?;
+
+                // Establish the default outline of our simple glyph
+                let Some(base_glyph) = instances.get(default_location) else {
                     return Err(Error::GlyphError(ir_glyph.name.clone(), GlyphProblem::MissingDefault));
                 };
-                let base_glyph = SimpleGlyph::from_kurbo(path).map_err(|e| Error::KurboError {
-                    glyph_name: self.glyph_name.clone(),
-                    kurbo_problem: e,
-                    context: path.to_svg(),
-                })?;
-                context.set_glyph(name, base_glyph.into());
+                context.set_glyph(name.clone(), base_glyph.clone().into());
+
+                (name, point_seqs_for_simple_glyph(ir_glyph, instances))
             }
-        }
+        };
+
+        // Everybody loves gvar!
+        let deltas = var_model
+            .deltas(&point_seqs)
+            .map_err(|e| Error::GlyphDeltaError(self.glyph_name.clone(), e))?;
+        context.set_gvar_fragment(name, GvarFragment { deltas });
 
         Ok(())
     }
 }
 
 fn cubics_to_quadratics(glyph: CheckedGlyph) -> CheckedGlyph {
-    let CheckedGlyph::Contour { name, contours } = glyph else {
+    let CheckedGlyph::Contour { name, paths: contours } = glyph else {
         return glyph;  // nop for composite
     };
 
@@ -256,7 +331,7 @@ fn cubics_to_quadratics(glyph: CheckedGlyph) -> CheckedGlyph {
 
     CheckedGlyph::Contour {
         name,
-        contours: new_contours,
+        paths: new_contours,
     }
 }
 
@@ -273,7 +348,7 @@ enum CheckedGlyph {
     },
     Contour {
         name: GlyphName,
-        contours: HashMap<NormalizedLocation, BezPath>,
+        paths: HashMap<NormalizedLocation, BezPath>,
     },
 }
 
@@ -356,7 +431,10 @@ impl TryFrom<&ir::Glyph> for CheckedGlyph {
                     (location.clone(), path)
                 })
                 .collect();
-            CheckedGlyph::Contour { name, contours }
+            CheckedGlyph::Contour {
+                name,
+                paths: contours,
+            }
         } else {
             let components = glyph
                 .sources()
--- a/fontbe/src/gvar.rs
+++ b/fontbe/src/gvar.rs
@@ -87,15 +87,17 @@ impl Work<Context, Error> for GvarWork {
         // We built the gvar fragments alongside glyphs, now we need to glue them together into a gvar table
         let static_metadata = context.ir.get_final_static_metadata();
 
+        let mut gids_and_deltas = Vec::new();
+
         let variations: Vec<_> = static_metadata
             .glyph_order
             .iter()
             .enumerate()
             .map(|(gid, gn)| {
-                (
-                    GlyphId::new(gid as u16),
-                    to_deltas(&context.get_gvar_fragment(gn)),
-                )
+                let gid = GlyphId::new(gid as u16);
+                let gvar_fragment = context.get_gvar_fragment(gn);
+                gids_and_deltas.push((gid, gvar_fragment.deltas.clone()));
+                (gid, to_deltas(&gvar_fragment))
             })
             .map(|(gid, deltas)| GlyphVariations::new(gid, deltas))
             .collect();
--- a/fontbe/src/gvar.rs
+++ b/fontbe/src/gvar.rs
@@ -1,21 +1,15 @@
 //! Generates a [gvar](https://learn.microsoft.com/en-us/typography/opentype/spec/gvar) table.
 
-use log::trace;
-
-use font_types::{F2Dot14, GlyphId};
+use font_types::GlyphId;
 use fontdrasil::orchestration::Work;
-use fontir::variations::VariationRegion;
 use write_fonts::{
     dump_table,
-    tables::{
-        gvar::{GlyphDeltas, GlyphVariations, Gvar},
-        variations::Tuple,
-    },
+    tables::gvar::{GlyphVariations, Gvar},
 };
 
 use crate::{
     error::Error,
-    orchestration::{BeWork, Bytes, Context, GvarFragment},
+    orchestration::{BeWork, Bytes, Context},
 };
 
 struct GvarWork {}
@@ -24,70 +18,11 @@ pub fn create_gvar_work() -> Box<BeWork> {
     Box::new(GvarWork {})
 }
 
-/// <https://learn.microsoft.com/en-us/typography/opentype/spec/otvaroverview#variation-data>
-#[derive(Debug, Default)]
-struct TupleBuilder {
-    axis_names: Vec<String>,
-    min: Vec<F2Dot14>,
-    peak: Vec<F2Dot14>,
-    max: Vec<F2Dot14>,
-}
-
-impl TupleBuilder {
-    fn build(self) -> (Tuple, Tuple, Tuple) {
-        (
-            Tuple::new(self.min),
-            Tuple::new(self.peak),
-            Tuple::new(self.max),
-        )
-    }
-}
-
-impl From<&VariationRegion> for TupleBuilder {
-    fn from(region: &VariationRegion) -> Self {
-        let mut builder = TupleBuilder::default();
-        for (axis_name, tent) in region.iter() {
-            builder.axis_names.push(axis_name.clone());
-            builder.min.push(F2Dot14::from_f32(tent.lower.to_f32()));
-            builder.peak.push(F2Dot14::from_f32(tent.peak.to_f32()));
-            builder.max.push(F2Dot14::from_f32(tent.upper.to_f32()));
-        }
-        trace!("{builder:?}");
-        builder
-    }
-}
-
-fn to_deltas(fragment: &GvarFragment) -> Vec<GlyphDeltas> {
-    fragment
-        .deltas
-        .iter()
-        .filter_map(|(region, deltas)| {
-            if region.is_default() {
-                return None;
-            }
-
-            // Variation of no point has limited entertainment value
-            if deltas.is_empty() {
-                return None;
-            }
-
-            // TODO: nice rounding on deltas
-            let deltas: Vec<_> = deltas.iter().map(|v| (v.x as i16, v.y as i16)).collect();
-
-            let tuple_builder: TupleBuilder = region.into();
-            let (min, peak, max) = tuple_builder.build();
-            Some(GlyphDeltas::new(peak, deltas, Some((min, max))))
-        })
-        .collect()
-}
-
 impl Work<Context, Error> for GvarWork {
     /// Generate [gvar](https://learn.microsoft.com/en-us/typography/opentype/spec/gvar)
     fn exec(&self, context: &Context) -> Result<(), Error> {
         // We built the gvar fragments alongside glyphs, now we need to glue them together into a gvar table
         let static_metadata = context.ir.get_final_static_metadata();
-
-        let mut gids_and_deltas = Vec::new();
 
         let variations: Vec<_> = static_metadata
             .glyph_order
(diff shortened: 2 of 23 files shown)


fontc/src/lib.rs
```diff
@@ -799,21 +839,26 @@ mod tests {
...
-                BeWorkIdentifier::Glyph("plus".into()).into(),
                 BeWorkIdentifier::Avar.into(),
                 BeWorkIdentifier::Cmap.into(),
                 BeWorkIdentifier::Fvar.into(),
                 BeWorkIdentifier::Glyf.into(),
+                BeWorkIdentifier::GlyfFragment("bar".into()).into(),
+                BeWorkIdentifier::GlyfFragment("plus".into()).into(),
+                BeWorkIdentifier::Gvar.into(),
+                BeWorkIdentifier::GvarFragment("bar".into()).into(),
+                BeWorkIdentifier::GvarFragment("plus".into()).into(),
```
anthrotype, line 861: why are these duplicated here?

fontbe/src/glyphs.rs
```diff
@@ -123,33 +127,100 @@ impl Work<Context, Error> for GlyphWork {
...
+
+    fn curve_to(&mut self, cx0: f32, cy0: f32, cx1: f32, cy1: f32, x: f32, y: f32) {
+        self.points.push(as_kurbo_point(cx0, cy0));
+        self.points.push(as_kurbo_point(cx1, cy1));
+        self.points.push(as_kurbo_point(x, y));
+    }
+
+    fn close(&mut self) {
+        if let Some(last_move) = self.last_move {
+            self.points.push(last_move);
```
anthrotype, line 217: why are you unconditionally adding this extra move point for closed contours? this deviates from the way fonttools 
converter pens (i.e. `fontTools.pointPen.{PointToSegmentPen,SegmentToPointPen}`) translate between segments and points...
Is that because you want to make implied closing line always explicit?

anthrotype: E.g. say a contour has a triangle with three on-curve points at each corner. After converting it to a BezPath with our GlyphPathBuilder and then back to points using your current PointPen, we end up with an additional point which was not there in the source!

```rust

#[cfg(test)]
mod tests {
    use fontir::ir::GlyphPathBuilder;
    use kurbo::Point;

    use super::points;

    #[test]
    fn test_roundtrip_points_to_segments_to_points() {
        let mut path_builder = GlyphPathBuilder::new("a".into());
        path_builder.line_to((0.0, 0.0)).unwrap();
        path_builder.line_to((1.0, 0.0)).unwrap();
        path_builder.line_to((1.0, 1.0)).unwrap();
        path_builder.close_path().unwrap();
        let path = path_builder.build();
        println!("{:?}", path);
        let path_points = points(path);
        println!("{:?}", path_points);

        assert_eq!(
            path_points,
            vec![
                Point::new(0.0, 0.0),
                Point::new(1.0, 0.0),
                Point::new(1.0, 1.0),
            ]
        );
    }
}
```

```sh
$ cargo test test_roundtrip_points
   Compiling fontbe v0.0.1 (/usr/local/google/home/clupo/oss/fontmake-rs/fontbe)
   Compiling fontc v0.0.1 (/usr/local/google/home/clupo/oss/fontmake-rs/fontc)
    Finished test [unoptimized + debuginfo] target(s) in 9.16s
     Running unittests src/lib.rs (target/debug/deps/fontbe-e2c6386d7cb37542)

running 1 test
test glyphs::tests::test_roundtrip_points_to_segments_to_points ... FAILED

failures:

---- glyphs::tests::test_roundtrip_points_to_segments_to_points stdout ----
BezPath([MoveTo((0.0, 0.0)), LineTo((1.0, 0.0)), LineTo((1.0, 1.0)), ClosePath])
[(0.0, 0.0), (1.0, 0.0), (1.0, 1.0), (0.0, 0.0)]
thread 'glyphs::tests::test_roundtrip_points_to_segments_to_points' panicked at 'assertion failed: `(left == right)`
  left: `[(0.0, 0.0), (1.0, 0.0), (1.0, 1.0), (0.0, 0.0)]`,
 right: `[(0.0, 0.0), (1.0, 0.0), (1.0, 1.0)]`', fontbe/src/glyphs.rs:632:9


failures:
    glyphs::tests::test_roundtrip_points_to_segments_to_points

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 9 filtered out; finished in 0.00s

error: test failed, to rerun pass `-p fontbe --lib`
```

fontbe/src/glyphs.rs
```diff
@@ -123,33 +127,100 @@ impl Work<Context, Error> for GlyphWork {
...
                     glyph_name: self.glyph_name.clone(),
                     kurbo_problem: e,
                     context: path.to_svg(),
                 })?;
-                context.set_glyph(name, base_glyph.into());
+                context.set_glyph(name.clone(), base_glyph.into());
+
+                let point_seqs: HashMap<_, _> = paths
+                    .into_iter()
+                    .map(|(loc, bez)| (loc, points(bez)))
```
anthrotype, line 149: since we are already converting BezPath segments to points inside `SimpleGlyph::from_kurbo` immediately before this, it would make sense to do that once (inside write_fonts::tables::glyf) and reuse the same list of points for the gvar deltas.

By the way, I think you are missing the additional four phantom points at the end.
cf. fonttools' `_getCoordinatesAndControls` method of glyf table, used by varLib when building gvar: https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L369

fontbe/src/glyphs.rs
anthrotype: please take a look at this wip branch of mine (based off yours) for a way to resolve the issue of roundtripping points => BezPath => points.
To match fonttools behavior, we need to change the GlyphPathBuilder to emit the implied closing line, and then in your PointPen (or in the SimpleGlyph::from_kurbo, wherever we decide to do this), we need to pop the last point of a closed contour when it's the same as the move point:

https://github.com/googlefonts/fontmake-rs/compare/gvar2...gvar3-wip?expand=1

anthrotype: (I don't plan to have that branch merged, feel free to cherry pick or adapt from it)

anthrotype: I actually extracted some commits from that gvar3-wip branch of mine and made https://github.com/googlefonts/fontmake-rs/pull/239

rsheeter: Ha, by the time I read it you fixed it, that's awesome. #239 lgtm, I'm going to rebase on your branch.

fontc/src/lib.rs
rsheeter: Do you mean Glyf  vs Gvar fragments? Or am I misinterpreting what's beng duplicated?

fontbe/src/glyphs.rs
rsheeter: Switched to convert everything to SimpleGlyph then compute deltas. Haven't had time to look at 

I can now write a gvar that FontTools can't understand.

> missing the additional four phantom points at the end

Confirmed, not implemented yet :)

fontc/src/lib.rs
anthrotype: oh sorry, I misread.

fontbe/src/gvar.rs
```diff
@@ -0,0 +1,101 @@
...
+fn to_deltas(fragment: &GvarFragment) -> Vec<GlyphDeltas> {
+    fragment
+        .deltas
+        .iter()
+        .filter_map(|(loc, deltas)| {
+            if !loc.has_any_non_zero() {
+                return None;
+            }
+
+            // TODO: add phantoms properly
```
anthrotype, line 44: this is where fontTools computes the phantom points (needs hmtx)

https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L335-L367

fontbe/src/gvar.rs
```diff
@@ -0,0 +1,101 @@
...
+                (x, y)
+            }).collect();
+
+            // TODO: add phantoms properly
+            deltas.push((min_x, 0));
+            deltas.push((max_x, 0));
+            deltas.push((0, min_y));
+            deltas.push((0, max_y));
+
+            // TODO: what exactly is an intermediate region? Can we ignore it and let Gvar solve for it?
```
anthrotype, line 68: > an intermediate region

it's the most general one, with three start/peak/end values; the special case with a single peak value is basically equivalent to start=0 and peak=end.
it's all described here: https://learn.microsoft.com/en-us/typography/opentype/spec/otvaroverview#variation-data

fontbe/src/gvar.rs
```diff
@@ -0,0 +1,101 @@
...
+            Some(GlyphDeltas::new(to_tuple(loc), deltas, None))
```
anthrotype, line 69: no this is not correct. You can't determine the regions from the locations, the variation model does that, you should be able to get them from it. These are stored in the VariationModel.supports attribute, I think you already have computed those. Deltas are associated with regions, not with locations (the latter convey only the peaks, not the start/end which are only present in the regions' tent definitions), only the absolute master values can be said to have a location.
The python VariationModel has a getDeltasAndSupports method that returns a 2-tuple with the detals (from getDeltas) and the model.supports. You need _both_ to construct gvar tuple variations, as well as any other variation table.

fontc/src/lib.rs
```diff
@@ -347,6 +348,39 @@ fn add_fvar_be_job(
...
+) -> Result<(), Error> {
+    let glyphs_changed = change_detector.glyphs_changed();
+
+    // If no glyph has changed there isn't a lot of merging to do
+    if !glyphs_changed.is_empty() {
+        let mut dependencies: HashSet<_> = glyphs_changed
+            .iter()
+            .map(|gn| BeWorkIdentifier::GvarFragment(gn.clone()).into())
+            .collect();
+        dependencies.insert(FeWorkIdentifier::FinalizeStaticMetadata.into());
```
anthrotype, line 363: you'll probably need to add the glyph metrics to be able to compute phantom points

fontbe/src/glyphs.rs
```diff
@@ -123,24 +123,51 @@ impl Work<Context, Error> for GlyphWork {
         // Hopefully in time https://github.com/harfbuzz/boring-expansion-spec means we can drop this
         let glyph = cubics_to_quadratics(glyph);
 
-        // TODO refine (submodel) var model if glyph locations is a subset of var model locations
```
anthrotype, line 126: you removed this TODO but it doesn't seem like you implemented the [getSubModel](https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/varLib/__init__.py#L273) method to support sparse masters

fontbe/src/glyphs.rs
```diff
@@ -123,24 +123,51 @@ impl Work<Context, Error> for GlyphWork {
...
+                            .iter()
+                            .flat_map(|c| c.iter())
+                            .map(|cp| Point::new(cp.x as f64, cp.y as f64))
+                            .collect();
+                        (loc, points)
+                    })
+                    .collect();
+
+                let deltas = var_model
+                    .deltas(&point_seqs)
```
anthrotype, line 167: (I thought I had commented on this but it got lost, I'll rephrase)

We want to pass the phantom points to the var_model to compute deltas for them.

Also before computing deltas, we want to check that the on/off-curve flags are identical for simple glyphs; for composites, the component glyph name and the 2x2 transform mast be the same (only x,y offset can vary).
Fonttools currently skips and issues a warning for incompatible glyphs, we might do the same.
https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/varLib/__init__.py#L276-L280

anthrotype: not just the on/off curve flags, actually, but also the number of contours and the list of end point indices for each contour. Basically what fontTools calls _GlyphControls here:
https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L375-L383

fontbe/src/glyphs.rs
anthrotype: oh I see, the VariationModel::deltas method takes a map of master values keyed by location, which in theory can be "sparse" (not contain values for all the locations, but only some), and then only those actually participate in the delta computation, so perhaps you don't actually need to make a sub-model; is that it?

anthrotype: I just realised this is even documented in the `deltas` method's docstring:

```
/// Note that it is NOT required to provide a point sequence for every location known
/// to the variation model.
```

By the way, should that method raise an error if any of the locations in `point_seqs` parameter are not actually present among the VariationModel's master locations? It seems that right now they are simply ignored.

fontbe/src/gvar.rs
```diff
@@ -0,0 +1,101 @@
...
+
+fn to_tuple(loc: &NormalizedLocation) -> Tuple {
+    Tuple::new(
+        loc.iter()
+            .map(|(_, nc)| F2Dot14::from_f32(nc.to_f32()))
+            .collect(),
+    )
+}
+
+fn to_deltas(fragment: &GvarFragment) -> Vec<GlyphDeltas> {
```
anthrotype, line 35: I believe the order in which you collect the vector of deltas is not deterministic, because VariationModel::deltas currently returns a `HashMap<NormalizedLocation, Vec<V>>` (whose order is not insertion-order, nor ordered, just random I believe); in fonttools VariationModel.getDeltas returns an ordered list of deltas, and the order matches the internal ordering of the master locations (as sorted by the variation model) and this order is deterministic. We want to keep that order, so maybe instead of returning a `HashMap<NormalizedLocation, Vec<V>>` from `deltas` method, you should return a vector or iterator of tuples `(NormalizedLocation, Vec<V>)`.
Even better, as I was saying in https://github.com/googlefonts/fontmake-rs/pull/236/files#r1159650526, the `deltas` method could return the regions instead of the locations for each deltas, i.e. `(VariationRegion, Vec<V>)`.

anthrotype: probably the rendering won't change if TupleVariations for a given glyph are ordered differently, since in the end the final set of deltas is going to result from a summation of the respective (scaled) deltas; but keeping the same order as fonttools and not throwing away the deterministic ordering that we already have are good things by themselves.

fontbe/src/gvar.rs
```diff
@@ -0,0 +1,101 @@
...
+    orchestration::{BeWork, Bytes, Context, GvarFragment},
+};
+
+struct GvarWork {}
+
+pub fn create_gvar_work() -> Box<BeWork> {
+    Box::new(GvarWork {})
+}
+
+fn to_tuple(loc: &NormalizedLocation) -> Tuple {
```
cmyr, line 27: can this be a method on `NormalizedLocation`?

fontbe/src/gvar.rs
cmyr: Do we generally have intermediate regions, or are they often elided? I inferred from the docs that they are often elided, and I reflected that in the API, but maybe this was a mistake?

fontbe/src/gvar.rs
```diff
@@ -0,0 +1,101 @@
...
+            let mut max_y = i16::MIN;
+            
+            
+            // TODO: nice rounding on deltas
+            let mut deltas: Vec<_> = deltas.iter().map(|v| {
+                let (x, y) = (v.x as i16, v.y as i16);
+                min_x = min(min_x, x);
+                max_x = max(max_x, x);
+                min_y = min(min_y, y);
+                max_y = max(max_y, y);
```
cmyr, line 57: just a style thing but I would prefer to not import `min` and `max` and just do,

```suggestion
                min_x = min_x.min(x);
                max_x = max_x.max(x);
                min_y = min_y.min(y);
                max_y = max_y.max(y);                
```

fontbe/src/gvar.rs
cmyr, line 35: ditto that this looks like it wants to be a method?

fontbe/src/gvar.rs
anthrotype: no, because I don't think a NormalizedLocation by itself has enough information to fully define a Tuple's start/peak/end region, it's only the peak. Unless the intention is to construct such an incomplete Tuple and modify afterwards

fontbe/src/gvar.rs
anthrotype: of course, intermediate regions are very common. Think of a single-axis setup with two masters in addition to the default master: e.g. {Weight: 0.0, Weight: 0.5, Weight: 1.0}. You end up with the following intermediate regions:

```python
>>> model = VariationModel([{"Weight": 0.0}, {"Weight": 0.5}, {"Weight": 1.0}], ["Weight"])
>>> model.supports
[{}, {'Weight': (0, 0.5, 1.0)}, {'Weight': (0.5, 1.0, 1.0)}]
```

Only with very simple setups where you have only a master at the maximum ends of an axis, you get the simple (non-intermediate) sort:
```
>>> model = VariationModel([{"Weight": 0.0}, {"Weight": 1.0}], ["Weight"])
>>> model.supports
[{}, {'Weight': (0, 1.0, 1.0)}]
```

fontbe/src/glyphs.rs
rsheeter: It should definitely error if you pass a location that isn't defined!

fontbe/src/gvar.rs
rsheeter: Oh I see, I think I confused Tuple and Tent. I tend to think we will want to change the API to more explicitly model the tent structure.

fontbe/src/gvar.rs
rsheeter: I believe the tents are in `influence: Vec<VariationRegion>`.

fontbe/src/gvar.rs
rsheeter: I think I had in mind the order didn't matter. I agree that's not going to be great wrt matching FontTools. Will change to something ordered.

fontbe/src/gvar.rs
rsheeter: It's completely not the right way to compute the Tuple, let us pretend it never existed :D

fontbe/src/gvar.rs
rsheeter: Updated.

fontc/src/lib.rs
rsheeter: Per-glyph work now emits glyph and gvar fragments, and pushes phantom points into the deltas. `leftSideX = glyph.xMin - leftSideBearing` seems like it would always be 0 since I believe we set lsb = xMin?

fontbe/src/gvar.rs
rsheeter: Updated to use influence to get the tents.

fontbe/src/glyphs.rs
rsheeter: > We want to pass the phantom points to the var_model to compute deltas for them.

Done, albeit using a dumbed down phantom calculator.

> before computing deltas, we want to ...

Good point, not yet done.

fontbe/src/gvar.rs
rsheeter: Copied the ref in, ty. I don't have the full capability yet.

rsheeter:
`cargo run -p fontc -- --source resources/testdata/mov_xy.designspace` produces a VF the browser renders but setting movy does not move the glyph. Baby steps.

fontbe/src/glyphs.rs
rsheeter: Font now works. Punting some of the work to https://github.com/googlefonts/fontmake-rs/issues/240.

fontbe/src/glyphs.rs
rsheeter: Punted to https://github.com/googlefonts/fontmake-rs/issues/241

fontbe/src/glyphs.rs
rsheeter: Added a check.

fontbe/src/glyphs.rs
```diff
@@ -110,46 +116,119 @@ fn create_composite(
...
+    // We currently always set lsb to xMin so leftSideX = 0, rightSideX = advance.
+    points.push(Point::new(0.0, 0.0)); // leftSideX, 0
+    points.push(Point::new(advance as f64, 0.0)); // rightSideX, 0
+
+    // TODO: vertical phantom points
+    points.push(Point::new(0.0, 0.0));
+    points.push(Point::new(0.0, 0.0));
+}
+
+fn point_seqs_for_simple_glyph(
```
cmyr, line 134: Would it make sense for these to be methods on `ir::Glyph`? I guess we don't own that type, here?

fontbe/src/glyphs.rs
```diff
@@ -110,46 +116,119 @@ fn create_composite(
...
-                })?;
-                context.set_glyph(name, base_glyph.into());
+                context.set_glyph(name.clone(), base_glyph.clone().into());
+
+                (name, point_seqs_for_simple_glyph(ir_glyph, instances))
             }
-        }
+        };
+
+        // Everybody loves gvar!
```
cmyr, line 219: I worry this comment may be misleading to people who are unfamiliar with gvar

fontbe/src/gvar.rs
```diff
@@ -0,0 +1,114 @@
...
+            if deltas.is_empty() {
+                return None;
+            }
+
+            // TODO: nice rounding on deltas
+            let deltas: Vec<_> = deltas.iter().map(|v| (v.x as i16, v.y as i16)).collect();
+
+            let tuple_builder: TupleBuilder = region.into();
+            let (min, peak, max) = tuple_builder.build();
+            Some(GlyphDeltas::new(peak, deltas, Some((min, max))))
```
cmyr, line 79: if we are always passing the full min/max/peak then maybe I should change this API?

fontdrasil/src/orchestration.rs
```diff
@@ -17,6 +17,8 @@ pub enum Access<I> {
     All,
     /// Access to one specific resource is permitted
     One(I),
+    /// Access to two specific resource is permitted
+    Two(I, I),
```
cmyr, line 21: I can see where this is going

fontir/src/variations.rs
```diff
@@ -29,6 +30,10 @@ const ONE: OrderedFloat<f32> = OrderedFloat(1.0);
 pub struct VariationModel {
     default: NormalizedLocation,
 
+    /// Non-point axes
+    axes: Vec<Axis>,
+    axis_names: HashSet<String>,
```
cmyr, line 35: I will continue to register my dislike for storing axis names as `String` 🤷

fontir/src/variations.rs
```diff
@@ -39,6 +44,17 @@ pub struct VariationModel {
     delta_weights: Vec<Vec<(usize, OrderedFloat<f32>)>>,
 }
 
+fn only_known_axes(
```
cmyr, line 47: this looks like it wants to a method called `retain` on `NormalizedLocation`, which would be similar to existing `std` apis for filtering items from a collection in place.

fontir/src/variations.rs
```diff
@@ -364,74 +401,93 @@ struct LocationSortKey {
...
 /// Visualize as a tent of influence, starting at lower, peaking at peak,
 /// and dropping off to zero at upper.
-#[derive(Serialize, Deserialize, Clone, PartialEq, Eq)]
-struct Tent {
-    lower: NormalizedCoord,
-    peak: NormalizedCoord,
-    upper: NormalizedCoord,
+#[derive(Serialize, Deserialize, Default, Clone, PartialEq, Eq)]
+pub struct Tent {
+    pub lower: NormalizedCoord,
```
cmyr, line 488: is this terminology borrowed from somewhere else? In the spec I see these called min/max/peak iirc.

cmyr (approved):
A quick run-through for me, but overall seems sane? Some notes inline, I imagine its worth getting cosimo's input here as well.

fontbe/src/glyphs.rs
rsheeter: Interesting question. I think I would argue IR shouldn't directly implement translation to BE constructs so no.

fontir/src/variations.rs
rsheeter: Terminology was from the FontTools code itw as ported from. I agree min/peak/max is more common, will update.

fontir/src/variations.rs
rsheeter: I agree! Updating so.

fontir/src/variations.rs
rsheeter: Resolving as I would like to tackle that in it's own PR

fontbe/src/gvar.rs
rsheeter: Yes please, I managed to confuse myself about whether that was the "normal" min/peak/max or not.

fontbe/src/glyphs.rs
rsheeter: Ha, that is a bit cryptic. Updated to more clearly indicate gvar applies to both types of glyph.

fontbe/src/gvar.rs
rsheeter: Good catch, making it so.

fontir/src/variations.rs
anthrotype: In the spec, I actually see these named "start" and "end", not "min" and "max", see the discussion about intermediate regions in https://learn.microsoft.com/en-us/typography/opentype/spec/otvaroverview

arguably "max" could be a bit ambiguous when placed next to "peak".

fontir/src/coords.rs
```diff
@@ -248,6 +252,10 @@ impl<T: Copy> Location<T> {
         self
     }
 
+    pub fn rm_pos(&mut self, axis: &str) {
```
anthrotype, line 255: where is this `rm_pos` being used? Also, the name is a bit cryptic

fontir/src/ir.rs
```diff
@@ -41,32 +41,40 @@ pub struct StaticMetadata {
     /// See <https://learn.microsoft.com/en-us/typography/opentype/spec/name>.
     pub names: HashMap<NameKey, String>,
 
-    /// Every axis used by the font being compiled
+    /// Every axis used by the font being compiled, including point axes.
```
anthrotype, line 44: I suggest using "discrete axes" as that's the name used in DesignSpace format

anthrotype: wait, maybe is this about something else?

anthrotype: "an axis of variation defined only at a single point"... It's not immediately clear to me why we need to special case these

anthrotype: I see you want to exclude these so-called "point axes" when constructing the VariationModel. In reality, these kind of setups (a variaiton axis where min=default=max) do not actually occur in real world font sources because they do not make sense to have, they would be no op. They may occur in your made-up test data (e.g. static.designspace) but I don't think it is worth having to different `Vec<Axis>` in the static metadata, one for axes in general and one for actually variable ones (which is 100% of them in an actually variable font project). You can filter on demand when needed, no?

anthrotype: FWIW fontmake does not treat these point-axes specially; e.g. if you feed your static.designspace to fontmake-py and ask it to make a "variable" font, `fontmake resources/testdata/static.designspace -o variable`, it will make one with fvar containing that point axis, and all the other variation tables basically empty.
If you ask it to create static ttfs from the same designspace (-o ttf), it will make one for each of the masters (or one for each of the instances with additional -i option), and would be equivalent to running fontmake directly on the master UFO as input. I'm not saying we should replicate this UX in fontmake-rs, I think the latter should be designed to build variable fonts only.
Just saying that a font like our test static.designspace does not normally occur in the wild; if one wants to build a static font, one builds that one, doesn't create a dummy designspace to wrap it.

anthrotype (commented):
it's nice to finally see variable glyphs render in browser, great work reaching this milestone.
A belated LGTM!

Issue #250, opened 2023-04-10
Figure out why our Oswald has significantly different glyphs than fontmake's


rsheeter:
So significantly so it makes me wonder if we are using the wrong master as default or something of that nature.  Also the point order seems ... rotated?

![image](https://user-images.githubusercontent.com/6466432/231013852-727cea09-7942-4bea-a9a7-d25b7f347fc9.png)

The diff is *huge*; it will probably be helpful to start with a version of the source where most of the glyphs are deleted.

anthrotype:
I like how the spell checker corrects fontmake as "don't make"

anthrotype:
For better comparisons you can disable glyph renaming with --no-production-names, and also --keep-direction to not reverse contour direction.

rsheeter:
The flags are very handy short-term, ty. Should we ultimatley just implement the renaming and direction shifting?

rsheeter:
fontmake

```
  <avar>
    <segment axis="wght">
      <mapping from="-1.0" to="-1.0"/>
      <mapping from="-0.5" to="-0.625"/>
      <mapping from="0.0" to="0.0"/>
      <mapping from="0.3333" to="0.5161"/>
      <mapping from="0.6667" to="0.7871"/>
      <mapping from="1.0" to="1.0"/>
    </segment>
  </avar>

  <fvar>

    <!-- Weight -->
    <Axis>
      <AxisTag>wght</AxisTag>
      <Flags>0x0</Flags>
      <MinValue>200.0</MinValue>
      <DefaultValue>400.0</DefaultValue>
      <MaxValue>700.0</MaxValue>
      <AxisNameID>256</AxisNameID>
    </Axis>

    <!-- ExtraLight -->
    <NamedInstance flags="0x0" subfamilyNameID="257">
      <coord axis="wght" value="200.0"/>
    </NamedInstance>

    <!-- Light -->
    <NamedInstance flags="0x0" subfamilyNameID="258">
      <coord axis="wght" value="300.0"/>
    </NamedInstance>

    <!-- Regular -->
    <NamedInstance flags="0x0" subfamilyNameID="259">
      <coord axis="wght" value="400.0"/>
    </NamedInstance>

    <!-- Medium -->
    <NamedInstance flags="0x0" subfamilyNameID="260">
      <coord axis="wght" value="500.0"/>
    </NamedInstance>

    <!-- SemiBold -->
    <NamedInstance flags="0x0" subfamilyNameID="261">
[...]

PR #252, opened 2023-04-12
Load .glyphs instances and use them to inform axis rigging

rsheeter:
Improve, though not yet to parity with fontmake, handling of glyphs 2/3 instances:

* Load glyphs instances for glyphs v2 and v3
   * Upgrade v2 to v3 per our std practice
   * We do not yet support all fields. Baby steps.
   * The code for this is tiresome and lengthy
* Fix typo that lead to avar not being merged into final font
* Include instance mappings in user:design mappings, which improves avar production
   * We now match the Oswald avar noted in https://github.com/googlefonts/fontmake-rs/issues/250#issuecomment-1504057038
* Figure out the default instance and use it to assign axis defaults
   * We now match the Oswald fvar axes noted in https://github.com/googlefonts/fontmake-rs/issues/250#issuecomment-1504057038
   * We do not yet emit any named instances

Step toward #250

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -706,6 +706,7 @@ mod tests {
         fs::{self, File},
         io::Read,
         path::{Path, PathBuf},
+        str::FromStr,
     };
 
     use fontbe::{
@@ -1291,6 +1292,7 @@ mod tests {
         assert_eq!(
             vec![
                 Tag::new(b"OS/2"),
+                Tag::new(b"avar"),
                 Tag::new(b"cmap"),
                 Tag::new(b"fvar"),
                 Tag::new(b"glyf"),
@@ -1421,5 +1423,82 @@ mod tests {
         fn close(&mut self) {
             // nop
         }
+    }
+
+    fn axes(font: &FontRef) -> Vec<(Tag, f32, f32, f32)> {
+        font.fvar()
+            .unwrap()
+            .axes()
+            .unwrap()
+            .iter()
+            .map(|a| {
+                (
+                    a.axis_tag(),
+                    a.min_value.get().to_f64() as f32,
+                    a.default_value.get().to_f64() as f32,
+                    a.max_value.get().to_f64() as f32,
+                )
+            })
+            .collect::<Vec<_>>()
+    }
+
+    #[test]
+    fn compile_glyphs_font_with_weight_axis() {
+        let temp_dir = tempdir().unwrap();
+        let build_dir = temp_dir.path();
+        compile(Args::for_test(build_dir, "glyphs2/WghtVar.glyphs"));
+
+        let font_file = build_dir.join("font.ttf");
+        assert!(font_file.exists());
+        let buf = fs::read(font_file).unwrap();
+        let font = FontRef::new(&buf).unwrap();
+
+        assert_eq!(
+            vec![(Tag::from_str("wght").unwrap(), 400.0, 400.0, 700.0)],
+            axes(&font),
+        );
+    }
+
+    #[test]
+    fn compile_glyphs_font_with_avar() {
+        let temp_dir = tempdir().unwrap();
+        let build_dir = temp_dir.path();
+        compile(Args::for_test(build_dir, "glyphs2/WghtVar_Avar.glyphs"));
+
+        let font_file = build_dir.join("font.ttf");
+        assert!(font_file.exists());
+        let buf = fs::read(font_file).unwrap();
+        let font = FontRef::new(&buf).unwrap();
+
+        // Default 400 is important, it means we found the index of Regular
+        assert_eq!(
+            vec![(Tag::from_str("wght").unwrap(), 300.0, 400.0, 700.0)],
+            axes(&font),
+        );
+
+        let axis_maps: Vec<_> = font
+            .avar()
+            .unwrap()
+            .axis_segment_maps()
+            .iter()
+            .map(|m| m.unwrap())
+            .map(|m| {
+                m.axis_value_maps()
+                    .iter()
+                    .map(|e| (e.from_coordinate().to_f32(), e.to_coordinate().to_f32()))
+                    .collect::<Vec<_>>()
+            })
+            .collect();
+
+        // 0 is 400, 7 is 700, .33 => .66 is mapping 500 => 600 as per instance definition
+        assert_eq!(
+            vec![vec![
+                (-1.0, -1.0),
+                (0.0, 0.0),
+                (0.333313, 0.666687),
+                (1.0, 1.0)
+            ]],
+            axis_maps
+        );
     }
 }
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -1490,7 +1490,7 @@ mod tests {
             })
             .collect();
 
-        // 0 is 400, 7 is 700, .33 => .66 is mapping 500 => 600 as per instance definition
+        // 0 is 400, 1 is 700, .33 => .66 is mapping 500 => 600 as per instance definition
         assert_eq!(
             vec![vec![
                 (-1.0, -1.0),
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -103,7 +103,7 @@ struct RawFont {
     pub axes: Option<Vec<Axis>>,
     pub glyphs: Vec<RawGlyph>,
     pub font_master: Vec<RawFontMaster>,
-    pub instances: Vec<RawInstance>,
+    pub instances: Option<Vec<RawInstance>>,
     pub feature_prefixes: Option<Vec<RawFeature>>,
     pub features: Option<Vec<RawFeature>>,
     pub classes: Option<Vec<RawFeature>>,
@@ -347,11 +347,7 @@ struct RawInstance {
     pub exports: Option<i64>,
     pub active: Option<i64>,
     pub type_: Option<String>,
-    pub interpolation_weight: Option<i64>,
-    pub interpolation_width: Option<i64>,
-    pub weight_class: Option<String>,
-    pub width_class: Option<String>,
-    pub custom_value: Option<i64>,
+    pub axes_values: Option<Vec<OrderedFloat<f64>>>,
 
     #[rest]
     pub other_stuff: BTreeMap<String, Plist>,
@@ -608,15 +604,18 @@ fn custom_param_mut<'a>(
     None
 }
 
-fn glyphs_v2_field_name_and_default(nth_axis: usize) -> Result<(&'static str, f64), Error> {
+fn glyphs_v2_field_name_and_default(
+    nth_axis: usize,
+) -> Result<(&'static [&'static str], f64), Error> {
     // Per https://github.com/googlefonts/fontmake-rs/pull/42#pullrequestreview-1211619812
     // the field to use is based on the order in axes NOT the tag.
     // That is, whatever the first axis is, it's value is in the weightValue field. Long sigh.
     // Defaults per https://github.com/googlefonts/fontmake-rs/pull/42#discussion_r1044415236.
+    // v2 instances use novel field names so send back several for linear probing.
     Ok(match nth_axis {
-        0 => ("weightValue", 100_f64),
-        1 => ("widthValue", 100_f64),
-        2 => ("customValue", 0_f64),
+        0 => (&["weightValue", "interpolationWeight"], 100_f64),
+        1 => (&["widthValue", "interpolationWidth"], 100_f64),
+        2 => (&["customValue"], 0_f64),
         _ => {
             return Err(Error::StructuralError(format!(
                 "We don't know what field to use for axis {nth_axis}"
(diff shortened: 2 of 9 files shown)


fontc/src/lib.rs
```diff
@@ -1422,4 +1424,81 @@ mod tests {
...
+            .map(|m| m.unwrap())
+            .map(|m| {
+                m.axis_value_maps()
+                    .iter()
+                    .map(|e| (e.from_coordinate().to_f32(), e.to_coordinate().to_f32()))
+                    .collect::<Vec<_>>()
+            })
+            .collect();
+
+        // 0 is 400, 7 is 700, .33 => .66 is mapping 500 => 600 as per instance definition
```
dfrg, line 1493: Should this be “1 is 700”?

glyphs-reader/src/font.rs
```diff
@@ -314,6 +316,51 @@ impl RawGlyph {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Hash)]
+pub struct Instance {
+    pub name: String,
+    pub active: bool,
+    pub type_: InstanceType,
```
dfrg, line 323: nit: we commonly spell this as `ty` in Rust land. Alternatively, `kind` and `InstanceKind`.

glyphs-reader/src/font.rs
```diff
@@ -314,6 +316,51 @@ impl RawGlyph {
...
+        }
+    }
+}
+
+#[derive(Debug, Clone, FromPlist, PartialEq, Eq, Hash)]
+struct RawInstance {
+    pub name: String,
+    pub exports: Option<i64>,
+    pub active: Option<i64>,
+    pub type_: Option<String>,
```
dfrg, line 349: This one might be using magic in `FromPlist`?

glyphs-reader/src/font.rs
```diff
@@ -1032,18 +1133,71 @@ fn user_to_design_from_axis_location(
...
+        return;
+    }
+    mappings.push((user, design));
+}
+
+/// * <https://github.com/googlefonts/glyphsLib/blob/6f243c1f732ea1092717918d0328f3b5303ffe56/Lib/glyphsLib/builder/axes.py#L128>
+/// * <https://github.com/googlefonts/glyphsLib/blob/6f243c1f732ea1092717918d0328f3b5303ffe56/Lib/glyphsLib/builder/axes.py#L353>
+fn add_mappings_from_instances(
+    instances: &[Instance],
+    mut mappings: BTreeMap<String, RawUserToDesignMapping>,
```
dfrg, line 1151: Why don’t we take mappings by mutable reference here? Also, the parameter order is different from `add_mapping_if_new` which is confusing.

glyphs-reader/src/font.rs
```diff
@@ -1032,18 +1133,71 @@ fn user_to_design_from_axis_location(
...
+            let mappings = mappings.entry(axis_name.clone()).or_default();
+            for (user, design) in inst_mapping.iter() {
+                add_mapping_if_new(mappings, *user, *design);
+            }
+        }
+    }
+    mappings
+}
+
+fn add_mappings_from_masters(
```
dfrg, line 1167: Ditto. Maybe these mappings should be promoted to full types with methods.

glyphs-reader/src/font.rs
```diff
@@ -1032,18 +1133,71 @@ fn user_to_design_from_axis_location(
...
+    from: &RawFont,
+    mut mappings: BTreeMap<String, RawUserToDesignMapping>,
+) -> BTreeMap<String, RawUserToDesignMapping> {
+    for master in from.font_master.iter() {
+        let Some(axes) = from.axes.as_ref() else { continue; };
+        for (axis, value) in axes.iter().zip(master.axes_values.as_ref().unwrap()) {
+            let mappings = mappings.entry(axis.name.clone()).or_default();
```
dfrg, line 1174: Maybe use different names here for clarity?

dfrg (approved):
LGTM with nits. I can’t comment on glyphs format specific things but will approve to unblock until our expert returns.

glyphs-reader/src/font.rs
```diff
@@ -557,15 +604,18 @@ fn custom_param_mut<'a>(
...
     // Per https://github.com/googlefonts/fontmake-rs/pull/42#pullrequestreview-1211619812
     // the field to use is based on the order in axes NOT the tag.
     // That is, whatever the first axis is, it's value is in the weightValue field. Long sigh.
     // Defaults per https://github.com/googlefonts/fontmake-rs/pull/42#discussion_r1044415236.
+    // v2 instances use novel field names so send back several for linear probing.
     Ok(match nth_axis {
-        0 => ("weightValue", 100_f64),
-        1 => ("widthValue", 100_f64),
-        2 => ("customValue", 0_f64),
+        0 => (&["weightValue", "interpolationWeight"], 100_f64),
```
cmyr, line 616: the compiler should be able to infer the float type, here?

glyphs-reader/src/font.rs
```diff
@@ -632,6 +682,34 @@ fn v2_to_v3_name(properties: &mut Vec<RawName>, v2_prop: &Option<String>, v3_nam
...
+        // Per https://github.com/googlefonts/fontmake-rs/pull/42#pullrequestreview-1211619812
+        // the field to use is based on the order in axes NOT the tag.
+        // That is, whatever the first axis is, it's value is in the weightValue field. Long sigh.
+        let (field_names, default_value) = glyphs_v2_field_name_and_default(idx)?;
+        let value = field_names
+            .iter()
+            .filter_map(|field_name| other_stuff.remove(*field_name))
+            .next()
+            .unwrap_or_else(|| Plist::Float(default_value.into()))
+            .as_f64()
```
cmyr, line 701: Looking at this more closely, the flow is a little funny. We want to return an error if there *is* a value in the plist for this field name, but it is the wrong type (not a number?) but we want to use a default value if there is no plist value for this field name?

If that is true, I would prefer to write this as something like,

```rust
        let value = field_names
            .iter()
            .find_map(|field_name| {
                other_stuff.remove(*field_name).map(|x| {
                    x.as_f64().ok_or_else(|| {
                        Error::StructuralError(format!(
                            "Invalid value {x:?} for field {field_name}"
                        ))
                    })
                })
            })
            .unwrap_or(Ok(default_value))?;
        axis_values.push(value.into());
```

the basic idea being to generate the error where it occurs, instead of unwrapping with a known good value and validating at the end?


This is also just for illustrative purposes, no change required.

glyphs-reader/src/font.rs
cmyr: I see `ty`, `typ` and `type_` in various places, happy to standardize on one or the other 🤷

fontc/src/lib.rs
rsheeter: Yes, oops

glyphs-reader/src/font.rs
```diff
@@ -1032,18 +1133,71 @@ fn user_to_design_from_axis_location(
...
+fn add_mapping_if_new(
+    mappings: &mut RawUserToDesignMapping,
+    user: OrderedFloat<f32>,
+    design: OrderedFloat<f32>,
+) {
+    if mappings.iter().any(|(u, d)| *u == user || *d == design) {
+        return;
+    }
+    mappings.push((user, design));
+}
```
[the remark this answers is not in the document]
rsheeter: In re-reading I think I agree ... that deserves to be a "real" type now.

glyphs-reader/src/font.rs
```diff
@@ -929,7 +1019,18 @@ fn default_master_idx(raw_font: &RawFont) -> usize {
...
+        .unwrap_or_else(|| {
+            raw_font
+                .font_master
+                .iter()
+                .position(|m| match m.other_stuff.get("name") {
+                    Some(Plist::String(name)) => name == "Regular",
+                    _ => false,
+                })
+                .unwrap_or_default()
+        })
```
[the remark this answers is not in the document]
rsheeter: Oh nice, I'm not used to if in matches yet so I underuse it.

glyphs-reader/src/font.rs
```diff
@@ -1172,6 +1326,110 @@ fn raw_feature_to_feature(feature: RawFeature) -> Result<FeatureSnippet, Error>
     Ok(FeatureSnippet(code))
 }
 
+/// <https://github.com/googlefonts/glyphsLib/blob/6f243c1f732ea1092717918d0328f3b5303ffe56/Lib/glyphsLib/classes.py#L220-L249>
+fn lookup_class_value(axis_tag: &str, user_class: &Option<&String>) -> Option<f32> {
```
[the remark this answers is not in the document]
rsheeter: I think I confused myself and it should be &str. Fixed.

glyphs-reader/src/font.rs
```diff
@@ -1172,6 +1326,110 @@ fn raw_feature_to_feature(feature: RawFeature) -> Result<FeatureSnippet, Error>
...
+        }
+    }
+}
+
+fn add_mapping_if_present(
+    axis_mappings: &mut BTreeMap<String, RawUserToDesignMapping>,
+    axes: &[Axis],
+    axis_tag: &str,
+    axes_values: &Option<Vec<OrderedFloat<f64>>>,
+    value: &Option<&Plist>,
```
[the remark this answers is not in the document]
rsheeter: Updated to use `Option<&whatever>`

glyphs-reader/src/font.rs
rsheeter: yup, ty

glyphs-reader/src/font.rs
```diff
@@ -632,6 +682,34 @@ fn v2_to_v3_name(properties: &mut Vec<RawName>, v2_prop: &Option<String>, v3_nam
...
+    let mut axis_values = Vec::new();
+    for idx in 0..axes.len() {
+        // Per https://github.com/googlefonts/fontmake-rs/pull/42#pullrequestreview-1211619812
+        // the field to use is based on the order in axes NOT the tag.
+        // That is, whatever the first axis is, it's value is in the weightValue field. Long sigh.
+        let (field_names, default_value) = glyphs_v2_field_name_and_default(idx)?;
+        let value = field_names
+            .iter()
+            .filter_map(|field_name| other_stuff.remove(*field_name))
+            .next()
```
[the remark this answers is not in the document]
rsheeter: TIL! Yes :)

glyphs-reader/src/font.rs
```diff
@@ -314,6 +316,51 @@ impl RawGlyph {
...
+
+#[derive(Debug, Clone, FromPlist, PartialEq, Eq, Hash)]
+struct RawInstance {
+    pub name: String,
+    pub exports: Option<i64>,
+    pub active: Option<i64>,
+    pub type_: Option<String>,
+    pub axes_values: Option<Vec<OrderedFloat<f64>>>,
+
+    #[rest]
```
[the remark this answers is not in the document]
rsheeter: TIL. I think I will fix that in it' sown PR.

glyphs-reader/src/font.rs
rsheeter: In re-reading I think I will change to issue a warning if it's a weird value

glyphs-reader/src/font.rs
rsheeter: This one I believe we need to name so to make FromPlist work, will add comment

glyphs-reader/src/font.rs
rsheeter: These names have to match the glyphs data and work with FromPlist, will add a comment

glyphs-reader/src/font.rs
rsheeter: I wanted to be able to write `add_mappings(blah)` in return position. In the light of a new day i don't think it's a win, will change to mut ref.

rsheeter: ...actually to it's own type

glyphs-reader/src/font.rs
```diff
@@ -314,6 +316,51 @@ impl RawGlyph {
...
+    }
+}
+
+#[derive(Debug, Clone, FromPlist, PartialEq, Eq, Hash)]
+struct RawInstance {
+    pub name: String,
+    pub exports: Option<i64>,
+    pub active: Option<i64>,
+    pub type_: Option<String>,
+    pub axes_values: Option<Vec<OrderedFloat<f64>>>,
```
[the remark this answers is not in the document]
rsheeter: Correct. It would be really nice if FromPlist could deserialize to an empty vec but I don't want to tackle that in this PR. 

Filed as https://github.com/googlefonts/fontmake-rs/issues/257

glyphs-reader/src/font.rs
rsheeter: Became nicer as a result of adding types

glyphs-reader/src/font.rs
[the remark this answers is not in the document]
rsheeter: https://github.com/googlefonts/fontmake-rs/pull/259

fontc/src/lib.rs
```diff
@@ -706,6 +706,7 @@ mod tests {
         fs::{self, File},
         io::Read,
         path::{Path, PathBuf},
+        str::FromStr,
```
anthrotype, line 709: noob question: why is this needed?

dfrg: A couple of the tests use `Tag::from_str` and the `FromStr` trait needs to be in scope for that method to be available.

anthrotype: Interesting. AFAIU, the Tag struct implements `FromStr` trait, which comes from `std::str::FromStr`; what it's not immediately clear to me is why the users of that method need to import that trait, and it is not enough to simply import Tag.

anthrotype (commented):
it must have been not fun to digest the messy logic from glyphsLib... LGTM!

fontc/src/lib.rs
dfrg: Just the rules of the language. The trait itself must be in scope in order to call its methods. You can implement multiple traits with the same method name(s) for a single type and this is one way of disambiguating those cases.

This is sometimes muddled because Rust has a [prelude](https://doc.rust-lang.org/std/prelude/index.html) that automatically imports common types and traits to all modules. So you’ll not see traits imported explicitly if they appear there.

glyphs-reader/src/font.rs
```diff
@@ -929,7 +1020,14 @@ fn default_master_idx(raw_font: &RawFont) -> usize {
                 .map(|(idx, _)| idx)
                 .unwrap_or(0)
         })
-        .unwrap_or(0)
+        // TODO: implement searching for "a base style shared between all masters" as FontTools does
+        // Still nothing? - just look for one called Regular
```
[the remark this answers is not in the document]
rsheeter: For Oswald fontmake falls back to Regular after failing at base style matching so this is incrementally closer. I need to find - or create - a source we build wrong due to not having the full find a shared base style. Will file an issue.

glyphs-reader/src/font.rs
```diff
@@ -929,7 +1020,14 @@ fn default_master_idx(raw_font: &RawFont) -> usize {
                 .map(|(idx, _)| idx)
                 .unwrap_or(0)
         })
-        .unwrap_or(0)
+        // TODO: implement searching for "a base style shared between all masters" as FontTools does
```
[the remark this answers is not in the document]
rsheeter: Fixed by https://github.com/googlefonts/fontmake-rs/pull/265

glyphs-reader/src/font.rs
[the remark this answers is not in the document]
rsheeter: https://github.com/googlefonts/fontmake-rs/issues/264

PR #259, opened 2023-04-13
Use `#[fromplist(rest)]` instead of `#[rest]`

rsheeter:
As requested in https://github.com/googlefonts/fontmake-rs/pull/252#discussion_r1165739577


dfrg (approved):
LGTM

PR #260, opened 2023-04-14
Pick short or long loca

rsheeter:
It's very handy to know from the loca data what the format is so (edit) added a new context slot just for that.

### DIFF
--- a/fontbe/src/font.rs
+++ b/fontbe/src/font.rs
@@ -28,20 +28,20 @@ enum TableType {
     Variable,
 }
 
-const TABLES_TO_MERGE: &[(WorkId, Tag, TableType, usize)] = &[
-    (WorkId::Avar, Avar::TAG, TableType::Variable, 0),
-    (WorkId::Cmap, Cmap::TAG, TableType::Static, 0),
-    (WorkId::Fvar, Fvar::TAG, TableType::Variable, 0),
-    (WorkId::Head, Head::TAG, TableType::Static, 0),
-    (WorkId::Hhea, Hhea::TAG, TableType::Static, 0),
-    (WorkId::Hmtx, Hmtx::TAG, TableType::Static, 0),
-    (WorkId::Glyf, Glyf::TAG, TableType::Static, 0),
-    (WorkId::Gvar, Gvar::TAG, TableType::Variable, 0),
-    (WorkId::Loca, Loca::TAG, TableType::Static, 1),
-    (WorkId::Maxp, Maxp::TAG, TableType::Static, 0),
-    (WorkId::Name, Name::TAG, TableType::Static, 0),
-    (WorkId::Os2, Os2::TAG, TableType::Static, 0),
-    (WorkId::Post, Post::TAG, TableType::Static, 0),
+const TABLES_TO_MERGE: &[(WorkId, Tag, TableType)] = &[
+    (WorkId::Avar, Avar::TAG, TableType::Variable),
+    (WorkId::Cmap, Cmap::TAG, TableType::Static),
+    (WorkId::Fvar, Fvar::TAG, TableType::Variable),
+    (WorkId::Head, Head::TAG, TableType::Static),
+    (WorkId::Hhea, Hhea::TAG, TableType::Static),
+    (WorkId::Hmtx, Hmtx::TAG, TableType::Static),
+    (WorkId::Glyf, Glyf::TAG, TableType::Static),
+    (WorkId::Gvar, Gvar::TAG, TableType::Variable),
+    (WorkId::Loca, Loca::TAG, TableType::Static),
+    (WorkId::Maxp, Maxp::TAG, TableType::Static),
+    (WorkId::Name, Name::TAG, TableType::Static),
+    (WorkId::Os2, Os2::TAG, TableType::Static),
+    (WorkId::Post, Post::TAG, TableType::Static),
 ];
 
 impl Work<Context, Error> for FontWork {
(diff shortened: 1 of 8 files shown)


fontbe/src/font.rs
```diff
@@ -28,20 +28,20 @@ enum TableType {
...
-    (WorkId::Hhea, Hhea::TAG, TableType::Static),
-    (WorkId::Hmtx, Hmtx::TAG, TableType::Static),
-    (WorkId::Glyf, Glyf::TAG, TableType::Static),
-    (WorkId::Gvar, Gvar::TAG, TableType::Variable),
-    (WorkId::Loca, Loca::TAG, TableType::Static),
-    (WorkId::Maxp, Maxp::TAG, TableType::Static),
-    (WorkId::Name, Name::TAG, TableType::Static),
-    (WorkId::Os2, Os2::TAG, TableType::Static),
-    (WorkId::Post, Post::TAG, TableType::Static),
+const TABLES_TO_MERGE: &[(WorkId, Tag, TableType, usize)] = &[
```
cmyr, line 31: This deserves a comment/apology

fontbe/src/orchestration.rs
```diff
@@ -219,8 +219,67 @@ impl From<&VariationRegion> for TupleBuilder {
...
+
+impl GlyfLoca {
+    pub fn new(glyf: Vec<u8>, loca: Vec<u32>) -> Self {
+        // https://github.com/fonttools/fonttools/blob/1c283756a5e39d69459eea80ed12792adc4922dd/Lib/fontTools/ttLib/tables/_l_o_c_a.py#L37
+        let mut format = SHORT_LOCA;
+        if let Some(max_offset) = loca.last() {
+            if *max_offset >= 0x20000 || loca.iter().any(|offset| offset % 2 != 0) {
+                format = LONG_LOCA;
+            }
+        }
```
cmyr, line 240: I really go out of my way to avoid this pattern ("declare a mut default and set it in a branch") where possible, because I think in general it is harder to follow.

In this case I think it probably makes sense to just have a standalone `compute_loca_format` fn?

cmyr (commented):
I do quite dislike the leading-byte hack, and I would personally prefer an approach where you use something like `bincode` to serialize an enum or a struct that encodes the format information.

If there's a good reason that this isn't possible or is a major headache then it's not the end of the world, but it doesn't feel great.

fontbe/src/orchestration.rs
rsheeter: Fixed per IM code golf

fontbe/src/font.rs
rsheeter: I agree. Fixed by shoving loca format into it's own slot instead of being sneaky.

fontbe/src/orchestration.rs
```diff
@@ -219,8 +220,99 @@ impl From<&VariationRegion> for TupleBuilder {
     }
 }
 
+#[repr(u8)]
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum LocaFormat {
+    Short,
+    Long,
+}
```
cmyr, line 228: could be explicit about the values here if you like:
```suggestion
    Short = 0,
    Long = 1,
}
```

and then you can avoid matching later and just do `format as u8 as _`. You could also do `repr(u16)` and then just cast directly.

fontbe/src/orchestration.rs
```diff
@@ -219,8 +220,99 @@ impl From<&VariationRegion> for TupleBuilder {
...
+            _ if value == LocaFormat::Long as u8 => Ok(LocaFormat::Short),
+            _ => Err(()),
+        }
+    }
+}
+
+// Free function of specific form to fit macro
+fn loca_format_from_file(file: &Path) -> LocaFormat {
+    let bytes = fs::read(file).unwrap();
+    LocaFormat::try_from(*bytes.first().unwrap()).unwrap()
```
cmyr, line 258: since we really aren't using the error value here at all I would be more tempted to remove `TryFrom` and just do,

```suggestion
    match bytes.first() {
        Some(0) => LocaFormat::Short,
        Some(1) => LocaFormat::Long,
        _ => panic!("serialized LocaFormat is invalid"),
    }
```

 which might be a bit easier to follow and a bit easier to debug if anything actually *does* go wrong, while also being less code.

fontbe/src/orchestration.rs
```diff
@@ -219,8 +220,99 @@ impl From<&VariationRegion> for TupleBuilder {
...
+    pub fn read(format: LocaFormat, paths: &Paths) -> Self {
+        let glyf = read_entire_file(&paths.target_file(&WorkId::Glyf));
+        let raw_loca = read_entire_file(&paths.target_file(&WorkId::Loca));
+        let loca = if format == LocaFormat::Short {
+            raw_loca[1..]
+                .chunks_exact(std::mem::size_of::<u16>())
+                .map(|bytes| u16::from_be_bytes(bytes.try_into().unwrap()) as u32 * 2)
+                .collect()
+        } else {
+            raw_loca[1..]
```
cmyr, line 286: should we still be skipping the first byte, here? is there a test that would catch this?

cmyr (approved):
I think this is a much more consistent approach, even if it might be doing slightly more disk access.

A few things inline, but overall lgtm :)

fontbe/src/orchestration.rs
cmyr: (to clarify: this is sort of redundant since this is how enums work in general, but since we do actually care about these values I think there's some value in being explicit)

fontbe/src/orchestration.rs
rsheeter: We should not, that's a bug leftover from my initial hackery. Evidently there is no such test, will add.

fontbe/src/orchestration.rs
rsheeter: That does seem easier to read, will change.

fontbe/src/orchestration.rs
```diff
@@ -219,8 +221,91 @@ impl From<&VariationRegion> for TupleBuilder {
...
+pub enum LocaFormat {
+    Short = 0,
+    Long = 1,
+}
+
+impl LocaFormat {
+    pub fn new(loca: &[u32]) -> LocaFormat {
+        // https://github.com/fonttools/fonttools/blob/1c283756a5e39d69459eea80ed12792adc4922dd/Lib/fontTools/ttLib/tables/_l_o_c_a.py#L37
+        if loca.last().copied().unwrap_or_default() < 0x20000
+            && loca.iter().all(|offset| offset % 2 == 0)
```
anthrotype, line 235: I wonder if it's really worth serializing this single-byte loca.format to disk (along with all the accompanying logic that this PR adds), when it can be computed using this simple formula based only on the number of locations and whether they are all 2-byte padded?

anthrotype: oh right, if we want to persist the loca table itself, we need to be able to deserialize it and for that we need the format. nevermind

PR #261, opened 2023-04-14
Obey emit_ir more closely

### LINKED ISSUE #224: [fontc] --emit-ir=false leads to TasksFailed([(Be(Font), "IO failure")])

rsheeter:
Add a test for compile w/o emitting ir, fix resulting problems. Fixes #224. Aside: discovered https://github.com/googlefonts/fontations/pull/367 along the way.

### DIFF
--- a/fontbe/src/avar.rs
+++ b/fontbe/src/avar.rs
@@ -1,5 +1,6 @@
 //! Generates a [avar](https://learn.microsoft.com/en-us/typography/opentype/spec/avar) table.
 
+use font_types::F2Dot14;
 use fontdrasil::orchestration::Work;
 use fontir::{
     coords::{CoordConverter, DesignCoord},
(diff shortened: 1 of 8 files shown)


anthrotype:
> fix resulting problems.

it would be enlightening if this PR description listed what these problems actually were, otheriwse the onus is on the reviewer to infer those from the code changes

rsheeter:
> what these problems actually were

Most notably that it failed (#224) if IR was off :)

Generally, places where we assumed the files would exist, for example we always created directories and files that are only required for subsequent incremental compiles and the font aggregation was reading files that don't exist when IR is off.

fontbe/src/avar.rs
```diff
@@ -25,8 +26,14 @@ fn to_segment_map(axis: &Axis) -> SegmentMaps {
     // because, per spec, "If the segment map for a given axis has any value maps,
     // then it must include at least three value maps: -1 to -1, 0 to 0, and 1 to 1"
     // so three value maps *must* produce identity.
+    //
+    // 0 entry avar may confuse some clients so emit a default one
```
anthrotype, line 30: this change to fontbe::avar isn't really related to --emit-ir flag, is it?

anthrotype: i see, it's to fix/workaround https://github.com/googlefonts/fontations/pull/367

anthrotype (approved):
LGTM! +1

PR #266, opened 2023-04-18
Only convert to contours if basis vectors vary

rsheeter:
Only consider basis vectors - not translation - when deciding if we need to convert to contours. Numerous Oswald glyphs we previously unnecessarily converted to contours are now kept as components, ref https://github.com/googlefonts/fontmake-rs/issues/250#issuecomment-1507845219, 2.

Credit @anthrotype for pointing out the problem in IM.

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -386,6 +386,10 @@ impl TryFrom<&ir::Glyph> for CheckedGlyph {
             })
             .collect();
         if path_els.len() > 1 {
+            warn!(
+                "{} has inconsistent path elements: {path_els:?}",
+                glyph.name
+            );
             return Err(Error::GlyphError(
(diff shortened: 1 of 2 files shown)


fontbe/src/glyphs.rs
```diff
@@ -386,6 +386,10 @@ impl TryFrom<&ir::Glyph> for CheckedGlyph {
             })
             .collect();
         if path_els.len() > 1 {
+            warn!(
+                "{} has inconsistent path elements: {path_els:?}",
+                glyph.name
```
anthrotype, line 391: can you not put also `glyph.name` inside the string-formatting braces, like you did for `path_els`?

```suggestion
                "{glyph.name} has inconsistent path elements: {path_els:?}",
```

anthrotype (approved):
LGTM

PR #268, opened 2023-04-18
Support component flattening

rsheeter:
Improves Oswald build alignment with fontmake as run by gftools. See https://github.com/googlefonts/fontmake-rs/issues/250#issuecomment-1513807429.

Builds on #267.

Once this plus the PRs it builds on are merged the largest source of difference in `glyf` for Oswald is that we don't remove implicit on-curves.

### DIFF
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -21,12 +21,15 @@ pub struct Args {
     #[arg(long, default_value = "false")]
     pub emit_debug: bool,
 
-    /// Whether to Try Hard(tm) to match fontmake (Python) behavior in cases where there are other options.
-    ///
-    /// See <https://github.com/googlefonts/fontmake-rs/pull/123> for an example of
-    /// where this matters.
+    /// In cases where a source glyph uses a mixture of components and contours, convert
+    /// all the components to contours.
     #[arg(long, default_value = "true")]
-    pub match_legacy: bool,
+    pub prefer_simple_glyphs: bool,
+
+    /// Eliminate component references to other glyphs using components, emitting only
+    /// component references to simple (contour) glyphs.
+    #[arg(long, default_value = "true")]
+    pub flatten_components: bool,
 
     /// Working directory for the build process. If emit-ir is on, written here.
     #[arg(short, long, default_value = "build")]
@@ -44,7 +47,8 @@ impl Args {
 
         flags.set(Flags::EMIT_IR, self.emit_ir);
         flags.set(Flags::EMIT_DEBUG, self.emit_debug);
-        flags.set(Flags::MATCH_LEGACY, self.match_legacy);
+        flags.set(Flags::PREFER_SIMPLE_GLYPHS, self.prefer_simple_glyphs);
+        flags.set(Flags::FLATTEN_COMPONENTS, self.flatten_components);
 
         flags
     }
@@ -65,7 +69,8 @@ impl Args {
             emit_ir: true,
             emit_debug: false,
             build_dir: build_dir.to_path_buf(),
-            match_legacy: true,
+            prefer_simple_glyphs: Flags::default().contains(Flags::PREFER_SIMPLE_GLYPHS),
+            flatten_components: Flags::default().contains(Flags::FLATTEN_COMPONENTS),
         }
     }
 }
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -26,8 +26,8 @@ pub struct Args {
     #[arg(long, default_value = "true")]
     pub prefer_simple_glyphs: bool,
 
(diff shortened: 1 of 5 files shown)


fontc/src/args.rs
```diff
@@ -21,12 +21,15 @@ pub struct Args {
...
 
-    /// Whether to Try Hard(tm) to match fontmake (Python) behavior in cases where there are other options.
-    ///
-    /// See <https://github.com/googlefonts/fontmake-rs/pull/123> for an example of
-    /// where this matters.
+    /// In cases where a source glyph uses a mixture of components and contours, convert
+    /// all the components to contours.
     #[arg(long, default_value = "true")]
-    pub match_legacy: bool,
+    pub prefer_simple_glyphs: bool,
```
anthrotype, line 27: how about `decompose_mixed_glyphs`

fontc/src/args.rs
```diff
@@ -21,12 +21,15 @@ pub struct Args {
...
+
+    /// Eliminate component references to other glyphs using components, emitting only
```
anthrotype, line 29: I like the word "nested" if you can place it somewhere in this docstring

fontc/src/args.rs
```diff
@@ -21,12 +21,15 @@ pub struct Args {
...
+    /// component references to simple (contour) glyphs.
+    #[arg(long, default_value = "true")]
+    pub flatten_components: bool,
```
anthrotype, line 32: I would very much prefer to stick to vanilla fontmake default behavior which doesn _not_ flatten components by default, see https://github.com/googlefonts/fontmake-rs/issues/250

fontir/src/glyph.rs
```diff
@@ -288,56 +368,36 @@ impl Work<Context, WorkError> for FinalizeStaticMetadataWork {
...
-                        glyph_to_fix.name
+                        glyph.name
                     );
-                } else {
+                    convert_components_to_contours(context, &glyph)?;
+                } else if context.flags.contains(Flags::PREFER_SIMPLE_GLYPHS) {
                     debug!(
-                        "Coalescing'{0}' into a simple glyph because it has contours and components and that's how fontmake handles it",
-                        glyph_to_fix.name
+                        "Coalescing'{0}' into a simple glyph because it has contours and components and prefer simple glyphs is set",
```
anthrotype, line 383:
```suggestion
                        "Coalescing '{0}' into a simple glyph because it has contours and components and prefer simple glyphs is set",
```

anthrotype: also missing a space in the previous `debug!("Coalescing'{0}'...`

fontir/src/glyph.rs
```diff
@@ -288,56 +368,36 @@ impl Work<Context, WorkError> for FinalizeStaticMetadataWork {
...
+                        glyph.name
                     );
+                    convert_components_to_contours(context, &glyph)?;
```
anthrotype, line 386: this else if branch is doing exactly the same thing as the previous `if inconsistent_components` branch, is it just for the different debugging message?

fontir/src/orchestration.rs
```diff
@@ -69,8 +72,9 @@ macro_rules! context_accessors {
 }
 
 impl Default for Flags {
+    /// Match the way gftools configures fontmake by default
     fn default() -> Self {
-        Flags::MATCH_LEGACY
+        Flags::PREFER_SIMPLE_GLYPHS | Flags::FLATTEN_COMPONENTS
```
anthrotype, line 77: again, if you ask me, it should not Flags::FLATTEN_COMPONENTS by default

anthrotype:
it just occurred to me that, after flattening components, we may end up with orphan glyphs that are no longer used by composite glyphs and may also have no direct cmapping nor be substituted to from GSUB rules so they are basically dead weight, which a subsetter --unicodes='*' would surely get rid of, but maybe the compiler could/should be smarter and avoid them to be left in.
see https://github.com/googlefonts/ufo2ft/issues/692

fontc/src/args.rs
rsheeter: Per IM I agree, will so update.

fontir/src/glyph.rs
rsheeter: Yes.

rsheeter:
> after flattening components, we may end up with orphan glyphs ...

Good point, filed https://github.com/googlefonts/fontmake-rs/issues/270

fontir/src/orchestration.rs
rsheeter: I was convinced of this in IM, will update the default.

fontc/src/args.rs
rsheeter: on it!

PR #269, opened 2023-04-19
Preallocate delta buffer

rsheeter:
Purely out of curiosity - we're far from ready for any real perf tuning - I ran a flamegraph of Oswald building (`rm -rf build/ && cargo flamegraph -p fontc --  --source ../OswaldFont/sources/Oswald.glyphs --emit-ir false`). It showed an inordinate amount of time in variationmodel deltas, `fontir::variations::VariationModel::deltas (404 samples, 10.75%)`, dominated by `alloc::raw_vec::RawVec<T,A>::reserve_for_push`. Reserve ahead, reducing deltas enough to be hard to find on the flamegraph.

### DIFF
--- a/fontir/src/variations.rs
+++ b/fontir/src/variations.rs
@@ -191,6 +191,8 @@ impl VariationModel {
         {
             let master_influences = &self.delta_weights[model_idx];
             let mut deltas = Vec::new();
+            deltas.reserve(points.len());
+
             for (idx, point) in points.iter().enumerate() {
                 let initial_vector: V = *point - Default::default();
                 deltas.push(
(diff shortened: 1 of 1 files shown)


fontir/src/variations.rs
```diff
@@ -191,6 +191,8 @@ impl VariationModel {
         {
             let master_influences = &self.delta_weights[model_idx];
             let mut deltas = Vec::new();
+            deltas.reserve(points.len());
```
cmyr, line 194: try `Vec::with_capacity(points.len())` instead of `Vec::new` instead of this. It's a good habit in general anytime you know the collection length ahead of time :)

rsheeter: Done

anthrotype (approved):

dfrg (approved):

PR #271, opened 2023-04-20
Make arg and flag defaults match

rsheeter:
This makes emit ir default on so I had to turn it off for tests where I don't want that.


dfrg (approved):
LGTM

PR #272, opened 2023-04-20
Centralize versions of shared dependencies

rsheeter:
It's unlikely these wouldn't match on purpose. If we do need inconsistency it becomes very visible: everything but that uses workspace, why doesn't that?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,40 @@
+[workspace.dependencies]
+
+bincode = "1.3.3"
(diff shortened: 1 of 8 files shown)


anthrotype:
i'm no cargo expert so I'll leave the LGTM to others, but I like the idea!

cmyr (approved):
this workspace dependencies stuff is new to me, seems like a nice improvement.

PR #273, opened 2023-04-27
Don't crash when a glyph bbox x_max is < 0

rsheeter:
Amusingly the sample compile at https://github.com/googlefonts/fontmake-rs#sources-to-play-with triggers this bug.

### DIFF
--- a/fontbe/src/metrics_and_limits.rs
+++ b/fontbe/src/metrics_and_limits.rs
@@ -25,7 +25,7 @@ pub fn create_metric_and_limit_work() -> Box<BeWork> {
 #[derive(Debug, Default)]
 struct GlyphLimits {
     min_left_side_bearing: Option<i16>,
-    min_right_side_bearing: Option<u16>,
+    min_right_side_bearing: Option<i16>,
     x_max_extent: Option<i16>,
     advance_width_max: u16,
     max_points: u16,
@@ -44,10 +44,10 @@ impl GlyphLimits {
         let bbox = glyph.bbox();
         let left_side_bearing = bbox.x_min;
         // aw - (lsb + xMax - xMin) ... but if lsb == xMin then just advance - xMax?
-        let right_side_bearing = if bbox.x_max > 0 {
-            advance.saturating_sub(bbox.x_max.try_into().unwrap())
-        } else {
-            0
+        let right_side_bearing: i16 = match advance as i32 - bbox.x_max as i32 {
+            value if value < i16::MIN as i32 => i16::MIN,
+            value if value > i16::MAX as i32 => i16::MAX,
+            value => value as i16,
         };
         self.min_left_side_bearing = self
             .min_left_side_bearing
@@ -139,14 +139,7 @@ impl Work<Context, Error> for MetricAndLimitWork {
             .min_left_side_bearing
             .unwrap_or_default()
             .into();
-        let min_right_side_bearing: i16 = glyph_limits
-            .min_right_side_bearing
-            .unwrap_or_default()
-            .try_into()
-            .map_err(|_| Error::OutOfBounds {
-                what: "min_right_side_bearing".into(),
-                value: format!("{:?}", glyph_limits.min_right_side_bearing),
-            })?;
+        let min_right_side_bearing = glyph_limits.min_right_side_bearing.unwrap_or_default();
         let min_right_side_bearing = min_right_side_bearing.into();
         let x_max_extent = glyph_limits.x_max_extent.unwrap_or_default().into();
         let hhea = Hhea {
@@ -209,6 +202,13 @@ mod tests {
                 )
                 .unwrap(),
             ),
+        );
+        assert_eq!(
+            (Some(-437), Some(334)),
+            (
+                glyph_limits.min_left_side_bearing,
+                glyph_limits.min_right_side_bearing
+            )
         );
     }
 }
--- a/fontbe/src/metrics_and_limits.rs
+++ b/fontbe/src/metrics_and_limits.rs
@@ -44,7 +44,11 @@ impl GlyphLimits {
         let bbox = glyph.bbox();
         let left_side_bearing = bbox.x_min;
         // aw - (lsb + xMax - xMin) ... but if lsb == xMin then just advance - xMax?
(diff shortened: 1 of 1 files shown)


fontbe/src/metrics_and_limits.rs
```diff
@@ -44,7 +44,11 @@ impl GlyphLimits {
         let bbox = glyph.bbox();
         let left_side_bearing = bbox.x_min;
         // aw - (lsb + xMax - xMin) ... but if lsb == xMin then just advance - xMax?
-        let right_side_bearing = advance.saturating_sub(bbox.x_max.try_into().unwrap());
+        let right_side_bearing = if bbox.x_max > 0 {
+            advance.saturating_sub(bbox.x_max.try_into().unwrap())
+        } else {
+            0
```
anthrotype, line 50: why can't the bbox.x_max and thus the right side bearing be negative? hhea.minRightSideBearing is an FWORD and as such signed

this is how ufo2ft computes the hhea min_{left,right}_side_bearings
https://github.com/googlefonts/ufo2ft/blob/0d2688cd847d003b41104534d16973f72ef26c40/Lib/ufo2ft/outlineCompiler.py#L778-L797

fontbe/src/metrics_and_limits.rs
```diff
@@ -184,3 +188,27 @@ impl Work<Context, Error> for MetricAndLimitWork {
...
+    use write_fonts::tables::glyf::SimpleGlyph;
+
+    use super::GlyphLimits;
+
+    // advance 0, bbox (-437,611) => (-334, 715) encountered in NotoSansKayahLi.designspace
+    #[test]
+    fn negative_xmax_does_not_crash() {
+        let mut glyph_limits = GlyphLimits::default();
+        // path crafted to give the desired bbox
+        glyph_limits.update(
```
anthrotype, line 204: should you not assert the result is as expected instead of simply not crashing?

anthrotype:
If you take this test font I made, which only contains one .notdef glyph with 0 advance and containing a rectangle that is on the left hand side thus its xmax is negative: [NegXMax.glyphs.zip](https://github.com/googlefonts/fontmake-rs/files/11342979/NegXMax.glyphs.zip)

If I compare building that with fontmake vs fontc with this PR, the hhea minRightSideBearing differs (`-` is fontmake-py, `+` is fontc):

```diff
-    <minRightSideBearing value="137"/>
+    <minRightSideBearing value="0"/>
```

fontbe/src/metrics_and_limits.rs
anthrotype: GlyphLimits::min_right_side_bearing is currently typed `Option<u16>` but it should be `Option<i16>` instead (i.e. FWORD)

anthrotype:
something like this matches ufo2ft output for hhea (ascent and caretSlopeRise are still different but that's a different issue)

```diff
diff --git a/fontbe/src/metrics_and_limits.rs b/fontbe/src/metrics_and_limits.rs
index 704fc24..803fabe 100644
--- a/fontbe/src/metrics_and_limits.rs
+++ b/fontbe/src/metrics_and_limits.rs
@@ -22,10 +22,10 @@ pub fn create_metric_and_limit_work() -> Box<BeWork> {
     Box::new(MetricAndLimitWork {})
 }

-#[derive(Debug, Default)]
+#[derive(Debug, Default, PartialEq, Eq)]
 struct GlyphLimits {
     min_left_side_bearing: Option<i16>,
-    min_right_side_bearing: Option<u16>,
+    min_right_side_bearing: Option<i16>,
     x_max_extent: Option<i16>,
     advance_width_max: u16,
     max_points: u16,
@@ -44,11 +44,7 @@ impl GlyphLimits {
         let bbox = glyph.bbox();
         let left_side_bearing = bbox.x_min;
         // aw - (lsb + xMax - xMin) ... but if lsb == xMin then just advance - xMax?
-        let right_side_bearing = if bbox.x_max > 0 {
-            advance.saturating_sub(bbox.x_max.try_into().unwrap())
-        } else {
-            0
-        };
+        let right_side_bearing: i16 = (advance as i32 - bbox.x_max as i32).try_into().unwrap();
         self.min_left_side_bearing = self
             .min_left_side_bearing
             .map(|v| min(v, left_side_bearing))
@@ -139,15 +135,10 @@ impl Work<Context, Error> for MetricAndLimitWork {
             .min_left_side_bearing
             .unwrap_or_default()
             .into();
-        let min_right_side_bearing: i16 = glyph_limits
+        let min_right_side_bearing = glyph_limits
             .min_right_side_bearing
             .unwrap_or_default()
-            .try_into()
-            .map_err(|_| Error::OutOfBounds {
-                what: "min_right_side_bearing".into(),
-                value: format!("{:?}", glyph_limits.min_right_side_bearing),
-            })?;
-        let min_right_side_bearing = min_right_side_bearing.into();
+            .into();
         let x_max_extent = glyph_limits.x_max_extent.unwrap_or_default().into();
         let hhea = Hhea {
             ascender: FWord::new(default_metrics.ascender.into_inner().ot_round()),
@@ -210,5 +201,17 @@ mod tests {
                 .unwrap(),
             ),
         );
+        assert_eq!(
+            glyph_limits,
+            GlyphLimits {
+                min_left_side_bearing: Some(-437),
+                min_right_side_bearing: Some(334),
+                x_max_extent: Some(-334),
+                advance_width_max: 0,
+                max_points: 3,
+                max_contours: 1,
+                max_component_elements: 0
+            }
+        )
     }
 }
```

anthrotype:
For NotoSansKayahLi, the hhea minRightSideBearing as built by ufo2ft via fontmake-py is -109, whereas with your current patch is clamped to 0.

```diff
-    <minRightSideBearing value="0"/>
+    <minRightSideBearing value="-109"/>
```

rsheeter:
Oh good catch, I mistakenly assumed the u16 was the correct destination type.

fontbe/src/metrics_and_limits.rs
rsheeter: Fixed

anthrotype (approved):
LGTM

PR #274, opened 2023-04-28
From plist fixups

### LINKED ISSUE #257: FromPlist support for empty vec
We end up with things like `axes_values: Option<Vec<OrderedFloat<f64>>>,` when we're prefer to just get an empty Vec if there aren't any values. See if we can make FromPlist do that and if we can remove Option from various things in glyphs-reader/src/font.rs

cmyr:
I had a small block of time so I decided to take a look at the fromplist macro, and fix #257.

I also tried to add a bit more error support. This macro is pretty hacky and I started trying to rationalize it but that is a project beyond the scope of my afternoon, so I'm backing away. There's definitely a whole whack of unnecessary panicking in there, though.

I've updated some of the code to use the new macro attribute, but i'm not totally sure where all we would want it; I figure after this gets merged y'all can fix up the code to take advantage of it as needed.

### DIFF
--- a/glyphs-reader/plist_derive/src/attrs.rs
+++ b/glyphs-reader/plist_derive/src/attrs.rs
@@ -1,0 +1,28 @@
+use syn::{spanned::Spanned, Attribute};
+
+/// Attributes on fields
+#[derive(Clone, Default)]
+pub(crate) struct FieldAttrs {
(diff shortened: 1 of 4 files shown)


rsheeter (approved):
LGTM, please make clippy happy

cmyr:
oops i'm running an old rustc locally

PR #277, opened 2023-05-04
Filter implied on curves for interpolatable glyphs

anthrotype:
~~requires yet to be released write-fonts from https://github.com/googlefonts/fontations/pull/368~~

~~EDIT: [write-fonts 0.2.0](https://crates.io/crates/write-fonts/0.2.0) has been released. However fea-rs also needs to bump its own write-fonts dependency before this can work (https://github.com/cmyr/fea-rs/pull/133)~~

Requires write-fonts 0.2.0 and fea-rs 0.3.2


rsheeter (approved):
lgtm

PR #279, opened 2023-05-16
Fire IUP optimization per glyph

rsheeter:
First tilt at the windmill.

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -8,15 +8,18 @@ use std::collections::{BTreeSet, HashMap, HashSet};
 use fontdrasil::{orchestration::Work, types::GlyphName};
 use fontir::{coords::NormalizedLocation, ir};
 use kurbo::{cubics_to_quadratic_splines, Affine, BezPath, CubicBez, PathEl, Point, Rect};
-use log::{trace, warn};
+use log::{log_enabled, trace, warn};
 
 use read_fonts::{
     tables::glyf::{self, Anchor, Transform},
     types::{F2Dot14, GlyphId},
 };
 use write_fonts::{
-    tables::glyf::{
-        simple_glyphs_from_kurbo, Bbox, Component, ComponentFlags, CompositeGlyph, SimpleGlyph,
+    tables::{
+        glyf::{
+            simple_glyphs_from_kurbo, Bbox, Component, ComponentFlags, CompositeGlyph, SimpleGlyph,
+        },
+        variations::iup_delta_optimize,
     },
     OtRound,
 };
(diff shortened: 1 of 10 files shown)


fontbe/src/glyphs.rs
```diff
@@ -258,14 +265,49 @@ impl Work<Context, Error> for GlyphWork {
...
+                )
             }
         };
 
+        let coords = point_seqs.get(default_location).ok_or_else(|| {
+            Error::GlyphError(ir_glyph.name.clone(), GlyphProblem::MissingDefault)
+        })?;
+
+        // TODO real values
+        let tolerance = 0.1;
```
anthrotype, line 287: I think fonttools varLib uses tolerance=0.5, and it doesn't look like it allows to override that from the public varLib.build(), so we should match that I think:

https://github.com/fonttools/fonttools/blob/65bc6105f7aec3478427525d23ddf2e3c8c4b21e/Lib/fontTools/varLib/__init__.py#L239

fontbe/src/glyphs.rs
```diff
@@ -258,14 +265,49 @@ impl Work<Context, Error> for GlyphWork {
                 };
                 context.set_glyph(name.clone(), base_glyph.clone().into());
 
-                (name, point_seqs_for_simple_glyph(ir_glyph, instances))
+                let mut contour_end = 0;
+                let mut contour_ends = Vec::with_capacity(base_glyph.contours().len());
+                for contour in base_glyph.contours() {
```
anthrotype, line 270: maybe write-fonts' SimpleGlyph could have a method to get a vec of contour end-point indices

fontbe/src/glyphs.rs
```diff
@@ -258,14 +265,49 @@ impl Work<Context, Error> for GlyphWork {
...
+                    iup_delta_optimize(deltas, coords.clone(), tolerance, &contour_ends)
+                        .map(|iup_deltas| (region.clone(), iup_deltas))
+                } else {
+                    Ok((region, deltas.into_iter().map(Some).collect()))
+                }
+            })
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(|e| Error::IupError(ir_glyph.name.clone(), e))?;
 
         context.set_gvar_fragment(name, GvarFragment { deltas });
```
anthrotype, line 312: I think you should check if all deltas are None (because they are all implied, meaning that the deltas were originally all no-op zeros), and if so we could just skip the GvarFragment since it doesn't do anything.. Unless something else down the pipeline is already taking care of that?

At least for simple glyphs.. remember we do _not_ want to drop these all-zero gvar tuples for the composite glyphs to match the fonttools workaround at https://github.com/googlefonts/fontmake-rs/issues/283

fontbe/src/glyphs.rs
rsheeter: good catch, making it so

fontbe/src/glyphs.rs
rsheeter: How about if it comes up 1-2 more times we do that? - for now I'm not sure it will.

fontbe/src/glyphs.rs
rsheeter: I'm not doing iup for composites at all which I think should suffice?

Good point wrt all None.

rsheeter: I'm merrily producing empty deltas. I think I should be filtering them out indeed.

rsheeter: Fixed

PR #284, opened 2023-05-18
Use font-types serde support

### LINKED ISSUE #278: Take advantage of font-types serde
https://github.com/googlefonts/fontations/pull/393 adds serde for font-types so the gymnastics in https://github.com/googlefonts/fontmake-rs/pull/215 to work around this could be undone

rsheeter:
Fixes #278. Includes minimal changes to accomodate updated signature for iup. JMM.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -4,8 +4,8 @@ use crate::{
     coords::{CoordConverter, NormalizedCoord, NormalizedLocation, UserCoord},
     error::{PathConversionError, VariationModelError, WorkError},
     serde::{
-        deserialize_name_id, deserialize_tag, serialize_name_id, serialize_tag,
-        GlobalMetricsSerdeRepr, GlyphSerdeRepr, StaticMetadataSerdeRepr,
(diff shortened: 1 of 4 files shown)


fontir/src/ir.rs
```diff
@@ -4,8 +4,8 @@ use crate::{
     coords::{CoordConverter, NormalizedCoord, NormalizedLocation, UserCoord},
     error::{PathConversionError, VariationModelError, WorkError},
     serde::{
-        deserialize_name_id, deserialize_tag, serialize_name_id, serialize_tag,
-        GlobalMetricsSerdeRepr, GlyphSerdeRepr, StaticMetadataSerdeRepr,
+        deserialize_name_id, serialize_name_id, GlobalMetricsSerdeRepr, GlyphSerdeRepr,
```
dfrg, line 7: Can we drop the extra serialization support for name id as well?

dfrg (approved):
LGTM with a comment.

fontir/src/ir.rs
rsheeter: I believe we can!

Issue #285, opened 2023-05-18
hmtx for Oswald differs from the one fontmake (py) emits


rsheeter:
A major source of fontc vs fontmake diffs is hmtx, mostly (though not entirely) due to lsb 0 for some composites:

![image](https://github.com/googlefonts/fontmake-rs/assets/6466432/c9e10bed-6f76-4236-a587-9def291abb1c)

rsheeter:
Seems to be due to ordering issues that are non-obvious courtesy of https://github.com/googlefonts/fontmake-rs/issues/141.

rsheeter:
After #286 biggest issue is some glyphs have different x-min in the rust version. At a glance the Rust version is just wrong, for example a glyph containing `<pt x="15" y="535" on="0"/>` is given `xMin="17"`.

rsheeter:
Looks like we end up calling https://github.com/linebender/kurbo/blob/master/src/param_curve.rs#L214 which is a tighter box than the simple control point box. Per https://learn.microsoft.com/en-us/typography/opentype/spec/glyf#glyph-headers we actually want the cbox.

rsheeter:
See https://github.com/googlefonts/fontmake-rs/pull/286#issuecomment-1553168738 for probable cause of some hmtx entries getting width=0 from fontmake and not from fontc.

rsheeter:
Using unicode data and codepoints I don't seem to get the same answer as fontmake-py for what is/isn't a non-spacing mark. It seems glyphsLib bundles https://github.com/googlefonts/glyphsLib/blob/main/Lib/glyphsLib/data/GlyphData.xml for this.

khaledhosny:
Yes, this and the category/subCategory set in the .glyphs files are the wayy to handle this. Any glyph can be made a mark glyph.

PR #286, opened 2023-05-18
Adjust the dependencies for hmtx to ensure bbox is available

rsheeter:
Helps with #285, after this hmtx has just a handful of diffs. Most are small diffs in lsb, for example 15 vs 18, plus one weird one `<mtx name="brevecomb-cy" width="600" lsb="167"/>` that gets a width of 0 in the fontmake-py version.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -589,18 +589,14 @@ fn add_metric_and_limits_job(
 
     // If no glyph has changed there isn't a lot to do
     if !glyphs_changed.is_empty() {
-        let mut dependencies: HashSet<_> = glyphs_changed
-            .iter()
-            .flat_map(|gn| {
-                [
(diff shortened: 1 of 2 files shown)


dfrg (approved):
LGTM. Combining breve should be a NSM so maybe advance width is being intentionally zeroed somewhere in fontmake-py? If so, seems weird that this hasn’t come up before.

khaledhosny:
> Combining breve should be a NSM so maybe advance width is being intentionally zeroed somewhere in fontmake-py? If so, seems weird that this hasn’t come up before.

Are these Glyphs sources? GlyphsApp and glyphsLib zero the width of nonspacing marks https://github.com/googlefonts/glyphsLib/blob/6372da0c1f77cad3152364c2473b7a380ac6c301/Lib/glyphsLib/builder/glyph.py#L129-L134

PR #287, opened 2023-05-18
Explore using the glyphsLib datafile to figure out if things are nonspacing marks

### LINKED ISSUE #285: hmtx for Oswald differs from the one fontmake (py) emits

rsheeter:
As noted on https://github.com/googlefonts/fontmake-rs/issues/285#issuecomment-1553653295, just using unicode properties - my initial theory - doesn't produce the same result. `python resources/scripts/ttx_diff.py ../OswaldFont/sources/Oswald.glyphs` now reports identical hmtx for fontmake vs fontc.

Per https://github.com/googlefonts/fontmake-rs/issues/285#issuecomment-1553710896 there are further cases that won't be handled correctly yet; tracked in #290.

Fixes #285.


resources/scripts/non_spacing_marks.py
```diff
@@ -0,0 +1,58 @@
...
+from lxml import etree
+from pathlib import Path
+import sys
+
+
+def main(_):
+    glyphs_lib = Path("../glyphsLib")
+    if not glyphs_lib.is_dir():
+        sys.exit(f"Missing {glyphs_lib}")
+    glyphdata_file = glyphs_lib / "Lib/glyphsLib/data/GlyphData.xml"
```
anthrotype, line 20: instead of expecting that glyphsLib source is cloned in a sibling folder, you could rely on the fact that these data files are shipped alongside the python modules as resources and use importlib to get to them, like glyphsLib itself does:

https://github.com/googlefonts/glyphsLib/blob/6372da0c1f77cad3152364c2473b7a380ac6c301/Lib/glyphsLib/glyphdata.py#L106-L108C7

this way, it doesn't matter where glyphsLib is, it's sufficient that it is importable. You could even get the `glyphsLib.__version__` and write that out to the generated .rs file so this is recorded (the glyphdata xml may change from time to time and it'd be good to track exactly which one we are using).

resources/scripts/non_spacing_marks.py
```diff
@@ -0,0 +1,58 @@
...
+    with open("glyphs2fontir/src/glyphdata.rs", "w") as f:
+        f.write("//! Generated module with Glyphs-specific data\n")
+        f.write("//!\n")
+        f.write("//! Run resources/scripts/non_spacing_marks.py to update\n")
+        
+        f.write("\n");
+        f.write("use std::collections::HashSet;\n")
+        f.write("\n")
+        f.write("pub(crate) fn is_nonspacing_mark(codepoints: &HashSet<u32>, name: &str) -> bool {\n")
+        f.write("    match codepoints.iter().next() {\n")
```
anthrotype, line 44: wouldn't it be preferable to use HashSets instead of a match here? you already know for which codepoints or glyph names GlyphData.xml sets those two attributes, you could hard-code them in this file (maybe using https://crates.io/crates/lazy_static) and then just check if any of the inputs is in those sets in O(1) time, I believe.

resources/scripts/non_spacing_marks.py
anthrotype: Don't know if you noticed, for CJK glyphs there's an additional XML data file as well.
Actually, instead of parsing it yourself with etree, since you are already requiring glyphsLib to be present, why don't you simply import glyphsLib.glyphdata and let it load both the GlyphDatas for you. Then you'd query them for the attributes that you are interested in. E.g. something like this:

```python
from glyphsLib.glyphdata import GlyphData

with (
    open_binary("glyphsLib.data", "GlyphData.xml") as f1,
    open_binary("glyphsLib.data", "GlyphData_Ideographs.xml") as f2,
):
    GLYPHDATA = GlyphData.from_files(f1, f2)

non_spacing_marks_by_name = set()
non_spacing_marks_by_unicode = set()

for name, attrs in GLYPHDATA.names.items():
    category = attrs.get("category")
    sub_category = attrs.get("subCategory")
    if category == "Mark" and sub_category == "Nonspacing":
        non_spacing_marks_by_name.add(name)
        if "production" in attrs:
            non_spacing_marks_by_name.add(attrs["production"])
        if "altNames" in attrs:
            non_spacing_marks_by_name.update(attrs["altNames"].replace(" ", "").split(","))
        if "unicode" in attrs:
            non_spacing_marks_by_unicode.add(attrs["unicode"])
```

resources/scripts/non_spacing_marks.py
anthrotype: notes that glyphsLib (and probably Glyphs.app too) looks things up by glyph name first and only when that's not found it looks them up by unicode so you should invert the ordering of the checks you have here.

Also I see with `codepoints.iter().next()` you are basically only checking one codepoint if any, the first one, and it is even a HashSet so I believe the order in which you iterate is random? glyphsLib looks up all the codepoints, not just the first, and breaks at the first found. I believe the `unicodes` is an ordered list rather than a set in in Glyphs (like in UFO, even if it doesn't make any sense because a "primary" codepoint is meaningless, cmap maps from codepoint to glyphs not viceversa).

Anyway, look up by name and if that's not found try to look up by unicode, all of them not just the random first -- this should match glyphsLib, at least for the data actually recorded in the XML files. There's other stuff that is made up at runtime when look up fails (see `_construct_production_name` and `_construct_category` in `glyphsLib.glyphdata`), but we can tackle that later)

rsheeter: > wouldn't it be preferable to use HashSets instead of a match here

I would expect the compiler output for the match to be _at least_ as good, probably better, than a general purpose hashset. We're giving it more to work with, it knows all the possible items. Also HashSet doesn't support const construction so we want something like [oncecell](https://doc.rust-lang.org/std/cell/struct.OnceCell.html) but that's still nightly/experimental.

> glyphsLib looks up all the codepoints, not just the first, and breaks at the first found

That's a bug, will fix. I assumed unicodes to always contain *one* codepoint for no particularly good reason.

resources/scripts/non_spacing_marks.py
rsheeter: Apparently openbinary is deprecated, will use the mechanism suggested by https://docs.python.org/3/library/importlib.resources.html#importlib.resources.open_binary

rsheeter: Done

resources/scripts/non_spacing_marks.py
rsheeter: I don't have the original ordering handy, if we need that I'd prefer to tackle it in a separate issue.

resources/scripts/non_spacing_marks.py
anthrotype: you no longer need to check that the glyphdata_file is  at `glyphs_lib / "Lib/glyphsLib/data/GlyphData.xml"`, because you are now using importlib.resources to get to the data file, glyphsLib need only be importable now, doesn't matter where it is located on the filesystem.

anthrotype (approved):
LGTM -- but see my comment about no longer needing to assert that glyphsLib directory or data file exists on some hard-coded path since we're using importlib to find the data file.

resources/scripts/non_spacing_marks.py
anthrotype: it should be fine

resources/scripts/non_spacing_marks.py
rsheeter: Oops, will fix.

rsheeter: Hm. I think I already did.

rsheeter: Ha, but only locally. Push is an important step.

PR #288, opened 2023-05-19
Helper for frequent diff pattern

rsheeter:
Builds on #287 because I was using it to help test that. Submit #287 first. Current result:

```
COMPARISON
  Only fontmake produced 'GDEF', 'GPOS', 'GSUB', 'HVAR', 'STAT'
  Identical 'GlyphOrder'
  DIFF 'OS_2', build/fontc.OS_2.ttx build/fontmake.OS_2.ttx
  Identical 'avar'
  DIFF 'cmap', build/fontc.cmap.ttx build/fontmake.cmap.ttx
  DIFF 'fvar', build/fontc.fvar.ttx build/fontmake.fvar.ttx
  DIFF 'glyf', build/fontc.glyf.ttx build/fontmake.glyf.ttx
  DIFF 'gvar', build/fontc.gvar.ttx build/fontmake.gvar.ttx
  DIFF 'head', build/fontc.head.ttx build/fontmake.head.ttx
  DIFF 'hhea', build/fontc.hhea.ttx build/fontmake.hhea.ttx
  Identical 'hmtx'
  Identical 'loca'
  DIFF 'maxp', build/fontc.maxp.ttx build/fontmake.maxp.ttx
  DIFF 'name', build/fontc.name.ttx build/fontmake.name.ttx
  DIFF 'post', build/fontc.post.ttx build/fontmake.post.ttx
```

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -9,7 +9,9 @@ fontmake should be installed in an active virtual environment.
 Usage:
     # rebuild with fontmake and fontc and compare
     python resources/scripts/ttx_diff.py ../OswaldFont/sources/Oswald.glyphs
+
     # rebuild the fontc copy, reuse the prior fontmake copy (if present), and compare
+    # useful if you are making changes to fontc meant to narrow the diff
     python resources/scripts/ttx_diff.py --rebuild fontc ../OswaldFont/sources/Oswald.glyphs
 """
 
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -0,0 +1,157 @@
...
+    cmd = [
+        "fontmake",
+        "-o",
+        "variable",
+        "--no-production-names",
+        "--keep-direction",
+        "--filter",
+        "FlattenComponentsFilter",
+        "--filter",
+        "DecomposeTransformedComponentsFilter",
```
anthrotype, line 88: note that this optional filter is more eager than our current implementation in fontc. This one will decompose any transformed components anywhere, not just those components that have different 2x2 in different masters like we currently do. We could add an option to do the former (off by default because fontmake doesn't enable this by default).

resources/scripts/ttx_diff.py
```diff
@@ -0,0 +1,157 @@
...
+
+def build_fontmake(source, build_dir):
+    cmd = [
+        "fontmake",
+        "-o",
+        "variable",
+        "--no-production-names",
+        "--keep-direction",
+        "--filter",
+        "FlattenComponentsFilter",
```
anthrotype, line 86: isn't --flatten_components false by default?

https://github.com/googlefonts/fontmake-rs/blob/f4a9940e2a225a04726d5d923cc01e884d730b54/fontc/src/args.rs#L32

it doesn't look like you have enabled that in the call to fontc. Not saying that you should, just that the fontc vs fontmake calls don't match my expectation given my current understanding as to where we currently stand in regards default options.

anthrotype (approved):
this is very useful, thanks!

resources/scripts/ttx_diff.py
rsheeter: I wonder if we need to test with/without flatten (for both tools) separately? - we do care it can work in gftools (with it on) and vs default fontmake behavior (with it off)

anthrotype: maybe add options to this script to forward to each compiler the relevant filters/options

rsheeter: Adding a flag for this, `flags.DEFINE_enum('compare', 'both', ['both', 'default', 'gftools'], 'Compare results with default flags, with the flags gftools uses, or both. Default both.')`

resources/scripts/ttx_diff.py
rsheeter: Good point, moving to a standalone issue.

rsheeter: https://github.com/googlefonts/fontmake-rs/issues/291

PR #289, opened 2023-05-19
Use latest write-fonts so we produce identical cmap's for Oswald.

### LINKED ISSUE #251: cmap formats
The default build of Oswald with fontmake has a ` <cmap_format_4 platformID="0" platEncID="3" language="0">` _and_ a `<cmap_format_4 platformID="3" platEncID="1" language="0">`. fontmake-rs emits only the latter. We appear to agree on the content so it appears we just need to decide if we should emit the extra one.

rsheeter:
Fixes #251. Do not just merge because the key fontations change, in https://github.com/googlefonts/fontations/pull/417, is not yet published.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ parking_lot = "0.12.1"
 fea-rs = "= 0.3.3"
 font-types = { version = "0.1.6", features = ["serde"] }
 read-fonts = "0.2.0"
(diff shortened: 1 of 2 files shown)


anthrotype (approved):
LGTM in advance of the yet unreleased write-fonts 0.3.5 ;)

PR #293, opened 2023-05-21
Add basic support for named instances for .glyphs files

rsheeter:
Sufficient to make Oswald fvar report to match the one from fontmake:

```shell
python resources/scripts/ttx_diff.py ../OswaldFont/sources/Oswald.glyphs | grep "'fvar'"
  Identical 'fvar'
  Identical 'fvar'
```

New test files created by loading WghtVar.glyphs in Glyphs, running add instance for each master, and saving as v2/3.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -1,7 +1,7 @@
 //! Font IR types.
 
 use crate::{
-    coords::{CoordConverter, NormalizedCoord, NormalizedLocation, UserCoord},
+    coords::{CoordConverter, NormalizedCoord, NormalizedLocation, UserCoord, UserLocation},
     error::{PathConversionError, VariationModelError, WorkError},
     serde::{GlobalMetricsSerdeRepr, GlyphSerdeRepr, StaticMetadataSerdeRepr},
     variations::VariationModel,
@@ -46,6 +46,9 @@ pub struct StaticMetadata {
     /// If empty this is a static font.
     pub variable_axes: Vec<Axis>,
 
+    /// Named locations in variation space
+    pub named_instances: Vec<NamedInstance>,
+
     /// The name of every glyph, in the order it will be emitted
     ///
     /// <https://rsheeter.github.io/font101/#glyph-ids-and-the-cmap-table>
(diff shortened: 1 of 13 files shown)


glyphs2fontir/src/toir.rs
```diff
@@ -116,9 +116,9 @@ pub(crate) fn to_ir_features(features: &[FeatureSnippet]) -> Result<ir::Features
     Ok(ir::Features::Memory(fea_snippets.join("\n\n")))
 }
 
-fn design_location(axes: &[ir::Axis], master: &FontMaster) -> DesignLocation {
+fn design_location(axes: &[ir::Axis], axes_values: &Vec<OrderedFloat<f64>>) -> DesignLocation {
```
anthrotype, line 119:
```suggestion
fn design_location(axes: &[ir::Axis], axes_values: &[OrderedFloat<f64>]) -> DesignLocation {
```

resources/scripts/ttx_diff.py
```diff
@@ -138,6 +138,25 @@ def copy(old, new):
...
+def name_id_to_name(ttx, xpath, attr):
+    id_to_name = {
+        el.attrib["nameID"]: el.text.strip()
+        for el in ttx.xpath("//name/namerecord[@platformID='3' and @platEncID='1' and @langID='0x409']")
+    }
+    for el in ttx.xpath(xpath):
+        if attr not in el.attrib:
+            continue
+        print(f"{el.attrib[attr]} => {id_to_name[el.attrib[attr]]}")
+        el.attrib[attr] = id_to_name[el.attrib[attr]]
```
anthrotype, line 150: I don't understand why are you modifying the tree here? you want a NamedInstance's the subfamilyNameID attribute to contain the text string itself instead of the nameID index? Doesn't fonttools ttx already add that very same name string (as a xml comment) next to every nameID that is referenced?

fontir/src/ir.rs
```diff
@@ -444,6 +453,13 @@ impl Axis {
     }
 }
 
+/// IR for a named position in variation space
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct NamedInstance {
+    pub name: String,
+    pub location: NormalizedLocation,
```
anthrotype, line 460: I think it's better if the ir for NamedInstance location has a UserLocation, not a NormalizedLocation. That's what fvar contains after all. It would be confusing if the ir had internal normalized location because then, without knowing exactly the context, one could interpret as either as default-normalized or after-avar-normalized.. Better to store as user-space locations.

resources/scripts/ttx_diff.py
```diff
@@ -138,6 +138,25 @@ def copy(old, new):
...
+
+
+def reduce_diff_noise(fontc, fontmake):
+    for ttx in (fontc, fontmake):
+        # different name ids with the same value is fine
```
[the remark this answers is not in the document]
rsheeter: I should narrow it to ids > 255; it's just for things like the order we assigned name ids to axis names and/or instance names isn't quite identical which afaik is completely fine as long as we agree on the value.

resources/scripts/ttx_diff.py
rsheeter: For font-specific ids that are not assigned by the spec (e.g. >255) I don't want to care if the id differs. Concretely, I don't care about this:
image

So instead compare this:
image

fontir/src/ir.rs
rsheeter: Oh good point, let me see about adjusting that.

resources/scripts/ttx_diff.py
rsheeter: Added the check to not adjust names <= 255. Closing, as always please reopen if you disagree.

fontir/src/ir.rs
rsheeter: That makes fontbe fvar nicer too.

rsheeter: Done.

resources/scripts/ttx_diff.py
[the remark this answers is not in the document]
rsheeter: Resolving as so far as I can tell for >255 it doesn't matter what the value is.

anthrotype (approved):
LGTM!

Are you going to add support for DS instances here or in a separate PR?

rsheeter:
DS can come separately. Per IM this morning I've added comparison of the ufo+ds version to the milestone so we don't forget.

PR #294, opened 2023-05-23
Make Oswald OS/2 from fontc closer to that from fontmake

rsheeter:
Makes most, though not all, of OS/2 match. The following differences are not addressed in this PR:

image

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -155,23 +155,103 @@ pub struct GlobalMetrics(
 pub enum GlobalMetric {
     Ascender,
     Descender,
+    Os2TypoAscender,
+    Os2TypoDescender,
+    Os2TypoLineGap,
+    Os2WinAscent,
+    Os2WinDescent,
     CapHeight,
     XHeight,
+    ItalicAngle,
+    YSubscriptXSize,
+    YSubscriptYSize,
+    YSubscriptXOffset,
+    YSubscriptYOffset,
+    YSuperscriptXSize,
+    YSuperscriptYSize,
+    YSuperscriptXOffset,
+    YSuperscriptYOffset,
+    YStrikeoutSize,
+    YStrikeoutPosition,
+}
+
+/// Adjust Y offset based on italic angle, to get X offset.
+///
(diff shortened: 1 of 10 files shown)


fontir/src/ir.rs
```diff
@@ -146,23 +146,103 @@ pub struct GlobalMetrics(
...
         set(GlobalMetric::CapHeight, 0.7 * units_per_em as f32);
-        set(GlobalMetric::XHeight, 0.5 * units_per_em as f32);
+        let x_height = x_height.unwrap_or(0.5 * units_per_em as f32);
+        set(GlobalMetric::XHeight, x_height);
+
+        // https://github.com/googlefonts/ufo2ft/blob/0d2688cd847d003b41104534d16973f72ef26c40/Lib/ufo2ft/fontInfoData.py#L360
+        let italic_angle = 0.0;
+        set(GlobalMetric::ItalicAngle, italic_angle);
+
+        // https://github.com/googlefonts/ufo2ft/blob/0d2688cd847d003b41104534d16973f72ef26c40/Lib/ufo2ft/fontInfoData.py#L360
```
anthrotype, line 222: the link is the same as the one before for italiAngle, copypasta?

anthrotype (approved):

ufo2fontir/src/source.rs
```diff
@@ -614,14 +614,25 @@ impl Work<Context, WorkError> for GlobalMetricsWork {
...
+            static_metadata.default_location().clone(),
+            static_metadata.units_per_em,
+            font_infos
+                .get(&default_master.filename)
+                .ok_or_else(|| {
+                    WorkError::FileExpected(designspace_dir.join(&default_master.filename))
+                })?
+                .x_height
+                .map(|v| v as f32),
+        );
```
anthrotype, line 635: you should add a TODO somewhere here for adding support to reading OS/2 typo/win metrics from UFO fontinfo.plist.

Maybe we need an issue in which we track the differences in the support of various features between .glyphs vs .designspace input. Since lately you're been advancing the .glyphs side of things (e.g. #293) there's a risk that the UFO+DS one lags behind.

rsheeter: Will add todo.

fontir/src/ir.rs
rsheeter: That was exceptionally wrong, good catch. Fixed.

PR #295, opened 2023-05-23
Support ttx_diff.py run on a .designspace

rsheeter:
Tweak paths slightly. JMM.


cmyr (approved):
I do not know what this is doing, but I trust you will fix it if you broke it 🤝

PR #298, opened 2023-05-25
Make fsSelection match for Oswald

rsheeter:
In both .glyphs and .designspace form

### DIFF
--- a/fontbe/src/os2.rs
+++ b/fontbe/src/os2.rs
@@ -78,13 +78,6 @@ fn apply_metrics(os2: &mut Os2, metrics: &GlobalMetricsInstance) {
     os2.s_typo_line_gap = metrics.os2_typo_line_gap.ot_round();
     os2.us_win_ascent = metrics.os2_win_ascent.ot_round();
     os2.us_win_descent = metrics.os2_win_descent.ot_round();
-
-    // Avoid "field must be present for version 2"
-    os2.ul_code_page_range_1 = Some(0);
-    os2.ul_code_page_range_2 = Some(0);
-    os2.us_default_char = Some(0);
-    os2.us_break_char = Some(0);
-    os2.us_max_context = Some(0);
 }
 
 impl Work<Context, Error> for Os2Work {
@@ -98,9 +91,17 @@ impl Work<Context, Error> for Os2Work {
             .at(static_metadata.default_location());
 
         let mut os2 = Os2 {
-            ach_vend_id: static_metadata.vendor_id,
-            fs_selection: static_metadata.selection_flags,
+            ach_vend_id: static_metadata.misc.vendor_id,
+            fs_selection: static_metadata.misc.selection_flags,
             x_avg_char_width: x_avg_char_width(context)?,
+
+            // Avoid "field must be present for version 2" caused by default to None
+            ul_code_page_range_1: Some(0),
+            ul_code_page_range_2: Some(0),
+            us_default_char: Some(0),
+            us_break_char: Some(0),
+            us_max_context: Some(0),
+
             ..Default::default()
         };
         apply_metrics(&mut os2, &metrics);
--- a/fontbe/src/os2.rs
+++ b/fontbe/src/os2.rs
@@ -2,7 +2,7 @@
 
 use fontdrasil::orchestration::Work;
 use fontir::ir::GlobalMetricsInstance;
-use read_fonts::{tables::hmtx::Hmtx, types::Tag, FontData, TopLevelTable};
+use read_fonts::{tables::hmtx::Hmtx, FontData, TopLevelTable};
 use write_fonts::{tables::os2::Os2, OtRound};
 
 use crate::{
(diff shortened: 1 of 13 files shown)


fontir/src/ir.rs
```diff
@@ -32,6 +33,9 @@ pub struct StaticMetadata {
     /// See <https://learn.microsoft.com/en-us/typography/opentype/spec/head>.
     pub units_per_em: u16,
 
+    /// See <https://learn.microsoft.com/en-us/typography/opentype/spec/os2#fsselection>
+    pub selection_flags: SelectionFlags,
```
anthrotype, line 37: I think the `fsSelection` name expands to "font _style_ selection" , so maybe it would read better if you name it `style_selection_flags` ("font" is implied of course); just `selection_flags` was unclear to me until I looked at the comment

anthrotype: I well I see write_fonts calls them simply SelectionFlags, nevermind

glyphs-reader/src/font.rs
```diff
@@ -46,6 +46,7 @@ pub struct RawAxisUserToDesignMap(Vec<(OrderedFloat<f32>, OrderedFloat<f32>)>);
 #[derive(Debug, PartialEq, Hash)]
 pub struct Font {
     pub units_per_em: u16,
+    pub use_typo_metrics: Option<bool>,
```
anthrotype, line 49: I don't know.. `use_typo_metrics` in second position in this struct definition looks a bit too prominent for such a relatively insignificant bit that's basically almost always set nowadays...

glyphs2fontir/src/source.rs
```diff
@@ -300,8 +303,21 @@ impl Work<Context, WorkError> for StaticMetadataWork {
             .filter(|gn| self.glyph_names.contains(gn))
             .collect();
 
+        let selection_flags = match font.use_typo_metrics.unwrap_or_default() {
+            true => SelectionFlags::USE_TYPO_METRICS,
+            false => SelectionFlags::empty(),
+        } |
+        // https://github.com/googlefonts/glyphsLib/blob/42bc1db912fd4b66f130fb3bdc63a0c1e774eb38/Lib/glyphsLib/builder/names.py#L27
+        match font.default_master().name.to_ascii_lowercase().as_str() {
```
anthrotype, line 311: I'm afraid simply looking at the default master.name is not sufficient, because glyphsLib contains additional logic, see `build_stylemap_names` (that is called when exporting glyphs->ufo by fontmake because `minimize_glyphs_diffs` is False when compiling):

https://github.com/googlefonts/glyphsLib/blob/42bc1db912fd4b66f130fb3bdc63a0c1e774eb38/Lib/glyphsLib/builder/names.py#L27-L43

the final ufo.info.styleMapStyleName in the end is what matters to decide the style flags for ufo2ft, and the default UFO master that glyphsLib produces is going to be the default master TTF which varLib will duplicate to use as basis for the new variable font.

ufo2fontir/src/source.rs
```diff
@@ -626,8 +629,30 @@ impl Work<Context, WorkError> for StaticMetadataWork {
...
+        let selection_flags = font_info_at_default
+            .open_type_os2_selection
+            .as_ref()
+            .map(|flags| {
+                flags.iter().fold(SelectionFlags::empty(), |acc, e| {
+                    acc | SelectionFlags::from_bits_truncate(1 << e)
+                })
+            })
+            .unwrap_or_default()
+            | font_info_at_default
```
anthrotype, line 642: i don't immeditately understand what is going on with this `|` bitwise or operator here after the unwrap_or_default().. What is the intention of joining these two statements with `|` here?

anthrotype: We want to match this: https://github.com/googlefonts/ufo2ft/blob/0d2688cd847d003b41104534d16973f72ef26c40/Lib/ufo2ft/outlineCompiler.py#L676-L687

note that whenever you see `getAttrWithFallback` you need to look at the ufo2ft.fontInfoData module and see what is the fallback value (either simple hardcoded default or generated from other attributes) that ufo2ft has.
In the case of `openTypeOS2Selection` it's an empty list `[]`, whereas in the case of `styleMapStyleName`  it has a dynamic fallback that checks for `openTypeNamePreferredSubfamilyName` (which in turn defaults to fontinfo's `styleName` if missing) and checks if the latter is in the RIBBI (Regular, Italic, Bold, Bold Italic) set of names otherwise defaults to "regular".

```python
        # style mapping
        selection = list(getAttrWithFallback(font.info, "openTypeOS2Selection"))
        styleMapStyleName = getAttrWithFallback(font.info, "styleMapStyleName")
        if styleMapStyleName == "regular":
            selection.append(6)
        elif styleMapStyleName == "bold":
            selection.append(5)
        elif styleMapStyleName == "italic":
            selection.append(0)
        elif styleMapStyleName == "bold italic":
            selection += [0, 5]
        os2.fsSelection = intListToNum(selection, 0, 16)
```

anthrotype: ah ok I see, you are OR-ing the SelectionFlags from `open_type_os2_selection` (when missing or empty it would be 0 thanks to unwrap_or_default()) with the style flags from style_map_style_name, which is pretty much what ufo2ft is doing. I think you're only missing the fallback treatment for styleMapStyleName

glyphs-reader/src/font.rs
```diff
@@ -1471,6 +1473,9 @@ impl TryFrom<RawFont> for Font {
         let glyph_order = parse_glyph_order(&from);
         let glyph_to_codepoints = parse_codepoints(&mut from, radix);
 
+        let use_typo_metrics =
+            custom_param_int(&from.other_stuff, "Use Typo Metrics").map(|v| v == 1);
```
anthrotype, line 1477: you may also want to support "Has WWS Names" custom parameter (flag 8 of fsSelection) since you're touching this.
It doesn't seem like Glyphs.app allow one to set arbitrary fsSelection flags like UFO fontinfo.plist does, the style-linking ones are somewhat implied by the master name and then there's "Use Typo Metrics" and "Has WWS Names", the rest of the fsSelection flags can't be turned on apparently (or at least I haven't found how.. I suspect nobody cares about those).

fontbe/src/os2.rs
```diff
@@ -56,43 +56,35 @@ fn x_avg_char_width(context: &Context) -> Result<i16, Error> {
...
+    os2.y_strikeout_size = metrics.y_strikeout_size.ot_round();
+    os2.y_strikeout_position = metrics.y_strikeout_position.ot_round();
+
+    os2.s_typo_ascender = metrics.os2_typo_ascender.ot_round();
+    os2.s_typo_descender = metrics.os2_typo_descender.ot_round();
+    os2.s_typo_line_gap = metrics.os2_typo_line_gap.ot_round();
+    os2.us_win_ascent = metrics.os2_win_ascent.ot_round();
+    os2.us_win_descent = metrics.os2_win_descent.ot_round();
+
+    // Avoid "field must be present for version 2"
```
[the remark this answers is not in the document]
rsheeter: The default for an Option is None. I agree this is the wrong place to be setting them, will move.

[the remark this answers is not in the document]
rsheeter: fixed and added comment to clarify

ufo2fontir/src/source.rs
rsheeter: Oops, fixed to default to regular and added a comment to make the logic clear, that | is a bit easy to miss

glyphs-reader/src/font.rs
rsheeter: Hm, I think you are right and the presence at the top of static metadata does send a message, the wrong one.

rsheeter: Moved to StaticMetadata::misc.

glyphs-reader/src/font.rs
rsheeter: Added.

glyphs2fontir/src/source.rs
rsheeter: This part I would prefer to kick down the road slightly, pushed to https://github.com/googlefonts/fontmake-rs/issues/300

anthrotype: fair enough

fontir/src/serde.rs
```diff
@@ -81,11 +81,11 @@ impl From<StaticMetadata> for StaticMetadataSerdeRepr {
         let glyph_locations = from.variation_model.locations().cloned().collect();
         StaticMetadataSerdeRepr {
             units_per_em: from.units_per_em,
-            selection_flags: from.selection_flags.bits(),
             axes: from.axes,
             named_instances: from.named_instances,
             glyph_locations,
             names: from.names,
+            platform_metadata: from.misc.into(),
```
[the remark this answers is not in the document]
anthrotype: i also wonder what platform_metadata means

ufo2fontir/src/source.rs
```diff
@@ -639,29 +640,31 @@ impl Work<Context, WorkError> for StaticMetadataWork {
                 })
             })
             .unwrap_or_default()
-            | font_info_at_default
+            // Also set any bits implied by the style map style name
+            | match font_info_at_default
                 .style_map_style_name
```
anthrotype, line 645: you're now defaulting to StyleMapStyle::Regular, but you still aren't checking for openTypePreferredSubfamilyName as a fallback when missing styleMapStyleName, see my earlier comment:

> in the case of styleMapStyleName it has a dynamic fallback that checks for openTypeNamePreferredSubfamilyName (which in turn defaults to fontinfo's styleName if missing) and checks if the latter is in the RIBBI (Regular, Italic, Bold, Bold Italic) set of names otherwise defaults to "regular".

you may argue "garbage in, garbage out" but if we want to match 100% ufo2ft, then we need to implement similar fallback logic

anthrotype (approved):
LGTM but see comments

fontir/src/serde.rs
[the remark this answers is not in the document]
rsheeter: It's a terrible name I thought I had fully scrubbed out, will update to match misc.

ufo2fontir/src/source.rs
rsheeter: I agree we should match but I want to land this chunk first so I will add ^ to #300

PR #299, opened 2023-05-25
Set OS/2 unicode ranges

rsheeter:
OS/2 flags are more fiddly than I gave them credit for


fontbe/src/os2.rs
```diff
@@ -10,6 +12,187 @@ use crate::{
     orchestration::{BeWork, Context},
 };
 
+/// Used to build a bitfield.
+///
+/// Intended use: use the outer range to determine if a given bit might apply. If it might,
+/// check the inner ranges.
```
anthrotype, line 18: the comment is no longer in sync with the implementation below I think, I don't see outer/inner ranges anymore, just a single array of 3-tuples

anthrotype (approved):
LGTM!

fontbe/src/os2.rs
```diff
@@ -10,6 +12,187 @@ use crate::{
...
+///
+/// Generated by resources/scripts/os2_unicode_ranges.py.
+///
+/// <https://learn.microsoft.com/en-us/typography/opentype/spec/os2#ur>
+/// <https://github.com/fonttools/fonttools/blob/47813b217c1f8bc343c094776020b4f32fc024b0/Lib/fontTools/ttLib/tables/O_S_2f_2.py#L351-L557>
+const UNICODE_RANGES: &[(usize, usize, usize)] = &[
```
cmyr, line 24: is `usize` the best type to represent a codepoint? `char` might be semantically more accurate, and `u32` would be a more obvious choice if we want to use an integer type.

Also I wonder if it makes sense to represent the ranges as an actual range? maybe not, but just a thought.

fontbe/src/os2.rs
```diff
@@ -80,6 +263,51 @@ fn apply_metrics(os2: &mut Os2, metrics: &GlobalMetricsInstance) {
...
+    let maybe_idx = UNICODE_RANGES
+        .binary_search_by(|(from, to, _)| match codepoint {
+            _ if codepoint < *from => Ordering::Greater,
+            _ if codepoint > *to => Ordering::Less,
+            _ => Ordering::Equal,
+        })
+        .ok();
+    if let Some(idx) = maybe_idx {
+        add_to.insert(UNICODE_RANGES[idx].2);
+    }
```
cmyr, line 277: code golf:

```suggestion
    // Is there any range this codepoint _might_ fit?
    if let Ok(idx) = UNICODE_RANGES
        .binary_search_by(|(from, to, _)| match (from.cmp(&codepoint), to.cmp(&codepoint)) {
        (Ordering::Greater, _) => Ordering::Greater,
        (_, Ordering::Less) => Ordering::Less,
        _ => Ordering::Equal,
        }) {
            add_to.insert(UNICODE_RANGES[idx].2);
        }      
    }
```

cmyr (approved):
Some notes inline that you are very welcome to ignore, overall looks reasonable!

fontbe/src/os2.rs
rsheeter: Fixed

fontbe/src/os2.rs
rsheeter: I tried that but I find it harder to read than doing it in separate steps and afaik there is no downside?

rsheeter: ...the comment is out of date!

rsheeter: I tried it again adn I still think it's harder to read so I'm inclined to leave this as-is.

PR #301, opened 2023-05-26
Set codepage range bits in OS/2

rsheeter:
After this and the PRs it is built on OS/2 for Oswald differs on {usLastCharIndex, usBreakChar, usMaxContext}


anthrotype (approved):
LGTM, thanks!

PR #305, opened 2023-05-26
Match outlineCompiler.py default value for break char

rsheeter:
Baby steps.


dfrg (approved):

anthrotype:
and that's it really, nobody actually cares about overriding that usBreakChar default value. another piece of useless legacy stuff

PR #306, opened 2023-05-26
Improve handling of fea-rs output

rsheeter:
Split the single feature font from fea-rs into pieces and include those pieces in the final font instead of running fea-rs and then just abandoning the results if it succeeded.

This reveals that for Oswald GSUB doesn't match and we aren't ending up with a GPOS at all; filed #308.

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -12,14 +12,15 @@ use fea_rs::{
     Compiler, GlyphMap, GlyphName as FeaRsGlyphName,
 };
 use fontir::{ir::Features, orchestration::Flags};
-use log::{debug, error, trace, warn};
-use write_fonts::FontBuilder;
+use log::{debug, error, warn};
+use read_fonts::{FontRef, TableProvider};
+use write_fonts::{from_obj::ToOwnedTable, FontBuilder};
 
 use fontdrasil::orchestration::Work;
 
 use crate::{
     error::Error,
-    orchestration::{BeWork, Context},
+    orchestration::{BeWork, Context, WorkId},
 };
 
 pub struct FeatureWork {}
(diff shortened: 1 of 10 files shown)


fontbe/src/features.rs
```diff
@@ -111,29 +112,42 @@ fn write_debug_fea(context: &Context, is_error: bool, why: &str, fea_content: &s
...
+                "Built features, gpos? {} gsub? {}",
+                font.gpos().is_ok(),
+                font.gsub().is_ok()
+            );
+            if let Ok(gpos) = font.gpos() {
+                context.set_gpos(gpos.to_owned_table());
             }
+            if let Ok(gsub) = font.gsub() {
+                context.set_gsub(gsub.to_owned_table());
+            }
```
cmyr, line 143: this is a very convoluted way of trying to do what we want, I think? Wouldn't it be nicer if we just had some way of retrieving the bytes of a specific table from the `FontBuilder`? (we'd need to add a method there, but nothing crazy)

cmyr (approved):
one note inline.

fontbe/src/features.rs
rsheeter: This way it's typed which is nice perhaps? - I agree it feels a bit clunky but ... it does express the intent in fairly straight forward terms

cmyr: Maybe this is a sign that fea-rs should be returning something other than a `FontBuilder`?

It would also be nice if `fea-rs` did not return a type from `write-fonts` directly, since this is a reason why every breaking change in write-fonts is currently a breaking change in fea-rs...

although maybe there's no good work around from this? it certainly doesn't change if I'm returning a set of write-fonts tables instead of a write-fonts TableBuilder. But it still might be a better API.

rsheeter: > Maybe this is a sign that fea-rs should be returning something other than a FontBuilder?
>
> It would also be nice if fea-rs did not return a type from write-fonts directly, since this is a reason why every breaking change in write-fonts is currently a breaking change in fea-rs...

Returning any type from write-fonts has the same effect. Raw bytes, or something close to raw bytes (map of tag to raw bytes?) would work. I don't want to have multiple write-fonts in my dependency tree but perhaps it would bumping through versions simpler. Or we could just live with it; I would expect the rate of write-fonts breaks to decline over time as it stabilizes.

PR #314, opened 2023-05-31
Make Oswald 'name' match fontmake

rsheeter:
ttx_diff in default mode:

```shell
  Only fontmake produced 'GDEF', 'GPOS', 'HVAR', 'STAT'
  Identical 'GlyphOrder'
  Identical 'OS_2'
  Identical 'avar'
  Identical 'cmap'
  Identical 'fvar'
  DIFF 'head', build/default/fontc.head.ttx build/default/fontmake.head.ttx
  DIFF 'hhea', build/default/fontc.hhea.ttx build/default/fontmake.hhea.ttx
  Identical 'hmtx'
  Identical 'loca'
  Identical 'name'
  DIFF 'post', build/default/fontc.post.ttx build/default/fontmake.post.ttx
```

Note that I modified ttx_diff.py to eliminate unwanted diffs due to https://github.com/googlefonts/fontmake/issues/1003

### DIFF
--- a/fontbe/src/fvar.rs
+++ b/fontbe/src/fvar.rs
@@ -32,6 +32,9 @@ fn generate_fvar(static_metadata: &StaticMetadata) -> Option<Fvar> {
     let reverse_names: HashMap<_, _> = static_metadata
         .names
         .iter()
+        // To match fontmake we should use the font-specific name range and not reuse
+        // a well-known name, even if the name matches.
+        .filter(|(key, _)| key.name_id.to_u16() > 255)
         .map(|(key, name)| (name, key.name_id))
         .collect();
 
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -21,7 +21,8 @@ use std::{
 };
 use write_fonts::tables::os2::SelectionFlags;
 
-const DEFAULT_VENDOR_ID: Tag = Tag::new(b"NONE");
+pub const DEFAULT_VENDOR_ID: &str = "NONE";
+const DEFAULT_VENDOR_ID_TAG: Tag = Tag::new(b"NONE");
 
 /// Global font info that cannot vary across the design space.
 ///
@@ -95,7 +96,8 @@ impl StaticMetadata {
         let mut name_id_gen = 255;
         let mut key_to_name = names;
         let mut visited = HashSet::new();
-        axes.iter()
+        variable_axes
+            .iter()
             .map(|axis| &axis.name)
             .chain(named_instances.iter().map(|ni| &ni.name))
             .for_each(|name| {
@@ -138,7 +140,7 @@ impl StaticMetadata {
             variable_axes_default,
             misc: MiscMetadata {
                 selection_flags: Default::default(),
-                vendor_id: DEFAULT_VENDOR_ID,
+                vendor_id: DEFAULT_VENDOR_ID_TAG,
             },
         })
     }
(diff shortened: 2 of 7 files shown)


cmyr (approved):

fontir/src/ir.rs
```diff
@@ -383,13 +384,33 @@ pub struct NameBuilder {
...
 
+impl Default for NameBuilder {
+    fn default() -> Self {
+        let mut builder = Self {
+            names: Default::default(),
+            name_to_key: Default::default(),
+            version_major: Default::default(),
+            version_minor: Default::default(),
+        };
+        // Based on diffing fontmake Oswald build
```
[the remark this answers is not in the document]
anthrotype: this is the logic we need to match, including the fallbacks when some parameters are omitted, and not whatever Oswald ends up with which depends on the particular configuration of settings it contains:

https://github.com/googlefonts/ufo2ft/blob/0d2688cd847d003b41104534d16973f72ef26c40/Lib/ufo2ft/outlineCompiler.py#L367-L445

fontbe/src/fvar.rs
```diff
@@ -32,6 +32,9 @@ fn generate_fvar(static_metadata: &StaticMetadata) -> Option<Fvar> {
     let reverse_names: HashMap<_, _> = static_metadata
         .names
         .iter()
+        // To match fontmake we should use the font-specific name range and not reuse
```
anthrotype, line 35: it's not just to match fontmake, but to match the spec (search for "255") https://learn.microsoft.com/en-us/typography/opentype/spec/fvar

fontir/src/ir.rs
```diff
@@ -480,13 +501,24 @@ impl NameBuilder {
...
-            let vendor = "NONE";
             let postscript_name = self.get(NameId::POSTSCRIPT_NAME).unwrap();
             self.add(
                 NameId::UNIQUE_ID,
-                format!("{version};{vendor};{postscript_name}"),
+                format!("{version};{vendor_id};{postscript_name}"),
             );
         }
+
+        // based on diffing with fontmake
```
anthrotype, line 511: I'd avoid comments like these, it suggests we're treating fontmake as a black box which shouldn't be the case. We should either point directly to the source code that does the equivalent of what is below, or add a link to the spec, or if it is not self evident try to come up with some rationale. And if it still doesn't make any sense, raise an issue.

resources/scripts/ttx_diff.py
```diff
@@ -157,11 +157,49 @@ def name_id_to_name(ttx, xpath, attr):
...
+    for drop in drops:
+        drop.getparent().remove(drop)
+
+
 def reduce_diff_noise(fontc, fontmake):
     for ttx in (fontc, fontmake):
         # different name ids with the same value is fine
         name_id_to_name(ttx, "//NamedInstance", "subfamilyNameID")
 
+        # different order of records is fine
```
anthrotype, line 197: well, different order may not be not fine, fonttools orders the table according to the spec (see `NameRecord.__lt__` method), we should make sure we are doing the same before massaging the diff like that.

rsheeter: Good catch, will fix.

rsheeter: Fixed in https://github.com/googlefonts/fontc/pull/321

fontir/src/ir.rs
rsheeter: Fair point, fixed in #323

PR #315, opened 2023-06-01
Match glyphsLib/ufo2ft defaults for underline thickness/position

rsheeter:
Oswald post matches after:

```
  Only fontmake produced 'GDEF', 'GPOS', 'HVAR', 'STAT'
  Identical 'GlyphOrder'
  Identical 'OS_2'
  Identical 'avar'
  Identical 'cmap'
  Identical 'fvar'
  DIFF 'head', build/default/fontc.head.ttx build/default/fontmake.head.ttx
  DIFF 'hhea', build/default/fontc.hhea.ttx build/default/fontmake.hhea.ttx
  Identical 'hmtx'
  Identical 'loca'
  Identical 'name'
  Identical 'post'
```

### DIFF
--- a/fontbe/src/post.rs
+++ b/fontbe/src/post.rs
@@ -1,5 +1,6 @@
 //! Generates a [post](https://learn.microsoft.com/en-us/typography/opentype/spec/post) table.
 
+use font_types::FWord;
 use fontdrasil::orchestration::Work;
 use write_fonts::tables::post::Post;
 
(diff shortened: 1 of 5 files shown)


dfrg (approved):
LGTM! Does glyphs not allow setting these values or are we just assuming defaults for now and those just happen to match Oswald?

rsheeter:
> Does glyphs not allow setting these values

It does but I don't have a sample font yet. I'll file an issue so I don't forget.

fontbe/src/post.rs
```diff
@@ -21,7 +22,9 @@ impl Work<Context, Error> for PostWork {
         // TODO optionally drop glyph names with format 3.0.
         // TODO a more serious post
         let static_metadata = context.ir.get_final_static_metadata();
-        let post = Post::new_v2(static_metadata.glyph_order.iter().map(|g| g.as_str()));
+        let mut post = Post::new_v2(static_metadata.glyph_order.iter().map(|g| g.as_str()));
+        post.underline_position = FWord::new(static_metadata.misc.underline_position.0 as i16);
+        post.underline_thickness = FWord::new(static_metadata.misc.underline_thickness.0 as i16);
```
anthrotype, line 27: ufo2ft does otRound, does not truncate. In general we use that when converting stuff that in ufo is allowed to be a plist "real" to an integer.

Issue #149, opened 2023-03-15
should add .notdef when missing from the source


anthrotype:
.notdef (GID=0) is required glyph in a font (a glyph at index 0 is special for cmap, for example).
Sometimes sources do not contain one and expect the compiler to synthesize one for them.
fontmake (via ufo2ft) adds one when the input doesn't contain it, using font dimensions like upem and vertical metrics to draw a nice tofu box. I'd expect fontmake-rs to do something similar.

See https://github.com/googlefonts/ufo2ft/blob/5b22f201b4650ab9ef1169357702a732e3127a5e/Lib/ufo2ft/outlineCompiler.py#L241-L270

anthrotype:
even when .notdef glyph may be present, we need to make sure it is the first in the glyph order.
some font sources do contain an explicit public.glyphOrder list but forget that .notdef should be the placed as the first in the list. Ufo2ft always ensures that gid=0 is a .notdef glyph whether or not the sources contain a glyph named like that, and whether they do place it or not as the first

e.g. https://github.com/googlefonts/googlesans/issues/590

PR #320, opened 2023-06-02
[WIP] Make maxp match for Oswald

rsheeter:
Many fewer zeroes in the fontc (left) copy, but composite points/contours still don't align at time of writing:
image


fontbe/src/metrics_and_limits.rs
```diff
@@ -31,57 +35,178 @@ struct GlyphLimits {
...
+                components.extend(
+                    glyph_info
+                        .components
+                        .as_ref()
+                        .unwrap()
+                        .iter()
+                        .map(|gid| self.glyph_info.get(gid).unwrap()),
+                );
+
+                // Update max points, contours while we're here
```
anthrotype, line 146: I find it strange that in a method called `compute_component_limits` you update max_points and max_contours which are about non-composite glyphs... Have you not done that already in GlyphLimits::update for simple glyphs?

anthrotype: sorry stupid me, you _are_ updating max_composite_points and max_composite_contours 😅

fontbe/src/metrics_and_limits.rs
```diff
@@ -31,57 +35,178 @@ struct GlyphLimits {
...
+        let mut depths = HashMap::<GlyphId, u16>::new();
+        let mut pending = self
+            .glyph_info
+            .iter()
+            .filter_map(|(gid, gi)| if gi.is_component() { Some(*gid) } else { None })
+            .collect::<Vec<_>>();
+        let mut components = Vec::with_capacity(8);
+        let mut max_points = 0;
+        let mut max_contours = 0;
+        while !pending.is_empty() {
```
anthrotype, line 132: a recursive algorithm like fonttools does would be much simpler to implement, is that a no-no in Rust? why?

fontbe/src/metrics_and_limits.rs
```diff
@@ -31,57 +35,178 @@ struct GlyphLimits {
...
+
+                // Update max points, contours while we're here
+                (max_points, max_contours) = components
+                    .iter()
+                    .filter_map(|c| {
+                        if !c.is_component() {
+                            Some((c.num_points, c.num_contours))
+                        } else {
+                            None
+                        }
```
anthrotype, line 154: the problem is here I believe, with the filter_map you are skipping components that are in turn components and only folding the num_points/num_contours for the components that are simple glyphs.
Try to make a composite glyph that contains two components with different depths, i.e. one component points to a simple glyph, whereas the other one points to a component that in turn points to a simple glyph.
Your code would not consider the points/contours of the second nested component when computing the total number of points/contours for the composite as a whole

anthrotype: I added a glyph called "c3" inside your NestedComponent.glyphs test file that has three components: "period" (simple glyph, depth=1), "c1" (in turn a composite referencing "period", depth=2) and "c2" (composite referencing "c1", depth=3). The expected num_points for "c3" as a whole should be 12, and the expected num_contours should be 3. It's max_depth should be 3. The maxp_with_deep_components test fails currently because I have adjusted the expectations.

resources/testdata/glyphs3/NestedComponent.glyphs
```diff
@@ -39,6 +45,15 @@ name = Regular;
 );
 glyphs = (
 {
+glyphname = .notdef;
```
anthrotype, line 48: I also added a .notdef glyph to aid comparisons with fontmake, since ufo2ft always adds a .notdef as the first glyph if it is not already there (we should do the same! #149)

fontbe/src/metrics_and_limits.rs
rsheeter: (facepalm) duh! Good catch, ty.

fontbe/src/metrics_and_limits.rs
rsheeter: Recursion is fine in Rust. I might argue the simplicity of recursive code to be subjective ;)

resources/testdata/glyphs3/NestedComponent.glyphs
rsheeter: Good call, ty

fontbe/src/metrics_and_limits.rs
rsheeter: I tidied up the loop, I believe it is now much clearer. Please reopen if you disagree.

fontbe/src/metrics_and_limits.rs
rsheeter: Fixed but result still doesn't match. It is much closer, now the only diff is maxCompositePoints 103 vs 104, presumably due to more vigorous dropping of implicit oncurves in fontc. I'd like to land this and then wait for your fontmake update to get `glyf` to be identical to see if maxp then matches.

fontbe/src/metrics_and_limits.rs
```diff
@@ -27,23 +27,21 @@ pub fn create_metric_and_limit_work() -> Box<BeWork> {
 }
 
 #[derive(Debug, Default)]
-struct GlyphLimits {
+struct Limits {
```
anthrotype, line 30: I find confusing that there's an unqualified "Limits" struct and also one called "GlyphLimits" in the same module... why are these limits just "Limits" whereas the other are "GlyphLimits"? aren't they all about glyphs?

anthrotype (approved):

fontbe/src/metrics_and_limits.rs
rsheeter: This one is global and that one is for one glyph. I will rename this one FontLimits in hopes of clarifying the distinction.

PR #321, opened 2023-06-02
Use the real sort order for 'name'

rsheeter:
As noted in https://github.com/googlefonts/fontc/pull/314#pullrequestreview-1455005475, apparently I invented a sort order for 'name'. Turns out the spec says something else; do that instead.

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -163,22 +163,6 @@ def find_table(ttx, tag):
     return el[0]
 
 
-def sort_name(ttx):
-    name = find_table(ttx, "name")
-    records = sorted(
-        list(name),
-        key=lambda el: (
-            int(el.attrib["nameID"]),
-            el.attrib["platformID"],
-            el.attrib["platEncID"],
-            el.attrib["langID"],
-        ),
-    )
-    name.clear()
-    for record in records:
-        name.append(record)
-
-
 def drop_weird_names(ttx):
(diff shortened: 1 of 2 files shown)


dfrg (approved):
To be fair, your invented sort order makes significantly more sense to me 🤷

cmyr:
Hmm, I thought that write-fonts handled sorting when it generated the table? Is this skipping that somehow?

rsheeter:
@cmyr the output changes when I change my sorting so evidently write-fonts is not handling it. To your point, it probably should so perhaps we can fix that then remove sorting here?

cmyr:
Huh I'm very confused, `Name` stores the records using a `BTreeSet` which should have a consistent ordering by definition? Something funny is going on, will take a look

resources/scripts/ttx_diff.py
```diff
@@ -163,22 +163,6 @@ def find_table(ttx, tag):
...
-            el.attrib["platformID"],
-            el.attrib["platEncID"],
-            el.attrib["langID"],
-        ),
-    )
-    name.clear()
-    for record in records:
-        name.append(record)
-
-
```
cmyr, line 181: is it possible that this was the only change required for this patch to work, and that the changes above are unnecessary?

rsheeter: It is! https://github.com/googlefonts/fontc/pull/326.

PR #322, opened 2023-06-02
Appease clippy

rsheeter:
JMM

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -810,7 +810,7 @@ mod tests {
             let mut loc = UserLocation::new();
             loc.set_pos("Optical Size", UserCoord::new(*opsz));
             loc.set_pos("Weight", UserCoord::new(*wght));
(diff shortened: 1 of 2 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -810,7 +810,7 @@ mod tests {
             let mut loc = UserLocation::new();
             loc.set_pos("Optical Size", UserCoord::new(*opsz));
             loc.set_pos("Weight", UserCoord::new(*wght));
-            let loc: _ = loc;
+            let loc = loc;
```
dfrg, line 813: Is this actually necessary?

dfrg (approved):
LGTM with a comment expressing my confusion about rebinding a variable to the same name.

glyphs2fontir/src/source.rs
rsheeter: Not really; I just like that it clearly ends mutable time.

Issue #300, opened 2023-05-25
Fully match glyphsLib build_stylemap_names


rsheeter:
https://github.com/googlefonts/fontmake-rs/pull/298#discussion_r1205332559 notes that we do not completely match:

> glyphsLib contains additional logic, see build_stylemap_names (that is called when exporting glyphs->ufo by fontmake because minimize_glyphs_diffs is False when compiling):
> 
> https://github.com/googlefonts/glyphsLib/blob/42bc1db912fd4b66f130fb3bdc63a0c1e774eb38/Lib/glyphsLib/builder/names.py#L27-L43
> 
> the final ufo.info.styleMapStyleName in the end is what matters to decide the style flags for ufo2ft, and the default UFO master that glyphsLib produces is going to be the default master TTF which varLib will duplicate to use as basis for the new variable font.

rsheeter:
Note also we need to check openTypePreferredSubfamilyName as a fallback when missing styleMapStyleName, as per @anthrotype on https://github.com/googlefonts/fontc/pull/298#discussion_r1206737846:

>you're now defaulting to StyleMapStyle::Regular, but you still aren't checking for openTypePreferredSubfamilyName as a fallback when missing styleMapStyleName, see my earlier comment:
>
>> in the case of styleMapStyleName it has a dynamic fallback that checks for openTypeNamePreferredSubfamilyName (which in turn defaults to fontinfo's styleName if missing) and checks if the latter is in the RIBBI (Regular, Italic, Bold, Bold Italic) set of names otherwise defaults to "regular".
>
> you may argue "garbage in, garbage out" but if we want to match 100% ufo2ft, then we need to implement similar fallback logic

PR #324, opened 2023-06-03
Set more of head correctly

### LINKED ISSUE #233: head.created should respect GSFont.date / UFO.info.openTypeHeadCreated
as of #232, we set head.created to the current timestamp (or SOURCE_DATE_EPOCH if any).
To match fontmake-py, we need to check if the source fontains either "date" font attribute for .glyphs, or "openTypeHeadCreated" in fontinfo.plist for UFO input, and use that for head.created (head.modified is fine like it currently is, always points to current time).

### LINKED ISSUE #249: Populate missing 'head' fields
We are not setting:

* fontRevision
* checkSumAdjustment
* flags
* xMin, yMin, xMax, yMax
* lowestRecPPEM
* indexToLocFormat (ref #155)

rsheeter:
Specifically:

* lowest rec ppm
* flags
* font revision
* created/modified

Fixes #233. Fixes #249.

Requires https://github.com/googlefonts/fontations/pull/469 which is in write-fonts 0.5.1.

The only Oswald head diff I'm left with is checksum. https://github.com/googlefonts/fontations/issues/466 requests fontations set checksum adjustment. It won't match until all tables do ofc.

### DIFF
--- a/fontbe/src/head.rs
+++ b/fontbe/src/head.rs
@@ -3,7 +3,7 @@
 use std::env;
 
 use chrono::{DateTime, TimeZone, Utc};
-use font_types::LongDateTime;
+use font_types::{Fixed, LongDateTime};
 use fontdrasil::orchestration::Work;
 use log::warn;
 use write_fonts::tables::head::Head;
@@ -23,7 +23,7 @@ pub fn create_head_work() -> Box<BeWork> {
 // Equivalent to Utc.with_ymd_and_hms(1904, 1, 1, 0, 0, 0).unwrap().timestamp()
 const MACINTOSH_EPOCH: i64 = -2082844800;
 
-fn timestamp_since_mac_epoch(datetime: DateTime<Utc>) -> i64 {
+fn seconds_since_mac_epoch(datetime: DateTime<Utc>) -> i64 {
     let mac_epoch = Utc.timestamp_opt(MACINTOSH_EPOCH, 0).unwrap();
     datetime.signed_duration_since(mac_epoch).num_seconds()
 }
@@ -45,15 +45,14 @@ fn current_timestamp() -> i64 {
             );
         }
     }
-    timestamp_since_mac_epoch(src_date.unwrap_or_else(Utc::now))
+    seconds_since_mac_epoch(src_date.unwrap_or_else(Utc::now))
 }
 
-fn build_head(units_per_em: u16, loca_format: LocaFormat) -> Head {
-    let now = LongDateTime::new(current_timestamp());
+fn init_head(units_per_em: u16, loca_format: LocaFormat, flags: u16, lowest_rec_ppem: u16) -> Head {
     Head {
         units_per_em,
-        created: now, // TODO: support GSFont.date / UFO.info.openTypeHeadCreated
-        modified: now,
+        lowest_rec_ppem,
+        flags,
         index_to_loc_format: match loca_format {
             LocaFormat::Short => 0,
             LocaFormat::Long => 1,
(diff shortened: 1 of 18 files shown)


fontir/src/ir.rs
```diff
@@ -81,6 +82,18 @@ pub struct MiscMetadata {
 
     pub underline_thickness: OrderedFloat<f32>,
     pub underline_position: OrderedFloat<f32>,
+
+    /// UFO appears to allow negative major versions.
```
anthrotype, line 86: hm I believe that's just a typo.. It doesn't make any sense to have version -1.0... 🤯

fontir/src/serde.rs
```diff
@@ -110,6 +117,12 @@ impl From<MiscSerdeRepr> for MiscMetadata {
             vendor_id: from.vendor_id,
             underline_thickness: from.underline_thickness.into(),
             underline_position: from.underline_position.into(),
+            version_major: from.version_major,
+            version_minor: from.version_minor,
+            head_flags: from.head_flags,
+            lowest_rec_ppm: from.lowest_rec_ppm,
+            created: from.created,
+            modified: from.created,
```
anthrotype, line 125: hm `modified: from.created` feels suspicious..
Do we actually need to store `modified` in the StaticMetadata IR? None of the source formats allow to override this currently (only head.created), the only way for user to _not_ set it to the current time is to export SOURCE_DATE_EPOCH right now.

anthrotype (commented):
the commit message "Make head a little closer for Oswald" is a bit too vague, would be better if you could summarise the changes in a couple of sentences

fontbe/src/head.rs
```diff
@@ -62,12 +61,56 @@ fn build_head(units_per_em: u16, loca_format: LocaFormat) -> Head {
...
+    head.modified = LongDateTime::new(modified.map(seconds_since_mac_epoch).unwrap_or(now));
+}
+
 impl Work<Context, Error> for HeadWork {
     /// Generate [head](https://learn.microsoft.com/en-us/typography/opentype/spec/head)
     fn exec(&self, context: &Context) -> Result<(), Error> {
         let static_metadata = context.ir.get_final_static_metadata();
         let loca_format = context.get_loca_format();
-        let head = build_head(static_metadata.units_per_em, *loca_format);
+        let mut head = init_head(
```
anthrotype, line 98: since you're touching head, would be nice if you could also set the head.macStyle bits (for bold/italic, work similarly to OS/2 fsSelection),

https://github.com/googlefonts/ufo2ft/blob/0d2688cd847d003b41104534d16973f72ef26c40/Lib/ufo2ft/outlineCompiler.py#L346-L354

or maybe you can punt until #300

fontir/src/ir.rs
rsheeter: norad concurs so while I agree -1.0 is a very very weird version it does appear possible.

fontir/src/serde.rs
rsheeter: Oh good call, will remove modified.

Reproducible builds would require modified *not* get updated every cycle but that can wait.

fontir/src/ir.rs
anthrotype: feel free to bump https://github.com/unified-font-object/ufo-spec/issues/215

fontir/src/serde.rs
anthrotype: > Reproducible builds would require modified not get updated every cycle but that can wait.

I thought we did support SOURCE_DATE_EPOCH for repro builds, don't we?

rsheeter: Fixed.

anthrotype (approved):
LGTM!

PR #329, opened 2023-06-05
Write a rudimentary STAT

rsheeter:
ttx_diff claims we match Oswald. Not cheating at all to massage the results right?

```shell
  Only fontmake produced 'GDEF', 'GPOS', 'HVAR'
  Identical 'GlyphOrder'
  Identical 'OS_2'
  Identical 'STAT'
  Identical 'avar'
  Identical 'cmap'
  Identical 'fvar'
  Identical 'head'
  Identical 'hhea'
  Identical 'hmtx'
  Identical 'loca'
  Identical 'name'
  Identical 'post'

```


rsheeter:
> when we implement DSv5 support

I entirely agree but I would not that I was unable to identify an actual font using DSv5

anthrotype:
> I was unable to identify an actual font using DSv5

most fonts on GF are built using gftools which has its own module that uses fontTools buildStatTable API to construct a full STAT using the metadata from axisregistry. We may have to look at non-GF projects for examples of DSv5 usage for STAT building.

Issue #308, opened 2023-05-26
No GPOS for Oswald


rsheeter:
As noted on #306, we aren't producing a GPOS for Oswald at all.

anthrotype:
that's because there are no handwritten GPOS features in Oswald, only auto-generated ones (kern from higher level kerning dictionaries, possibly also mark + mkmk from glyph anchors), the stuff that ufo2ft featureWriters take care of.
We need to start thinking how to go about implementing these in fontc.
To be faithful to our 'variable-first' approach, instead of compiling one GPOS for each master and merge those at the end (as it's done currently in fontmake/ufo2ft/fonttools) we should generate features in variable syntax and build those as such (see https://github.com/googlefonts/ufo2ft/pull/635).
fea-rs will  to support variable FEA syntax (https://github.com/cmyr/fea-rs/issues/131) for this.

rsheeter:
Taking notes  in https://docs.google.com/document/d/1xsFkaa_G-wdhGW4I5zL4HoQw71LAt698kRXCwoq7hdE/edit

rsheeter:
@anthrotype notes that the autogenerated parts of fea generation are to be found in [ufo2ft.featureCompiler](https://github.com/googlefonts/ufo2ft/blob/main/Lib/ufo2ft/featureCompiler.py), which runs a series of so-called featureWriters (defined in [ufo2ft.featureWriters](https://github.com/googlefonts/ufo2ft/tree/main/Lib/ufo2ft/featureWriters)) on each master font; these generate FEA for kern (kernFeatureWriter), mark, mkmk (markFeatureWriter), which gets appended or inserted in the hand-written FEA and then sent off for compilation to feaLib

rsheeter:
See also https://github.com/googlefonts/ufo2ft/pull/635 which gives a small example of variable kerning and some potentially informative related python to handle it.

anthrotype:
As @simoncozens and @madig noted in https://github.com/googlefonts/ufo2ft/pull/635, in order to make sure the variable-first kernFeatureWriter produces the same result (when compiled to GPOS) as the multiple-masters-then-varlib-merged one, the kerning dictionaries for each masters need to be "unioned" or "aligned"  to properly take care of differing kerning "exceptions" in the various masters. In other words, kerns missing from some masters need to be "backfilled", not just with zeros but using the more "general" value available for the given glyph pair in each master (falling back to zeros if none), see:

[...]

PR #330, opened 2023-06-05
Add example .glyphs kern and ability to read it

rsheeter:
Tiptoing toward #308

The existing test to confirm reading glyphs2 wghtvar == read glyphs3 wghtvar verifies the field rename works as intended.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -62,6 +62,9 @@ pub struct Font {
     pub version_major: i32,
     pub version_minor: u32,
     pub date: Option<String>,
+
+    // master id => { (name or class, name or class) => adjustment }
+    pub kerning_ltr: BTreeMap<String, BTreeMap<(String, String), i32>>,
 }
 
 #[derive(Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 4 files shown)


glyphs-reader/src/font.rs
```diff
@@ -968,13 +971,21 @@ impl RawFont {
         Ok(())
     }
 
+    fn v2_to_v3_kerning(&mut self) -> Result<(), Error> {
+        if let Some(kerning) = self.other_stuff.remove("kerning") {
+            self.other_stuff.insert("kerningLTR".to_string(), kerning);
```
anthrotype, line 976: amusingly in glyphsLib we ended up actually doing the opposite i.e. down-convert the kerningLTR/kerningRTL split dictionaries into a single kerning dict, like in was in glyphs v2 and in UFO.. the history of this is complicated as everything in font-land..
You may start from https://github.com/googlefonts/glyphsLib/pull/865

rsheeter: I explicitly didn't handle the kerningRTL set yet. My intent wasthat this is valid for both v2 and v3 LTR kerning. I _think_ you are saying it is? Perhaps I should file a bug for kerningRTL.

anthrotype (approved):
LGTM

glyphs-reader/src/font.rs
rsheeter: #333

PR #331, opened 2023-06-06
Add/update test files with kerning data

rsheeter:
Files used in #331; split out here in hopes of making review simpler.

Step toward #308


glyphs-reader/src/font.rs
```diff
@@ -975,6 +979,14 @@ impl RawFont {
         if let Some(kerning) = self.other_stuff.remove("kerning") {
             self.other_stuff.insert("kerningLTR".to_string(), kerning);
         };
+        for glyph in self.glyphs.iter_mut() {
+            if let Some(Plist::String(group)) = glyph.other_stuff.remove("leftKerningGroup") {
+                glyph.kern_left = Some(group.clone());
```
cmyr, line 984: I don't think this clone is necessary? `remove` should take ownership.

glyphs-reader/src/font.rs
```diff
@@ -80,6 +80,8 @@ impl FeatureSnippet {
 pub struct Glyph {
     pub glyphname: String,
     pub layers: Vec<Layer>,
+    pub left_kern: Option<String>,
+    pub right_kern: Option<String>,
```
cmyr, line 84: I would probably either call these `left/right_kern_group` or else I would add a doc comment that these strings are the name of the group the glyph belongs to, if any?

cmyr (approved):
looks good, two little notes inline but nothing blocking :)

glyphs-reader/src/font.rs
rsheeter: Broadly I'm trying to match modern glyphs naming so I will add a comment

anthrotype: in that case it may be worth sticking with the original "kern_left" / "kern_right" even if it the reverse may look more correct in English

glyphs-reader/src/font.rs
rsheeter: Quite right, ty. Originally I had `.get`. I'm a bit surprised clippy didn't fault me for that one.

PR #334, opened 2023-06-07
Model kerning in IR and capture it for .glyphs and .designspace

rsheeter:
Intent is to output variable features we compile with fea-rs. IR model is meant to accomodate glyphs and designspace. Most interesting part should be `fontir/src/ir.rs` which has the IR changes, esp. `struct Kerning` which holds a variable-first representation of kerning.

left/right KerningGroup all the things. Or kern left/right ... it's all very confusing.

https://gist.github.com/rsheeter/4c8e7822d276f871eef974f8c1975ace shows an example of kerning IR (for Oswald).

Step toward #308. I want to leave filling in the blanks and aligning kerning groups to a subsequent PR.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -3,20 +3,23 @@
 use crate::{
     coords::{CoordConverter, NormalizedCoord, NormalizedLocation, UserCoord, UserLocation},
     error::{PathConversionError, VariationModelError, WorkError},
-    serde::{GlobalMetricsSerdeRepr, GlyphSerdeRepr, MiscSerdeRepr, StaticMetadataSerdeRepr},
+    serde::{
+        GlobalMetricsSerdeRepr, GlyphSerdeRepr, KerningSerdeRepr, MiscSerdeRepr,
+        StaticMetadataSerdeRepr,
+    },
     variations::VariationModel,
 };
 use chrono::{DateTime, Utc};
 use font_types::NameId;
 use font_types::Tag;
-use fontdrasil::types::GlyphName;
+use fontdrasil::types::{GlyphName, GroupName};
 use indexmap::IndexSet;
 use kurbo::{Affine, BezPath, PathEl, Point};
 use log::{trace, warn};
 use ordered_float::OrderedFloat;
 use serde::{Deserialize, Serialize};
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     fmt::Debug,
     path::PathBuf,
 };
@@ -93,6 +96,38 @@ pub struct MiscMetadata {
     pub lowest_rec_ppm: u16,
 
     pub created: Option<DateTime<Utc>>,
+}
+
+/// In logical (reading) order
+type KernPair = (KernParticipant, KernParticipant);
+type KernValues = BTreeMap<NormalizedLocation, OrderedFloat<f32>>;
+
+/// IR representation of kerning.
+///
+/// In UFO terms, roughly [groups.plist](https://unifiedfontobject.org/versions/ufo3/groups.plist/)
+/// and [kerning.plist](https://unifiedfontobject.org/versions/ufo3/kerning.plist/) combined.
+#[derive(Serialize, Deserialize, Debug, Default, Clone, PartialEq, Eq)]
+#[serde(from = "KerningSerdeRepr", into = "KerningSerdeRepr")]
+pub struct Kerning {
+    pub groups: HashMap<GroupName, HashSet<GlyphName>>,
+    /// An adjustment to the space *between* two glyphs in logical order.
+    ///
+    /// Maps (side1, side2) => a mapping location:adjustment.
+    ///
+    /// Used for both LTR and RTL. The BE application differs but the concept
+    /// is the same.
+    pub kerns: HashMap<KernPair, KernValues>,
+}
+
+/// A participant in kerning, one of the entries in a kerning pair.
+///
+/// Concretely, a glyph or a group of glyphs.
+///
+/// <https://unifiedfontobject.org/versions/ufo3/kerning.plist/#kerning-pair-types>
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum KernParticipant {
+    Glyph(GlyphName),
+    Group(GroupName),
 }
 
 impl StaticMetadata {
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -1,12 +1,12 @@
 use chrono::{TimeZone, Utc};
 use font_types::{NameId, Tag};
 use fontdrasil::orchestration::Work;
-use fontdrasil::types::GlyphName;
+use fontdrasil::types::{GlyphName, GroupName};
 use fontir::coords::NormalizedCoord;
 use fontir::error::{Error, WorkError};
 use fontir::ir::{
-    self, GlobalMetric, GlobalMetrics, GlyphInstance, NameBuilder, NameKey, NamedInstance,
-    StaticMetadata, DEFAULT_VENDOR_ID,
+    self, GlobalMetric, GlobalMetrics, GlyphInstance, KernParticipant, Kerning, NameBuilder,
+    NameKey, NamedInstance, StaticMetadata, DEFAULT_VENDOR_ID,
 };
 use fontir::orchestration::{Context, IrWork};
 use fontir::source::{Input, Source};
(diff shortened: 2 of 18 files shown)


ufo2fontir/src/source.rs
```diff
@@ -845,6 +863,79 @@ impl Work<Context, WorkError> for FeatureWork {
...
+        let mut kerning = Kerning::default();
+        for source in self.designspace.sources.iter() {
+            let pos = master_locations.get(&source.name).unwrap();
+            let ufo_dir = designspace_dir.join(&source.filename);
+            let data_request = norad::DataRequest::none().groups(true).kerning(true);
+            let font = norad::Font::load_requested_data(&ufo_dir, data_request)
+                .map_err(|e| WorkError::ParseError(ufo_dir, format!("{e}")))?;
+
+            kerning.groups = font
+                .groups
```
anthrotype, line 883: note that UFO's groups.plist is not just for _kerning_ groups, technically there can be any named list of glyphs and the use depend on the authoring tool. The kerning groups proper are prefixed with "public.kern1." and "public.kern2.", you should ignore anything else

ufo2fontir/src/source.rs
```diff
@@ -845,6 +863,79 @@ impl Work<Context, WorkError> for FeatureWork {
...
+                                }
+                            })
+                            .collect(),
+                    )
+                })
+                .collect();
+
+            let resolve_participant = |name: &norad::Name| {
+                let group_name = GroupName::from(name.as_str());
+                if kerning.groups.contains_key(&group_name) {
```
anthrotype, line 906: it's not the mere fact that the name of a kerning pair's member matches the name of one of the UFO groups that makes it a "kerning" group. You don't want to accidentally consider a kerning participant named "Rod" to be a "group" just because there's a group inside groups.plist that's named like that.
There's the public.kern{1,2}. prefix I mentioned above that has special meaning/restrictions (kern1 can only occur in first logical position, kern2 only on the second).
You should take a look at how ufo2ft's KernFeatureWriter collects UFO's kerning groups and pairs:
https://github.com/googlefonts/ufo2ft/blob/3e0563814cf541f7d8ca2bb7f6e446328e0e5e76/Lib/ufo2ft/featureWriters/kernFeatureWriter.py#L302-L357

fontir/src/ir.rs
```diff
@@ -95,6 +98,37 @@ pub struct MiscMetadata {
...
 }
 
+/// IR representation of kerning.
+///
+/// In UFO terms, roughly [groups.plist](https://unifiedfontobject.org/versions/ufo3/groups.plist/)
+/// and [kerning.plist](https://unifiedfontobject.org/versions/ufo3/kerning.plist/) combined.
+#[derive(Serialize, Deserialize, Debug, Default, Clone, PartialEq, Eq)]
+#[serde(from = "KerningSerdeRepr", into = "KerningSerdeRepr")]
+pub struct Kerning {
+    pub groups: HashMap<GroupName, HashSet<GlyphName>>,
```
anthrotype, line 108: since these are only the groups that have to do with kerning, we don't actually need to keep the "public.kern1." and "public.kern2." prefixes in the GroupName, there's no risk of clashing with non-kerning groups

fontir/src/ir.rs
```diff
@@ -95,6 +98,37 @@ pub struct MiscMetadata {
...
+    /// An adjustment to the space *between* two glyphs in logical order.
+    ///
+    /// Maps (side1, side2) => a mapping location:adjustment.
+    ///
+    /// Used for both LTR and RTL. The BE application differs but the concept
+    /// is the same.
+    pub kerns: HashMap<
+        (KernParticipant, KernParticipant),
+        BTreeMap<NormalizedLocation, OrderedFloat<f32>>,
+    >,
```
cmyr, line 118: this is a pretty nasty type, maybe an alias or two would be helpful?

glyphs2fontir/src/source.rs
```diff
@@ -458,6 +468,138 @@ impl Work<Context, WorkError> for FeatureWork {
     }
 }
 
+/// What side of the kern is this, in logical order
+enum KernSide {
+    L,
+    R,
```
cmyr, line 474: I'd prefer `Left` and `Right` 🤷

ufo2fontir/src/source.rs
rsheeter: TIL! ty, will fix

glyphs2fontir/src/source.rs
rsheeter: Ha, fair point, will rename.

fontir/src/ir.rs
rsheeter: That IS gross, adding!

fontir/src/ir.rs
rsheeter: Nice to be able to directly lookup what you get from kerning.plist so I'm inclined to retain. Please reopen if you disagree.

ufo2fontir/src/source.rs
rsheeter: Good catch, fixed, ty.

fontir/src/ir.rs
```diff
@@ -98,6 +98,9 @@ pub struct MiscMetadata {
     pub created: Option<DateTime<Utc>>,
 }
 
+type KernParticipants = (KernParticipant, KernParticipant);
```
anthrotype, line 101: how about `KernPair`?

glyphs2fontir/src/source.rs
```diff
@@ -548,8 +548,13 @@ impl Work<Context, WorkError> for KerningWork {
                 glyph
                     .left_kern
                     .iter()
-                    .map(|group| (KernSide::L, group))
-                    .chain(glyph.right_kern.iter().map(|group| (KernSide::R, group)))
+                    .map(|group| (KernSide::Side1, group))
```
anthrotype, line 551: I commented on the meaning of left/right in kernLeft/kernRight glyph attributes here: https://github.com/googlefonts/fontc/pull/331#discussion_r1223305938

I believe you have confused the two here. You are associating `left_kern` (which you renamed `kern_left`  in #331) to `KernSide::Side1`, but actually `kernLeft` (`leftKerningGroup` in Glyphs v2 terminology) is the group of glyphs that have the same kerning towards their left, so for LTR kerning that means logical side2, not side1...

ufo2fontir/src/source.rs
```diff
@@ -845,6 +866,107 @@ impl Work<Context, WorkError> for FeatureWork {
...
+
+        let designspace_dir = self.designspace_file.parent().unwrap();
+        let static_metadata = context.get_final_static_metadata();
+        let master_locations = master_locations(&static_metadata.axes, &self.designspace.sources);
+
+        let mut kerning = Kerning::default();
+
+        // Pass 1: find all the groups used by any source
+        // Woe unto ye who define the same group but inconsistently; we effectively union the members
+        for source in self.designspace.sources.iter() {
```
anthrotype, line 882: similarly to how we do for the global metrics, we need to skip for kerning those sources that have not-None `layer` attribute and that only contribute glyph data:

https://github.com/googlefonts/fontc/blob/04465968e5a7d68dfb1c80b058ccfc2d421286b7/ufo2fontir/src/source.rs#L752-L759

ufo2ft similarly doesn't compile features for sources whose layerName is not None

glyphs2fontir/src/source.rs
anthrotype:
```python
    if side == 1:
        return "rightKerningGroup"
    else:
        return "leftKerningGroup"
```

https://github.com/googlefonts/glyphsLib/blob/42bc1db912fd4b66f130fb3bdc63a0c1e774eb38/Lib/glyphsLib/builder/groups.py#L192-L195

as used in 

https://github.com/googlefonts/glyphsLib/blob/42bc1db912fd4b66f130fb3bdc63a0c1e774eb38/Lib/glyphsLib/builder/groups.py#L115-L123

rsheeter: Ugh indeed I misunderstood and didn't take it as "toward their left" but rather "goes on the left. Will fix, ty.

ufo2fontir/src/source.rs
rsheeter: Ew. Fixed.

glyphs2fontir/src/source.rs
rsheeter: I will copy https://github.com/googlefonts/fontc/compare/kern3...kern4?expand=1 in and update the glyphs2 copy.

rsheeter: Take 2 is posted.

fontir/src/ir.rs
rsheeter: Much better!

resources/testdata/glyphs3/WghtVar.glyphs
```diff
@@ -313,20 +313,20 @@ unicode = 61;
...
 m01 = {
+"@MMK_L_bracketleft_R" = {
+exclam = -165;
+};
 bracketleft = {
 bracketright = -300;
 };
 exclam = {
-"@MMK_R_brackets" = -160;
+"@MMK_R_bracketright_L" = -160;
```
anthrotype, line 323: this kern pair is lost (intentionally?) because you haven't renamed the bracketright glyph kerning groups accordingly, they are still called "brackets" for both L and R

anthrotype (approved):

resources/testdata/glyphs3/WghtVar.glyphs
rsheeter: Not intentionally! Fixed.

PR #336, opened 2023-06-12
ttx_diff.py: pass --drop-implied-oncurves fontmake option to reduce glyf diff

anthrotype:
Requires [fontmake 3.6.0](https://github.com/googlefonts/fontmake/releases/tag/v3.6.0) just released


rsheeter:
Awesome!

rsheeter (approved):

PR #337, opened 2023-06-12
Resolve fea includes for UFO against dir containing .ufo

### LINKED ISSUE #296: Included feature files not found
1. Try and compile https://github.com/madig/FirerSans/blob/main/sources/FiraSans.designspace
2. :fire: 

```
2023-05-23T21:03:02.584914Z: ThreadId(1): ERROR: Be(Features) failed Be(FeaCompileError(ParseFail(DiagnosticSet { messages: [Diagnostic { message: Message { text: "Failed to load source at 'family.fea': 'No such file or directory (os error 2)'", file: FileId(2), span: Span { start: 8, end: 18 } }, level: Error }], tree: "ParseTree" })))
Error: TasksFailed([(Be(Features), "Fea compilation failure")])
```

The default source just includes a shared family.fea: https://github.com/madig/FirerSans/blob/main/sources/FiraSans-Regular.ufo/features.fea -> https://github.com/madig/FirerSans/blob/main/sources/family.fea

The include path logic is explained in https://github.com/googlefonts/ufo2ft/blob/0d2688cd847d003b41104534d16973f72ef26c40/Lib/ufo2ft/featureCompiler.py#L47-L51.

rsheeter:
First crack at resolving includes. For the purposes of incremental compilation all fea files in the .designspace dir will be considered for feature file change. An include pointed outside the .designspace dir won't be detected as changed when attempting incremental compile.

Fixes #296. After this change I can compile https://github.com/madig/FirerSans w/o error. I make no claim that the font will actually _work_ ofc.

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -73,19 +73,26 @@ impl FeatureWork {
     }
 
     fn compile(&self, features: &Features, glyph_order: GlyphMap) -> Result<Compilation, Error> {
-        let root_path = if let Features::File(file) = features {
-            OsString::from(file)
-        } else {
-            OsString::new()
+        let compiler = match features {
+            Features::File {
+                fea_file,
+                include_dir,
+            } => {
+                let mut compiler = Compiler::new(OsString::from(fea_file), &glyph_order);
+                if let Some(include_dir) = include_dir {
+                    compiler = compiler.with_project_root(include_dir)
+                }
+                compiler
+            }
+            Features::Memory(fea_content) => {
+                let root = OsString::new();
+                Compiler::new(root.clone(), &glyph_order).with_resolver(InMemoryResolver {
+                    content_path: root,
+                    content: Arc::from(fea_content.as_str()),
+                })
+            }
+            Features::Empty => panic!("compile isn't supposed to be called for Empty"),
         };
-        let mut compiler = Compiler::new(root_path.clone(), &glyph_order);
-        if let Features::Memory(fea_content) = features {
-            let resolver = InMemoryResolver {
-                content_path: root_path,
-                content: Arc::from(fea_content.as_str()),
-            };
-            compiler = compiler.with_resolver(resolver);
-        }
         compiler.compile().map_err(Error::FeaCompileError)
     }
 }
--- a/resources/testdata/fea_include.designspace
+++ b/resources/testdata/fea_include.designspace
@@ -1,0 +1,35 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!-- Derived from Noto Sans Adlam -->
+<designspace format="4.1">
+  <axes>
+    <axis tag="wght" name="Weight" minimum="400" maximum="700" default="400"/>
+  </axes>
+  <sources>
+    <source filename="fea_include_ufo/FeaInc-Regular.ufo" name="Fea Inc Regular" familyname="Fea Inc" stylename="Regular">
+      <lib copy="1"/>
+      <groups copy="1"/>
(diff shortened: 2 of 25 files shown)


fontbe/src/features.rs
```diff
@@ -73,19 +73,26 @@ impl FeatureWork {
...
+            } => {
+                let mut compiler = Compiler::new(OsString::from(fea_file), &glyph_order);
+                if let Some(include_dir) = include_dir {
+                    compiler = compiler.with_project_root(include_dir)
+                }
+                compiler
+            }
+            Features::Memory(fea_content) => {
+                let root = OsString::new();
+                Compiler::new(root.clone(), &glyph_order).with_resolver(InMemoryResolver {
```
anthrotype, line 89: `include_dir` isn't used for in-memory compilation, does it mean we only properly support resolving external .fea includes when emitting IR and thus compiling features from a file?

ufo2fontir/src/source.rs
```diff
@@ -836,7 +858,11 @@ impl Work<Context, WorkError> for FeatureWork {
         }
 
         if !fea_files.is_empty() {
-            context.set_features(Features::from_file(&fea_files[0]));
+            let designspace_dir = self.designspace_file.parent().unwrap();
+            context.set_features(Features::from_file(
+                fea_files[0].clone(),
+                Some(designspace_dir.to_path_buf()),
```
anthrotype, line 864: no, it's not "relative to the directory of the .designspace", it's relative to the directory of the UFO. The designspace spec doesn't say anything about features.fea, it's the UFO spec that mandates that:
https://unifiedfontobject.org/versions/ufo3/features.fea/

it happens to work with FirerSans because in that particular case the source UFOs which contain features.fea with include statements happen to be in the same directory as the .designspace file that glues them together, but they may well be in a different sub-directory

rsheeter: Note to self: directory of the UFO means the directory containing it, not the actual blah.ufo/ directory

fontbe/src/features.rs
rsheeter: I'll double-check and add a test; it should work.

resources/testdata/fea_include.designspace
```diff
@@ -0,0 +1,35 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!-- Derived from Noto Sans Adlam -->
+<designspace format="4.1">
+  <axes>
+    <axis tag="wght" name="Weight" minimum="400" maximum="700" default="400"/>
+  </axes>
+  <sources>
+    <source filename="FeaInc-Regular.ufo" name="Fea Inc Regular" familyname="Fea Inc" stylename="Regular">
```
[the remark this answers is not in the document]
rsheeter: Ty, will fix.

anthrotype (approved):
LGTM provided we change it relative to UFO path and verify this works for in-memory compilation as well

ufo2fontir/src/source.rs
rsheeter: I will modify my test files to expose the bug & fix.

fontbe/src/features.rs
rsheeter: (in memory should be when the source is in memory, not just when emit ir is off)

anthrotype: oh, like when compiling off .glyphs source (which embeds a features string in its single file source) as opposed to designspace+UFO where features.fea are distinct files?

rsheeter:
> verify this works for in-memory compilation as well

Added a test with IR turned off

PR #338, opened 2023-06-13
Take the default master as authoritative in designspace kerning

rsheeter:
Based on discussion on https://github.com/googlefonts/ufo2ft/pull/635, treat the group names in the default master as authorititive. Remap alternate names for the same {glyphs} in other masters. Ignore all groups from non-default masters that do not identify a set of glyphs given a name in the default.

### DIFF
--- a/ufo2fontir/src/source.rs
+++ b/ufo2fontir/src/source.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{BTreeMap, HashMap, HashSet},
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
     path::{Path, PathBuf},
     str::FromStr,
     sync::Arc,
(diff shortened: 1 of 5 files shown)


ufo2fontir/src/source.rs
```diff
@@ -877,47 +929,64 @@ impl Work<Context, WorkError> for KerningWork {
...
-                            })
-                            .collect(),
-                    )
-                })
-                .collect();
+            for (name, entries) in
+                groups_for(designspace_dir, &static_metadata.glyph_order, source)?
+            {
+                let Some(real_name) = reverse_groups.get(&(KernSide::of(&name), &entries)) else {
+                    trace!("{name} exists only in {} and will be ignored", source.name);
```
anthrotype, line 974: I suggest you turn this to a warning level message because, in my understanding, non-default masters aren't supposed to define different groups from the default

ufo2fontir/src/source.rs
```diff
@@ -939,15 +1008,15 @@ impl Work<Context, WorkError> for KerningWork {
                         return None;
                     }
                     let group_name = GroupName::from(name);
-                    if !kerning.groups.contains_key(&group_name) {
-                        debug!("'{name}' references non-existent group {group_name}; ignored");
+                    let Some(group_name) = old_to_new_group_name.get(&group_name) else {
+                        debug!("'{name}' is not a valid group name; ignored");
```
anthrotype, line 1012: better to issue a warning since we are effectively dropping a kerning pair value here, better to alert the user so they fix their group names

ufo2fontir/src/source.rs
```diff
@@ -939,15 +1008,15 @@ impl Work<Context, WorkError> for KerningWork {
...
                         return None;
-                    }
-                    Some(KernParticipant::Group(group_name))
+                    };
+                    Some(KernParticipant::Group((**group_name).clone()))
                 } else {
                     let glyph_name = GlyphName::from(name);
                     if !static_metadata.glyph_order.contains(&glyph_name) {
-                        debug!("'{name}' references non-existent glyph {glyph_name}; ignored");
+                        debug!("'{name}' is not a valid glyph name; ignored");
```
anthrotype, line 1019: again, warning would be better

the message should make it clear that the glyph name is not simply "invalid", but the glyph referenced in the kerning does't exist in the font

anthrotype (approved):

ufo2fontir/src/source.rs
rsheeter: Done, ty

ufo2fontir/src/source.rs
rsheeter: Done

ufo2fontir/src/source.rs
rsheeter: Updated several similar debugs to warn

PR #340, opened 2023-06-14
only test for use_my_metrics component flags for the default instance

anthrotype:
Initially I thought of modifying fonttools to match fontc and make sure USE_MY_METRICS flag is only set on the VF when all the masters have it set (https://github.com/fonttools/fonttools/pull/3163), but then I changed my mind, because it doesn't matter what the non default masters do, and in fact I wasn't able to see any differences whatsoever whether that flag was set only when all masters would set it (fontc's current logic), vs the flag being set when the default master has it set (ufo2ft's logic).
Note that we only ever set that flag automatically on the first component whose base glyph's advance is _the same_ as the composite glyph that references it, so whether it is set or not doesn't actually produce any visual differences, but probably saves a fraction of a microsecond as freetype or skrifa don't need to obtain the phantom points for the composite glyphs but can copy those from the component, that's it really. But remember if the flag is set by us, then it means the advances of component/composite are the same; if it is not set, they are different and they will be fetched from hmtx accordingly. In any case, the flag doesn't change anything, and since the goal here is to match fontmake's output, the easiest is to modify our own logic and set the component flag only if the _default instance_'s composite glyph matches the width of any of its components, without also checking other non-default instances.

With this patch, Oswald's glyf table matches almost 100% except for one last difference (a duplicate last point due to reverse pen issue https://github.com/googlefonts/fontations/issues/370)

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -37,7 +37,7 @@ pub fn create_glyf_work(glyph_name: GlyphName) -> Box<BeWork> {
     Box::new(GlyphWork { glyph_name })
 }
 
-/// Can glyph reuse the metrics of other?
+/// Can glyph instance reuse the metrics of other?
 ///
 /// To be safe the component should have:
 ///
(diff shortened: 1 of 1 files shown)


anthrotype:
It doesn't matter for the non default masters because when we get the phantom points for those, as we build the point list to be used for computing gvar deltas, we don't care about whether that flag is set or not, but always consult the metrics tables for the actual values

anthrotype:
It really doesn't make any difference at runtime whether it is automatically set they way we do (i.e. composite width is same as the component's width), or not set at all, hence it's simpler to just do what the python pipeline does in this case, the least meaningless changes the better

anthrotype:
> it should be possible to construct a case where it matters?

I tried but I wasn't able to

rsheeter:
>> it should be possible to construct a case where it matters?
>
> I tried but I wasn't able to

Huh, fascinating. In that case LGTM until such time as we can prove it matters :)

anthrotype:
It doesn't otherwise we would have known by now

PR #346, opened 2023-06-27
[fontir] Use common rust method names in Location

cmyr:
This makes the API more discoverable and intuitive to users familiar with other Rust collection types.


rsheeter:
I'm in the midst of moving these types so I might manually make an equivalent change after that.

rsheeter (approved):
LGTM, ty

cmyr:
This will be easy to fixup after #348 so let's get that in first.

PR #347, opened 2023-06-27
Move some types to fontdrasil. No edits, just move.

rsheeter:
Meant to allow access to types to fea-rs. I prefer to make non-move changes in a separate follow-on.


rsheeter:
I'm still of the opinion this is a good idea but enough has changed I'd probably want to do it in a new PR

PR #348, opened 2023-06-27
Use tag instead of name in Location

rsheeter:
Name matches how some sources do it but Tag feels much more natural so translate to tag more aggressively.

IMHO #347 makes more sense after this. I think this is a good change regardless of whether we proceed with #347.

JMM.

### DIFF
--- a/fontbe/src/orchestration.rs
+++ b/fontbe/src/orchestration.rs
@@ -2,7 +2,7 @@
 
 use std::{collections::HashMap, fs, io, path::Path, sync::Arc};
 
-use font_types::F2Dot14;
+use font_types::{F2Dot14, Tag};
 use fontdrasil::{
     orchestration::{Access, AccessControlList, Work, MISSING_DATA},
     types::GlyphName,
@@ -205,7 +205,7 @@ impl GvarFragment {
 /// <https://learn.microsoft.com/en-us/typography/opentype/spec/otvaroverview#variation-data>
 #[derive(Debug, Default)]
 struct TupleBuilder {
-    axis_names: Vec<String>,
+    axes: Vec<Tag>,
     min: Vec<F2Dot14>,
     peak: Vec<F2Dot14>,
     max: Vec<F2Dot14>,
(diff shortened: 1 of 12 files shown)


fontbe/src/orchestration.rs
```diff
@@ -705,7 +708,7 @@ mod tests {
     fn non_default_region() -> VariationRegion {
         let mut region = VariationRegion::default();
         region.insert(
-            "Weight".to_string(),
+            Tag::from_str("wght").unwrap(),
```
cmyr, line 711: I'd prefer writing this as,


```suggestion
            Tag::new(b"wght"),
```

glyphs2fontir/src/source.rs
```diff
@@ -996,8 +993,8 @@ mod tests {
             (72.0, 700.0),
         ] {
             let mut loc = UserLocation::new();
-            loc.set_pos("Optical Size", UserCoord::new(*opsz));
-            loc.set_pos("Weight", UserCoord::new(*wght));
+            loc.set_pos(Tag::from_str("opsz").unwrap(), UserCoord::new(*opsz));
```
cmyr, line 996: ditto these could just use`Tag::new`, which is infallible.

ufo2fontir/src/source.rs
```diff
@@ -1252,11 +1258,12 @@ mod tests {
     fn add_design_location(
         add_to: &mut HashMap<PathBuf, Vec<DesignLocation>>,
         glif_file: &str,
-        axis: &str,
+        tag: &str,
```
cmyr, line 1261: would it make sense to ask the caller to pass in a `Tag`?

cmyr (approved):
I think this makes sense, tags are used much more consistently than names, and means we don't need to worry about borrowing/cloning all over the place.

In this PR you're using `Tag::from_str` as a general constructor, but that method is really only intended for use when parsing; if you have a known-good input (a literal) you should prefer `Tag::new` which takes a `&[u8; 4]` and can't fail.

glyphs2fontir/src/source.rs
rsheeter: Ty, will fix.

ufo2fontir/src/source.rs
rsheeter: Yes, will fix.

rsheeter:
Force push to rebase after merging #346

Issue #349, opened 2023-06-28
Google Sans has outlines we don't like


rsheeter:
https://github.com/googlefonts/googlesans/issues/586

anthrotype:
The offending contour is in GoogleSans-opsz18-wght380-GRAD-50.ufo/glyphs/tzaa-ethiopic.glif, here's an excerpt:

```xml
    <contour>
      <point x="109" y="720"/>
      <point x="30" y="649"/>
      <point x="30" y="521" type="curve" smooth="yes"/>
[...]
      <point x="405" y="658"/>
      <point x="346" y="720"/>
      <point x="228" y="720" type="curve" smooth="yes"/>
    </contour>
```

looks like we don't add an inital MoveTo when the glif contour begin with off-curves like in this case, here's the generated IR generated from the above, which is missing the initial MoveTo and causing the panic:

```
    contours:
    - - !CurveTo
        - x: 109.0
          y: 720.0
        - x: 30.0
          y: 649.0
        - x: 30.0
          y: 521.0
[...]
      - !CurveTo
        - x: 405.0
          y: 658.0
        - x: 346.0
          y: 720.0
        - x: 228.0
          y: 720.0
```

anthrotype:
when compiling in debug mode, kurbo contains debug_assert! statements that ensure that a valid BezPath always starts with a MoveTo. The OP run fontc as `cargo run --release` hence these were skipped, but then we panic later on in fontbe/glyphs.rs when converting curves to quadratics

PR #350, opened 2023-06-28
Use display representations when logging errors

cmyr:
`Display` is intended to be human readable, and will generally produce better output than `Debug`.

This is particularly true in fea-rs, where we have fancy display impls.

Motivated by trying to debug #349. I have some fixes up, but there's a bunch of other nasty stuff in the FEA here that I need to look into.

### DIFF
--- a/fontbe/src/error.rs
+++ b/fontbe/src/error.rs
@@ -12,7 +12,7 @@ use write_fonts::tables::{glyf::BadKurbo, gvar::GvarInputError, variations::IupE
 pub enum Error {
     #[error("IO failure")]
     IoError(#[from] io::Error),
-    #[error("Fea compilation failure")]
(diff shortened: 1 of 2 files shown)


fontbe/src/error.rs
```diff
@@ -12,7 +12,7 @@ use write_fonts::tables::{glyf::BadKurbo, gvar::GvarInputError, variations::IupE
 pub enum Error {
     #[error("IO failure")]
     IoError(#[from] io::Error),
-    #[error("Fea compilation failure")]
+    #[error("FEA compilation failure {0}")]
```
rsheeter, line 15: +1 to Display, -1 to FEA in all caps. Alternatives: ".fea compilation failure" or "Feature file compilation failure"

rsheeter (approved):
LGTM

Issue #355, opened 2023-07-02
Curious relationship between change detector and Source


rsheeter:
Migrating fixme to a comment:

```rust
                //FIXME: it's weird that we call a method on this type, passing
                //in an argument we also get from this type? In a world where there
                //is a general trait for 'create work' on Source we can have
                //a method on change_detector that calls this method with the current inputs?
                work: change_detector
                    .ir_source()
                    .create_global_metric_work(change_detector.current_inputs())?
                    .into(),
```

PR #356, opened 2023-07-02
Housekeeping: a work should know it's own id

rsheeter:
For context, initially I wasn't sure how our work model would play out. I feel like it's settled enough to be worth tidying up. I plan to move knowledge of what slots the work reads/writes to the work itself in a subsequent PR.

`fontdrasil/src/orchestration.rs` is the interesting part, to the extent there is an interesting part. The result is just shuffling things to get id from `Work::id()`

Along the way I deleted a fixme, filing https://github.com/googlefonts/fontc/issues/355 instead.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -96,27 +96,23 @@ fn add_init_static_metadata_ir_job(
     change_detector: &mut ChangeDetector,
     workload: &mut Workload,
 ) -> Result<(), Error> {
+    let work = change_detector
+        .ir_source()
+        .create_static_metadata_work(change_detector.current_inputs())?;
     if change_detector.init_static_metadata_ir_change() {
-        let id: AnyWorkId = FeWorkIdentifier::InitStaticMetadata.into();
+        let id: AnyWorkId = work.id().into();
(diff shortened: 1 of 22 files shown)


fontc/src/lib.rs
```diff
@@ -148,17 +145,18 @@ fn add_finalize_static_metadata_ir_job(
...
-                work: create_finalize_static_metadata_work().into(),
+                work: work.into(),
                 dependencies,
                 read_access: ReadAccess::Custom(read_access),
                 write_access,
             },
         );
     } else {
-        workload.mark_success(FeWorkIdentifier::FinalizeStaticMetadata);
+        workload.mark_success(work.id());
```
cmyr, line 159: cool, this code is now starting to look much more general which suggests maybe we can leverage generics & traits more, going forward.

cmyr (approved):
looks good! I started playing with this awhile ago but never took it very far, but it's nice to continue moving towards a place where more of the logic is in the trait implementations, which means we may be able to remove a lot of the redundant glue code at some point.

rsheeter:
> we may be able to remove a lot of the redundant glue code at some point.

That's the plan!

PR #358, opened 2023-07-02
Housekeeping: Let Work tell us what ids it completed

rsheeter:
Building on #356, let work tell us if it completes things other than itself (e.g. glyf completes loca)

JMM if happy

### DIFF
--- a/fontc/src/workload.rs
+++ b/fontc/src/workload.rs
@@ -20,6 +20,9 @@ pub struct Workload {
     job_count: usize,
     success: HashSet<AnyWorkId>,
     error: Vec<(AnyWorkId, String)>,
+
+    // When K completes also mark all entries in V complete
+    also_completes: HashMap<AnyWorkId, Vec<AnyWorkId>>,
     pub(crate) jobs_pending: HashMap<AnyWorkId, Job>,
 }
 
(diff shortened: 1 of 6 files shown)


fontc/src/workload.rs
```diff
@@ -65,6 +72,14 @@ impl Workload {
 
     fn handle_success(&mut self, fe_root: &FeContext, success: AnyWorkId) {
         log::debug!("{success:?} successful");
+
+        let also_completed = self.also_completes.get(&success).cloned();
```
cmyr, line 76: This tweaks me a little bit, but I get that the alternatives are annoying. I can think of two:

1. instead of storing a `Vec<_>` in `self.also_completes`, store a `Arc<[_]>`; that is, an atomically reference counted fixed-size slice, which doesn't require a deep copy.
2. Rewrite this in such a way that the compiler can tell that our borrows don't overlap. In this case I would move this into its own little helper fn, which would look like,

```rust
fn mark_also_completed(&mut self, done_id: &AnyWorkId) {
    let Some(also_completed) = self.also_completes.get(done_id) else { return };
    for id in also_completed {
        if self.success.insert(id.clone()) {
            self.pre_success += 1;
        }
    }
}
```

this works because within a function the compiler can tell that the borrows don't overlap, but one you start calling methods on self it stops doing fine-grained borrow tracking.

cmyr (approved):
looks good, one comment inline that you're free to ignore.

fontc/src/workload.rs
rsheeter: Option 2 works for me. Using an arc for this feels mildly silly.

rsheeter: Aside: Option 3: clone and don't worry about it unless it shows up in a profile also works for me.

rsheeter: Made it a function

cmyr: Option 3 was totally reasonable, should have made that more clear.

PR #359, opened 2023-07-02
Housekeeping: let work tell us dependencies and what it reads/writes

rsheeter:
Large but mostly boring cleanup. Took longer than I care to admit to get everything wired up and working again :D

Move the knowledge of read/write access to the work. Define dependencies in terms of read access: if you can read it you depend on it and therefore must happen after it. Eliminate separate storage of dependencies, it was duplicative and quite weird when inconsistent with read_access.

Eliminate the macro for Context items, preferring to instead write reusable types ContextItem and ContextMap, as @cmyr once suggested.

Files (there are several) named orchestration.rs are probably the most interesting part; the rest is essentially repetitive updates to match the new interfaces. I would start with `fontdrasil/src/orchestration.rs`, then `fontir/src/orchestration.rs` and finally `fontbe/src/orchestration.rs`.

### DIFF
--- a/fontbe/src/orchestration.rs
+++ b/fontbe/src/orchestration.rs
@@ -1,20 +1,26 @@
 //! Helps coordinate the graph execution for BE
 
-use std::{collections::HashMap, fs, io, path::Path, sync::Arc};
+use std::{
+    fs::File,
+    io::{self, BufReader, BufWriter, Read, Write},
+    path::{Path, PathBuf},
+    sync::Arc,
+};
 
 use font_types::{F2Dot14, Tag};
 use fontdrasil::{
-    orchestration::{Access, AccessControlList, Work, MISSING_DATA},
+    orchestration::{Access, AccessControlList, Identifier, Work},
     types::GlyphName,
 };
 use fontir::{
-    context_accessors,
-    orchestration::{Context as FeContext, ContextItem, Flags, WorkId as FeWorkIdentifier},
+    orchestration::{
+        Context as FeContext, ContextItem, ContextMap, Flags, IdAware, Persistable,
+        PersistentStorage, WorkId as FeWorkIdentifier,
+    },
     variations::VariationRegion,
 };
 use kurbo::Vec2;
 use log::trace;
-use parking_lot::RwLock;
 use read_fonts::{FontData, FontRead};
 use serde::{Deserialize, Serialize};
 
@@ -81,6 +87,8 @@ pub enum WorkId {
     Stat,
 }
 
+impl Identifier for WorkId {}
+
 // Identifies work of any type, FE, BE, ... future optimization passes, w/e.
 // Useful because BE work can very reasonably depend on FE work
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
@@ -88,6 +96,8 @@ pub enum AnyWorkId {
     Fe(FeWorkIdentifier),
     Be(WorkId),
 }
+
+impl Identifier for AnyWorkId {}
 
 impl AnyWorkId {
     pub fn unwrap_be(&self) -> &WorkId {
@@ -117,46 +127,123 @@ impl From<WorkId> for AnyWorkId {
     }
 }
 
+impl From<&FeWorkIdentifier> for AnyWorkId {
+    fn from(id: &FeWorkIdentifier) -> Self {
+        AnyWorkId::Fe(id.clone())
+    }
+}
+
+impl From<&WorkId> for AnyWorkId {
+    fn from(id: &WorkId) -> Self {
+        AnyWorkId::Be(id.clone())
+    }
+}
+
 /// <https://learn.microsoft.com/en-us/typography/opentype/spec/glyf>
 #[derive(Debug, Clone)]
 pub enum Glyph {
-    Simple(SimpleGlyph),
-    Composite(CompositeGlyph),
-}
-
-impl From<SimpleGlyph> for Glyph {
-    fn from(value: SimpleGlyph) -> Self {
-        Glyph::Simple(value)
-    }
-}
-
-impl From<CompositeGlyph> for Glyph {
-    fn from(value: CompositeGlyph) -> Self {
-        Glyph::Composite(value)
-    }
+    Simple(GlyphName, SimpleGlyph),
+    Composite(GlyphName, CompositeGlyph),
 }
 
 impl Glyph {
+    pub(crate) fn new_simple(glyph_name: GlyphName, simple: SimpleGlyph) -> Glyph {
+        Glyph::Simple(glyph_name, simple)
+    }
+
+    pub(crate) fn new_composite(glyph_name: GlyphName, composite: CompositeGlyph) -> Glyph {
+        Glyph::Composite(glyph_name, composite)
+    }
+
+    pub(crate) fn glyph_name(&self) -> &GlyphName {
+        match self {
+            Glyph::Simple(name, _) | Glyph::Composite(name, _) => name,
+        }
+    }
+
     pub fn to_bytes(&self) -> Vec<u8> {
         match self {
-            Glyph::Simple(table) => dump_table(table),
-            Glyph::Composite(table) => dump_table(table),
+            Glyph::Simple(_, table) => dump_table(table),
+            Glyph::Composite(_, table) => dump_table(table),
         }
         .unwrap()
     }
 
     pub fn bbox(&self) -> Bbox {
         match self {
-            Glyph::Simple(table) => table.bbox,
-            Glyph::Composite(table) => table.bbox,
+            Glyph::Simple(_, table) => table.bbox,
+            Glyph::Composite(_, table) => table.bbox,
         }
     }
 
     pub fn is_empty(&self) -> bool {
         match self {
-            Glyph::Simple(table) => table.contours().is_empty(),
-            Glyph::Composite(table) => table.components().is_empty(),
+            Glyph::Simple(_, table) => table.contours().is_empty(),
+            Glyph::Composite(_, table) => table.components().is_empty(),
         }
+    }
+}
+
+impl IdAware<AnyWorkId> for Glyph {
+    fn id(&self) -> AnyWorkId {
+        AnyWorkId::Be(WorkId::GlyfFragment(self.glyph_name().clone()))
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+struct GlyphPersistable {
+    name: GlyphName,
+    simple: bool,
+    glyph: Vec<u8>,
+}
+
+impl From<&Glyph> for GlyphPersistable {
+    fn from(value: &Glyph) -> Self {
+        match value {
+            Glyph::Simple(name, table) => GlyphPersistable {
+                name: name.clone(),
+                simple: true,
+                glyph: dump_table(table).unwrap(),
+            },
+            Glyph::Composite(name, table) => GlyphPersistable {
+                name: name.clone(),
+                simple: false,
+                glyph: dump_table(table).unwrap(),
+            },
+        }
+    }
+}
+
+impl From<GlyphPersistable> for Glyph {
+    fn from(value: GlyphPersistable) -> Self {
+        match value.simple {
+            true => {
+                let glyph =
+                    read_fonts::tables::glyf::SimpleGlyph::read(FontData::new(&value.glyph))
+                        .unwrap();
+                let glyph = SimpleGlyph::from_table_ref(&glyph);
+                Glyph::Simple(value.name, glyph)
+            }
+            false => {
+                let glyph =
+                    read_fonts::tables::glyf::CompositeGlyph::read(FontData::new(&value.glyph))
+                        .unwrap();
+                let glyph = CompositeGlyph::from_table_ref(&glyph);
+                Glyph::Composite(value.name, glyph)
+            }
+        }
+    }
+}
+
+impl Persistable for Glyph {
+    fn read(from: &mut dyn Read) -> Self {
+        bincode::deserialize_from::<&mut dyn Read, GlyphPersistable>(from)
+            .unwrap()
+            .into()
+    }
+
+    fn write(&self, to: &mut dyn Write) {
+        bincode::serialize_into::<&mut dyn Write, GlyphPersistable>(to, &self.into()).unwrap();
     }
 }
 
@@ -165,6 +252,7 @@ impl Glyph {
 /// <https://learn.microsoft.com/en-us/typography/opentype/spec/gvar>
 #[derive(Serialize, Deserialize, Debug)]
 pub struct GvarFragment {
+    pub glyph_name: GlyphName,
     /// None entries are safe to omit per IUP
     pub deltas: Vec<(VariationRegion, Vec<Option<Vec2>>)>,
 }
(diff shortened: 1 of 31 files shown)


fontdrasil/src/orchestration.rs
```diff
@@ -11,7 +11,10 @@ pub const MISSING_DATA: &str = "Missing data, dependency management failed us?";
 
 /// A rule that represents whether access to something is permitted.
 #[derive(Clone)]
-pub enum Access<I> {
+pub enum Access<I>
+where
+    I: Debug,
```
cmyr, line 16: I assume something happened, here? There's a major convention in Rust to avoid trait bounds on types wherever possible, and leave them for impl blocks; for instance even `HashMap` does not have a constraint that the key implement `Hash`. Looking through this I assume you ran into some problem that pushed you in this direction, but maybe there's an alternative?

fontir/src/orchestration.rs
```diff
@@ -34,83 +32,306 @@ bitflags! {
...
+/// Clones are cheap and reference the same wrapped item courtesy of Arc
 ///
-/// Assumes we are in an impl block for a Context and that
-/// self.$lock_name is a ContextItem<$ir_type>
+/// Courtesy of Arc this is Clone even if T isn't
+#[derive(Default)]
+pub struct ContextItem<I, T, Ir>
+where
+    I: Clone + Debug + PartialEq + Eq + Hash,
+    Ir: IrIo<I>,
```
cmyr, line 51: ditto, you can just delete this

fontir/src/orchestration.rs
```diff
@@ -34,83 +32,306 @@ bitflags! {
...
+{
+    id: I,
+    acl: Arc<AccessControlList<I>>,
+    irio: Arc<Ir>,
```
cmyr, line 55: maybe worth a comment noting what `irio` does? I assume serialization?

fontir/src/orchestration.rs
```diff
@@ -34,83 +32,306 @@ bitflags! {
...
+
+    /// Update the value of the item whether or not it has changed.
+    ///
+    /// The change will be logged and anything relying on change detection, such as
+    /// conditional execution of dependent tasks, will fire.
+    ///
+    /// [ContextItem::set] is preferable where possible.
+    ///
+    /// This exists largely because write types in fontations do not always implement PartialEq.
+    /// TODO: should they?
```
cmyr, line 120: maybe? I have a mechanism for selectively implementing eq/hash/ord, but maybe it makes sense to just always implement these traits?

fontir/src/orchestration.rs
```diff
@@ -34,83 +32,306 @@ bitflags! {
...
+
+        // nop?
+        if self
+            .value
+            .read()
+            .as_ref()
+            .map(|arc| **arc == value)
+            .unwrap_or(false)
+        {
+            eprintln!("set_if_changed {:?}, NO CHANGE!", self.id);
```
cmyr, line 154: would logging be more appropriate?

fontir/src/orchestration.rs
```diff
@@ -34,83 +32,306 @@ bitflags! {
...
+    Ir: IrIo<I>,
+{
+    acl: Arc<AccessControlList<I>>,
+    irio: Arc<Ir>,
+    value: Arc<RwLock<HashMap<I, Arc<T>>>>,
+}
+
+impl<I, T, Ir> ContextMap<I, T, Ir>
+where
+    I: Clone + Debug + PartialEq + Eq + Hash,
```
cmyr, line 180: At this point I'm starting to wonder if we should just define,

```rust
pub trait Identifier: Clone + Debug + Eq + Hash {}
```
?

fontir/src/orchestration.rs
```diff
@@ -34,83 +32,306 @@ bitflags! {
...
+pub trait IdAware<I> {
+    fn id(&self) -> I;
+}
+
+pub trait Persistable {
+    fn read(from: &mut dyn Read) -> Self;
+    fn write(&self, to: &mut dyn Write);
+}
+
+/// Reads and writes IR
```
cmyr, line 270: I'd love a bit more exposition, I assume this is going to be implemented by some type that is responsible for reading/writing... stuff? anything? specific things?

fontir/src/orchestration.rs
```diff
@@ -34,83 +32,306 @@ bitflags! {
...
+pub trait IrIo<I> {
+    fn emit_ir(&self) -> bool;
```
cmyr, line 272: is this something I call to find out if I *should* emit IR, or something that emits IR and returns a success flag, or something else?

fontir/src/orchestration.rs
```diff
@@ -34,83 +32,306 @@ bitflags! {
...
+    PreliminaryGlyphOrder,
+    /// The final glyph order. Most things that need glyph order should rely on this.
+    GlyphOrder,
     Features,
     Kerning,
 }
 
 pub type IrWork = dyn Work<Context, WorkId, WorkError> + Send;
 
+pub struct FeIrIo {
```
cmyr, line 305: All I see is *Fee-fi-fo-fum* 😬

fontir/src/orchestration.rs
```diff
@@ -119,22 +340,29 @@ pub type IrWork = dyn Work<Context, WorkId, WorkError> + Send;
...
 
     // work results we've completed or restored from disk
     // We create individual caches so we can return typed results from get fns
-    init_static_metadata: ContextItem<ir::StaticMetadata>,
-    final_static_metadata: ContextItem<ir::StaticMetadata>,
-    global_metrics: ContextItem<ir::GlobalMetrics>,
-    glyph_ir: Arc<RwLock<HashMap<GlyphName, Arc<ir::Glyph>>>>,
-    feature_ir: ContextItem<ir::Features>,
-    kerning: ContextItem<ir::Kerning>,
+    pub static_metadata: ContextItem<WorkId, ir::StaticMetadata, FeIrIo>,
```
cmyr, line 353: okay, nice to see how this fits together.

Looking at how similar these all are, would it make sense to have something like,

```rust
type FeContextItem<T> = ContextItem<WorkId, T, FeIoFum>;
```
?

fontir/src/serde.rs
```diff
@@ -87,15 +85,25 @@ impl From<StaticMetadata> for StaticMetadataSerdeRepr {
...
-                .glyph_order
-                .into_iter()
-                .map(|n| n.as_str().to_string())
-                .collect(),
         }
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct GlyphOrderSerdeRepr(Vec<String>);
```
cmyr, line 93: does this need to be public?

fontbe/src/orchestration.rs
```diff
@@ -117,54 +122,85 @@ impl From<WorkId> for AnyWorkId {
     }
 }
 
+impl From<&FeWorkIdentifier> for AnyWorkId {
```
cmyr, line 125: I imagine this is just being moved around? In any case just noticing it now, but I would _generally_ avoid doing `impl From<&T>` if you are just cloning in your impl; it's preferable to just `impl From<T>` and then force the caller to be explicit about the clone.

This is a broader idea in rust of avoiding writing methods that ask for things to be passed by value, but then clone internally; the caller can't tell that they're cloning, and it may lead to situations where the caller has owned data which they pass by reference, causing it to be cloned unnecessarily.

fontbe/src/orchestration.rs
```diff
@@ -117,54 +122,85 @@ impl From<WorkId> for AnyWorkId {
...
 
+impl IdAware<AnyWorkId> for Glyph {
+    fn id(&self) -> AnyWorkId {
+        AnyWorkId::Be(WorkId::GlyfFragment(self.glyph_name().clone()))
+    }
+}
+
+impl Persistable for Glyph {
+    fn read(from: &mut dyn Read) -> Self {
+        todo!()
```
cmyr, line 190: was this intended to be filled in by now?

cmyr (commented):
Did a quick read through the various orchestration files, and have a few notes inline; overall this is a nice improvement, and it's nice to see you being more comfortable mucking about with the trait system.

fontir/src/orchestration.rs
rsheeter: Oh taht should just go away. For context, I'm fond of using eprint for logs I want while playing that I don't intend to keep.

fontir/src/serde.rs
rsheeter: The entire module is pub(crate) in lib.rs so I thought I didn't need to care. However, I have subsequently come to appreciate that pub(crate) here would let the compiler check for unused items so I think I should indeed go set things here pub(crate).

This PR already has a lot of noise so I'm inclined to tackle separately; deferred to https://github.com/googlefonts/fontc/issues/362.

fontbe/src/orchestration.rs
rsheeter: It is on my local copy :D

fontdrasil/src/orchestration.rs
rsheeter: I don't need to commit that, will fix. I underappreciated the power of where clauses until the PR. I did run into the traditional desire for a negative clause, but I can see how that might have some sharp edges if added.

> even HashMap does not have a constraint that the key implement Hash

I agree with your point but am inclined to pointlessly dispute this example. The key not having to implement hash is pretty intrinsic to a design where you have a separate hasher instead of keys that that know how to hash themselves.

fontir/src/orchestration.rs
rsheeter: I landed on maybe too! Deferred to https://github.com/googlefonts/fontations/issues/532

fontbe/src/orchestration.rs
rsheeter: Good catch. There may be more instances of this, deferred to https://github.com/googlefonts/fontc/issues/363.

fontir/src/orchestration.rs
rsheeter: I like it!

fontir/src/orchestration.rs
rsheeter: Hm, I think I will try to rename to better reflect purpose

rsheeter: Hopefully better

fontir/src/orchestration.rs
rsheeter: Added

fontir/src/orchestration.rs
rsheeter: renamed active to hopefully more clearly capture meaning

fontir/src/orchestration.rs
rsheeter: I was going to rename but now I don't want to!

fontir/src/orchestration.rs
rsheeter: Probably; it's a pretty common idea. Per your other comment perhaps debug isn't strictly necessary although ... I think being able to debug your ids is going to end up ubiquitous.

rsheeter: I like it, conveys the intent well.

fontdrasil/src/orchestration.rs
cmyr: Even with a custom hasher, keys need to implement the `Hash` trait. The [impl block that constains methods like `insert`](http://doc.rust-lang.org/1.70.0/src/std/collections/hash/map.rs.html#732-1193) requires that keys impl `Hash`.

fontir/src/ir.rs
```diff
@@ -98,6 +95,75 @@ pub struct MiscMetadata {
     pub created: Option<DateTime<Utc>>,
 }
 
+/// The name of every glyph, in the order it will be emitted
+///
+/// <https://rsheeter.github.io/font101/#glyph-ids-and-the-cmap-table>
+#[derive(Serialize, Deserialize, Default, Debug, Clone, PartialEq, Eq)]
+#[serde(from = "GlyphOrderSerdeRepr", into = "GlyphOrderSerdeRepr")]
```
cmyr, line 102: I'm surprised this needs a custom serde impl? It feels simple enough to Just Work.

Also this could potentially be shared with fea-rs at some point, although fea-rs also has to support CID identifiers for CFF, e.g. sometimes we don't get a glyph name, just an integer CID.

fontdrasil/src/orchestration.rs
```diff
@@ -9,9 +9,14 @@ use std::{
...
 
+/// A type that affords identity.
+///
+/// Frequently copied, used in hashmap/sets and printed to logs, hence the trait list.
+pub trait Identifier: Debug + Clone + Eq + Hash {}
+
 /// A rule that represents whether access to something is permitted.
 #[derive(Clone)]
-pub enum Access<I> {
+pub enum Access<I: Identifier> {
```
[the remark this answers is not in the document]
cmyr: Also you can derive default for this type, with

```rust
#[derive(Default)]
pub enum Access<I> {
    #[default]
    None,
    ...
}
```

which then lets you derive `Default` on the acl type as well.

fontir/src/orchestration.rs
```diff
@@ -34,83 +32,321 @@ bitflags! {
...
-                }
-            }
-            set_cached(&self.$lock_name, $restore_fn(&self.paths.target_file(&id)));
-            let rl = self.$lock_name.read();
-            rl.as_ref().expect(MISSING_DATA).clone()
+/// Courtesy of Arc this is Clone even if T isn't
+#[derive(Default)]
+pub struct ContextItem<I, T, P>
+where
+    I: Identifier,
```
cmyr, line 48: ditto: you can delete this bound and nothing breaks.

fontir/src/orchestration.rs
```diff
@@ -34,83 +32,321 @@ bitflags! {
...
+pub trait Persistable {
+    fn read(from: &mut dyn Read) -> Self;
+    fn write(&self, to: &mut dyn Write);
+}
+
+/// Reads and writes to somewhere that lives longer than processes.
+///
+/// This enables the compiler to restore state from prior executions which is
+/// crucial to incremental operation.
+pub trait PersistentStorage<I> {
```
cmyr, line 281: much clearer, thanks!

cmyr (approved):
A few more little non-blocking things, but overall this feels like a big improvement. I'm happy to see this get merged, and then we can iterate from here.

fontir/src/ir.rs
rsheeter: Having a shared glyph identifier would be nice.

rsheeter: I think this is unnecessary indeed, deleted.

fontdrasil/src/orchestration.rs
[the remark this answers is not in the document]
rsheeter: I like the Default change, so updated. I'm less a fan of the removal of Identifier; you aren't *supposed* to use access for anything that isn't an Identifier and this encodes that intent.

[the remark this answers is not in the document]
rsheeter: That I think I'd prefer to leave though we can revisit if there is some reason that's a Really Bad Idea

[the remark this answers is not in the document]
cmyr: The best explanation I can find on this convention is here: https://stackoverflow.com/a/66369912/1529922

Issue #365, opened 2023-07-11
compiling GoogleSans with fontc gets stuck in an infinite loop


anthrotype:
Running `cargo run -- --source ../googlesans/source/GoogleSans/GoogleSans.designspace` off latest `main` branch never stops.
A thread panics because of https://github.com/googlefonts/fontc/issues/349 `'no open subpath (missing MoveTo)'`, but I don't think that's the problem because even if I fix the latter issue in a WIP branch of mine (where kurbo no longer panics because all BezPaths start with valid MoveTo), fontc continues forever until I CTRL-C.

I suspect it might have to do something with #359, because if I revert that PR it no longer hangs -- in my WIP branch at least.

anthrotype:
> even if I fix the latter issue in a WIP branch of mine (where kurbo no longer panics because all BezPaths start with valid MoveTo), fontc continues forever until I CTRL-C.

correction, in my WIP branch (where the thread does not panic) fontc does complete eventually. But it takes more than 2 minutes and is almost 4 times slower than when I also revert #359.

cmyr:
will take a look at this in a little bit. :)

cmyr:
Okay I dug in a bit: my best guess right now is that we aren't really handling panics at all, so we're getting into a deadlock where some job depends on another job that can't finish.

We have a few options for fixing this. The easiest in the short-term is probably just to track the time of the last completed job, and then timeout if we don't make progress after some time. We should probably do this anyway, as a backup. More specifically, though, I think we need some way to signal a panic has occurred, probably by using rayon's `ThreadPoolBuilder` to set a custom panic handler, and maybe we can set a flag in the panic handler and check that in the main loop.

anthrotype:
> is almost 4 times slower than when I also revert https://github.com/googlefonts/fontc/pull/359

maybe this warrants its own issue. Something is going on with #359 that has slowed down things noticeably. Building GoogleSans.designspace takes 27 seconds after #359, and only 5 seconds if I revert it (timings are with fontc built in release mode and with --emit-ir=false, debug or emit-ir=true make it even slower). I am not able to figure out what in #359 would produce such a major (5x) slow-down.

PR #367, opened 2023-07-12
Read with timeout instead of blocking

cmyr:
Previously we were using a blocking read to wait for completions, which would deadlock if a worker thread had panicked.

Now we will read in a loop, and return an error if we timeout.

I've arbitrarily decided that ten seconds is a reasoanble timeout interval, but it might not be! We also may wish to make this setting configurable.

With this, we will no longer hang (#365) although I do think we should look into better handling of panics (and I'm going to poke at that a bit more, now)

### DIFF
--- a/fontc/src/error.rs
+++ b/fontc/src/error.rs
@@ -23,4 +23,8 @@ pub enum Error {
     BadRegex(#[from] regex::Error),
     #[error("Unable to proceed")]
     UnableToProceed,
+    // Not sure what the best message is, for this; it probably means that
+    // a thread has panicked?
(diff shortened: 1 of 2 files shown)


rsheeter:
I'm not a fan of adding random timeouts, tends to lead to very confusing
interactions. I suggest we install a proper panic handler and see if that
suffices

cmyr:
I tend to agree, if you see a timeout that doesn't involve IO it probably means someone made an architectural mistake.

Going to look into the more principled options, and depending on how easy that is we can decide if this is a useful stopgap.

cmyr:
superseded by #371

PR #371, opened 2023-07-13
Halt runloop when jobs panic

### LINKED ISSUE #365: compiling GoogleSans with fontc gets stuck in an infinite loop

cmyr:
This involved quite a bit of fiddling. I couldn't get rayon's built-in panic_handler to work, so I ended up rolling my own solution, where we manually catch unwinds that occur in spawned jobs.

When a panic occurs, we will set an atomic flag and return an error. We check this flag before beginning any scheduled job, and return immediately if it is set.

---

I learned a bunch in writing this patch, and got to dig into a lot of the internals of both fontc and rayon. We may at some point wish to have more control over our threadpool and runloop, but I think this is adequate for now.

I spent the better part of a day trying to figure out how rayon propogated panics, without much luck; I opened https://github.com/rayon-rs/rayon/issues/1076 to see if I'm just missing something.

I also learned a bunch about unwinding and panics in general. It would be nice if we could capture backtraces where the panic occurred, and then print them at the end as needed, but this isn't really necessary and would be a bunch more code (it involves overriding the [global panic hook](http://doc.rust-lang.org/1.70.0/std/panic/fn.set_hook.html) and then we'd have to figure out how to smuggle panics out to where we need them.

- fixes #365
- closes #367


fontc/src/workload.rs
```diff
@@ -191,7 +196,15 @@ impl<'a> Workload<'a> {
         // Async work will send us it's ID on completion
         let (send, recv) = crossbeam_channel::unbounded::<(AnyWorkId, Result<(), AnyWorkError>)>();
 
-        rayon::in_place_scope(|scope| {
+        // a flag we set if we panic
+        let abort_queued_jobs = Arc::new(AtomicBool::new(false));
+        // build a custom threadpool. we use this to ensure all threads are named
+        let threadpool = ThreadPoolBuilder::new()
+            .thread_name(|n| format!("tp{n}"))
```
rsheeter, line 203: Unless we can assign useful names I wonder how much better than thread ids this is? Not worse I suppose.

fontc/src/workload.rs
```diff
@@ -228,8 +241,38 @@ impl<'a> Workload<'a> {
...
+                        // references:
+                        // <https://doc.rust-lang.org/nomicon/exception-safety.html#exception-safety>
+                        // <https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html>
+                        let result = match std::panic::catch_unwind(AssertUnwindSafe(|| {
+                            work.exec(work_context)
+                        })) {
+                            Ok(result) => result,
+                            Err(err) => {
+                                let msg = get_panic_message(err);
+                                abort.store(true, std::sync::atomic::Ordering::Release);
```
rsheeter, line 272: Strongly recommend importing Release (`abort.store(true, Release)`) rather than writing the fully qualified name; it's uninformative noise.

rsheeter: Hm. Also are you quite sure Relaxed is insufficient? If so, why, and we should comment as much.

rsheeter (approved):
LGTM with a few minor comments

fontc/src/workload.rs
cmyr: I was finding it annoying that the `ThreadId` type didn't impl Display, and the Debug impl was longer than necessary, which made our log preambles longer. With this, we will print `tp3` instead of `ThreadId(3)`. 🤷

PR #372, opened 2023-07-13
Remove most impls of From<&T>

cmyr:
These impls fell into one of two categories: they were either wrapping a call to clone, or they were doing acting as general constructors.

In the cases where they were wrapping a clone call, I just moved the 'clone' to the callsite. In the cases where they were acting as general constructors, I moved them into `new` methods.


dfrg (approved):
I like that this makes the clones more explicit. LGTM

PR #373, opened 2023-07-19
Generate kern fea and pass to fea-rs

rsheeter:
#345 with exploration of sparseness removed. Produces fea akin to:

```
# fontc generated kerning

@public.kern1.bracketleft_R = [bracketleft];
@public.kern2.bracketleft_L = [bracketleft];
@public.kern1.bracketright_R = [bracketright];
@public.kern2.bracketright_L = [bracketright];


feature kern {
  pos bracketleft bracketright (wght=400:-300 wght=700:-150);
  pos exclam exclam (wght=400:-360 wght=700:-100);
  pos hyphen hyphen (wght=400:-150 wght=700:-50);
  pos @public.kern1.bracketleft_R exclam (wght=400:-165 wght=700:0);
  pos exclam @public.kern2.bracketright_L (wght=400:-160 wght=700:0);
  pos exclam hyphen (wght=400:20 wght=700:0);
} kern;
```

Note use of user coords as fea-rs does not support unit suffixes at time of writing.

There are a fair # of TODOs remaining, plus the variable kerning built doesn't quite work (#402). Even so, I think this moves us forward enough to be worth capturing. Update: #406 helps a lot with having variable kerning that works by not throwing away data :)


fontbe/src/features.rs
```diff
@@ -72,12 +76,153 @@ impl Display for NotSupportedError {
...
+            .deltas(&point_seqs)?
+            .into_iter()
+            .map(|(region, values)| {
+                assert!(values.len() == 1, "{} values?!", values.len());
+                (region, values[0])
+            })
+            .collect();
+
+        // Compute the default on the unrounded deltas
+        let default_value = deltas
```
anthrotype, line 174: > // Compute the default

hm I'd expect the "default" to _not_ be computed, but already defined in the input (as we discussed in https://github.com/googlefonts/fontc/pull/345)

fontbe/src/features.rs
```diff
@@ -72,12 +76,157 @@ impl Display for NotSupportedError {
...
+        values: &HashMap<BTreeMap<Tag, Fixed>, i16>,
+    ) -> Result<
+        (
+            i16,
+            Vec<(write_fonts::tables::variations::VariationRegion, i16)>,
+        ),
+        Box<(dyn StdError + 'static)>,
+    > {
+        // WARNING: this will fail if the fea location isn't also a glyph location. In time we may wish to fix that.
+        let var_model = &self.static_metadata.variation_model;
```
anthrotype, line 148: I expect this reuse of a single VariationModel for all kerning pairs to create similar issues as in the sparse glyph variations #400 

The correct way to do this is to make specific VariationModels for each set of locations, fonttools does that and caches the models for the paris referencing the same set of locations, but am not sure if the latter optimization is needed for the Rust port. I think we should prefer correctness at first and only do the memoization if the profiler confirms it's worth it.

rsheeter: SG but I would like to defer this, will file an issue.

fontbe/src/features.rs
```diff
@@ -72,12 +76,157 @@ impl Display for NotSupportedError {
     }
 }
 
+// TODO: ask Colin if we need to be threadsafe
```
cmyr, line 79: I think if rust lets you compile the code then it's fine. I'm not doing anything sneaky on my side. fea-rs itself is single-threaded.

cmyr (approved):
A couple notes inline, nothing major. :)

fontbe/src/features.rs
rsheeter: Ah good point, I forgot how good the compiler is when writing that :)

fontbe/src/features.rs
```diff
@@ -72,12 +76,157 @@ impl Display for NotSupportedError {
...
+                let user = UserLocation::from_iter(
+                    pos.iter()
+                        .map(|(tag, value)| (*tag, UserCoord::new(value.to_f64() as f32))),
+                );
+                (user.to_normalized(&self.axes), vec![*value as f64])
+            })
+            .collect();
+
+        // Only 1 value per region for our input
+        // TODO is that actually guaranteed?
```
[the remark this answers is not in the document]
rsheeter: Ty. I _thought_ so but wasn't 100% confident

fontbe/src/features.rs
rsheeter: https://github.com/googlefonts/fontc/issues/407

PR #376, opened 2023-07-26
[ufo2fontir] Fix for optional designspace names

cmyr:
The upstream API has changed to make the 'name' fields on instance/source types optional. This patch is the minimal fix, where we check all names on load and insert placeholders if they are missing.

An alternative solution might be to use something other than the name to uniquely identify these objects.

This is a draft as the upstream PR (https://github.com/linebender/norad/pull/313) has not yet been merged.

### DIFF
--- a/ufo2fontir/src/source.rs
+++ b/ufo2fontir/src/source.rs
@@ -152,8 +152,20 @@ impl DesignSpaceIrSource {
         }
     }
     fn load_designspace(&self) -> Result<DesignSpaceDocument, Error> {
-        DesignSpaceDocument::load(&self.designspace_file)
-            .map_err(|e| Error::UnableToLoadSource(Box::new(e)))
+        let mut doc = DesignSpaceDocument::load(&self.designspace_file)
+            .map_err(|e| Error::UnableToLoadSource(Box::new(e)))?;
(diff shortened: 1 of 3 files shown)


ufo2fontir/src/source.rs
```diff
@@ -696,9 +708,14 @@ impl Work<Context, WorkId, WorkError> for StaticMetadataWork {
             .instances
             .iter()
             .map(|inst| NamedInstance {
-                name: match inst.name.strip_prefix(family_prefix.as_str()) {
+                name: match inst
+                    .name
```
anthrotype, line 712: I just noticed that we are using the instance `name` attribute to set the fvar's NamedInstance's subfamilyNameID names, but this is wrong, varLib uses the `stylename` attribute of an instance, not its `name` for that purpose. The latter is similar to source.name and only meant to uniquely identify a given DS source/instance element for cross referencing within the doc (should have been called `id` perhaps but it's too late).

https://github.com/fonttools/fonttools/blob/57fb47de3a41c724c3bf1051d00715e08d86141e/Lib/fontTools/varLib/__init__.py#L103-L118

If we use `stylename` here we no longer need to do inst.name.strip_prefix(family_prefix).

anthrotype (approved):

PR #380, opened 2023-08-08
Restore fine-grained glyph work dependencies

### LINKED ISSUE #369: Investigate performance regression
> is almost 4 times slower than when I also revert https://github.com/googlefonts/fontc/pull/359

maybe this warrants its own issue. Something is going on with #359 that has slowed down things noticeably. Building GoogleSans.designspace takes 27 seconds after #359, and only 5 seconds if I revert it (timings are with fontc built in release mode and with --emit-ir=false, debug or emit-ir=true make it even slower). I am not able to figure out what in #359 would produce such a major (5x) slow-down.

_Originally posted by @anthrotype in https://github.com/googlefonts/fontc/issues/365#issuecomment-1632981997_

rsheeter:
#359 changed BE glyph work from depending on only the IR glyphs used to all IR glyphs. This forces the can run check to take a slow path (table scan) due to use of Access::Custom and inhibits concurrent execution of FE and BE glyphs.

This PR late-binds dependencies for BE glyph work. Initially access unknown - never runnable - is used and then, once the corresponding IR glyph is done, a set of the specific dependencies is used. This should be faster to evaluate runnable on and re-enable concurrent FE/BE glyph work executions.

Intended to partially address #369. 

#369 doesn't offer complete steps to reproduce but it appears to be roughly `time fontc GS designspace`. My tests and results follow (some noise from cli omitted, measured on an M1 Mac):

**Oswald 2.415s => 1.302s**

```shell
$ for br in conc main; do git checkout $br && cargo build --release && rm -rf build/ && time target/release/fontc --source ../OswaldFont/sources/Oswald.glyphs --emit-ir false; done

Switched to branch 'conc'

real	0m2.451s
user	0m0.104s
sys	0m0.019s
Switched to branch 'main'

real	0m1.302s
user	0m0.212s
sys	0m0.026s
```

Fixes #369 per https://github.com/googlefonts/fontc/pull/380#issuecomment-1669920774

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -240,15 +240,13 @@ impl Work<Context, AnyWorkId, Error> for GlyphWork {
         WorkId::GlyfFragment(self.glyph_name.clone()).into()
     }
 
+    /// We need to block on all our components, but we don't know them yet.
+    ///
+    /// We could block on ALL IR glyphs, but that triggers inefficient behavior in workload.rs.
+    /// Instead, start in a hard block and update upon success of the corresponding IR job.
+    /// See fontc, workload.rs, handle_success.
     fn read_access(&self) -> Access<AnyWorkId> {
-        Access::Custom(Arc::new(|id| {
-            matches!(
-                id,
-                AnyWorkId::Fe(FeWorkId::StaticMetadata)
-                    | AnyWorkId::Fe(FeWorkId::GlyphOrder)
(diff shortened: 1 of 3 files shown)


rsheeter:
Cloning Google Sans on airplane wifi is not going particularly well, hopefully the Oswald results will suffice for now.

anthrotype:
I just tried to build GS and with this patched, it goes down from 25 seconds to 5!

anthrotype (approved):

PR #382, opened 2023-08-08
Use fontations Glyph enum

cmyr:
A bit of cleanup on top of  #379.

 This renames the backend 'Glyph' type to 'NamedGlyph', and turns it into a simple struct, containing a `write-fonts` Glyph and the associated name.

This also lets us delete the GlyphPersistable type, since we don't need to store a flag indicating whether the glyph is simple or composite.

### DIFF
--- a/fontbe/src/orchestration.rs
+++ b/fontbe/src/orchestration.rs
@@ -27,26 +27,13 @@ use serde::{Deserialize, Serialize};
 use write_fonts::{
     dump_table,
     tables::{
-        avar::Avar,
-        cmap::Cmap,
-        fvar::Fvar,
-        glyf::{Bbox, SimpleGlyph},
-        gpos::Gpos,
-        gsub::Gsub,
-        gvar::GlyphDeltas,
-        head::Head,
-        hhea::Hhea,
-        maxp::Maxp,
-        name::Name,
-        os2::Os2,
-        post::Post,
-        stat::Stat,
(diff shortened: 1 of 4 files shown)


anthrotype (approved):

fontbe/src/orchestration.rs
```diff
@@ -127,107 +114,47 @@ impl From<WorkId> for AnyWorkId {
...
-/// <https://learn.microsoft.com/en-us/typography/opentype/spec/glyf>
+/// A glyph and its associated name
 #[derive(Debug, Clone)]
-pub enum Glyph {
-    Simple(GlyphName, SimpleGlyph),
-    Composite(GlyphName, CompositeGlyph),
+pub struct NamedGlyph {
+    pub name: GlyphName,
+    pub glyph: Glyph,
 }
```
rsheeter, line 122: NamedGlyph suggests the existence of other types but there really aren't any. I'd rather call this Glyph and rename the Glyph nested within, e.g.

```rust
/// A glyph and its associated name
///
/// See <https://learn.microsoft.com/en-us/typography/opentype/spec/glyf>
#[derive(Debug, Clone)]
pub struct Glyph {
    pub name: GlyphName,
    pub data: GlyphData,
}
```

rsheeter (approved):
LGTM with one nit re naming (the hardest part)

PR #383, opened 2023-08-08
Use fontations glyf/loca builder

cmyr:
Based on #382 

-----

This also uses the LocaFormat type in fontations (which was copied from here) which then forces us to hack around orphan rules.

Playing around with this i have a better understanding of the annoyance we encountered previously about serializing loca. The main question this patch raised for me, was whether or not we should just have a `GlyfAndLoca` struct that wraps both those tables, as they're always generated together? This would let us simplify some of our serialization code.

### DIFF
--- a/fontbe/src/orchestration.rs
+++ b/fontbe/src/orchestration.rs
@@ -28,8 +28,8 @@ use write_fonts::{
     dump_table,
     tables::{
         avar::Avar, cmap::Cmap, fvar::Fvar, glyf::Glyph as RawGlyph, gpos::Gpos, gsub::Gsub,
-        gvar::GlyphDeltas, head::Head, hhea::Hhea, maxp::Maxp, name::Name, os2::Os2, post::Post,
-        stat::Stat, variations::Tuple,
+        gvar::GlyphDeltas, head::Head, hhea::Hhea, loca::LocaFormat, maxp::Maxp, name::Name,
+        os2::Os2, post::Post, stat::Stat, variations::Tuple,
     },
     validate::Validate,
     FontWrite, OtRound,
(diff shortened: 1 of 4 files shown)


fontbe/src/orchestration.rs
```diff
@@ -248,41 +248,39 @@ impl TupleBuilder {
...
-        let mut buf = Vec::new();
-        from.read_to_end(&mut buf).unwrap();
-        match buf.first() {
-            Some(0) => LocaFormat::Short,
-            Some(1) => LocaFormat::Long,
-            _ => {
-                panic!("serialized LocaFormat is invalid")
-            }
-        }
+        bincode::deserialize_from(from).unwrap()
```
anthrotype, line 279: just curious, why do we need to use bincode to (de)serialize the loca?

anthrotype:
it does make sense to glue the two tables together, you always build them both at the same time, and can't read glyf without loca, and loca by itself is similarly useless

fontbe/src/orchestration.rs
cmyr: This isn't loca itself, but the loca format flag; we serialize it separately so that we can serialize loca in its final representation, even though we might not have the head table yet.

fontbe/src/orchestration.rs
```diff
@@ -248,41 +248,39 @@ impl TupleBuilder {
...
-#[repr(u8)]
-#[derive(Copy, Clone, Debug, PartialEq)]
-pub enum LocaFormat {
-    Short = 0,
-    Long = 1,
+// work around orphan rules.
+//
+// FIXME: Clarify if there's a good reason not to treat glyf/loca as a single
+// entity, for the purpose of persistence? like a struct that contains both
+// tables, and from which the format can be retrieved
```
rsheeter, line 255: Given that we compute them together I think this might make more sense and might be a better change. Want to try that?

rsheeter: Also I really dislike the format wrapper; hopefully use of a single entity would let us avoid that

cmyr: yea, it felt very bad writing this code. That said I think it makes sense to merge as is and open an issue, and that work can be a separate patch.

PR #387, opened 2023-08-09
must follow fvar axes order when building gvar GlyphDeltas

anthrotype:
TupleBuilder was iterating over VariationRegion's axis_tents which is a BTreeMap, but it should instead iterate over the region's tents following the fvar's axis order, that's what the gvar Tuples must follow. Otherwise we store deltas for the wrong axes!

### DIFF
--- a/fontbe/src/gvar.rs
+++ b/fontbe/src/gvar.rs
@@ -52,7 +52,9 @@ impl Work<Context, AnyWorkId, Error> for GvarWork {
         Access::Custom(Arc::new(|id| {
             matches!(
                 id,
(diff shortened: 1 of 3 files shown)


dfrg (approved):
Good catch! I wonder if `VariationRegion` should be changed to store tents in axis order rather than using a map but I see at least one place where lookup is done by tag and changing that might be more invasive so this solution looks good to me.

anthrotype:
> I wonder if VariationRegion should be changed to store tents in axis order

I wondered the same, maybe we can revise this later. I'll merge this for now

Issue #388, opened 2023-08-10
invalid gvar glyph count != font numGlyphs


anthrotype:
when building gvar table, fontc omits those glyphs that contain no variations, it even has a test for this:

https://github.com/googlefonts/fontc/blob/097ab92c30836f46abc8299526218bb89465f88c/fontbe/src/gvar.rs#L102-L122

however this is incorrect and leads to invalid gvar table that cannot be decompiled by either fonttools or read-fonts.
Gvar must contain one GlyphVariationData entry for each glyph in the font, whether this contains any variation deltas or is empty.

see gvar OT spec: https://learn.microsoft.com/en-us/typography/opentype/spec/gvar#gvar-header

> uint16 | glyphCount | The number of glyphs in this font. This must match the number of glyphs stored elsewhere in the font.

Write-fonts' Gvar::new constructor also notes this requirement in the docstring but doesn't enforce (maybe it should? but to do that it would need to know the maxp.numGlyphs..):

```
/// Variations must be present for each glyph, but may be empty.
```

https://github.com/googlefonts/fontations/blob/ed1603854e79906a48c5e056795d91099911ef11/write-fonts/src/tables/gvar.rs#L47C9-L47C69

PR #389, opened 2023-08-10
fontbe/gvar: don't skip glyphs without variations

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/388


rsheeter (approved):
LG, ty

PR #391, opened 2023-08-10
Restore the ability to produce a meaningful flamegraph

rsheeter:
Assignment of thread names causes each to root individually in flamegraph, making it essentially useless. Note the tp0, 1, etc:

image

If we don't name them the flamegraph is very useful:

image

So don't name the threads. If we *really* like thread names we can make it optional - even on by default - in a followon while also documenting how to get a good flamegraph.

JMM.

### DIFF
--- a/fontc/src/workload.rs
+++ b/fontc/src/workload.rs
@@ -1,6 +1,5 @@
 //! Tracking jobs to run
 
-use rayon::ThreadPoolBuilder;
 use std::{
     collections::{HashMap, HashSet},
     panic::AssertUnwindSafe,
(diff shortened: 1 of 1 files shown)


anthrotype (approved):

cmyr:
Huh that's funny behaviour, I wonder what the rationale is? I did a quick skim of the flamegraph docs and didn't see an option to turn off the thread grouping, so 🤷 

In any case, this was part of a patch that included one other change, to how we format log messages, and you'll want to back that out too (see https://github.com/googlefonts/fontc/commit/eadad)

PR #393, opened 2023-08-10
Preallocate some buffers based on flamegraph

rsheeter:
Before, `fontc`glyphs2fontir::toir::to_ir_contours_and_components (1 samples, 0.92%)` showing the usual vec growth fns.
After, can't find to_ir_contours_and_components in the flamegraph for GlyphIrWork at all.

Also avoid the clone in `src_path.nodes.clone()`.


cmyr (approved):
Looks reasonable!

PR #395, opened 2023-08-10
[WIP] Tweak is_nonspacing_mark_name based on flamegraph

rsheeter:


glyphs2fontir/src/glyphdata.rs
```diff
@@ -4,8 +4,705 @@
 
 use std::collections::HashSet;
 
-pub(crate) fn is_nonspacing_mark(codepoints: &HashSet<u32>, name: &str) -> bool {
-    // Try first by name
+#[inline(always)]
+pub(crate) fn might_be_a_nonspacing_mark_name(name: &str) -> bool {
```
cmyr, line 8: Can we capture the rationale and the process here, as a comment?

glyphs2fontir/src/glyphdata.rs
```diff
@@ -4,8 +4,705 @@
...
+    // shove the first 4 chars into a u32 and see if they could possibly be a nonspacing mark
+    let head = name
+        .chars()
+        .enumerate()
+        .take(4)
+        .fold(0u32, |acc, (i, ch)| acc | (ch as u32) << (i * 8));
```
cmyr, line 14: we're taking chars but then truncating them, could we just use `bytes()`?

cmyr (commented):
Okay so to check my understanding:

- we have, from fonttools, a list of the names of all possible nonspacing marks
- for all our glyph names, we need to check if they're on that list

I wonder what other approaches might work for us? One option would be a compile-time perfect hash function, using a crate such as https://docs.rs/phf/latest/phf/? Apparently this also has a codegen mode, so we wouldn't necessarily even need to bring it in as a dep.

It might be worth at least trying that, and seeing what the perf looks like?

glyphs2fontir/src/glyphdata.rs
rsheeter: Yes.

glyphs2fontir/src/glyphdata.rs
rsheeter: Added a terse comment. We also refer the reader to the generating python at the top.

rsheeter:
I have deferred exploration of phf, which does seem potentially promising, to https://github.com/googlefonts/fontc/issues/399

anthrotype (approved):

PR #403, opened 2023-08-16
Fix glyph variations for sparse and intermediate masters

### LINKED ISSUE #400: incorrect glyph variation deltas for sparse masters
fontc currently creates a single VariationModel that is initialized with the union of all the glyph locations and then uses this same model for computing the deltas of each glyph. But sparse glyphs may not contain sources for all the locations, and in those cases we need to create a sub-model like fontTools.varLib does, where the locations and regions that do not apply are omitted and the min/max partition and weights of the remaining regions are recomputed for that particular glyph's constellation of locations.
If we don't do that, we may end up associating the wrong region min/max to given glyph's locations, leading to unexpected interpolations that look weird and do not match fontTools.
Besides, fontTools keeps a caches of these submodels since it's common for several glyphs to define sources for the same set of locations, maybe we should do the same as well.

anthrotype:
Fixes #400

I'm having troubles coming up with a minimal reproducer that proves that this fix is the correct one. The changes were motivated by my attempts at building a private font of ours that contains loads of axes and masters, which resulted in incorrect variations using our current code. So I compared again the original fonttools Python code with our Rust port and noticed two major differences.

1) fonttools uses a specific sub-model for each individual glyph that contains only the locations that a glyph defines explicitly; the super-model from which the sub-model is derived contains _all_ the designspace locations (i.e. the union of all the glyphs' locations). When one calls `VariationModel.getSubModel` method with a list of masters that may contain some `None`s, a new VariationModel is built (or a cached one returned) containing only the subset of locations for which a glyph master is not `None`. The sorted order of the master locations in the sub-model ends up being the same as in the parent model; however, the absence of some locations in a sparse sub-model can produce regions with different min or max from those from the super-model. In our Rust code, we make a single VariationModel with the union of all the glyph locations and store that in the StaticMetadata and use for all the individual glyph variations. That means, for some glyphs that do not define all the locations, we end up with incorrect master supports.

2) the `_computeMasterSupports` method (called `master_influence` in our Rust port) iterates over previous regions to see if they overlap with the current one and may require splitting; but in the Python code, if the current region is modified (some of its tents' min or max may be changed) this is done _in-place_ such that successive iterations use the modified object; in our Rust code, in order to being able to, at the same time, iterate over regions, modify some of them, and also iterate over previous regions up to `i` (while not irritating the borrow checker), we resort to cloning the current region but then subsequent ones will not see those modifications...

In this PR I fix these two issues, and verified locally that that big private font of our ours finally renders correctly throughout the designspace (big win).

As I think I understand this issue better now (also thanks to @behdad's comment in #400), I will now try to work on creating a minimal reproducer.

EDIT: Reproducer is https://github.com/googlefonts/fontc/pull/404, cherry picked test onto this PR branch as https://github.com/googlefonts/fontc/pull/403/commits/a908a46cf87cdc7cc35472799749d64fdfb0e8a1


anthrotype:
the test that was failing on `main` from https://github.com/googlefonts/fontc/pull/404, now passes when applied onto this PR branch

rsheeter:
Per f2f, lets upload the .glyphs and put the derived .designspace into `resources/testdata/designspace_from_glyphs`

anthrotype:
> lets upload the .glyphs and put the derived .designspace into resources/testdata/designspace_from_glyphs

done

fontbe/src/glyphs.rs
```diff
@@ -331,6 +332,12 @@ impl Work<Context, AnyWorkId, Error> for GlyphWork {
             Error::GlyphError(ir_glyph.name.clone(), GlyphProblem::MissingDefault)
         })?;
 
+        // build a VariationModel specific to this glyph's master locations, upon which the region
+        // of influence and the delta weights associated to each master in turn depend.
+        let locations: HashSet<_> = ir_glyph.sources().keys().cloned().collect();
+        let var_model = VariationModel::new(locations, static_metadata.variable_axes.clone())
+            .map_err(|e| Error::VariationModelError(self.glyph_name.clone(), e))?;
```
rsheeter, line 339: My intuition is that we would wind up building the same variation model over and over again, that if you made a set of all glyph instance locations it's size would be far lower than the number of glyphs.

Maybe that's best kicked into a follow-on PR given this does fix a major problem for us. I will file an issue to investigate.

rsheeter: https://github.com/googlefonts/fontc/issues/415

rsheeter: Checking if the glyph uses the same set of locations as the global variation model reduces the number of models created for Oswald from 884 to 1. #415 will generalize this but for now I'm inclined to sneak that in here, will push a commit.

rsheeter (approved):
LGTM, exploration of creating less models deferred to #415

fontbe/src/glyphs.rs
rsheeter: Oops, I merged and then realized I should have let you review my changes first. Sorry about that; if you see problems I am happy to fix them in a follow-on.

PR #404, opened 2023-08-17
add test for intermediate layer/sparse source in designspace

anthrotype:
Intended to reproduce the issue described at #400, whereby we currently use the same VariationModel for all glyphs whether or not they contain sources for all the locations or only some, leading to incorrect region definitions

this is the same test font I showed in https://github.com/googlefonts/fontc/issues/400#issuecomment-1682235603

DO NOT MERGE: this is for illustration purposes only. The actual test will be cherry-picked on the #403 branch


anthrotype:
the test fails on `main` as expected:
https://github.com/googlefonts/fontc/actions/runs/5893540102/job/15985303387?pr=404#step:5:169

```
---- tests::intermediate_layer_in_designspace stdout ----
thread 'tests::intermediate_layer_in_designspace' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
 [
     Rect { origin: (231.0, 0.0), size: 133.0W×700.0H },
<    Rect { origin: (195.0, 0.0), size: 205.0W×700.0H },
>    Rect { origin: (231.0, 0.0), size: 133.0W×700.0H },
     Rect { origin: (171.0, 0.0), size: 253.0W×700.0H },
 ]
```

PR #418, opened 2023-08-24
Update to fea-rs floating like a butterfly

rsheeter:
And fix bug that revealed where the ordering required by VariationRegion was not being enforced.

DO NOT SUBMIT until using a release that contains https://github.com/cmyr/fea-rs/pull/222; should be fine to review tho. EDIT: updated to fea-rs 0.15.0.

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -83,11 +83,7 @@ impl NoIncludePathError {
     }
 }
 
-impl std::error::Error for NoIncludePathError {
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
-        None
-    }
-}
+impl std::error::Error for NoIncludePathError {}
 
 impl Display for NoIncludePathError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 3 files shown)


fontbe/src/features.rs
```diff
@@ -143,19 +143,63 @@ impl<'a> FeaVariationInfo<'a> {
...
+    fn new(loc: NormalizedLocation) -> UnsupportedLocationError {
+        UnsupportedLocationError(loc)
+    }
+}
+
+impl std::error::Error for UnsupportedLocationError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
+    }
+}
```
cmyr, line 172: this is the default impl, so you can just write `impl std::error::Error for X {}`.

fontbe/src/features.rs
```diff
@@ -143,19 +143,63 @@ impl<'a> FeaVariationInfo<'a> {
...
+
+impl Display for UnsupportedLocationError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!("No variation model for {:?}", self.0))?;
+        Ok(())
```
cmyr, line 177: more idiomatic here to use the `write!` macro,

```suggestion
        write!(f, "No variation model for {:?}", self.0)
```

and since this fn returns a result you can skip the `?` and `Ok(())` and just return the output of the macro directly

fontbe/src/features.rs
```diff
@@ -143,19 +143,63 @@ impl<'a> FeaVariationInfo<'a> {
...
+    }
+}
+
+#[derive(Debug)]
+struct MissingTentError(Tag);
```
cmyr, line 182: if these two errors are always returned in the same place (like are two error conditions of the same logic) then maybe an enum makes sense 🤷

cmyr (approved):
Couple nits inline that you are free to ignore.

fontbe/src/features.rs
rsheeter: Fixed, ty

fontbe/src/features.rs
rsheeter: Fixed

fontbe/src/features.rs
rsheeter: Maaybe ... how about we hold that idea until we add a third

PR #423, opened 2023-08-28
Generate missing .notdef like ufo2ft

### LINKED ISSUE #149: should add .notdef when missing from the source

### LINKED ISSUE #420: fontc does not cmap latin captial A in Oswald
ttx_diff as instructed in the milestone; cmap doesn't match because fontc's version has no entry for `<map code="0x41" name="A"/><!-- LATIN CAPITAL LETTER A -->`

### LINKED ISSUE #421: notdef isn't gid 0 in Oswald
ttx_diff, run per milestone instructions on the .glyphs file, reports GlyphOrder doesn't match for Oswald. This appears to be due to fontc putting .notdef in slot 702.

rsheeter:
For WghtVar.glyphs, which does not define notdef, we get this:

image

```shell
$ python resources/scripts/ttx_diff.py ../OswaldFont/sources/Oswald.glyphs
# default parameter comparison
COMPARISON
  Only fontmake produced 'HVAR'
  DIFF 'GDEF', build/default/fontc.GDEF.ttx build/default/fontmake.GDEF.ttx
  DIFF 'GPOS', build/default/fontc.GPOS.ttx build/default/fontmake.GPOS.ttx
  Identical 'GlyphOrder'
  Identical 'OS_2'
  Identical 'STAT'
  Identical 'avar'
  Identical 'cmap'
  Identical 'fvar'
  Identical 'glyf'
  Identical 'head'
  Identical 'hhea'
  Identical 'hmtx'
  Identical 'loca'
  Identical 'maxp'
  Identical 'name'
  Identical 'post'

```

Fixes #149 
Fixes #420

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -312,7 +312,7 @@ mod tests {
     use fontbe::orchestration::{
         AnyWorkId, Context as BeContext, Glyph, LocaFormatWrapper, WorkId as BeWorkIdentifier,
     };
-    use fontdrasil::types::{GlyphName, NOTDEF};
+    use fontdrasil::types::GlyphName;
     use fontir::{
         ir::{self, KernParticipant},
         orchestration::{Context as FeContext, Persistable, WorkId as FeWorkIdentifier},
@@ -1273,10 +1273,14 @@ mod tests {
             .fe_context
             .preliminary_glyph_order
             .get()
-            .contains(&NOTDEF));
+            .contains(&GlyphName::NOTDEF));
         assert_eq!(
             Some(0),
-            result.fe_context.glyph_order.get().glyph_id(&NOTDEF)
+            result
+                .fe_context
+                .glyph_order
+                .get()
+                .glyph_id(&GlyphName::NOTDEF)
         );
 
         let font_file = build_dir.join("font.ttf");
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -312,7 +312,7 @@ mod tests {
     use fontbe::orchestration::{
         AnyWorkId, Context as BeContext, Glyph, LocaFormatWrapper, WorkId as BeWorkIdentifier,
     };
-    use fontdrasil::types::GlyphName;
+    use fontdrasil::types::{GlyphName, NOTDEF};
     use fontir::{
         ir::{self, KernParticipant},
         orchestration::{Context as FeContext, Persistable, WorkId as FeWorkIdentifier},
@@ -520,12 +520,14 @@ mod tests {
                 BeWorkIdentifier::Font.into(),
                 BeWorkIdentifier::Fvar.into(),
                 BeWorkIdentifier::Glyf.into(),
+                BeWorkIdentifier::GlyfFragment(".notdef".into()).into(),
                 BeWorkIdentifier::GlyfFragment("bar".into()).into(),
                 BeWorkIdentifier::GlyfFragment("plus".into()).into(),
                 BeWorkIdentifier::Gpos.into(),
                 BeWorkIdentifier::Gsub.into(),
                 BeWorkIdentifier::Gdef.into(),
                 BeWorkIdentifier::Gvar.into(),
+                BeWorkIdentifier::GvarFragment(".notdef".into()).into(),
                 BeWorkIdentifier::GvarFragment("bar".into()).into(),
                 BeWorkIdentifier::GvarFragment("plus".into()).into(),
                 BeWorkIdentifier::Head.into(),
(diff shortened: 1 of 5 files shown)


fontdrasil/src/types.rs
```diff
@@ -10,6 +10,9 @@ use smol_str::SmolStr;
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub struct GlyphName(SmolStr);
 
+/// The name of the undefined glyph
+pub static NOTDEF: GlyphName = GlyphName(SmolStr::new_inline(".notdef"));
```
cmyr, line 14: A common pattern here is to define the const in the impl of the type itself, like,

```rust
impl GlyphName {
    pub static NOTDEF: GlyphName = ...;
}
```

and then you can reference it anywhere that `GlyphName` is in scope, as `GlyphName::NOTDEF`.

fontir/src/glyph.rs
```diff
@@ -435,6 +471,11 @@ impl Work<Context, WorkId, WorkError> for GlyphOrderWork {
         } else {
             trace!("No new glyphs, final glyph order == preliminary glyph order");
         }
+
+        let Some(0) = new_glyph_order.glyph_id(&NOTDEF) else {
+            todo!(".notdef must be present and must be gid 0")
```
cmyr, line 476: is this intentional? If we want to leave in a panic I'd write this as an `assert` instead.

fontir/src/ir.rs
```diff
@@ -975,6 +986,59 @@ impl GlyphBuilder {
...
+        path.move_to((x_min, y_min));
+        path.line_to((x_min, y_max));
+        path.line_to((x_max, y_max));
+        path.line_to((x_max, y_min));
+        path.line_to((x_min, y_min));
+        path.close_path();
+
+        Self {
+            name: NOTDEF.clone(),
+            codepoints: HashSet::from([0]),
```
cmyr, line 1031: I'd double-check with @anthrotype but I'm not sure .notdef has any codepoints. Some fonts include a NUL glyph in addition to notdef, although I'm not sure about (and am curious to learn more!) the semantics/rationale for this.

cmyr (approved):
A couple little notes but looks good, happy to see this addition :)

fontir/src/glyph.rs
rsheeter: oh lol that's now guaranteed; this is leftover exporation that I will delete

fontir/src/ir.rs
rsheeter: Oswald cmap and glyphorder are now identical so I think this is right. I will double-check with Cosimo just in case it's only _sometimes_ right or not as simplistic as I made it. A lot of these little details turn out to be fiddly.

fontdrasil/src/types.rs
rsheeter: Much nicer, updated so.

fontc/src/lib.rs
```diff
@@ -1018,11 +1021,12 @@ mod tests {
                 .collect();
             assert_eq!(
                 vec![
-                    (0x002C, 1),
-                    (0x002E, 0),
-                    (0x0030, 2),
-                    (0x031, 3),
-                    (0x032, 4)
+                    (0x0000, 0),
```
anthrotype, line 1024: hm .notdef should not be mapped to anything, 0x0000 is the NULL character which is something else, a control character that is no longer required to be defined (there used to be a recommendation in the TrueType spec about the first 4 glyphs which was removed later on, so nowadays a font no longer needs to have a glyph for NULL 0x0000 character).

fontc/src/lib.rs
```diff
@@ -1259,6 +1263,37 @@ mod tests {
...
+        );
+
+        let font_file = build_dir.join("font.ttf");
+        assert!(font_file.exists());
+        let buf = fs::read(font_file).unwrap();
+        let font = FontRef::new(&buf).unwrap();
+
+        assert_eq!(
+            GlyphId::new(0),
+            font.cmap().unwrap().map_codepoint(0u32).unwrap()
```
anthrotype, line 1293: again, this is wrong, NULL character (0x0000) should not be mapped to ".notdef" glyph, i.e. the first glyph (gid=0) in the font. They are different things.

fontc/src/lib.rs
```diff
@@ -1280,9 +1315,10 @@ mod tests {
                 GlyphId::new(0),
                 GlyphId::new(1),
                 GlyphId::new(2),
-                GlyphId::new(5),
+                GlyphId::new(3),
+                GlyphId::new(6),
             ],
-            [0x20, 0x21, 0x2d, 0x3d]
+            [0x00, 0x20, 0x21, 0x2d, 0x3d]
```
anthrotype, line 1321: nope, {0x00: GlyphId(0)} should be absent from the cmap

fontir/src/ir.rs
anthrotype: Colin was right, the `codepoints` for the generated .notdef glyph must be left empty, .notdef should be understood as the absence of a character mapping; 0x0000 is a different unicode codepoint denoting the NULL character (like in NULL-terminated C strings), it used to be among the recommended first 4 glyphs in old TrueType but current OT spec no longer does that and modern fonts don't contain NULL (and neither the carriage return CR character), but they all contain .notdef (un-cmapped) glyph as the first one (index=0).

PR #424, opened 2023-08-29
Prove no outlines take no glyf space

### LINKED ISSUE #378: Glyphs with no contours should not be written.
Currently if we have an empty simple glyph, we write the glyph header to the `glyf` table. In this case we should be writing nothing in the `glyf` table, and just adding a duplicate of the last entry in `loca`.

Although not part of the spec, it might be nice if we had a way of representing the empty glyph in read/write fonts? See https://github.com/googlefonts/fontations/issues/561

              In fontc, currently, if a glyph contains no outlines we still write the header.

_Originally posted by @cmyr in https://github.com/googlefonts/fontations/issues/559#issuecomment-1665984699_

rsheeter:
#378 reports no contours should take no space. Evidently that has been made true already so add a test to make sure it stays true.

Fixes #378.

JMM

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -1756,4 +1756,29 @@ mod tests {
             ]
         );
     }
+
+    #[test]
+    fn empty_glyph_is_zero_length() {
+        let temp_dir = tempdir().unwrap();
+        let build_dir = temp_dir.path();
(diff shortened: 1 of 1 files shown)


cmyr (approved):
👍

PR #427, opened 2023-09-05
Fix MissingAxisTentError when building Oswald

### LINKED ISSUE #426: compiling Oswald fails with 'Missing a tent for wdth'
As I noted in https://github.com/googlefonts/fontc/issues/422#issuecomment-1706310342, if I try to run fontc off main HEAD (currently at https://github.com/googlefonts/fontc/commit/23ce03f24d76039bc2f175812b1412c09b03be1f) on OswaldFont/sources/Oswald.glyphs, I'm getting loads of `failed to compute deltas: 'Missing a tent for wdth'`

I think I know what's going on, will send a PR soon.

JMM

anthrotype:
Fixes #426 

I first push a reproducer to demostrate the bug, then follow up with the fix


anthrotype:
fontbe test failed as expected

```
thread 'features::tests::resolve_kern' panicked at 'called `Result::unwrap()` on an `Err` value:
MissingTentError(Tag(wdth))', fontbe/src/features.rs:681:14
```

PR #428, opened 2023-09-05
leave generated .notdef glyph unmapped in cmap

anthrotype:
.notdef glyph represent the absence of a character, so it should not be explicitly mapped in the cmap. Currently we were adding a no-op map from 0x0000 (the NULL character) to glyph index 0 (.notdef) but this is unnecessary.

https://learn.microsoft.com/en-us/typography/opentype/spec/cmap#overview

> Regardless of the encoding scheme, character codes that do not correspond to any glyph in the font should be mapped to glyph index 0. The glyph at this location must be a special glyph representing a missing character, commonly known as .notdef.

NULL is for something else entirely. Basically _all_ character codepoints that are not suppoted in a given font (including NULL in the particular case of the test font as well as most font nowadays) get _implicitly_ mapped to glyph id 0, aka ".notdef" glyph.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -1021,7 +1021,6 @@ mod tests {
                 .collect();
             assert_eq!(
                 vec![
-                    (0x0000, 0),
                     (0x002C, 2),
                     (0x002E, 1),
                     (0x0030, 3),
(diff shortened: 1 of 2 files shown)


anthrotype:
FWIW, I verified with ttx_diff.py that Oswald's cmap is unchanged and matches fontmake even after this PR.

dfrg (approved):
Good catch. Lgtm

PR #431, opened 2023-09-06
Rename static metadata fields for clarity

rsheeter:
Based on observed confusion, rename the fields that hold axes and update comments. Key change is in `fontir/src/ir.rs`, the rest is just refactoring updating things.

Note that I am deliberately not fixing the clippy error about datetimes, will do that in it's own PR

JMM


cmyr (approved):
👍

PR #439, opened 2023-09-14
Script generating designspaces from .glyphs

rsheeter:
Why readme when you can shell

Also ran script and included updated files, notably including a designspace with anchors

### DIFF
--- a/resources/testdata/designspace_from_glyphs/make_designspace.sh
+++ b/resources/testdata/designspace_from_glyphs/make_designspace.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -o nounset
+set -o errexit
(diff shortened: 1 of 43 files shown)


resources/testdata/designspace_from_glyphs/make_designspace.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -o nounset
+set -o errexit
+set -x
+
+function generate_designspace() {
+    rm -rf variable_ttf/ master_ufo/
```
anthrotype, line 8:
```suggestion
    rm -rf master_ufo/
```

anthrotype (approved):

PR #440, opened 2023-09-14
Capture anchors in glyphs and ufo 2 fontir

rsheeter:
Anchors are very helpful if you want ot eventually build marks.

An unfortunate implication of anchors living in glyphs is features will end up unable to start until all FE glyphs are done. This is hard to avoid given that the anchors live with the glyphs in both major source formats.

It would be possible to just bundle anchors with glyph ir but for now I didn't.

### DIFF
--- a/fontir/src/paths.rs
+++ b/fontir/src/paths.rs
@@ -17,7 +17,7 @@ pub struct Paths {
 impl Paths {
     pub fn new(build_dir: &Path) -> Paths {
         let build_dir = build_dir.to_path_buf();
-        let anchor_ir_dir = build_dir.join("glyph_ir");
+        let anchor_ir_dir = build_dir.join("anchor_ir");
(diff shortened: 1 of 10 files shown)


fontir/src/paths.rs
```diff
@@ -9,17 +9,20 @@ use crate::orchestration::WorkId;
...
     build_dir: PathBuf,
+    anchor_ir_dir: PathBuf,
     glyph_ir_dir: PathBuf,
     ir_input_file: PathBuf,
 }
 
 impl Paths {
     pub fn new(build_dir: &Path) -> Paths {
         let build_dir = build_dir.to_path_buf();
+        let anchor_ir_dir = build_dir.join("glyph_ir");
```
anthrotype, line 20: anchor_ir?

rsheeter: Fixed, ty

anthrotype (approved):

PR #443, opened 2023-09-17
Add a visualization of thread activity

rsheeter:
How might we understand what our threads are up to? - it's hard to tell from most profiles. By visualizing it ourselves of course! First hacky draft:

image

Above each row is a thread and time goes left to right. The pink area (color/label strategy tbd) shows very fine grained FE glyph work being distributed among threads, this is what we want. Next we see a period where nothing can be done until we have final glyph order, understandable but not what we want. We can see any speedup to glyph order production is a big win as it's delaying everything. Next we see a lot of per-glyph BE work in light blue, then a period where fea-rs is the long pole as we push toward completion. 

We can see that kern, which is executable as soon as glyph order is done and blocks backend fea, doesn't start for a quite a while which can work out poorly:

image

https://github.com/rayon-rs/rfcs/blob/master/accepted/rfc0001-scope-scheduling.md#current-behavior-per-thread-lifo suggests the order you submit to rayon can matter a lot. When I test with Oswald it seems kern is submitted with a large batch of backend glyph work and often sits in spot 600+ in the batch. If I hack to move it to the front it executes almost immediately after glyph order:

image

^ is significantly better; we will want to land some sort of prioritization. However, while this works for kern because it becomes executable at the same time as a large set of things that block on glyph order it doesn't work for fea so we still end up with fea not starting until all the be glyphs are done and then spending time as the only active task; if it had started as soon as kern finished we'd be in better shape:

image

Plenty of room for refinement but already seems useful so I will try to prepare a less hacky version to land.

### DIFF
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -31,6 +31,12 @@ pub struct Args {
     #[arg(long, default_value = "false")]
     pub flatten_components: bool,
 
+    /// Whether to write a visualization of threadpool execution of tasks.
+    ///
+    /// See <https://github.com/googlefonts/fontc/pull/443>
+    #[arg(long, default_value = "false")]
+    pub emit_threads: bool,
+
     /// Working directory for the build process. If emit-ir is on, written here.
     #[arg(short, long, default_value = "build")]
     pub build_dir: PathBuf,
@@ -53,6 +59,7 @@ impl Args {
         flags.set(Flags::EMIT_DEBUG, self.emit_debug);
         flags.set(Flags::PREFER_SIMPLE_GLYPHS, self.prefer_simple_glyphs);
         flags.set(Flags::FLATTEN_COMPONENTS, self.flatten_components);
+        flags.set(Flags::EMIT_THREADS, self.emit_threads);
 
         flags
     }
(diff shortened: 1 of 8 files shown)


rsheeter:
Impact of long poles is more significant the more cores you have, on a 72-way system:

image

fontc/src/args.rs
```diff
@@ -31,6 +31,12 @@ pub struct Args {
     #[arg(long, default_value = "false")]
     pub flatten_components: bool,
 
+    /// Whether to write a visualization of threadpool execution of tasks.
+    ///
+    /// See <https://github.com/googlefonts/fontc/pull/443>
+    #[arg(long, default_value = "false")]
+    pub emit_threads: bool,
```
anthrotype, line 38: maybe `emit_timing` would be a clearer name for this option? Of course the timing is per-thread since we run tasks in parallel..

anthrotype (approved):
This is great! 👍

fontc/src/args.rs
rsheeter: Naming is the hardest part :D I like your naming better, will update.

PR #446, opened 2023-09-18
add --compile-features=false to disable FEA compilation

### LINKED ISSUE #430: Ability to disable features
In https://github.com/googlefonts/googlesans/issues/586#issuecomment-1706365071 @madig asked "Is there a way to tell fontc to not bother with features in general? They can be done with the current Python stack." 

I like the idea, lets add a flag?

anthrotype:
Fixes #430

Adds a --compile-features={true,false} bool flag (true by default) to optionally skip OT layout features compilation.

I need to add tests and convince myself the new option is wired up in all the places it needs be, but judging from casual local testing it seems to produce the desired effect, i.e. optionally skip BE job for FEA compilation, observe font contains no GDEF/GPOS/GSUB tables.

### DIFF
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -38,6 +38,10 @@ pub struct Args {
     /// Glyph names must match this regex to be processed
     #[arg(short, long, default_value = None)]
     pub glyph_name_filter: Option<String>,
+
+    /// Set to false to skip compilation of OpenType Layout features
+    #[arg(long, default_value = "true", action = ArgAction::Set)]
+    pub compile_features: bool,
 }
 
 impl Args {
@@ -68,6 +72,7 @@ impl Args {
             build_dir: build_dir.to_path_buf(),
             prefer_simple_glyphs: Flags::default().contains(Flags::PREFER_SIMPLE_GLYPHS),
             flatten_components: Flags::default().contains(Flags::FLATTEN_COMPONENTS),
+            compile_features: true,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


fontc/src/args.rs
```diff
@@ -38,6 +38,10 @@ pub struct Args {
     /// Glyph names must match this regex to be processed
     #[arg(short, long, default_value = None)]
     pub glyph_name_filter: Option<String>,
+
+    /// Set to false to skip compilation of OpenType Layout features
+    #[arg(long, default_value = "true", action = ArgAction::Set)]
+    pub compile_features: bool,
```
cmyr, line 44: total bikeshed, but would it make more sense to have this be `--no-compile-features` or `--`--no-layout` or something? basically having it false by default and then the presence of the flag disables? just feels *slightly* simpler.

cmyr (approved):
looks good with one optional idea, feel free to merge either way.

fontc/src/args.rs
anthrotype: I've followed the pattern we used for --emit-ir=false and --prefer-simple-glyphs=false (also default=true bool options), I don't remember the exact details but we were having problems at the time to accomplish exactly what you suggested and ended up doing tihs.. @rsheeter may remember

cmyr: okay that makes sense, thanks for the rationale :)

anthrotype: maybe this https://github.com/googlefonts/fontc/issues/224#issuecomment-1490600494

PR #450, opened 2023-09-20
Improve plist ergonomics

cmyr:
This is an attmept to smooth over some papercuts, motivated by reviewing the glyphspackage PR.

- This makes the error type an actual error type
- adds type aliases for Dictionary/Array
- Adds simple methods for trying to cast from a generic plist enum to dict/array/string, returning a useable error if the types are wrong


anthrotype (approved):
i like this, thanks!

PR #452, opened 2023-09-22
glyphs-reader: ignore nodes userData

### LINKED ISSUE #449: Error parsing Inter.glyphs when nodes have userData
Glyphs nodes can have a 'userData' dictionary as their fourth element:

https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md

right now fontc's glyphs-reader asserts the node arrays contain only three elements and fail otherwise.

E.g. take Inter.glyphs from https://github.com/rsms/inter

```
$ fontc -s src/Inter-Roman.glyphspackage
thread 'main' panicked at 'Invalid node content Array([Integer(96), Integer(1490), String("l"), Dictionary({"name": String("hr00")})])', glyphs-reader/src/font.rs:435:21
```

anthrotype:
Fixes #449 

Turns out that for compilation we don't actually need to parse and hold on to these optional userData dictionaries that may be associated with Nodes. They are likely the result of conversion from UFOs.
If in the future we find a use case for them it's easy to add them.

Pushing the repro test files first, will follow up with the actual fix

JMM


anthrotype:
Rod gave his LGTM via chat

Issue #458, opened 2023-09-25
glyph order is surprisingly slow and blocks everything


rsheeter:
#443 shows that the glyph order task blocks the entire system because:

1. It can't run until FE work is complete
1. It blocks all BE work

PR #462, opened 2023-09-26
Move consistent transform computation to glyph and avoid allocation

rsheeter:
As suggested by #458.

```shell
$ rm -rf build/ perf.data && cargo flamegraph -p fontc --  --source ../OswaldFont/sources/Oswald.glyphs --emit-ir false --emit-timing
$ perf script | ../FlameGraph/stackcollapse-perf.pl | grep GlyphOrderWork | ../FlameGraph/flamegraph.pl > glyphorder-flames.svg
```

Causes has_consistent_2x2_transforms to vanish from a flamegraph zoomed in on GlyphOrder; this is before, after just doesn't have that section:

image

In the `--emit-timing` thread view glyphorder goes from 2.34% to 1.35% (subject to noise, but consistently less)

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -44,32 +44,6 @@ fn has_components_and_contours(glyph: &Glyph) -> bool {
         .any(|inst| !inst.components.is_empty() && !inst.contours.is_empty())
 }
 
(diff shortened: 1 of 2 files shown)


behdad:
LGTM.

Issue #454, opened 2023-09-24
feature properties


schriftgestalt:
in .glyphs files, features (and classes and prefixes) have more properties than currently supported. 
This is a feature in format 2
```
features = (
{
automatic = 1;
code = "lookupflag IgnoreMarks;\012sub f f i by f_f_i;\012";
disabled = 1;
name = liga;
}
);
```
and the same in format 3:
```
features = (
{
automatic = 1;
code = "lookupflag IgnoreMarks;
sub f f i by f_f_i;
";
disabled = 1;
tag = liga;
}
);
```

The most important thing here is the `disabled` property. It means the feature should be ignored when compiling the fea code. Often there are some version of a feature that is kept for reference but is not compilable any more.

(The code checks for `"# automatic\n"` but that is not Glyphs syntax. That is a way to store the `automatic` property in .ufos. So I don’t think it is really of concern for fontc unless you plan to regenerate the fea code.)

And when we are at it, this is the syntax to store feature names:
```
{
code = "sub a by a.double;";
labels = (
{
language = dflt;
value = "double story a";
},
{
language = DEU;
value = "Doppelstöckiges a";
}
);
tag = ss01;
}
```

rsheeter:
Thank you, compiling disabled features seems like it could be a really nasty non-obvious source of bugs.

cmyr:
@rsheeter confirm this was fixed by the PR above?

rsheeter:
We should support disabled. I believe we don't yet support the feature names part.

rsheeter:
Then I believe we can close :)

anthrotype:
well, not quite close. We should do the conversion from Glyphs.app feature labels => featureNames FEA block in order for fea-rs to be able to compile them

anthrotype:
fixed by https://github.com/googlefonts/fontc/pull/969

PR #463, opened 2023-09-26
Filter out disabled features in glyphs files

rsheeter:
Per https://github.com/googlefonts/fontc/issues/454

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -70,11 +70,21 @@ pub struct Font {
 }
 
 #[derive(Debug, PartialEq, Eq, Hash)]
-pub struct FeatureSnippet(String);
+pub struct FeatureSnippet {
+    pub content: String,
+    pub disabled: bool,
+}
 
 impl FeatureSnippet {
-    pub fn as_str(&self) -> &str {
-        self.0.as_str()
+    pub fn new(content: String, disabled: bool) -> Self {
+        FeatureSnippet { content, disabled }
+    }
+
+    pub fn str_if_enabled(&self) -> Option<&str> {
+        match self.disabled {
+            true => None,
+            false => Some(&self.content),
+        }
     }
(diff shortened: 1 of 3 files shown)


glyphs-reader/src/font.rs
```diff
@@ -70,11 +70,21 @@ pub struct Font {
...
 
 impl FeatureSnippet {
-    pub fn as_str(&self) -> &str {
-        self.0.as_str()
+    pub fn new(content: String, disabled: bool) -> Self {
+        FeatureSnippet { content, disabled }
+    }
+
+    pub fn str_if_enabled(&self) -> Option<&str> {
+        match self.disabled {
```
dfrg, line 84: Maybe `self.disabled.then_some(&self.content)`?

glyphs-reader/src/font.rs
```diff
@@ -1397,6 +1408,10 @@ impl RawFeature {
             ))),
         }
     }
+
+    fn disabled(&self) -> bool {
+        self.disabled.map(|v| v == 1).unwrap_or_default()
```
dfrg, line 1413: `self.disabled == Some(1)` looks a bit more clear to me.

dfrg (approved):
LGTM with a couple suggestions

glyphs-reader/src/font.rs
dfrg: Or, more correctly: `(!self.disabled).then_some…`

glyphs-reader/src/font.rs
rsheeter: Agreed, updated

glyphs-reader/src/font.rs
rsheeter: You get a str when it's false so it would be (!self.disabled).then_some I suppose. It's less noisy, will update.

PR #467, opened 2023-09-28
Parse plists directly into the target structures instead of building a graph of enums first

rsheeter:
Generate a plist => structure parser instead of producing a deep Plist::whatever graph and then traversing it. Use Result.

The interesting part is in `glyphs-reader/plist_derive/src/`.

Motivations:

1. All code that directly interacts with Plist enums is nightmareish to both read and write
1. We're doing unnecessary work to produce a Plist we don't actually want


cmyr:
Maybe we should just implement the serde `Deserializer` trait?

rsheeter:
> Maybe we should just implement the serde Deserializer trait?

I _think_ this will be exciting because of the nuances of the .glyphs format. When I've found more time to bang on this I'll have a better sense of that.

glyphs-reader/src/plist.rs
```diff
@@ -472,6 +506,387 @@ impl From<BTreeMap<String, Plist>> for Plist {
...
+        if tokenizer.consume(delim.end) {
+            return Ok(list);
+        }
+        if !tokenizer.consume(delim.sep) {
+            return Err(Error::ExpectedChar(delim.sep as char));
+        };
+    }
+}
+
+pub(crate) fn parse_vec<T>(tokenizer: &Tokenizer<'_>) -> Result<Vec<T>, crate::plist::Error>
```
cmyr, line 708: any reason not to have these two functions be methods on `Tokenizer`?

glyphs-reader/src/plist.rs
```diff
@@ -472,6 +506,387 @@ impl From<BTreeMap<String, Plist>> for Plist {
...
+impl TryFrom<&Tokenizer<'_>> for Option<String> {
+    type Error = Error;
+
+    fn try_from(tokenizer: &Tokenizer<'_>) -> Result<Self, Self::Error> {
+        let val: String = tokenizer.try_into()?;
+        Ok(Some(val))
+    }
+}
+
+impl TryFrom<&crate::plist::Tokenizer<'_>> for Vec<String> {
```
cmyr, line 865: As you've noted we should definitely have blanket impls for `Vec` and `Option`, and this should definitely be possible.

glyphs-reader/plist_derive/src/attrs.rs
```diff
@@ -26,3 +29,66 @@ impl FieldAttrs {
...
 }
+
+static IGNORE: &str = "ignore";
+static KEY: &str = "key";
+static FROM_PLIST: &str = "fromplist";
+static ALT_NAME: &str = "alt_name";
+static DOC: &str = "doc";
+
+#[derive(Clone, Default, Debug)]
+pub(crate) struct TryFromTokenizerFieldAttrs {
```
cmyr, line 40: so presumably we're keeping both of these traits?

glyphs-reader/src/font.rs
```diff
@@ -125,129 +231,342 @@ struct RawFont {
...
+    Int(i64),
+    String(String),
+    Axes(Vec<Axis>),
+    AxesMappings(Vec<AxisMapping>),
+    AxisLocations(Vec<AxisLocation>),
+    GlyphOrder(Vec<String>),
+}
+
+/// Hand-parse these because they take multiple shapes
+impl TryFrom<&Tokenizer<'_>> for CustomParameters {
```
cmyr, line 310: we *could* make our derive work on enums, if we wanted..

glyphs-reader/src/plist.rs
```diff
@@ -472,6 +506,387 @@ impl From<BTreeMap<String, Plist>> for Plist {
...
+//     }
+// }
+
+pub(crate) struct VecDelimiters {
+    start: u8,
+    end: u8,
+    sep: u8,
+}
+
+pub(crate) const CSV_IN_PARENS: VecDelimiters = VecDelimiters {
```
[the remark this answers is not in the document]
rsheeter: Good catch, done. For context I think originally it was just a tuple and I missed moving it when I made a struct.

glyphs-reader/src/plist.rs
```diff
@@ -472,6 +506,387 @@ impl From<BTreeMap<String, Plist>> for Plist {
...
+
+    pub(crate) fn consume(&self, delim: u8) -> bool {
+        let Some(idx) = Token::expect(self.content, *self.idx.borrow(), delim) else {
+            return false;
+        };
+        *self.idx.borrow_mut() = idx;
+        true
+    }
+
+    pub(crate) fn skip_rec(&self) -> Result<(), Error> {
```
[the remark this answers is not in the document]
rsheeter: The name mirrors the existing parse_rec, added docs.

glyphs-reader/src/plist.rs
rsheeter: That would be better!

glyphs-reader/src/plist.rs
rsheeter: I agree but my on-plane attempt did not compile. Let me tidy up per your other comments nad then I may ping you to help figure out why not if it still won't cooperate.

rsheeter:
@cmyr agreed re TryFrom. Will fix. I didn't modify FromPlist because I needed both to coexist but now I don't so I can reuse the name.

glyphs-reader/plist_derive/src/attrs.rs
rsheeter: ...lets not

glyphs-reader/src/plist.rs
rsheeter: After cleanup blanket impl is fine. I don't entirely understand why it wasn't before but I don't have infinite time to pursue that.

glyphs-reader/src/font.rs
```diff
@@ -66,7 +61,114 @@ pub struct Font {
     pub date: Option<String>,
 
     // master id => { (name or class, name or class) => adjustment }
-    pub kerning_ltr: BTreeMap<String, BTreeMap<(String, String), i32>>,
+    pub kerning_ltr: Kerning,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Hash)]
+pub struct Kerning(BTreeMap<String, BTreeMap<(String, String), i32>>);
```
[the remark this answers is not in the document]
rsheeter: Done.

glyphs-reader/src/plist.rs
```diff
@@ -472,6 +506,387 @@ impl From<BTreeMap<String, Plist>> for Plist {
...
+    pub fn new(content: &'a str) -> Tokenizer<'a> {
+        Tokenizer {
+            content,
+            idx: RefCell::new(0),
+        }
+    }
+
+    pub(crate) fn peek(&self) -> Result<Token<'a>, Error> {
+        let (tok, _) = Token::lex(self.content, *self.idx.borrow())?;
+        //eprintln!("{tok:?}");
```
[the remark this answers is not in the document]
rsheeter: Fixed some of them

glyphs-reader/src/font.rs
rsheeter: My gut says to leave it until another example pops up, this is currently a nasty unique case.

glyphs-reader/src/font.rs
```diff
@@ -66,7 +60,114 @@ pub struct Font {
...
+        *self
+            .0
+            .entry(master_id)
+            .or_default()
+            .entry((lhs_class_or_group, rhs_class_or_group))
+            .or_default() = kern as i32;
+    }
+}
+
+/// Hand-parse because it's a bit weird
```
anthrotype, line 99: curious, what is weird about kerning dictionary that requires it to be hand-parsed?

glyphs-reader/plist_derive/src/attrs.rs
```diff
@@ -1,26 +1,60 @@
...
-}
+static IGNORE: &str = "ignore";
+static KEY: &str = "key";
+static ALT_NAME: &str = "alt_name";
+static DOC: &str = "doc";
 
-static REST: &str = "rest";
-static DEFAULT: &str = "default";
+#[derive(Clone, Default, Debug)]
+pub(crate) struct FromPlistFieldAttrs {
```
cmyr, line 12: if we now only have the one item let's just keep the old name?

glyphs-reader/plist_derive/src/lib.rs
```diff
@@ -2,157 +2,102 @@ extern crate proc_macro;
...
-    Ok(proc_macro::TokenStream::from(expanded))
-}
-
-fn add_deser(input: &DeriveInput) -> syn::Result<TokenStream> {
+fn add_fieldcases(input: &DeriveInput) -> syn::Result<TokenStream> {
     let Data::Struct(data) = &input.data else {
         return Err(syn::Error::new(
             input.ident.span(),
-            "FromPlist only supports structs",
+            "TryFromPlistString only supports structs",
```
cmyr, line 59: Should this go back to the original?

glyphs-reader/src/font.rs
```diff
@@ -8,16 +8,18 @@ use std::borrow::Cow;
...
-use log::{debug, log_enabled, trace, warn};
+use log::{debug, warn};
 use ordered_float::OrderedFloat;
+use plist_derive::FromPlist;
 use regex::Regex;
 
 use crate::error::Error;
-use crate::from_plist::FromPlist;
-use crate::plist::{Array, Dictionary, Plist};
+use crate::plist::{try_parse_plist, Plist, Token, Tokenizer, VecDelimiters};
```
cmyr, line 22: not really this PR's fault but the imports in this file are a mess 🤷

glyphs-reader/src/font.rs
```diff
@@ -66,7 +60,114 @@ pub struct Font {
...
+        if !tokenizer.consume(b'{') {
+            return Err(Error::ExpectedOpenBrace);
+        }
+
+        loop {
+            if tokenizer.consume(b'}') {
+                break;
+            }
+
+            // master_id = {
```
cmyr, line 114: delete?

glyphs-reader/src/font.rs
```diff
@@ -125,129 +230,340 @@ struct RawFont {
...
+
+    fn string(&self, name: &str) -> Option<&str> {
+        let Some(CustomParameterValue::String(str)) = self.0.get(name) else {
+            return None;
+        };
+        Some(str)
+    }
+
+    fn axes(&self) -> Option<&Vec<Axis>> {
+        let Some(CustomParameterValue::Axes(axes)) = self.0.get("Axes") else {
```
cmyr, line 269: not being familiar with the format, I have to do a lot of reading between the lines to figure out what's probably going on here; I presume in some cases there can be arbitrary keys/values, but in other cases there are defined keys that have expected types? Is there any reason that `CustomParameters` doesn't have named fields for those items?

glyphs-reader/src/font.rs
```diff
@@ -125,129 +230,340 @@ struct RawFont {
...
+            return None;
+        };
+        Some(axes)
+    }
+
+    fn axis_mappings(&self) -> Option<&Vec<AxisMapping>> {
+        let Some(CustomParameterValue::AxesMappings(mappings)) = self.0.get("Axis Mappings") else {
+            return None;
+        };
+        Some(mappings)
```
cmyr, line 279: just an observation, but as written this is very... loose feeling. I assume that it would be Quite Bad if this key existed but had a different type? If that is true should we just unwrap? Ideally we would be validating during parse time, and then it would be an invariant that if this key exists the type is correct.

glyphs-reader/plist_derive/src/lib.rs
```diff
@@ -2,157 +2,102 @@ extern crate proc_macro;
...
-        impl crate::from_plist::FromPlist for #name {
-            fn from_plist(plist: crate::plist::Plist) -> Self {
-                let mut map = plist.expect_dict().unwrap();
-                #name {
-                    #deser
+        impl FromPlist for #name {
+            fn parse(tokenizer: &mut Tokenizer<'_>) -> Result<Self, crate::plist::Error> {
+                use crate::plist::Error;
+
+                if !tokenizer.consume(b'{') {
```
cmyr, line 26: although I agree it is not worth the change at this point, I do think that having direct parsing code in all of these impls is suboptimal.

glyphs-reader/src/plist.rs
```diff
@@ -472,6 +505,289 @@ impl From<BTreeMap<String, Plist>> for Plist {
...
+    }
+    Err(Error::ExpectedNumber)
+}
+
+/// This type can be parsed from a Plist string
+pub trait FromPlist
+where
+    Self: Sized,
+{
+    fn parse(tokenizer: &mut Tokenizer<'_>) -> Result<Self, Error>;
```
cmyr, line 534: you can actually completely omit the `<'_>` here, which would save a little bit of noise.

glyphs-reader/src/plist.rs
```diff
@@ -472,6 +505,289 @@ impl From<BTreeMap<String, Plist>> for Plist {
...
+        Ok(Some(tokenizer.parse()?))
+    }
+}
+
+pub fn try_parse_plist<T>(plist: &str) -> Result<T, Error>
+where
+    T: FromPlist,
+{
+    Tokenizer::new(plist).parse()
+}
```
cmyr, line 560: An alternative to this would be to have an extra method on the `FromPlist` trait with an auto impl, like...

```rust
impl FromPlist: Sized {
    fn parse(tokenizer: &mut Tokenizer) -> Result<Self, crate::plist::Error> {
         ...
     }

    fn parse_plist_str(plist: &str) -> Result<Self, Error> {
        Tokenizer::new(plist).parse()
    }
}
```

And then you could just invoke it with, say, `String::parse_plist_str(&my_plist)`, without needing to think about another import.

cmyr (approved):
A bunch of little notes and nits but nothing major, happy to see this get in!

glyphs-reader/plist_derive/src/attrs.rs
rsheeter: done

glyphs-reader/plist_derive/src/lib.rs
rsheeter: The bulk of it is struct.field = parse value; I'm not sure how to avoid that. We could migrate the surrounding boilerplate out I suppose. Leaving it alone for now.

glyphs-reader/plist_derive/src/lib.rs
rsheeter: Yes, ty.

glyphs-reader/src/font.rs
rsheeter: It's not a structure with named fields and consistent form, e.g. kern.left_side = blah, we use blah as the field name.

glyphs-reader/src/font.rs
rsheeter: Named fields would be nicer, will fix in a follow-on. Initially I had in mind it was mostly bool/string/w/e with arbitrary name/values but in looking again a lot of it is not.

glyphs-reader/src/font.rs
rsheeter: That's fair, will fix.

glyphs-reader/src/plist.rs
rsheeter: Done

glyphs-reader/src/plist.rs
rsheeter: I like it, making it so

rsheeter:
Quick perf comparison

```shell
$ cargo build --release && hyperfine --warmup 3 --runs 250 --prepare 'rm -rf build/' 'target/release/fontc --source ../OswaldFont/sources/Oswald.glyphs --emit-ir false'
# main Time (mean ± σ):     154.8 ms ±   8.2 ms
# p2s Time (mean ± σ):     132.7 ms ±   6.4 ms
```

PR #470, opened 2023-09-28
skip avar segments when they are all identity mappings

### LINKED ISSUE #469: fontc produces no-op avar even when axes define no mappings
When a source file contains no axis mappings (thus user-space and design-space locations are meant to be identical and no "warping" occurs), e.g. this simple single-axis font:

```xml
<designspace format="5.0">
  <axes>
    <axis tag="wght" name="Weight" minimum="400" maximum="700" default="400"/>
  </axes>
```

... then fontc insists on building an avar table even when this does not do anything interesting besides the default {-1.0: -1.0, 0.0: 0.0, 1.0: 1.0} mappings, i.e. `ttx -t avar -q -o - build/font.ttf` gives:

```xml
  <avar>
    <version major="1" minor="0"/>
    <segment axis="wght">
    </segment>
  </avar>
```

FontTools in cases like this omits building an avar table, so we should match it.

anthrotype:
Fixes #469

although a few other tests are still failing, I'm not sure exactly how to fix them.

IIUC, because in the case of all-identity maps I'm no longer writing out an avar table, the tests that attempt to compile a second time are failing because the Avar backend task is being run again, whereas they expect that it not be run...

### DIFF
--- a/fontbe/src/avar.rs
+++ b/fontbe/src/avar.rs
@@ -97,13 +97,8 @@ impl Work<Context, AnyWorkId, Error> for AvarWork {
             .filter_map(to_segment_map)
             .filter(|sm| !sm.axis_value_maps.is_empty())
             .collect();
-        if axis_segment_maps.is_empty() {
-            debug!("Skip avar; no interesting mappings");
-            return Ok(());
-        }
-        context
-            .avar
-            .set_unconditionally(Avar::new(axis_segment_maps).into());
+        let avar = (!axis_segment_maps.is_empty()).then_some(Avar::new(axis_segment_maps));
+        context.avar.set_unconditionally(avar.into());
         Ok(())
     }
 }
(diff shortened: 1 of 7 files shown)


rsheeter:
We haven't had to deal with a BE choosing to *not* generate a table before. The simplest solution would be a sentinal for an empty table. Thinking "aloud" as these are stored as bytes a simple choice would be to use an empty byte array and update the Font assembly code to filter that out. I haven't tried it out yet tho.

anthrotype:
> use an empty byte array and update the Font assembly code to filter that out

that sounds good but I don't feel familiar enough with that.. maybe you can pick this one up?

rsheeter:
> that sounds good but I don't feel familiar enough with that.. maybe you can pick this one up?

I'll add a commit and then refer to you to review, thus achieving you reviewing your own PR :D

rsheeter (approved):
I approve @anthrotype's changes so if/when mine are approved (see commits) this is good to go :)

PR #471, opened 2023-09-28
Update for new delta IUP API

cmyr:
This updates our deps (mostly fontations/fea-rs, but also a few other things that have version bumps) and makes the minimal set of changes to get us compiling again. There is more work to do to improve our gvar compilation, but that will happen in follow-up work.


anthrotype (approved):
Thanks!

dfrg (approved):
LGTM!

Issue #422, opened 2023-08-27
fontc's gvar for Oswald has significantly fewer deltas


rsheeter:
```shell
$ python resources/scripts/ttx_diff.py ../OswaldFont/sources/Oswald.glyphs
...
  DIFF 'gvar', build/gftools/fontc.gvar.ttx build/gftools/fontmake.gvar.ttx
...
$ diffstat build/gftools/fontmake.gvar.ttx  build/gftools/fontc.gvar.ttx| diffstat
 1 file changed, 206 insertions(+), 5851 deletions(-)
```

Visually there are a few places where the result is significantly different but the vast majority is just fontmake having more deltas, ex

![image](https://github.com/googlefonts/fontc/assets/6466432/6f2be338-bb74-4718-b46a-fb0dab69d6f5)

If some of these are fine/wai I would like ttx_diff to eliminate those cases.

behdad:
I don't think these are fine.

rsheeter:
> I don't think these are fine.

Oops, we concluded the same in IM but evidently I didn't update the issue.

anthrotype:
I believe some of the the diffs are because fonttools varLib does not always use the IUP-optimized deltas (like fontc currently does) but sometimes prefers to store the unoptimized deltas if they happen to serialize to a smaller TupleVariation record. Apparently fewer deltas does not always mean fewer bytes for the complicated way gvar is encoded. The iup_delta_optimize algorithm works on the deltas themselves (replacing some with Nones to make them implied), but does not know about the specifics of gvar serialization.

anthrotype:
I just tried to repro but got loads of `failed to compute deltas: 'Missing a tent for wdth'` seemingly coming from fea-rs, this was using fontc off main HEAD (currently at https://github.com/googlefonts/fontc/commit/23ce03f24d76039bc2f175812b1412c09b03be1f)

anthrotype:
If I checkhout main at a few commits before that (e.g. c267b29), I can then run ttx_diff.py, not exactly sure what broke it.

Anyway, I think my hunch is correct because if I force fontTools to always use the IUP optimized deltas (instead of only when beneficial for size), then the only remaining diff in gvar between fontmake and fontc are the composite glyphs (https://github.com/googlefonts/fontc/issues/398), where fontmake omits the no-op (0,0) deltas whereas fontc does not.


```diff
diff --git a/Lib/fontTools/varLib/__init__.py b/Lib/fontTools/varLib/__init__.py
[...]

PR #472, opened 2023-09-28
Improve handling of composite glyph deltas

### LINKED ISSUE #398: should we compute IUP optimized deltas also for composite glyphs?
the OT spec says:

https://learn.microsoft.com/en-us/typography/opentype/spec/gvar#inferred-deltas-for-un-referenced-point-numbers

> Note: inferring of deltas for un-referenced points applies only to simple glyphs, not to composite glyphs.

And fontc, currently does not process the composite glyphs through `iup_delta_optimize`:

https://github.com/googlefonts/fontc/blob/cfd11e2ff32ab50259fb5f582d514ade66273b7f/fontbe/src/glyphs.rs#L344

https://github.com/googlefonts/fontc/blob/cfd11e2ff32ab50259fb5f582d514ade66273b7f/fontbe/src/glyphs.rs#L600-L606

However, fonttools varLib _does_ process the composite glyphs through its `iup_delta_optimize`, the deltas and origCoords arrays contain one (x,y) for each component and it treats each component as a distinct "contour" in the endPts list, see `_getCoordinatesAndControls` method of glyf table, as used in varLib.build:

https://github.com/fonttools/fonttools/blob/dd783ff95e97e020bf6a4215a20ea11707174f38/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L400-L412
https://github.com/fonttools/fonttools/blob/dd783ff95e97e020bf6a4215a20ea11707174f38/Lib/fontTools/varLib/__init__.py#L354-L357

Because of this discrepancy between fonttools and fontc, we get the following diffs when comparing the gvar of e.g. Oswald-VF.ttf for composite glyphs such as "A-cy" and others (note how in fonttools output, the (0,0) deltas are omitted, i.e. inferred):

```diff
--- Oswald-VF-fontmake.ttf
+++ Oswald-VF-fontc.ttf
     <glyphVariations glyph="A-cy">
       <tuple>
         <coord axis="wght" value="-1.0"/>
+        <delta pt="0" x="0" y="0"/>
+        <delta pt="1" x="0" y="0"/>
         <delta pt="2" x="-24" y="0"/>
+        <delta pt="3" x="0" y="0"/>
+        <delta pt="4" x="0" y="0"/>
       </tuple>
```

Question: should we match the fonttools output and therefore also include composites in the iup_delta_optimization, despite what the spec says about inferred deltas only applying to simple glyphs? I suppose yes but maybe @behdad has more insights into this.

cmyr:
This better matches the behaviour of fontmake:

- we mark (0, 0) all deltas as being optional
- we ensure we always write at least one delta for a composite glyph, per https://github.com/fonttools/fonttools/issues/1381

With this patch our gvar diff is down to about 300 lines, the majority of which looks to be something causing the two methods that are computing interpolatable points producing slightly different results in some cases; will investigate that next.

closes #398

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -277,7 +277,12 @@ fn compute_deltas(
             } else {
                 let deltas = deltas
                     .into_iter()
-                    .map(|delta| GlyphDelta::required(delta.x.ot_round(), delta.y.ot_round()))
+                    .map(|delta| match delta.to_point().ot_round() {
+                        // IUP only applies to simple glyphs; for composites we
+                        // just mark the zero deltas as being interpolatable.
+                        (0, 0) => GlyphDelta::optional(0, 0),
+                        (x, y) => GlyphDelta::required(x, y),
+                    })
                     .collect();
 
                 Ok((region, deltas))
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -18,7 +18,7 @@ use fontir::{
     orchestration::WorkId as FeWorkId,
     variations::{VariationModel, VariationRegion},
 };
-use kurbo::{cubics_to_quadratic_splines, Affine, BezPath, CubicBez, PathEl, Point, Rect};
+use kurbo::{cubics_to_quadratic_splines, Affine, BezPath, CubicBez, PathEl, Point, Rect, Vec2};
 use log::{log_enabled, trace, warn};
 
 use read_fonts::{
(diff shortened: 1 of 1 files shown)


anthrotype (approved):
at first glance, it looks good. Can we add a test as well?

cmyr:
added a couple tests, which motivated factoring out the code that processes the composite deltas (which is cleaner anyway, so win-win)

anthrotype (approved):

anthrotype:
This also fixes https://github.com/googlefonts/fontc/issues/422

no more gvar diffs when building Oswald with fontc vs fontmake! 🥳

Issue #174, opened 2023-03-20
should reverse path direction when going cu2qu


anthrotype:
by default fontmake (via ufo2ft and cu2qu) reverses the glyph contours' winding direction when building TrueType glyf outlines, assuming that sources contain PostScript-style cubic beziers that OT spec recommends to be set in _counter-clockwise_ order. 
If that is true (we don't actually validate that), then the resulting TrueType glyf outlines after the flipping become _clockwise_, which is what the TrueType spec in turn recommends.

https://github.com/googlefonts/ufo2ft/blob/fca66fe3ea1ea88ffb36f8264b21ce042d3afd05/Lib/ufo2ft/__init__.py#L223

To match fontmake-py, fontamake-rs could similarly apply its ReverseContourPen by default, and have an option to not to, when building glyf table. 

I've always thought that path direction didn't matter, until I stumbled on this: https://github.com/adobe-fonts/source-han-serif/issues/182

PR #475, opened 2023-09-29
Reverse TrueType glyph contour direction by default to match fontmake

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/174

~~but somewhat blocked on https://github.com/googlefonts/fontations/issues/640~~ Now using the latest kurbo's BezPath::reverse_subpaths() instead of ReverseContourPen

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -15,7 +15,7 @@ use fontdrasil::{
 use fontir::{
     coords::{Location, NormalizedCoord, NormalizedLocation},
     ir,
-    orchestration::WorkId as FeWorkId,
+    orchestration::{Flags, WorkId as FeWorkId},
     variations::{VariationModel, VariationRegion},
 };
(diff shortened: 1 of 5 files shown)


fontbe/src/glyphs.rs
```diff
@@ -686,6 +691,34 @@ impl CheckedGlyph {
...
+
+        let reversed = contours
+            .iter()
+            .map(|(loc, contour)| (loc.clone(), contour.reverse_subpaths()))
+            .collect();
+
+        *self = CheckedGlyph::Contour {
+            name: name.clone(),
+            paths: reversed,
+        }
```
dfrg, line 720:
```suggestion
        match self {
            CheckedGlyph::Contour { name, paths } => {
                trace!("Reverse '{name}' contour direction");
                for (_, contour) in paths {
                    *contour = contour.reverse_subpaths();
                }
            }
            _ => {} // nop for composite
        }
```
This might be a bit more clear? If you match on an enum value with `&mut` binding, the internal fields of the variants are also bound `&mut` in the pattern so you can modify them directly.

dfrg (approved):
LGTM with a small suggestion on style.

Issue #291, opened 2023-05-21
Support more eager decomposition of transforms OR alter fontmake-py


rsheeter:
@anthrotype notes:

> This optional filter is more eager than our current implementation in fontc. This one will decompose any transformed components anywhere, not just those components that have different 2x2 in different masters like we currently do. We could add an option to do the former (off by default because fontmake doesn't enable this by default).

https://github.com/googlefonts/fontmake-rs/pull/288#discussion_r1199176650

Alternately, update fontmake and gftools to support matching the fontc behavior.

anthrotype:
the fontc behavior is arguably the most correct of the three. By default we should only decompose transformed components that have different 2x2 transforms between masters. Right now I think fontmake/ufo2ft will let them pass through but [varLib will issue a logging warning](https://github.com/fonttools/fonttools/issues/2572) about the components not being interpolation compatible and leave their gvar blank.. not super nice.
Decomposing _all_ transformed components may be desirable if one wants to support some legacy environment that doesn't play well with those, but is not strictly necessary to support normal VF builds, hence I think it should not be done by default.

PR #480, opened 2023-10-03
Optionally eliminate components with transforms to match gftools

rsheeter:
https://github.com/googlefonts/fontc/issues/291#issuecomment-1557358538 notes "if we want to match gftools default behavior of decomposing unconditionally all transformed components we need to add a corresponding option to fontc"

So ... do that. After this the glyf and gvar diff in ttx_diff gftools mode is down to a very few points where we're off by one:

image

### DIFF
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -31,6 +31,12 @@ pub struct Args {
     #[arg(long, default_value = "false")]
     pub flatten_components: bool,
 
+    /// Whether a component is allowed to have a non-identity 2x2 transform.
+    ///
+    /// If not, any source component that does will be flattened
+    #[arg(long, default_value = "true", action = ArgAction::Set)]
+    pub allow_component_transforms: bool,
+
     /// Whether to out timing data, notably a visualization of threadpool execution of tasks.
     ///
     /// See <https://github.com/googlefonts/fontc/pull/443>
@@ -59,6 +65,10 @@ impl Args {
         flags.set(Flags::EMIT_DEBUG, self.emit_debug);
         flags.set(Flags::PREFER_SIMPLE_GLYPHS, self.prefer_simple_glyphs);
         flags.set(Flags::FLATTEN_COMPONENTS, self.flatten_components);
+        flags.set(
+            Flags::ALLOW_COMPONENT_TRANSFORM,
+            self.allow_component_transforms,
+        );
         flags.set(Flags::EMIT_TIMING, self.emit_timing);
 
         flags
@@ -80,6 +90,7 @@ impl Args {
             build_dir: build_dir.to_path_buf(),
             prefer_simple_glyphs: Flags::default().contains(Flags::PREFER_SIMPLE_GLYPHS),
             flatten_components: Flags::default().contains(Flags::FLATTEN_COMPONENTS),
+            allow_component_transforms: Flags::default().contains(Flags::ALLOW_COMPONENT_TRANSFORM),
             compile_features: true,
         }
     }
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -34,8 +34,9 @@ pub struct Args {
     /// Whether a component is allowed to have a non-identity 2x2 transform.
     ///
     /// If not, any source component that does will be flattened
-    #[arg(long, default_value = "true", action = ArgAction::Set)]
-    pub allow_component_transforms: bool,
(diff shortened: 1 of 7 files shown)


fontc/src/args.rs
```diff
@@ -31,6 +31,12 @@ pub struct Args {
     #[arg(long, default_value = "false")]
     pub flatten_components: bool,
 
+    /// Whether a component is allowed to have a non-identity 2x2 transform.
+    ///
+    /// If not, any source component that does will be flattened
+    #[arg(long, default_value = "true", action = ArgAction::Set)]
+    pub allow_component_transforms: bool,
```
anthrotype, line 38: optional: we could use the same name as the ufo2ft's filter `--decompose-transformed-components` which is already familiar to our user, and have the bool flag be false by default (so we don't need to add a value `--allow-component-transforms=false`).

fontir/src/glyph.rs
```diff
@@ -402,6 +402,15 @@ impl Work<Context, WorkId, WorkError> for GlyphOrderWork {
             }
         }
 
+        if !context.flags.contains(Flags::ALLOW_COMPONENT_TRANSFORM) {
```
anthrotype, line 405: I think we want to apply this optional decomposing operation at the beginning, before the default mixed-glyphs decomposition (in ufo2ft this is usually set as a "pre" filter, one that precedes the default filters), and also before the flatten-components operation. It should not change the results, but I think you end up doing less work in the following component-only operations if you have already removed some of the components.

anthrotype (approved):
LGTM with a couple of comments

fontc/src/args.rs
anthrotype: gftools builder also calls this option `decomposeTransformedComponents` https://github.com/googlefonts/gftools/blob/b074d8fbb52631030e6c1f524513a5a2eb7b2d56/docs/gftools-builder/README.md?plain=1#L134

fontir/src/glyph.rs
anthrotype: actually nevermind this last comment, doing it after flatten-components is ok and in fact matches the order in which gftools builder requests these filters to be applied:

https://github.com/googlefonts/gftools/blob/b074d8fbb52631030e6c1f524513a5a2eb7b2d56/Lib/gftools/builder/__init__.py#L358-L371
https://github.com/googlefonts/gftools/blob/b074d8fbb52631030e6c1f524513a5a2eb7b2d56/Lib/gftools/builder/_ninja.py#L203-L206

anthrotype:
> After this the glyf and gvar diff in ttx_diff gftools mode is down to a very few points where we're off by one

we should file a separate issue for these and investigate the cause. It might be down to the order in which various operations are applied to contours. I think in ufo2ft the cu2qu filter is done before the transformed components are decomposed (whereas we do the former in the fontbe). In the end of the day those diffs could be acceptable

fontir/src/orchestration.rs
```diff
@@ -29,15 +29,15 @@ bitflags! {
         // If set, a composite that references another composite will replace that composite with the
         // glyph(s) it references until only simple (contour) glyphs are referenced
         const FLATTEN_COMPONENTS = 0b00001000;
+        const ALLOW_COMPONENT_TRANSFORM = 0b00010000;
         // If set a files reporting on timing will be emitted to disk
-        const EMIT_TIMING = 0b00010000;
+        const EMIT_TIMING = 0b00100000;
```
anthrotype, line 34: maybe we should make Flags a u64, looks like we may run out of empty slots soon

rsheeter: Agreed but as best I can tell we can defer that until we actually hit it because nothing outside fontc relies on these flags.

fontc/src/args.rs
```diff
@@ -31,6 +31,13 @@ pub struct Args {
     #[arg(long, default_value = "false")]
     pub flatten_components: bool,
 
+    /// Whether a component is allowed to have a non-identity 2x2 transform.
+    ///
+    /// If not, any source component that does will be flattened
+    /// Named to match the ufo2ft flag as suggested in <https://github.com/googlefonts/fontc/pull/480#discussion_r1343801553>
```
anthrotype, line 37: I don't think we need to be reminded of this every time we do `fontc --help`

anthrotype: dunno, move it inline with the code somewhere else than the docstring perhaps

rsheeter: #485

Issue #386, opened 2023-08-08
[ufo2fontir] Use `stylename` as instance name (not `name`)


cmyr:
This sounds good, but I'm not sure what we should do if there's no stylename specified for a given instance? Do we leave it blank, fallback on the actual name, or ??

> I just noticed that we are using the instance `name` attribute to set the fvar's NamedInstance's subfamilyNameID names, but this is wrong, varLib uses the `stylename` attribute of an instance, not its `name` for that purpose. The latter is similar to source.name and only meant to uniquely identify a given DS source/instance element for cross referencing within the doc (should have been called `id` perhaps but it's too late).

https://github.com/fonttools/fonttools/blob/57fb47de3a41c724c3bf1051d00715e08d86141e/Lib/fontTools/varLib/__init__.py#L103-L118

If we use `stylename` here we no longer need to do inst.name.strip_prefix(family_prefix).

_Originally posted by @anthrotype in https://github.com/googlefonts/fontc/pull/376#discussion_r1286960988_

anthrotype:
> what we should do if there's no stylename specified for a given instance? 

fonttools in the linked code above raises an error if it cannot find a suitable style name. It will be there for well-behaved real-world sources. An instance without a stylename makes little sense, to be honest, either as the target for an interpolated, static font (the legacy build pipeline) or for variable font builds.
I am not sure why designspaceLib doesn't mark it as required. Maybe because one could omit it if they do provide the localisedStyleName["en"] variant... (which we don't support either I think).

anthrotype:
Ah... yeah, it is optional because in DesignSpace v5 the instance style name can be synthesized using STAT stylistic attributes defined with axis labels. One private font of ours uses that approach actually so gets a bunch of "unnamed_instance_0" etc. in fvar currently.

anthrotype:
using `stylename` attribute instead of `name` should be trivial to do, but my impression is your font does not have those but relies on DSv5 instance stylenames generated from axis labels (those also used for STAT stylistic attributes). If that's the case, then I think it will take some time to implement in fontc and may be quicker if you patch the fontc-built fvar tables using some post-build script

anthrotype:
> using stylename attribute instead of name should be trivial to do

I'll start working on that

anthrotype:
@RickyDaMa I can confirm that with https://github.com/googlefonts/fontc/pull/481 we build fvar instances named as expected for those two Google private projects

PR #481, opened 2023-10-03
ufo2fontir: prefer use instance.stylename

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/386

~~Note the previous behavior of taking instance.name and stripping familyname prefix does not match fonttools behavior so I removed it. The instance.name (like similar element name attributes in DS format) is meant as an identifier for cross-referencing elements and is not used for named instances' style names. Only instance `stylename` attribute (and the localised names if any) are used by fonttools to set the fvar's NamedInstance subFamilyNameID.~~

The `stylename` attribute can be omitted in DSv5 and be generated using axis labels, but for most projects (including those two Google private projects that we wish to use this for) they are likely to be defined explicitly so with this PR we use them.

Also note that the `compile_named_instances_from_designspace` test in fontc/lib.rs is already covering the changes in this PR (the "wght_var.designspace" has instances with correct stylename attributes)

### DIFF
--- a/ufo2fontir/src/source.rs
+++ b/ufo2fontir/src/source.rs
@@ -712,7 +712,7 @@ impl Work<Context, WorkId, WorkError> for StaticMetadataWork {
                 // TODO: Also support localised names, and names inferred from axis labels
                 // (also used to build STAT table)
                 NamedInstance {
-                    name: inst.stylename.clone().unwrap_or(
+                    name: inst.stylename.clone().unwrap_or_else(|| {
                         match inst
                             .name
                             .as_ref()
@@ -721,8 +721,8 @@ impl Work<Context, WorkId, WorkError> for StaticMetadataWork {
                         {
                             Some(tail) => tail.to_string(),
                             None => inst.name.clone().unwrap(),
-                        },
-                    ),
+                        }
+                    }),
                     location: to_design_location(&tags_by_name, &inst.location)
                         .to_user(&axes_by_tag),
                 }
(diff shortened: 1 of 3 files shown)


ufo2fontir/src/source.rs
```diff
@@ -700,25 +700,18 @@ impl Work<Context, WorkId, WorkError> for StaticMetadataWork {
...
-                {
-                    Some(tail) => tail.to_string(),
-                    None => inst.name.clone().unwrap(),
-                },
-                location: to_design_location(&tags_by_name, &inst.location).to_user(&axes_by_tag),
+            .map(|inst| {
+                // TODO: Also support localised names, and names inferred from axis labels
+                // (also used to build STAT table)
+                NamedInstance {
+                    name: inst.stylename.clone().unwrap_or(inst.name.clone().unwrap()),
```
anthrotype, line 711: initially here I had a try_fold that would return an error when stylename was not set, but then I thought it wasn't worth the hassle and I'm falling back to the instance.name that we have already populated with `unnamed_instance_{0}`, etc. upon loading the designspace and thus we can unwrap().

anthrotype: I actually ended up restoring the family-prefix stipping of instance.name which I had initially removed, even though it doesn't match fonttools, may still have uses. In any case we shall revisit this when we will add support for instance stylenames from DSv5 axis labels

rsheeter (approved):

Issue #477, opened 2023-10-02
`--compile-features false` still checks feature files for equality


madig:
Try it on https://github.com/LettError/mutatorSans.

```
❯ ../fontc/target/release/fontc -s MutatorSans.designspace --compile-features false
[2023-10-02T09:51:46.702397Z main fontc::workload ERROR] Fe(Features) failed Expected to match: "MutatorSansLightCondensed.ufo/features.fea", "MutatorSansBoldCondensed.ufo/features.fea"
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', ufo2fontir/src/toir.rs:25:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
At least one work item failed
```

anthrotype:
it should work now after #482. Note that I have also renamed the flag to --skip-features (it no longer takes a true|false value), since it now does more than just skipping the backend compilation step, it also skips the font IR step for the features (where that check for equality is done)

PR #482, opened 2023-10-03
skip feature IR job if --compile-features=false

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/477

JMM


rsheeter (approved):
Maybe we should rename the flag since now it does more than skip compile? "skip-features" for example.

PR #484, opened 2023-10-04
use BezPath::reverse_subpaths() from latest kurbo

anthrotype:
Added with kurbo 0.10.2 in https://github.com/linebender/kurbo/pull/307

has same functionality as ReverseContourPen when applied to BezPath, but without the loss of precision from f64 to f32

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ bitflags = "2.0"
 chrono = { version = "0.4.24", features = ["serde"] }
 filetime = "0.2.18"
 indexmap = "1.9.2"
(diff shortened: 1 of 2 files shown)


anthrotype:
note the CI fails because one of fontc depedencies, fea-rs, is pulling in an old write-fonts which in turn depends on an old kurbo so we end up with two incompatible write-fonts and kurbos.. we need to first tag a release of write-fonts that uses the latest kurbo, then bump write-fonts in fea-rs, release fea-rs and finally bump write-fonts, fea-rs and kurbo here in fontc...

dfrg (approved):
Lgtm! Pre-approving for pending version bumps.

Issue #41, opened 2022-12-08
Naive glyphname => IR filename leads to glyphs overwritten on case insensitive filesystems


anthrotype:
right now we use {glyph_name}.yml but that only works on case sensitive fs like Linux.

Initially I thought I'd use norad's `user_name_to_file_name` method (which uses UFO filenaming conventions) for this, but it actually complicates things because we need to know all glyph names upfront and it's order dependent (we can't call that on demand because its output depends on the order in which duplicate clashes are handled).

we could try see if we can use a scheme like Just proposed in https://github.com/unified-font-object/ufo-spec/issues/164, which has an added bonus of being reversible

anthrotype:
I just re-stumbled on this while running fontc on a Mac. It happens with any font that has both uppercase and lowercase glyphs e.g. "A" and "a" as in the following test file: [Aa.glyphs.zip](https://github.com/googlefonts/fontc/files/12845246/Aa.glyphs.zip)

running `fontc -s Aa.glyphs` the first time produces a correct font, but then modifying one of the two glyphs and running `fontc -s Aa.glyphs` a second time leads to invalid results where one of the two glyphs gets lost.
The `build/glyph_ir` and `build/glyphs` folder only contain respectively `a.yml` and `a.glyf` (the filename is always normalized as lowecase, but if you inspect the glyph name stored in the yaml file it can be either "a" or "A" depending on which one wins..)

This makes using the default --emit-ir=true very risky when running fontc on a non-Linux.

anthrotype:
Maybe incremental build should be opt-in rather then opt out, and --emit-ir=false by default.

justvanrossum:
I've previously suggested [some ideas](https://github.com/unified-font-object/ufo-spec/issues/164) to enhance the UFO glyph name -> file name scheme, but today I would be inclined to just append a hash of the glyph name to the file name. `A.<hash-of-A>.ext`

Regardless of incremental build issues, this seems a problem worth fixing.

rsheeter:
How are case insensitive filesystems still a thing!

It's quite handy to have a human readable name when troubleshooting. Let's
implement a scheme to make it work.

anthrotype:
Thanks Just. I had linked to that other thread, contains useful hints but got sidetracked.

> I would be inclined to just append a hash of the glyph name to the file name. A.<hash-of-A>.ext

but hash cannot be reverted, can it? Ideally we'd be able to go in either direction with only the glyph name or the filename available.

[...]

PR #486, opened 2023-10-09
Add an underscore after capitals in glyph names

rsheeter:
Helps with #41. Doesn't deal with reserved / weird filenames.

### DIFF
--- a/fontdrasil/src/paths.rs
+++ b/fontdrasil/src/paths.rs
@@ -1,5 +1,45 @@
 pub fn glyph_file(glyph_name: &str, suffix: &str) -> String {
     // TODO handle names that are invalid for the filesystem
     // Ref https://github.com/unified-font-object/ufo-spec/issues/164
-    glyph_name.to_owned() + suffix
+    let mut filename = Vec::new();
+    for ch in glyph_name.chars() {
+        filename.push(ch);
+        if ch == '_' || ch.is_uppercase() {
+            filename.push('_');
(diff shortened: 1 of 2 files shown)


fontdrasil/src/paths.rs
```diff
@@ -1,5 +1,42 @@
 pub fn glyph_file(glyph_name: &str, suffix: &str) -> String {
     // TODO handle names that are invalid for the filesystem
     // Ref https://github.com/unified-font-object/ufo-spec/issues/164
-    glyph_name.to_owned() + suffix
+    let mut filename = Vec::new();
+    for ch in glyph_name.chars() {
+        filename.push(ch);
+
+        // if we wanted reversibility we could append _ when we see an _ in the name
+        // at time of writing we don't want reversibility, merely uniqueness
```
anthrotype, line 9: you also want to escape the underscore otherwise you may inadvertently confuse two valid glyph names like `Aa` and `a_a`. If both get translated to filename as `a_a` then one may overwrite the other, no?

rsheeter: Yes, I think I do!

anthrotype (approved):
LGTM once test pass

this might be ok for a first pass, but the illegal characters and reserved filename will surely come back biting us at some point

rsheeter:
> illegal characters and reserved filename will surely come back biting us at some point

Yup! Baby steps :)

Issue #248, opened 2023-04-10
Match fontmake glyph names


rsheeter:
Currently the diff for fontmake Oswald vs fontmake-rs Oswald is very noisy due to differing glyph names. Fix it, or update the instructions (on milestone) to avoid.

LHS fontmake, RHS fontmake-rs

![image](https://user-images.githubusercontent.com/6466432/231012428-9f849497-e6de-487f-a5ee-9f95773cf110.png)

rsheeter:
`--no-production-names` resolves this for the purpose of comparing compiles. Removing from the milestone but leaving open in case we do want to support this form of renaming.

anthrotype:
>  glyphs2ir support for postscript names

oh right I forgot that.. Then #498 too hastily got rid of the --no-production-names option. I think we want to continue passing --no-production-names flags to both fontc and fontmake in ttx_diff.py, until that is implemented for both DS and glyphs inputs. I'll make another PR

anthrotype:
> matching fontmake's name generation

implementing the full ufo2ft postProcessor glyph-renaming logic is lower priority in my opinion, I think for now we can content ourselves to writing the explicit postscriptNames which are already there in the UFOs (and the ones generated from GlyphData.xml for .glyphs inputs).

anthrotype:
> Main points that need to be addressed still to close this issue:
>
> - [ ] glyphs2ir support for postscript names

this requires glyphs_reader::glyphdata module to parse so-called "production" glyph names from GlyphData.xml, there's a linked issue at https://github.com/googlefonts/fontc/issues/780

anthrotype:
In addition to the work on 'synthetic' production names (for alternates and ligature glyphs) in #1430, fontc will also need to support the `Don't use Production Names` Glyphs custom parameter which sometimes is found in sources, e.g. NotoSerifAhom.glyphs.

python resources/scripts/ttx_diff.py 'https://github.com/notofonts/ahom#sources/NotoSerifAhom.glyphs'

The above source tells the compiler to _not_ rename glyphs to production names, but fontc (with #1430 patch) cannot read that yet and therefore proceeds to rename the glyphs, whereas fontmake abides by that and doesn't rename so we get a ttx diff. 

The DS/UFO equivalent is 

```
    <key>com.github.googlei18n.ufo2ft.useProductionNames</key>
    <false/>
```

glyphsLib turns the former custom parameter into the latter lib key when translating from .glyphs to DS/ufo.
[...]

PR #489, opened 2023-10-17
Postscript names for UFO sources

RickyDaMa:
Related to: #248

Reads postscript names from the `lib.plist` of UFO sources, and then uses them when writing the post table.

Compiling Oswald using its `Makefile` yields an identical post table to that of a fontc-compiled UFO-based Oswald.

Tests not yet updated to account for the extra field on `StaticMetadata`, or extra parameter of `StaticMetadata::new`. Very much just a proof of concept to gather feedback and see if there are any changes wanted in our approach before updating tests. Naturally, tests should also be added for reading/writing these names.

For discussion:

1. How do we want to handle duplicate postscript entry values?
2. What auto-generation do we need to implement for glyph names that are not included in AGLFN? What does `fontmake` do?
3. Do we want to add a CLI flag to use design names instead of production names, similar to `fontmake`'s `--no-production-names`?
4. Shall we implement this feature for `glyphs2fontir` in this PR?

Implementation notes/thoughts:

5. Do we want to check for equality of `public.postscriptNames` across all masters?
6. Should we read postscript names and glyph order at the same time? Or at least, without loading the plist into memory twice?
7. What error variant should we use for "public.postscriptNames is present, but not a dictionary"?
8. Does `postscript_names` belong in the top-level `StaticMetadata`, or in misc (or context, like `GlyphOrder`)?
9. Should we use a smaller type for postscript name values, i.e. `GlyphName` or `Box<str>` instead of `String`?

Tangent (though tests will currently fail if run): https://matklad.github.io/2022/10/24/actions-permissions.html

### DIFF
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -118,19 +118,22 @@ impl Args {
 mod tests {
     use clap::Parser;
     use fontir::orchestration::Flags;
+    use log::{debug, LevelFilter};
 
     use crate::Args;
 
     // It's awkward to get the Flags::default values into #[arg] so test for consistency
     #[test]
     fn arg_default_matches_flags_default() {
+        let _ = env_logger::builder()
+            .is_test(true)
+            .format_timestamp(None)
+            .filter_level(LevelFilter::Debug)
+            .try_init();
         let arg_default = Args::parse_from(vec!["program", "--source", "dont.care"]).flags();
         let flags_default = Flags::default();
-        println!(
-            "flags_default: {:#032b}\nargs_default:  {:#032b}",
-            flags_default.bits(),
-            arg_default.bits()
-        );
+        debug!("flags_default: {:#032b}", flags_default.bits());
+        debug!("args_default:  {:#032b}", arg_default.bits());
         assert_eq!(flags_default.bits(), arg_default.bits());
     }
 }
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -65,7 +65,7 @@ pub struct Args {
     pub keep_direction: bool,
 
     /// Don't rename glyphs with production names
-    #[arg(long, default_value = "true")]
+    #[arg(long, default_value = "false")]
     pub no_production_names: bool,
 }
 
@@ -109,6 +109,7 @@ impl Args {
                 .contains(Flags::DECOMPOSE_TRANSFORMED_COMPONENTS),
             skip_features: false,
             keep_direction: false,
+            no_production_names: false,
         }
     }
 }
@@ -125,6 +126,11 @@ mod tests {
     fn arg_default_matches_flags_default() {
         let arg_default = Args::parse_from(vec!["program", "--source", "dont.care"]).flags();
         let flags_default = Flags::default();
+        println!(
+            "flags_default: {:#032b}\nargs_default:  {:#032b}",
+            flags_default.bits(),
+            arg_default.bits()
+        );
         assert_eq!(flags_default.bits(), arg_default.bits());
     }
 }
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -63,6 +63,10 @@ pub struct Args {
     // https://github.com/googlefonts/fontmake/blob/6a8b2907/Lib/fontmake/__main__.py#L443
     #[arg(long, default_value = "false")]
     pub keep_direction: bool,
+
+    /// Don't rename glyphs with production names
+    #[arg(long, default_value = "true")]
+    pub no_production_names: bool,
 }
 
 impl Args {
@@ -80,6 +84,7 @@ impl Args {
         );
         flags.set(Flags::EMIT_TIMING, self.emit_timing);
         flags.set(Flags::KEEP_DIRECTION, self.keep_direction);
+        flags.set(Flags::PRODUCTION_NAMES, !self.no_production_names);
 
         flags
     }
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -65,6 +65,8 @@ pub struct Args {
     pub keep_direction: bool,
 
     /// Don't rename glyphs with production names
+    // Named to match fontmake's homonymous flag:
+    // https://github.com/googlefonts/fontmake/blob/6a8b2907/Lib/fontmake/__main__.py#L602
     #[arg(long, default_value = "false")]
     pub no_production_names: bool,
 }
@@ -118,22 +120,20 @@ impl Args {
 mod tests {
     use clap::Parser;
     use fontir::orchestration::Flags;
-    use log::{debug, LevelFilter};
 
     use crate::Args;
 
     // It's awkward to get the Flags::default values into #[arg] so test for consistency
     #[test]
     fn arg_default_matches_flags_default() {
-        let _ = env_logger::builder()
-            .is_test(true)
-            .format_timestamp(None)
-            .filter_level(LevelFilter::Debug)
-            .try_init();
         let arg_default = Args::parse_from(vec!["program", "--source", "dont.care"]).flags();
         let flags_default = Flags::default();
-        debug!("flags_default: {:#032b}", flags_default.bits());
-        debug!("args_default:  {:#032b}", arg_default.bits());
-        assert_eq!(flags_default.bits(), arg_default.bits());
+        assert_eq!(
+            flags_default.bits(),
+            arg_default.bits(),
+            "mismatch in defaults. flags_default: {:#032b}. args_default: {:#032b}",
+            flags_default.bits(),
+            arg_default.bits(),
+        );
     }
 }
--- a/fontir/src/serde.rs
+++ b/fontir/src/serde.rs
@@ -6,6 +6,7 @@ use std::{
 use chrono::{DateTime, Utc};
 use filetime::FileTime;
 use font_types::Tag;
+use fontdrasil::types::GlyphName;
 use ordered_float::OrderedFloat;
 use serde::{Deserialize, Serialize};
 use write_fonts::tables::os2::SelectionFlags;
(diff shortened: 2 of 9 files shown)


anthrotype:
> 7. What error variant should we use for "public.postscriptNames is present, but not a dictionary"?

`WorkError::ParseError(lib_plist_file, "Not a dictionary".to_string())` as in https://github.com/googlefonts/fontc/blob/6381929c162e2f97497a56daaa9a8d0b98065b04/ufo2fontir/src/source.rs#L478

> 8. Does postscript_names belong in the top-level StaticMetadata, or in misc (or context, like GlyphOrder)?

it's fine in StaticMetadata, I think

anthrotype:
> taking the one from the designspace.lib or, if that missing, the one from the default master UFO lib.plist should suffice.

it looks like ufo2ft only looks for postscriptNames defined in the default master UFO lib.plist, that's the one it uses in its postProcessor when compiling a VF. I confused this with the other "skipExportGlyphs" key which ufo2ft attempts to find in the designspace.lib first. Maybe it should do that as well for postscriptNames.. Anyway matching current fontmake.py is fine for now.

anthrotype:
> Duplicate as in the same value repeated more than once. There's currently no handling for this

I don't think the python toolchain checks for this, does it? In any case, feel free to add or not add the additional check. To me it feels more like the job of a linter

Hoolean:
Hello all, and many thanks for the speedy review Cosimo :)

> I don't think the python toolchain checks for this, does it? In any case, feel free to add or not add the additional check. To me it feels more like the job of a linter

IIRC in ufo2ft we do some sanitisation of names and [add a suffix to ensure uniqueness](https://github.com/googlefonts/ufo2ft/blob/8a512b89/Lib/ufo2ft/postProcessor.py#L229-L239), although I cannot see a uniqueness requirement in the [`post` table spec](https://learn.microsoft.com/en-us/typography/opentype/spec/post#version-20); I am not sure if this was added for platform support or just to cooperate with ttLib's use of names as unique IDs. In most real-world instances where I have seen duplicate `post` names in sources it has been unintentional, however, and so I am torn.

anthrotype:
Thanks Harry. Indeed, looks like that dedup logic was deliberately added in ufo2ft to address some fonttools quirk:
https://github.com/googlefonts/ufo2ft/pull/274

while the post spec may not require the names to be unique, it kind of defeats the purpose of having multiple glyphs mapped to the same postscript name if you use the name to identify a specific glyph or reverse-cmap, which is what AGLFN is primarily about. If cmap maps from unicode codepoint to glyph (ids), it's impossible for a given codepoint to map to more than one glyph at a time.

I think it's rare that sources have duplicate postscript glyph names, and if they do it's usually by mistake, I don't see a use case for it besides wanting to make a deliberately broken font for testing purposes. But sure let's add the dedup logic

anthrotype:
> sure let's add the dedup logic

also feel free to kick the can down the road and not do it in this PR (can file an issue)

anthrotype:
anyway, when decompiling post, fonttools seems to be able to deduplicate glyph names as needed:

https://github.com/fonttools/fonttools/blob/6d531fed449d299c31e4cf69957b3fdd81cb982d/Lib/fontTools/ttLib/tables/_p_o_s_t.py#L117-L122

since we are not using fonttools to compile the binary font, we don't have this problem, so it's only about whether or not we think a valid post should contain unique names. Given the post spec does not enforce this requirement, I believe we should not either.

anthrotype:
>  it's impossible for a given codepoint to map to more than one glyph at a time.

while that's true, if you also think about alternate glyphs substituted for via GSUB, then indirectly it is possible to say that both glyph "A" and glyph "A.ss01" are mapped to 0x0041, and in fact AGL spec says one should drop the suffix after the period when mapping from glyph name to character string: https://github.com/adobe-type-tools/agl-specification#2-the-mapping

yeah I think I have convinced myself that we do not need to do any checks for duplicates here

anthrotype:
> I don't see a use case

now I do see it. If you don't care about the nice debug glyph names and being able to distinguish glyph "A" from "A.ss01" (e.g. when decompiling the font in a font editor) but only care about AGL compliance and reverse-cmapping of names to characters for old PDF distillers, then it may be beneficial to actually reuse postscript names for alternate glyphs that ultimately are mapped to the same unicode codepoint(s) such that the post table is as compact as needed.

fontc/src/args.rs
```diff
@@ -63,6 +63,10 @@ pub struct Args {
     // https://github.com/googlefonts/fontmake/blob/6a8b2907/Lib/fontmake/__main__.py#L443
     #[arg(long, default_value = "false")]
     pub keep_direction: bool,
+
+    /// Don't rename glyphs with production names
```
rsheeter, line 67: Can you add a note that this is setup to match the fontmake flag and a reference to it similar to the one on keep_direction?

fontc/src/args.rs
```diff
@@ -112,14 +118,22 @@ impl Args {
...
     fn arg_default_matches_flags_default() {
+        let _ = env_logger::builder()
+            .is_test(true)
+            .format_timestamp(None)
+            .filter_level(LevelFilter::Debug)
+            .try_init();
         let arg_default = Args::parse_from(vec!["program", "--source", "dont.care"]).flags();
         let flags_default = Flags::default();
+        debug!("flags_default: {:#032b}", flags_default.bits());
+        debug!("args_default:  {:#032b}", arg_default.bits());
```
rsheeter, line 136: Instead of firing up the logger and writing to debug make the assert do what you want, either by altering the left/right values or by adding a custom panic string to show the `{:#032b}` strings

ufo2fontir/src/source.rs
```diff
@@ -519,6 +519,48 @@ fn glyph_order(
...
+                    }
+                })
+                .collect()
+        }
+        None => HashMap::new(),
+    };
+    // Warn about duplicate public.postscriptNames values
+    // Allocate space ahead of time for speed, and because in the happy path the set length will be
+    // the same as the map len
+    let mut duplicate_values_check = HashSet::with_capacity(postscript_names.len());
```
rsheeter, line 550: Optional: I might suggest something like "seen"; duplicate_values_check is what we are doing but what this stores is the values we've encountered already

ufo2fontir/src/source.rs
```diff
@@ -519,6 +519,48 @@ fn glyph_order(
...
+    let duplicate_values = postscript_names
+        .values()
+        .filter(|ps_name| {
+            let ps_name = (*ps_name).clone();
+            !duplicate_values_check.insert(ps_name)
+        })
```
rsheeter, line 556: Optional, no need for {}'s?

ufo2fontir/src/source.rs
```diff
@@ -519,6 +519,48 @@ fn glyph_order(
...
+        .collect::<BTreeSet<_>>();
+    if !duplicate_values.is_empty() {
+        warn!("public.postscriptNames: the following production names are used by multiple glyphs: {duplicate_values:?}");
```
rsheeter, line 559: Since we don't use duplicate_values at all if warn is suppressed I would wrap the dup checking in `if log::log_enabled!(log::Level::Warn) {`

ufo2fontir/src/source.rs
```diff
@@ -1733,4 +1800,186 @@ mod tests {
...
+            PostscriptNames::from_iter([(
+                GlyphName::from("manual-component"),
+                GlyphName::from("manualcomponent")
+            )]),
+        );
+    }
+
+    #[test]
+    fn static_metadata_disable_postscript_names() {
+        // Copied from build_static_metadata
```
rsheeter, line 1957: Can we have a shared function to reduce the noise and make it easier to see what's unique to this test instead of `// This is the line that's different from build_static_metadata`

rsheeter (approved):
LGTM subject to nits being fixed

ufo2fontir/src/source.rs
```diff
@@ -1346,11 +1345,17 @@ mod tests {
...
+    }
+
+    fn build_static_metadata(name: &str, flags: Flags) -> (impl Source, Context) {
         let _ = env_logger::builder().is_test(true).try_init();
         let (source, input) = load_designspace(name);
-        let mut flags = Flags::default();
-        flags.set(Flags::EMIT_IR, false); // we don't want to write anything down
+        assert!(
+            !flags.contains(Flags::EMIT_IR),
+            "we don't want to write anything"
```
anthrotype, line 1357: I don't think we need to assert flags never contains EMIT_IR here, one may actually want to call build_static_metadata with flags parameter that does contain EMIT_IR (e.g. for testing something related to writing out IR to disk). I suggest you simply remove this assert

ufo2fontir/src/source.rs
```diff
@@ -1346,11 +1345,17 @@ mod tests {
         (source, input)
     }
 
-    fn build_static_metadata(name: &str) -> (impl Source, Context) {
+    fn flags_without_emit_ir() -> Flags {
+        Flags::default() - Flags::EMIT_IR
+    }
+
+    fn build_static_metadata(name: &str, flags: Flags) -> (impl Source, Context) {
```
anthrotype, line 1352: you could even make `flags: Option<Flags>` and when caller pass `None` it means "I don't care, just use the test defaults" (which in our case means `flags_without_emit_ir`), so that in future if we want to change the test default flags we can do in one place

anthrotype: ok, you convinced me :)
can you maybe rename flags_without_emit_ir to something like default_flags_for_tests or similar?

fontir/src/serde.rs
```diff
@@ -70,6 +72,7 @@ impl From<StaticMetadataSerdeRepr> for StaticMetadata {
             from.axes,
             from.named_instances,
             from.glyph_locations.into_iter().collect(),
+            from.postscript_names,
         )
```
Hoolean, line 76: Hello again! A wee question, tangential to this PR: we noticed that `misc` is serialised in `StaticMetadata`→`StaticMetadataSerdeRepr` but not deserialised in `StaticMetadataSerdeRepr`→`StaticMetadata`; do we populate it elsewhere, or could this be a bug? 🐛

anthrotype: looks like a bug to me

anthrotype: technically it's a separate issue, might be worth tackling it separately

anthrotype (approved):

PR #490, opened 2023-10-18
[WIP] Build HVAR

anthrotype:
this depends on "direct-varstore-builder" branch from fontations:

https://github.com/googlefonts/fontations/compare/main...direct-varstore-builder


anthrotype:
And yes, with this PR, Oswald's HVAR has no more ttx diffs when compared with the one fontmake generates 🥳

rsheeter:
I don't object but in reading I was slightly surprised we need a standalone resources/testdata/HVAR; perhaps a readme md there to explain what each file contains and why it's split off?

anthrotype:
> was slightly surprised we need a standalone resources/testdata/HVAR

we don't _need_, I can move them up one level, it's just that they are related to one another and made for the specific purpose of testing HVAR compilation. I needed three of them because each is testing one branch of the code, as the comments in the tests try to explain..

fontbe/src/hvar.rs
```diff
@@ -0,0 +1,207 @@
...
+                None,
+            )
+        } else {
+            Hvar::new(
+                MajorMinor::VERSION_1_0,
+                indirect_store,
+                Some(varidx_map),
+                None,
+                None,
+            )
```
rsheeter, line 200: Seems verbose, suggested change in https://github.com/googlefonts/fontc/pull/524

anthrotype: applied

fontc/src/lib.rs
```diff
@@ -247,6 +248,26 @@ fn add_metric_and_limits_job(workload: &mut Workload) -> Result<(), Error> {
...
+
+    let work = create_hvar_work().into();
+    workload.add(
+        work,
+        // Static metadata contains VariationModel, if axes coordinates change
+        // we need to recompute variable tables such as HVAR.
+        // Glyph order matters because HVAR may choose to store variation items
+        // directly mapping to glyph indices. And glyph IR contains advance width
+        // among other things.
+        // TODO: ideally be more granular here e.g. by storing axes and advance widths
```
rsheeter, line 262: Suggest issue rather than TODO

fontbe/src/hvar.rs
```diff
@@ -0,0 +1,191 @@
...
+                        // Only 1 value per region for our input
+                        assert!(values.len() == 1, "{} values?!", values.len());
+                        Some((
+                            region.to_write_fonts_variation_region(axes),
+                            values[0].ot_round(),
+                        ))
+                    })
+                    .collect(),
+            )
+        }
```
rsheeter, line 135: I spent a reasonable amount of time staring at this loop trying to figure out what was happening and why. I don't see anything wrong with it but I think it might be possible to make easier on the reader.

My immediate thought is to make `all_glyph_width_deltas: Vec<Vec<(VariationRegion, i16)>>` a struct and then called a method on it to add glyphs, e.g.

```rust
        let mut deltas = Deltas::new();
        for (i, gn) in glyph_order.iter().enumerate() {
            let glyph = context.ir.glyphs.get(&FeWorkId::Glyph(gn.clone()));
            deltas.add(glyph.as_ref());
        }
```

Deltas would own the models map. The single_model bool is replaced by Deltas::is_single_model. `exec` gets a bit closer to being a short method you can read quickly.

fontc/src/lib.rs
```diff
@@ -2028,4 +2053,211 @@ mod tests {
...
+                [
+                    coords.start_coord().to_f32(),
+                    coords.peak_coord().to_f32(),
+                    coords.end_coord().to_f32(),
+                ]
+            })
+            .collect::<Vec<_>>();
+        assert_eq!(region_coords, vec![[0.0, 1.0, 1.0]]);
+        // we expect one ItemVariationData and one delta set per glyph
+        assert_eq!(varstore.item_variation_data_count(), 1);
```
rsheeter, line 2099: Suggest printing the varstore if it's unexpected, add `, "{varstore:?}" to the assert.

fontbe/src/hvar.rs
anthrotype: yeah, that for loop is pretty big/ugly, I will try to refactor like you suggest next week, thanks

fontc/src/lib.rs
```diff
@@ -2028,4 +2053,211 @@ mod tests {
...
+        assert_eq!(vardata.delta_set(4).collect::<Vec<_>>(), vec![100]);
+        assert_eq!(vardata.delta_set(5).collect::<Vec<_>>(), vec![120]);
+        assert_eq!(vardata.delta_set(6).collect::<Vec<_>>(), vec![120]);
+        assert_eq!(vardata.delta_set(7).collect::<Vec<_>>(), vec![70]);
+        assert_eq!(vardata.delta_set(8).collect::<Vec<_>>(), vec![70]);
+        assert_eq!(vardata.delta_set(9).collect::<Vec<_>>(), vec![120]);
+        assert_eq!(vardata.delta_set(10).collect::<Vec<_>>(), vec![70]);
+        assert_eq!(vardata.delta_set(11).collect::<Vec<_>>(), vec![0]);
+        assert_eq!(vardata.delta_set(12).collect::<Vec<_>>(), vec![0]);
+        assert_eq!(vardata.delta_set(13).collect::<Vec<_>>(), vec![0]);
```
rsheeter, line 2116: Optional, I find it easier to read as shown in https://github.com/googlefonts/fontc/pull/525

fontc/src/lib.rs
```diff
@@ -2028,4 +2053,211 @@ mod tests {
...
+                .unwrap(),
+            Fixed::from_f64(120.0)
+        );
+        // ... but also has an additional Medium (wght=500, normalized 0.3333) master with
+        // a delta of 70; so at normalized 0.5 (wght=550), its delta will *not* be 60
+        assert_eq!(
+            hvar.advance_width_delta(GlyphId::new(2), &[F2Dot14::from_f32(0.5)])
+                .unwrap(),
+            Fixed::from_f64(83.0)
+        );
```
rsheeter, line 2274: Optional, I found this hard to read due to types (F2Dot14 didn't scream normalized coord at me), the name is clearer than the gid, and the interesting part is a bit obscured by things like Fixed coversion, building arrays of F2Dot14, etc. For me at least it would be more readable if some of the noise went away, e.g.:

```rust
// glyph_order can be obtained from the result of the compile call

assert_hvar_advance(glyph_order, vec![
 // .notdef has no variation
  (".notdef", NormalizedCoord::new(0.5), 0.0),
  (".notdef", NormalizedCoord::new(1.0), 0.0),
  // 'space' has two masters, with Bold 100 units wider than Regular
  ("space", NormalizedCoord::new(1.0), 100.0),
   // ... so it will be exactly 50 units wider half-way in between
  ("space", NormalizedCoord::new(0.5), 50.0),
]
```

rsheeter (approved):
LGTM, only complaint is a few places I found a touch hard to read

fontc/src/lib.rs
anthrotype: https://github.com/googlefonts/fontc/issues/526

fontc/src/lib.rs
anthrotype: Done in https://github.com/googlefonts/fontc/pull/490/commits/7bc827faf2a0f406d57aeaaa8a8b92ba90147599

fontbe/src/hvar.rs
anthrotype: Done in https://github.com/googlefonts/fontc/pull/490/commits/3f5e6f8092918929382344a25fd8b72002e8c766

fontbe/src/hvar.rs
```diff
@@ -0,0 +1,233 @@
...
+#[derive(Debug)]
+struct HvarWork {}
+
+pub fn create_hvar_work() -> Box<BeWork> {
+    Box::new(HvarWork {})
+}
+
+/// Compute the final size of a table, after it has been serialized to bytes
+fn table_size<T>(table: &T, context: &str) -> Result<usize, Error>
+where
```
cmyr, line 40: instead of passing in a string here, you could get the type name with the [`type_name`](http://doc.rust-lang.org/1.73.0/std/any/fn.type_name.html) fn, which would probably be suitable for the purposes of an error message?

fontbe/src/hvar.rs
```diff
@@ -0,0 +1,233 @@
...
+            // In fonttools, all master TTFs have a .notdef glyph as their first glyph; in fontc,
+            // unless the input source defines a .notdef, only a default instance is generated.
+            // And that's ok for gvar, however for HVAR the order in which regions and associated
+            // deltas are added to VariationStoreBuilder, one glyph at a time, can produce
+            // different orderings of the ItemVariationStore.VariationRegionList (newly seen
+            // regions get appended, and existing regions reused).
+            // So, to match the VarRegionList produced by fontTools, we need to make the deltaset
+            // for the first .notdef glyph similarly "dense", by copying its default instance to
+            // all other glyph locations...
+            if i == 0 && name.as_str() == ".notdef" {
```
cmyr, line 102: Instead of comparing with a string you can use the `NOTDEF` associated context that is defined on `GlyphName`, e.g. `name == GlyphName::NOTDEF`.

fontbe/src/hvar.rs
```diff
@@ -0,0 +1,233 @@
...
+        let direct_store = if glyph_width_deltas.is_single_model() {
+            let mut direct_builder = VariationStoreBuilder::new_with_implicit_indices();
+            for deltas in glyph_width_deltas.iter() {
+                var_idxes.push(direct_builder.add_deltas(deltas.clone()));
+            }
+            // sanity check
+            assert_eq!(
+                mem::take(&mut var_idxes),
+                (0..glyph_order.len() as u32).collect::<Vec<_>>()
+            );
```
cmyr, line 192: `mem::take` replaces `var_idxes` with an empty vector, which feels weird. is `var_idxes` only constructed for the purposes of this assert?

In any case, there are likely more efficient ways to write this. One simple one would be to avoid the vector, and instead just do the comparison with a length check followed by direct iteration, e.g.

```rust
assert_eq!(var_idxes.len(), glyph_order.len());
assert!(var_idxes.iter().enumerate().all(|(i, idx)| i as u32 == *idx));
```

fontbe/src/hvar.rs
```diff
@@ -0,0 +1,233 @@
...
+            var_idxes.push(indirect_builder.add_deltas(deltas.clone()));
+        }
+        let (indirect_store, varidx_map) = indirect_builder.build();
+
+        let varidx_map: DeltaSetIndexMap = var_idxes
+            .into_iter()
+            .map(|idx| {
+                varidx_map
+                    .get(idx)
+                    .unwrap_or_else(|| panic!("varidx_map missing index {}", idx))
```
cmyr, line 211: I think just `expect` is fine here? it feels strange to use `unwrap_or_else` in this way although I get the motivation (I think? to avoid the allocation of using `format!` in `expect`?)

cmyr (approved):
Some Rust style/idiom nits & notes inline, but overall this looks great!

fontbe/src/hvar.rs
anthrotype: yes, exactly that. This should never trigger basically, if the VariationStoreBuilder holds on to its contract

fontbe/src/hvar.rs
anthrotype: thanks, I applied most of your suggestions in the cleaned up PR => https://github.com/googlefonts/fontc/pull/527/commits/c724502fb6d34cdff6e7c239b2f8276eda130829

fontbe/src/hvar.rs
cmyr: It's an invariant that any temporary index returned by `add_deltas` will exist in the returned map; if this is ever not true then something is seriously wrong, so I would personally just `unwrap` or `expect` here, but this is fine.

anthrotype: ok, I'll unwrap() then

Issue #407, opened 2023-08-21
incorrect feature deltas when sparse


rsheeter:
As noted in https://github.com/googlefonts/fontc/pull/373#discussion_r1299409825, #373 will have the same problem as #400 for kerning.

anthrotype:
the same or equivalent patch as above should be applied to the `resolve_variable_metric` standalone function, which is the one used for the autogenerated kern/mark which we care the most

anthrotype:
Fixed by #779

Issue #84, opened 2023-01-11
Why panic when you can Result


rsheeter:
As https://github.com/googlefonts/fontmake-rs/pull/79/files/8c03ae050bb29a9552c94a7d0cb9b165600ab668#r1066299668 suggests, we have accumulated a number of panic or known to be fallible unwrap cases we probably want to make into errors.

"probably" because panic has one really nice characteristic: you get a useful callstack. Tracking down problems from errors seems to be less convenient. Perhaps that means error should https://doc.rust-lang.org/std/backtrace/struct.Backtrace.html#method.capture?

PR #492, opened 2023-10-19
First step toward marks

rsheeter:
Starting to work toward marks based on discussion with @anthrotype. Likely to be step one in a series of PRs.

Helps with https://github.com/googlefonts/fontc/issues/308.

Builds on @cmyr's work to avoid generating fea strings in #496. Rebased to sit on top of fea-rs merge into fontc and the feature writer api (#517).

Includes API sketch from https://github.com/cmyr/fea-rs/pull/243, updated per https://github.com/cmyr/fea-rs/pull/243#discussion_r1370696462.

`fontbe/src/features.rs` is the heart of it.

### DIFF
--- a/fea-rs/src/compile/feature_writer.rs
+++ b/fea-rs/src/compile/feature_writer.rs
@@ -108,10 +108,10 @@ impl<'a> FeatureBuilder<'a> {
     /// Add lookups to every default language system.
     ///
     /// Convenience method for recurring pattern.
-    pub fn add_to_default_language_systems(&mut self, feature_tag: Tag, lookups: &Vec<LookupId>) {
+    pub fn add_to_default_language_systems(&mut self, feature_tag: Tag, lookups: &[LookupId]) {
         for langsys in self.language_systems() {
             let feature_key = langsys.to_feature_key(feature_tag);
-            self.add_feature(feature_key, lookups.clone());
+            self.add_feature(feature_key, lookups.to_vec());
         }
     }
 
--- a/fea-rs/src/compile/feature_writer.rs
+++ b/fea-rs/src/compile/feature_writer.rs
@@ -2,14 +2,15 @@
 
 use std::collections::{BTreeMap, HashMap};
 
-use smol_str::SmolStr;
-use write_fonts::tables::{
-    gpos::AnchorTable,
-    layout::{LookupFlag, PendingVariationIndex},
-    variations::VariationRegion,
+use write_fonts::{
+    tables::{
+        layout::{LookupFlag, PendingVariationIndex},
+        variations::VariationRegion,
+    },
+    types::Tag,
 };
 
-use crate::common::{GlyphClass, MarkClass};
+use crate::common::GlyphClass;
 
 use super::{
     features::FeatureLookups,
@@ -30,7 +31,6 @@ pub struct FeatureBuilder<'a> {
     pub(crate) tables: &'a mut Tables,
     pub(crate) lookups: Vec<(LookupId, PositionLookup)>,
     pub(crate) features: BTreeMap<FeatureKey, FeatureLookups>,
-    pub(crate) mark_classes: HashMap<SmolStr, MarkClass>,
     mark_filter_sets: HashMap<GlyphClass, FilterSetId>,
     // because there may already be defined filter sets from the root fea
     filter_set_id_start: usize,
@@ -61,7 +61,6 @@ impl<'a> FeatureBuilder<'a> {
             tables,
             lookups: Default::default(),
             features: Default::default(),
-            mark_classes: Default::default(),
             mark_filter_sets: Default::default(),
             filter_set_id_start,
         }
@@ -75,22 +74,6 @@ impl<'a> FeatureBuilder<'a> {
     /// If the FEA text contained an explicit GDEF table block, return its contents
     pub fn gdef(&self) -> Option<&GdefBuilder> {
         self.tables.gdef.as_ref()
-    }
-
-    /// Define a new mark class, for use in mark-base and mark-mark rules.
-    ///
-    /// TODO: do we want to ensure there are no redefinitions? do we want
-    /// return an error? do we want to uphold any other invariants?
-    ///
-    /// ALSO: mark class IDs depend on definition order in the source.
-    /// I have no idea how best to approximate that in this API :/
-    pub fn define_mark_class(
-        &mut self,
-        class_name: impl Into<SmolStr>,
-        members: Vec<(GlyphClass, Option<AnchorTable>)>,
-    ) {
-        self.mark_classes
-            .insert(class_name.into(), MarkClass { members });
     }
 
     /// Create a new lookup.
@@ -120,6 +103,16 @@ impl<'a> FeatureBuilder<'a> {
     ) -> PendingVariationIndex {
         let delta_set_id = self.tables.var_store().add_deltas(deltas);
         PendingVariationIndex { delta_set_id }
+    }
+
+    /// Add lookups to every default language system.
+    ///
+    /// Convenience method for recurring pattern.
+    pub fn add_to_default_language_systems(&mut self, feature_tag: Tag, lookups: &Vec<LookupId>) {
+        for langsys in self.language_systems() {
+            let feature_key = langsys.to_feature_key(feature_tag);
+            self.add_feature(feature_key, lookups.clone());
+        }
     }
 
     /// Create a new feature, registered for a particular language system.
--- a/fea-rs/src/compile/variations.rs
+++ b/fea-rs/src/compile/variations.rs
@@ -27,7 +27,7 @@ pub trait VariationInfo {
     ) -> Result<(i16, Vec<(VariationRegion, i16)>), AnyError>;
 }
 
-type AnyError = Box<dyn std::error::Error>;
+pub type AnyError = Box<dyn std::error::Error>;
 
 /// A type that implements [`VariationInfo`], for testing and debugging.
 #[derive(Clone, Debug, Default)]
--- a/fontbe/src/error.rs
+++ b/fontbe/src/error.rs
@@ -67,6 +67,8 @@ pub enum Error {
     ExpectedAnchor(FeWorkId),
     #[error("No glyph id for '{0}'")]
     MissingGlyphId(GlyphName),
+    #[error("No glyph class '{0}'")]
+    MissingGlyphClass(GlyphName),
     #[error("Multiple assignments for class: {0:?}")]
     PreviouslyAssignedClass(PreviouslyAssignedClass),
 }
--- a/fontbe/src/error.rs
+++ b/fontbe/src/error.rs
@@ -1,9 +1,11 @@
 use std::{fmt::Display, io, path::PathBuf};
 
-use fea_rs::compile::error::CompilerError;
+use fea_rs::compile::{error::CompilerError, PreviouslyAssignedClass};
 use font_types::Tag;
 use fontdrasil::types::GlyphName;
-use fontir::{error::VariationModelError, variations::DeltaError};
+use fontir::{
+    error::VariationModelError, orchestration::WorkId as FeWorkId, variations::DeltaError,
+};
 use read_fonts::ReadError;
 use thiserror::Error;
 use write_fonts::tables::{
@@ -61,6 +63,12 @@ pub enum Error {
     FileExpected(PathBuf),
     #[error("Missing {0}")]
     MissingTable(Tag),
+    #[error("Expected an anchor, got {0:?}")]
+    ExpectedAnchor(FeWorkId),
+    #[error("No glyph id for '{0}'")]
+    MissingGlyphId(GlyphName),
+    #[error("Multiple assignments for class: {0:?}")]
+    PreviouslyAssignedClass(PreviouslyAssignedClass),
 }
 
 #[derive(Debug)]
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -32,7 +32,6 @@ use fontdrasil::{
     orchestration::{Access, Work},
     types::GlyphName,
 };
-use smol_str::SmolStr;
 use write_fonts::{
     tables::layout::LookupFlag,
     tables::{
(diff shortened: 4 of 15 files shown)


fea-rs/src/compile/feature_writer.rs
```diff
@@ -86,11 +90,40 @@ impl<'a> FeatureBuilder<'a> {
     /// I have no idea how best to approximate that in this API :/
     pub fn define_mark_class(
         &mut self,
-        class_name: impl Into<SmolStr>,
-        members: Vec<(GlyphClass, Option<AnchorTable>)>,
+        members: impl Into<GlyphClass>,
+        anchors: HashMap<NormalizedLocation, (i16, i16)>,
+    ) {
```
rsheeter, line 95: @cmyr I wonder if something in this direction could work?

cmyr: a mark class can have members that have different anchor positions.

rsheeter: resolving as I will update per vc

fontbe/src/features.rs
```diff
@@ -113,39 +125,82 @@ impl<'a> FeaVariationInfo<'a> {
     }
 }
 
+/// When generating mark features it's handy to know what is a base and what is a mark
+#[derive(Copy, Clone, Debug, PartialEq)]
+enum AnchorType {
+    Base,
+    Mark,
+}
```
cmyr, line 133:
```suggestion
/// The type of an anchor, used when generating mark features
#[derive(Copy, Clone, Debug, PartialEq)]
enum AnchorType {
    Base,
    Mark,
}
```

fontbe/src/features.rs
```diff
@@ -113,39 +125,82 @@ impl<'a> FeaVariationInfo<'a> {
...
+
+impl AnchorType {
+    fn group_name(&self, glyph_name: GlyphName) -> GlyphName {
+        match self {
+            AnchorType::Base => glyph_name,
+            AnchorType::Mark => glyph_name.as_str()[1..].into(),
```
cmyr, line 139: so looking below, it seems that every time we have an `AnchorType` we call this method several times. Would it make sense to compute that once when we create the enum, and store it there?

fontbe/src/features.rs
```diff
@@ -113,39 +125,82 @@ impl<'a> FeaVariationInfo<'a> {
...
             glyph_map,
+            raw_anchors,
         }
     }
 
+    fn glyph_id(&self, glyph_name: &GlyphName) -> Option<GlyphId> {
+        self.glyph_map
+            .glyph_id(glyph_name)
+            .map(|val| Some(GlyphId::new(val as u16)))
+            .unwrap_or_default()
```
cmyr, line 192: There's an extra option or something in here, I think it should just be,
```suggestion
        self.glyph_map
            .glyph_id(glyph_name)
            .map(|val| GlyphId::new(val as u16))
```

fontbe/src/features.rs
```diff
@@ -113,39 +125,82 @@ impl<'a> FeaVariationInfo<'a> {
...
         }
 
         // a little helper closure used below
-        let name_to_gid = |name| {
-            self.glyph_map
-                .glyph_id(name)
-                .map(|val| GlyphId::new(val as u16))
-                .unwrap_or(GlyphId::NOTDEF)
-        };
+        let name_to_gid = |name| self.glyph_id(name).unwrap_or(GlyphId::NOTDEF);
```
cmyr, line 203: if this is the only rationale for the method than we can just delete this closure and make the method return `NOTDEF` if the glyph is missing (or error, since this doesn't seem like an expected outcome?)

fontbe/src/features.rs
```diff
@@ -219,14 +274,171 @@ impl<'a> FeatureWriter<'a> {
...
+            }
+        }
+
+        // Build the actual mark base and mark mark constructs using fea-rs builders
+
+        let mut mark_base_lookups = Vec::new();
+        let mut mark_mark_lookups = Vec::new();
+
+        for (group_name, group) in groups.iter() {
+            let mark_class_name: SmolStr = format!("MC_{group_name}").into();
```
cmyr, line 340: I think you should be able to just use the group name directly now.

cmyr (commented):
some observations inline but looks like we're very close!

fontbe/src/features.rs
rsheeter: Good catch, I think I modified it a couple times and left it weird

fontbe/src/features.rs
rsheeter: Making the method silently return NOTDEF seems off; this seems like it should be an error.

fontbe/src/features.rs
rsheeter: Changed to not take ownership

fontbe/src/features.rs
rsheeter: Done. Making it a SmolStr still ends up needing an into.

fontbe/src/features.rs
```diff
@@ -219,13 +317,153 @@ impl<'a> FeatureWriter<'a> {
...
+                };
+                if !glyph_anchors
+                    .anchors
+                    .iter()
+                    .any(|a| matches!(AnchorInfo::new(&a.name), AnchorInfo::Mark(..)))
+                {
+                    continue;
+                }
+                let mark_gid = self.glyph_id(mark_name)?;
+                let base_name = AnchorInfo::new(&mark_anchor.name).group_name();
```
anthrotype, line 400: isn't this `base_name` the same as `group_name` local variable defined in the outer for loop above?

fontbe/src/features.rs
```diff
@@ -322,13 +555,7 @@ impl<'a> VariationInfo for FeaVariationInfo<'a> {
...
         values: &HashMap<NormalizedLocation, i16>,
-    ) -> Result<
-        (
-            i16,
-            Vec<(write_fonts::tables::variations::VariationRegion, i16)>,
-        ),
-        Box<(dyn StdError + 'static)>,
-    > {
+    ) -> Result<(i16, Vec<(VariationRegion, i16)>), Box<(dyn StdError + 'static)>> {
         let var_model = &self.static_metadata.variation_model;
```
anthrotype, line 559: this is assuming that a single global VariationModel applies to all glyphs which isn't necessarily the case. We should make sparse models for each unique set of locations  (similarly to how we did for gvar and HVAR), otherwise the generated regions will be incorrect

anthrotype: Same as https://github.com/googlefonts/fontc/issues/407

anthrotype (commented):
nice to see this is making progress! left some comment

fontc/src/lib.rs
```diff
@@ -2289,4 +2290,89 @@ mod tests {
...
+        let buf = fs::read(font_file).unwrap();
+        let font = FontRef::new(&buf).unwrap();
+
+        let gpos = font.gpos().unwrap();
+
+        let base_gid = GlyphId::new(result.get_glyph_index("A") as u16);
+        let macroncomb_gid = GlyphId::new(result.get_glyph_index("macroncomb") as u16);
+        let brevecomb_gid = GlyphId::new(result.get_glyph_index("brevecomb") as u16);
+
+        // If only we had more indirections
```
cmyr, line 2319: *apologies that this turned into a largely unrelated essay*

so the types we use to represent lookups are annoying; it was one of the trickier parts of the whole scheme. They also haven't been exercised much, since we don't actually have (say) a shaper.

(One of the worst parts of this is the extensions, which have a different type and an *additional* level of indirection)

In any case, I definitely think there's room to add some methods here that would make this more convenient, and I'm trying to think what exactly would be best given the underlying structure.

I'm going to write up my thoughts here, and then i can convert it into an issue.

Ultimately all we really care about is:
- getting the lookupflag
- getting the markFilteringSet, if present
- getting access to the subtables
- being able unify extension and non-extension lookups (e.g. return the same thing for a pairpos lookup as an extension lookup with pairpos subtables)

Our current structure:

- a generic `Lookup<T>` type, where `T` is the type of the subtables
- a pair of generic `ExtensionLookup<T>` types (one each for GSUB/GPOS) where `T` is again the type of the subtable
- a pair of enums `PositionLookup` and `SubstitutionLookup` with a variant for each lookup type, with each variant wrapping a `Lookup<T>` where `T` is the subtable type for that variant.
- another pair of enums for each extension type


Let's imagine we want to simplify this so that, in the general case, the user only needs to think about the top-level `PositionLookup` and `SubstitutionLookup` types.

We can have lookupFlag & markFilteringSet fields on this enum.
Then to make accessing the subtables easy, I can current envision two versions of the API:

- we could just have one method for each format, e.g.

```rust
impl<'a> PositionLookup<'a> {
    single_pos_subtables(&self) -> Option<ArrayOfOffsets<'a, SinglePos<'a>, Offset16>> { .. }
}
```

(except this won't exactly work, since extension and non-extensions have a different offset type, which means we don't have a unified return type here; we'd need to introduce an enum to hide those two possibilities)

- *or* we could have *another* enum, with a variant for each non-extension lookup, that would look something like,

```rust
impl<'a> PositionLookup<'a> {
    fn subtables(&self) -> PositionSubtables<'a> { .. }
}

enum PositionSubtables<'a> {
    SinglePos(SomethingThatActsLikeAnArray<SinglePos<'a>>),
    PairPos(SomethingThatActsLikeAnArray<PairPos<'a>>),
    ...
}
```

The idea here is to basically change the level of granularity of the generics from the top-level lookup object down to the subtables themselves, at least from the perspective of the API.


How this would relate to the current code is unclear. It would be worth at least considering what this would look like as a systematic change, but it might be more pragmatic to treat it as syntactic sugar, on top of the existing types.

fontir/src/ir.rs
```diff
@@ -831,6 +831,16 @@ pub struct Anchor {
     pub positions: HashMap<NormalizedLocation, Point>,
 }
 
+impl Anchor {
+    pub fn default_pos(&self) -> Point {
+        self.positions
+            .iter()
+            .find(|(loc, _)| !loc.has_any_non_zero())
```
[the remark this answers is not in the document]
cmyr: Are these methods functionally equivalent? If they are we should remove one of them?

fontbe/src/features.rs
```diff
@@ -180,19 +268,27 @@ impl<'a> FeatureWriter<'a> {
...
+                    let gid0 = self.glyph_id(left)?;
+                    let gid1 = self.glyph_id(right)?;
                     ppos_subtables.insert_pair(gid0, x_adv_record, gid1, empty_record);
                 }
                 (KernParticipant::Group(left), KernParticipant::Group(right)) => {
-                    let left = glyph_classes.get(left).unwrap().clone();
-                    let right = glyph_classes.get(right).unwrap().clone();
+                    let left = glyph_classes
+                        .get(left)
+                        .ok_or_else(|| Error::MissingGlyphId(left.clone()))?
```
cmyr, line 278: I wonder if we could pre-validate the kerning info at an earlier point in the pipeline?

cmyr (approved):
Some notes inline but I think we should get this in and iterate from here!

fontc/src/lib.rs
rsheeter: Perhaps this should become a fontations issue?

It's largely not that bad in that 1) it matches the table structure and 2) it's typed ... EXCEPT the FontData you have to pass in all of a sudden.

fontbe/src/error.rs
```diff
@@ -61,6 +63,14 @@ pub enum Error {
     FileExpected(PathBuf),
     #[error("Missing {0}")]
     MissingTable(Tag),
+    #[error("Expected an anchor, got {0:?}")]
+    ExpectedAnchor(FeWorkId),
+    #[error("No glyph id for '{0}'")]
+    MissingGlyphId(GlyphName),
```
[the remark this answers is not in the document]
rsheeter: I would very much like our errors to capture backtraces, not having that makes troubleshooting much harder in some situations. I believe https://doc.rust-lang.org/std/backtrace/struct.Backtrace.html is now stable so we could do it.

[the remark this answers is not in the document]
rsheeter: See https://github.com/googlefonts/fontc/issues/84

fontc/src/lib.rs
```diff
@@ -2289,4 +2290,89 @@ mod tests {
...
     }
+
+    fn anchor_coords(at: AnchorTable) -> (i32, i32) {
+        match at {
+            AnchorTable::Format1(at) => (at.x_coordinate() as i32, at.y_coordinate() as i32),
+            AnchorTable::Format2(at) => (at.x_coordinate() as i32, at.y_coordinate() as i32),
+            AnchorTable::Format3(at) => (at.x_coordinate() as i32, at.y_coordinate() as i32),
+        }
+    }
+
```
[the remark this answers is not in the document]
rsheeter: I love the idea of doing that systematically. I don't want to block this PR on that so resolving here; we can adopt the new fontations methods in a follow-on.

fontc/src/lib.rs
rsheeter: It appears this si now https://github.com/googlefonts/fontations/issues/683

fea-rs/src/compile/variations.rs
```diff
@@ -27,7 +27,7 @@ pub trait VariationInfo {
     ) -> Result<(i16, Vec<(VariationRegion, i16)>), AnyError>;
 }
 
-type AnyError = Box<dyn std::error::Error>;
+pub type AnyError = Box<dyn std::error::Error>;
```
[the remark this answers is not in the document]
rsheeter: https://github.com/googlefonts/fontc/issues/535 for associated type for the error

fontbe/src/features.rs
```diff
@@ -219,13 +317,153 @@ impl<'a> FeatureWriter<'a> {
...
+
+        let mut mark_base_lookups = Vec::new();
+        let mut mark_mark_lookups = Vec::new();
+
+        for (group_name, group) in mark_groups.iter() {
+            // if we have bases *and* marks produce mark to base
+            if !group.bases.is_empty() && !group.marks.is_empty() {
+                for (base_name, base_anchor) in group.bases.iter() {
+                    let base_gid = self.glyph_id(base_name)?;
+                    let mark_class: SmolStr = format!("MC_{}", base_anchor.name).into();
```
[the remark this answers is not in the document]
rsheeter: Good catch. I think the MC_ prefix doesn't matter -  since we aren't generating fea anymore; we can just use group_name.0 directly.

fontbe/src/features.rs
rsheeter: Noted on #407. Baby steps :)

fontir/src/ir.rs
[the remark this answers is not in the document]
rsheeter: Ha, yes they are. I don't think is_default existed when I started roughing in marks.

fontbe/src/features.rs
rsheeter: Yes.

rsheeter: Fixed.

fontc/src/lib.rs
```diff
@@ -2289,4 +2290,89 @@ mod tests {
...
+            )
+            .collect::<Vec<_>>();
+
+        assert_eq!(
+            (
+                vec![(base_gid, (300, 700))],
+                vec![(macroncomb_gid, (300, 600)), (brevecomb_gid, (200, 500))]
+            ),
+            (bases, marks)
+        );
```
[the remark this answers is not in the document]
rsheeter: Added the class check, ty

PR #496, opened 2023-10-20
Use fea-rs feature writer API for kerning

cmyr:
This is a WIP; in particular I'm not sure how we should be passing errors around. That said this is working for features, and we can iterate from here.

@rsheeter I think this should illustrate the overall approach? I'll also update the fea-rs PR (https://github.com/cmyr/fea-rs/pull/241) to make the various mark builders public.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,3 +48,6 @@ members = [
     "ufo2fontir",
     "fontc",
 ]
+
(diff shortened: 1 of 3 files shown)


rsheeter:
Migrated to #517 which now has this and the fea-rs changes together.

PR #498, opened 2023-10-23
[ttx_diff.py] remove fontmake's --no-production-names flag

anthrotype:
After PR #489 got merged, fontc should use the same glyph postscript names as fontmake does, so we no longer need to pass the --no-production-names flag to compare the two sets of fonts


rsheeter:
Nice! LGTM (not logged in on phone so I can't give a proper approval)

anthrotype:
after this, the only remaining flag to help with ttx-diffing is `--drop-implied-oncurves` (and this is going to stay, as it's unlikely fontmake will change its default setting for backward compat reasons)

PR #500, opened 2023-10-23
[ttx_diff.py] pass --no-production-names until supported for both .glyphs & DS

anthrotype:
https://github.com/googlefonts/fontc/issues/248#issuecomment-1775429891

JMM

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -99,7 +99,7 @@ def build_fontc(source: Path, build_dir: Path, compare: str):
         "--",
         # uncomment this to compare output w/ fontmake --keep-direction
         # "--keep-direction",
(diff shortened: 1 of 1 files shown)


anthrotype:
there may be a case for keeping --no-production-names flag regardless of whether support for .glyphs input is implemented, since it aids with debugging. For the human reader, it is easier to indentify a diff related to a specific glyph if  the latter is named "acutecomb" instead of "uni0301"...

rsheeter:
As a human reader I would very much prefer to see acutecomb in the diff :)

rsheeter (approved):
LGTM

anthrotype:
> As a human reader I would very much prefer to see acutecomb in the diff :)

in that case I shall remove the TODO and explain why we prefer to always pass --no-production-names in ttx_diff.py

rsheeter (approved):
LGTM

PR #501, opened 2023-10-23
Don't lose MiscMetadata in serde

### LINKED ISSUE #499: `fontir::MiscMetadata` isn't deserialized as part of `StaticMetadata`
Lifting this out from [this PR comment](https://github.com/googlefonts/fontc/pull/489#discussion_r1368556358) by @Hoolean

`StaticMetadata::new` is used for the conversion from `StaticMetadataSerdeRepr` (contains `MiscMetadataSerdeRepr`) to `StaticMetadata`, but the `misc` field isn't copied across, or re-created within the constructor

https://github.com/googlefonts/fontc/blob/0cafae418b2965fe4bb25b67481d4ff023a18eb2/fontir/src/serde.rs#L67-L79

rsheeter:
Add a test to expose #499 and fix it. Fixes #499.

While we're here, coalesce  `axes_default: NormalizedLocation, variable_axes_default: NormalizedLocation,` to a single field with the default location as that's what the consuming code expects.

JMM

### DIFF
--- a/fontir/src/coords.rs
+++ b/fontir/src/coords.rs
@@ -242,6 +242,12 @@ impl<T: Copy> FromIterator<(Tag, T)> for Location<T> {
     }
 }
 
+impl<T: Copy> From<Vec<(Tag, T)>> for Location<T> {
+    fn from(value: Vec<(Tag, T)>) -> Self {
+        Location::<T>::from_iter(value)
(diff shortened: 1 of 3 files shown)


cmyr (approved):
looks reasonable!

PR #503, opened 2023-10-24
Move coords to fontdrasil so fea-rs can agree with us on what NormalozedLocation looks like

rsheeter:
Per discussion with @cmyr fea-rs will gain a dependency on fontdrasil and then we can share types. Meant to make builder and variation interaction easier.


cmyr (approved):
looks good, will update fea-rs shortly.

PR #505, opened 2023-10-25
Merge fea-rs repo

cmyr:
This merges the history of the fea-rs repo, which will be subsequently archived.

After this is merged I will:

- [ ] make sure that fea-rs is setting up its venv during testing correctly
- [ ] make fea-rs use the workspace deps where appropriate
- [ ] make crates here depend on the local version of fea-rs
- [ ] update the metadata & links in the fea-rs cargo.toml and readme
- [ ] fix anything else that's broken?


rsheeter (approved):
LGTM. I think you can delete gitmodules?

Also ignoring _experiments is extremely Colin-specific; I suggest we dont' keep that change.

PR #506, opened 2023-10-25
[ci] Remove 'no println' check

cmyr:
This was a nice idea but has caused several headaches, and the problem it solves is ultimately a fairly minor one.


rsheeter:
I find this check quite useful and would prefer not to give it up. When do we *need* eprints?

PR #513, opened 2023-10-25
[chore] Update manifests post fea-rs merge

cmyr:
This makes fea-rs use the workspace deps where possible, and makes fontbe use the local fea-rs.

It also makes a few other updates to update paths & urls in fea-rs/Cargo.toml, and it bumps a few out of date dependencies.

### DIFF
--- a/fea-rs/Cargo.toml
+++ b/fea-rs/Cargo.toml
@@ -4,30 +4,31 @@ version = "0.18.0"
 license = "MIT/Apache-2.0"
 authors = ["Colin Rofls <colin@cmyr.net>"]
 description = "Tools for working with Adobe OpenType Feature files."
-repository = "https://github.com/cmyr/fea-rs"
+repository = "https://github.com/googlefonts/fontc"
 categories = ["parsing", "text-processing"]
 keywords = ["fonts", "opentype"]
-readme = "../README.md"
+readme = "./README.md"
 edition = "2021"
 default-run = "fea-rs"
 exclude = ["test-data"]
 
 [dependencies]
 ansi_term = "0.12.1"
-smol_str = "0.2.0"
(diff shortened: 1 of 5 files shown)


fea-rs/Cargo.toml
```diff
@@ -4,30 +4,31 @@ version = "0.18.0"
...
-smol_str = "0.2.0"
 norad = { version = "0.12", optional = true } # Used in the compile binary
-write-fonts = { version = "0.17.0", features = ["read"] }
-chrono = "0.4.3"
 diff = { version = "0.1.12", optional = true }
 rayon = { version = "1.6", optional = true }
 serde = { version = "1.0.147", features = ["derive"], optional = true }
 serde_json = {version = "1.0.87", optional = true }
-thiserror = "1.0.37"
 clap = { version = "4.0.32", features = ["derive"], optional = true }
```
rsheeter, line 22: I think several more can move to workspace if we follow the current practice of doing so as soon as more than one crate uses something:

* fontc also uses both clap and rayon so they should move
* serde is in the workspace
* ufo2fontir uses norad

rsheeter (approved):
LGTM, just need a few more workspace entries

fea-rs/Cargo.toml
cmyr: the issue is that in fea-rs these are optional, whereas they aren't in the other crates, so I don't think this works?

rsheeter: You can combine workspace and optional, https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

cmyr: hey great, TIL.

PR #514, opened 2023-10-25
Add an allowlist for prints and permit a few

rsheeter:
A few remain; I believe these need to either go away or move to `fea-rs/src/bin/` (e.g. the compile and print warnings fn)

```shell
$ resources/scripts/check_no_println.sh
Error: The following Rust source files contain println! or eprintln! statements:
fea-rs/src/compile/compiler.rs:182:             eprintln!("{}", tree.format_diagnostic(&w, is_tty));
fea-rs/src/token_tree.rs:327:                     eprintln!(
fea-rs/src/token_tree/rewrite.rs:134:         eprintln!("\"");
Please remove or comment out the println! and eprintln! statements before pushing.
```

### DIFF
--- a/resources/scripts/check_no_println.sh
+++ b/resources/scripts/check_no_println.sh
@@ -79,6 +79,21 @@ else
     matches=$(diff_rust_lines_with_prints "$from_ref" "$to_ref")
 fi
 
+# https://stackoverflow.com/q/1527049
+function join() {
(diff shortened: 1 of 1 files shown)


cmyr (approved):
great, happy to get this in as-is and then i'll fixup the remaining issues

PR #515, opened 2023-10-25
[ci] Address remaining eprintln statements in fea-rs

cmyr:
- two of these were debug helpers, which I can comment out
- one of these is (imo) legitimate use, reporting warnings to the user, and I've added it to the allowlist

### DIFF
--- a/resources/scripts/check_no_println.sh
+++ b/resources/scripts/check_no_println.sh
@@ -89,6 +89,7 @@ function join() {
 allowlist=()
 allowlist+=("fea-rs/src/bin/")
 allowlist+=("fea-rs/src/util/ttx.rs")
+allowlist+=("fea-rs/src/tests/parse.rs")
 
 allowlist=$(join "|" "${allowlist[@]}")
 echo grep -v "($allowlist)"
--- a/fea-rs/src/compile/compiler.rs
+++ b/fea-rs/src/compile/compiler.rs
@@ -2,6 +2,7 @@
 
 use std::{
     ffi::OsString,
+    io::Write,
     path::{Path, PathBuf},
 };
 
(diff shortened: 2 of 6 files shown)


resources/scripts/check_no_println.sh
```diff
@@ -89,6 +89,7 @@ function join() {
 allowlist=()
 allowlist+=("fea-rs/src/bin/")
 allowlist+=("fea-rs/src/util/ttx.rs")
+allowlist+=("fea-rs/src/compile/compiler.rs")
```
rsheeter, line 92: I don't think I agree with this one; the compiler can be used as a library and the user doesn't necessarily want it writing to stderr. I suggest we migrate the print_warnings_return_errors to the binary instead.

cmyr: there is an option to disable printing of warnings when you configure the compiler, so the user can control this as desired.

I do agree that this doesn't feel like the perfect API, but I've struggled to figure out a better one, and I would like to get CI working again without doing a large refactor. I could just disable this printing altogether for the time being, but that also feels bad...

rsheeter: Oh wait, there is a flag.

What if we make the flag default to false and then use the `writeln!(std::io::stderr()` trick to avoid the allowlist?

rsheeter (approved):
LGTM with one concern about print_warnings_return_errors being on by default

resources/scripts/check_no_println.sh
cmyr: sure that works. I prefer default `true` out of what is basically an ethical belief in the importance of warnings, but not going to die on that hill..

PR #516, opened 2023-10-25
fea-rs => fontc coords

rsheeter:
* Make fontbe depend on the local fea-rs
* Share coord types
* Simplify the VariationInfo trait as a result. 

Fixes https://github.com/cmyr/fea-rs/issues/242

Note that I retained the fea-rs AxisLocation to represent the raw value parsed. It has to be turned into a fontdrasil coord to be able to change types.

### DIFF
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -8,6 +8,7 @@ use std::{
     ops::Range,
 };
 
+use fontdrasil::coords::{DesignCoord, NormalizedCoord, NormalizedLocation, UserCoord};
 use smol_str::SmolStr;
 use write_fonts::{
     tables::{
@@ -18,7 +19,7 @@ use write_fonts::{
             ConditionFormat1, ConditionSet, FeatureVariations, LookupFlag, PendingVariationIndex,
         },
     },
-    types::{Fixed, NameId, Tag},
+    types::{F2Dot14, NameId, Tag},
 };
 
 use crate::{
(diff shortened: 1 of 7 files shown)


fea-rs/src/compile/variations.rs
```diff
@@ -17,32 +15,29 @@ use write_fonts::{
...
     /// If the tag is an axis in this font, return the min/max values from fvar
-    fn axis_info(&self, axis_tag: Tag) -> Option<AxisInfo>;
-
-    /// Return the normalized value for a user coordinate for the given axis.
-    ///
-    /// NOTE: This is only used for computing the normalized values for ConditionSets.
-    ///
-    /// NOTE: if unit suffixes become a thing, and ConditionSets remain a thing, then
-    /// this should use the same AxisLocation enum as below.
-    fn normalize_coordinate(&self, axis_tag: Tag, value: Fixed) -> F2Dot14;
```
cmyr, line 28: this is still something that we would need for the proposed variable syntax extension for FEA, specifically to support feature variations (it's unrelated to variable value records etc)

fea-rs/src/compile/compile_ctx.rs
```diff
@@ -1129,18 +1130,29 @@ impl<'a> CompilationCtx<'a> {
...
-                (user_loc, value)
-            })
-            .collect::<HashMap<_, _>>();
+        let mut locations = HashMap::new();
+        for metric_loc in metric.location_values() {
+            let mut pos = NormalizedLocation::new();
+            for axis_value in metric_loc.location().items() {
+                let tag = axis_value.axis_tag().to_raw();
+                let Some((_, axis)) = var_info.axis(tag) else {
+                    self.error(metric.range(), format!("unrecognized tag '{tag}'"));
```
cmyr, line 1139: We don't need to error here, since we've already checked this condition in the validation pass. The idea that once you get to the compile pass you can (generally) just focus on building up the data structures.

There are a few exceptions regarding things that are not available in the validation pass, but this is not one of them.

fea-rs/src/compile/variations.rs
```diff
@@ -17,32 +15,29 @@ use write_fonts::{
...
+    fn axis(&self, axis_tag: Tag) -> Option<(usize, &Axis)>;
```
cmyr, line 18: I'd clarify the docs to say that it returns the axis as well as the index in of the axis as it will appear in fvar (and elsewhere?)

cmyr (approved):
A couple notes inline, but overall nice to see this simplification!

fea-rs/src/compile/variations.rs
rsheeter: Good catch, adding.

fea-rs/src/compile/compile_ctx.rs
rsheeter: Added a note it's definitely valid and unwrapped

fea-rs/src/compile/variations.rs
rsheeter: It's done slightly differently now, for example https://github.com/googlefonts/fontc/blob/2cfc562e421647628ac5669fca9c62fae805adbb/fea-rs/src/compile/compile_ctx.rs#L1143-L1151 supports all types of location and converts them to normalized. If we need that in several places we can hoist it out. I will leave such hoisting for future as I dont' currently have a second location that needs it.

PR #517, opened 2023-10-26
feature writer api

rsheeter:
Migrate https://github.com/cmyr/fea-rs/pull/241 to fontc. Rebase on #516; that should be merged first.

Fix a test that took up failing because we assumed no user features meant no features at all and that stopped being true when we added feature writers.

Copied from fea-rs, @cmyr's notes:

This PR is a start on providing API that allows the client to specify additional feature code at compile time, without needing to use FEA as an IR.

This involves:
- letting the client provide us with a `FeatureWriter` type during compilation
- calling this type after the compilation pass, but before assembling the final tables
- letting it add new lookups and features
- and merging these features and lookups into the overall set before building the final tables

It also involves making certain internal types public, such as some of our lookup builders.

This is a WIP. Currently unresolved:

- how we handle errors
- merging lookups in any way that is not just concatenation
- support for mark rules

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -1,7 +1,7 @@
 //! Feature binary compilation.
 
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     error::Error as StdError,
     ffi::{OsStr, OsString},
     fmt::Display,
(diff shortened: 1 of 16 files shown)


cmyr:
okay so this is basically the code from the https://github.com/cmyr/fea-rs/pull/241 plus the code from #496 ?

cmyr (approved):
Okay, looks good. Some remaining todos:

- figure out our error story here: it's sort of annoying since fea-rs is calling us back, and we can produce an error, that fea-rs is then going to want to return back to us? So maybe we want an associated error type on the trait.
- figure out an API that doesn't require the caller to manually add deltas etc?

fontbe/src/features.rs
```diff
@@ -111,6 +113,190 @@ impl<'a> FeaVariationInfo<'a> {
...
+                            empty_record.clone(),
+                        );
+                    }
+                }
+            }
+        }
+
+        // now we have a builder for the pairpos subtables, so we can make
+        // a lookup:
+        let lookup_id = builder.add_lookup(LookupFlag::empty(), None, vec![ppos_subtables]);
```
anthrotype, line 222: ufo2ft kernFeatureWriter sets IgnoreMarks lookupflag by default on its kerning lookups, unless there are kerning pairs between glyphs classified respectively as bases and marks (or marks vs marks); the latter are split into separate lookups where IgnoreMarks lookupflag is _not_ set.

fontbe/src/features.rs
```diff
@@ -111,6 +113,190 @@ impl<'a> FeaVariationInfo<'a> {
     }
 }
 
+struct FeatureWriter<'a> {
```
anthrotype, line 116: wouldn't it be better to have distinct feature writers for kern on the one hand and mark/mkmk on the other? As we implement the full logic of the python equivalent this single struct may get become too complex..

anthrotype (commented):
This looks good 👍  so much better than writing strings

fontbe/src/features.rs
cmyr: The way the API currently works, from rust, is that you need to provide a single type that gets the callback; but there's no reason we couldn't split it up on the fontc side.

fwiw this patch was really written as a proof of concept, just to see if the API would work. Happy to see this improved upon going forward :)

rsheeter: I'm hoping that some of the work can be done separately so the implementation here is basically go get the kerning (or w/e) we computed earlier and return it.

rsheeter: And a big +1 to refinement over time, hopefully in small increments now we've got fea-rs in the same repo so we can iterate on fontbe and fea-rs in a single PR

PR #518, opened 2023-10-26
Remove unused quick-xml

anthrotype:
it's a norad dependency but we don't import it directly anywhere in fontc (`git grep quick_xml` returns nothing) so there's no need to have it here, is there?


anthrotype:
println! checks are failing for unrelated reasons

cmyr (approved):

PR #534, opened 2023-10-31
[fea-rs] Rename gsub/gpos module to gsub/gpos{_builders}

cmyr:
The shorter names were okay when fea-rs was in its repo, but it was getting annoying to use things like fzf to try and find them here.

JMM


rsheeter (approved):
LGTM, didn't merge due to conflicts

PR #536, opened 2023-11-02
[fea-rs] Tweak logic for assigning glyph class ids

cmyr:
We previously matched otlLib, but that order is slightly funny and doesn't match what varLib does, and varLib is more relevant for fontc.

I expect that otlLib will be update shortly to match this behaviour in any case.

### DIFF
--- a/fea-rs/src/compile/lookups/helpers.rs
+++ b/fea-rs/src/compile/lookups/helpers.rs
@@ -58,11 +58,13 @@ impl ClassDefBuilder2 {
     /// to the final class ids
     pub(crate) fn build(self) -> (ClassDef, HashMap<GlyphClass, u16>) {
         let mut classes = self.classes.into_iter().collect::<Vec<_>>();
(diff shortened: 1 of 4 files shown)


fea-rs/src/compile/lookups/helpers.rs
```diff
@@ -59,10 +59,10 @@ impl ClassDefBuilder2 {
     pub(crate) fn build(self) -> (ClassDef, HashMap<GlyphClass, u16>) {
```
rsheeter, line 59: Where we deliberately match python code lets add a permalink to the python equivalent, I think that's true here and maybe elsewhere in this code?

For example,  https://github.com/googlefonts/fontations/pull/685/files does this.

rsheeter (approved):

anthrotype:
@cmyr fonttools 4.44.0 is out with the class order fix
https://github.com/fonttools/fonttools/releases/tag/4.44.0

PR #538, opened 2023-11-03
Add support for intermediate (aka "brace") layers from Glyphs sources

### LINKED ISSUE #62: Support sparse glyphs in .glyphs files
We see layer identifiers in Texturina that don't resolve. When this occurs it appears there is an associated master id:

```
associatedMasterId = "C4D96FDC-31F2-4CF4-BEA5-4CCE99FECD99";
layerId = "1FFA553E-9323-49A1-BA35-824C5E633887";
```

Not quite sure what this is; a "virtual layer"? Could also be associatedMasterId is unrelated, it just _sounds_ related and sits right beside layer ids :)

anthrotype:
The v2 notation is up-converted to equivalent layer attributes' coordinates as found in v3 sources.

We already have test files for glyphs{2,3}/IntermediateLayer.glyphs, because we checked them in when we added support for sparse DS sources, so the tests simply check that we get the same result whether we build from glyphs2, glyphs3 or DS sources (the latter were originally converted from .glyphs using `fontmake -o ufo`).

Fixes #62

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -192,7 +192,7 @@ impl Layer {
     }
 
     pub fn is_intermediate(&self) -> bool {
-        self.associated_master_id.is_some() && self.attributes.coordinates.is_some()
+        self.associated_master_id.is_some() && !self.attributes.coordinates.is_empty()
     }
 
     // TODO add is_alternate, is_color, etc.
@@ -200,14 +200,14 @@ impl Layer {
 
 #[derive(Clone, Default, Debug, PartialEq, Hash)]
 pub struct LayerAttributes {
-    pub coordinates: Option<Vec<OrderedFloat<f64>>>,
+    pub coordinates: Vec<OrderedFloat<f64>>,
     // TODO: add axisRules, color, etc.
 }
 
 // hand-parse because they can take multiple shapes
 impl FromPlist for LayerAttributes {
     fn parse(tokenizer: &mut Tokenizer<'_>) -> Result<Self, crate::plist::Error> {
-        let mut coordinates = None;
+        let mut coordinates = Vec::new();
 
         tokenizer.eat(b'{')?;
 
@@ -220,7 +220,7 @@ impl FromPlist for LayerAttributes {
             tokenizer.eat(b'=')?;
             match key.as_str() {
                 "coordinates" => {
-                    coordinates = Some(tokenizer.parse()?);
+                    coordinates = tokenizer.parse()?;
                 }
                 // skip unsupported attributes for now
                 // TODO: match the others
@@ -569,7 +569,7 @@ impl RawLayer {
                 .unwrap_or_default();
         }
         if !brace_coordinates.is_empty() {
-            self.attributes.coordinates = Some(brace_coordinates);
+            self.attributes.coordinates = brace_coordinates;
         }
         // TODO: handle 'bracket' layers and other attributes
     }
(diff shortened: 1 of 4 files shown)


glyphs-reader/src/font.rs
```diff
@@ -179,9 +179,58 @@ pub struct Glyph {
...
+    pub fn is_intermediate(&self) -> bool {
+        self.associated_master_id.is_some() && self.attributes.coordinates.is_some()
+    }
+
+    // TODO add is_alternate, is_color, etc.
+}
+
+#[derive(Clone, Default, Debug, PartialEq, Hash)]
+pub struct LayerAttributes {
+    pub coordinates: Option<Vec<OrderedFloat<f64>>>,
```
rsheeter, line 203: Unless an empty Vec has different meaning than None, which is rare and weird, use Vec rather than Option<Vec>

glyphs-reader/src/font.rs
```diff
@@ -482,17 +531,50 @@ pub struct RawGlyph {
...
     pub other_stuff: BTreeMap<String, Plist>,
 }
 
+impl RawLayer {
+    /// Return true if the layer is a draft that is not meant to be compiled.
+    ///
+    /// The presence of an associated master indicates this is not a simple 'master' instance.
+    /// Without 'attributes' that specify whether it's a special intermediate, alternate or
+    /// color layer, we can assume the non-master layer is a draft.
+    fn is_draft(&self) -> bool {
```
rsheeter, line 554: I would guess this corresponds to py code that does the same, can we reference that py code?

glyphs2fontir/src/source.rs
```diff
@@ -731,15 +731,38 @@ impl Work<Context, WorkId, WorkError> for GlyphIrWork {
 
         // Glyphs have layers that match up with masters, and masters have locations
         let mut axis_positions: HashMap<Tag, HashSet<NormalizedCoord>> = HashMap::new();
+        let axes_by_name = font_info.axes.iter().map(|a| (a.tag, a)).collect();
         for instance in glyph.layers.iter() {
-            let Some(master_idx) = font_info.master_indices.get(instance.layer_id.as_str()) else {
+            // skip not-yet-supported types of layers (e.g. alternate, color, etc.)
+            if !(instance.is_master() || instance.is_intermediate()) {
```
rsheeter, line 737: Optional, if the bools here are exclusive - you can be a master, you can be an intermediate, you can be neither, but you cannot be both - then I wonder if we should have something akin to enum LayerType.

rsheeter (approved):

glyphs2fontir/src/source.rs
anthrotype: > if the bools here are exclusive 

while that is true for master vs non-master layer types, however non-master layer types can combine with one another, e.g you _can_ have an intermediate, alternate, color layer in fact.. though it bogs my mind

glyphs-reader/src/font.rs
anthrotype: no, there isn't one. The inline comment above comes from observing that upon creating a new layer from Glyphs.app and immediately saving the file, you get exactly that: a layer with a not-None associated_master_id and no attributes set (and a name that is the current timestamp indicating a temporary draft). I don't think glyphsLib needs to explicitly filter "draft" layers as such, it just ignores them if they aren't master or one of the non-master attributes-bearing layers. Here I needed a way to exclude them when turning RawLayer instances into Layer ones so I created this `is_draft` to make the code more readable.

glyphs-reader/src/font.rs
anthrotype: good point, they ought to mean the same thing. An empty "coordinates" vector makes no sense; if it did, we could risk treating that not-quite-intermediate layer as a duplicate location (as it would overlap with its own associated master location). I'll remove the Option.

PR #539, opened 2023-11-03
Add 'GlyphSet' type to fea-rs to encode the property of a sorted & deduped set of glyphs

cmyr:
This is a funny patch, but it addresses a long-standing wart in fea-rs that regained my attention in the process of working through how our ClassDefs were assigning ids differently than was fontmake.

The tl;dr is that what FEA calls a 'glyphclass' is just an arbitrary sequence of glyphs. In several places, however, we use these classes in places where we are representing members of an OpenType ClassDef. In these cases, we need to do things like compare two sets for equality, and in this case we want to have a fixed order, and ignore duplicates.

The previous implementation was ad-hoc, using a `sort_and_dedupe` method in certain places, where it would take some arbitrary `GlyphClass` and turn it into one that conformed to these expectations; but we were not using it consistently, and it would have been easy to overlook.

With this change, I've added a new type, `GlyphSet`, which is sorted and deduped by construction. This type is used in our public API anywhere we are working with OpenType ClassDefs (really just in the PairPos builder and in mark filtering sets) and so the previous `GlyphClass` type can go back to being private API.

### DIFF
--- a/fea-rs/src/common.rs
+++ b/fea-rs/src/common.rs
@@ -9,7 +9,9 @@ pub use write_fonts::types::GlyphId;
 mod glyph_class;
 mod glyph_map;
 
-pub use glyph_class::GlyphClass;
+pub(crate) use glyph_class::GlyphClass;
+
+pub use glyph_class::GlyphSet;
 pub use glyph_map::GlyphMap;
 
 /// A glyph name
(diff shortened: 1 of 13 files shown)


dfrg (approved):
Functionally, this looks good to me and seems like a real improvement.

The names still feel somewhat confusing though. I’d expect both “class” and “set” to be set-like. Is there a reason why we don’t use `GlyphClass` for things that end up in classdef tables and `Glyph[Sequence|List|Array]` for the ordered/un-deduped type?

cmyr:
I'm also not happy with the names, and tried a few different things. The FEA spec uses the term 'glyph class' to refer to the sequence type, and so I wanted to match that naming; also the sequence type is used extensively, and is already named, and so changing it is more invasive. I'm not thrilled with the circumstances, though 🤷

anthrotype (approved):

dfrg:
Definitely empathize with being constrained by spec (and existing implementations :) LGTM

PR #543, opened 2023-11-06
New test file plus instructions to make more

rsheeter:
As discussed in f2f this morning. ttx_diff on this doesn't match and is tractably small. JMM.

### DIFF
--- a/resources/testdata/README.md
+++ b/resources/testdata/README.md
@@ -1,0 +1,5 @@
+* `resources/testdata/glyphs3/Oswald-AE-comb.glyphs`
+   * Oswald with most of the glyphs deleted
+   * After deletion go to Info > Features
+      * Update language systems
+      * Delete features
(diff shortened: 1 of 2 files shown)


resources/testdata/README.md
```diff
@@ -0,0 +1,5 @@
+* `resources/testdata/glyphs3/Oswald-AE-comb.glyphs`
+   * Oswald with most of the glyphs deleted
+   * After deletion go to Info > Features
+      * Update language systems
+      * Delete features
```
cmyr, line 5: this is all already done I imagine? Are these instructions intended as general instructions for how to make a test font like this?

rsheeter: For that or if you want to build a new version of this file with more glyphs

rsheeter: In general I've never regretted capturing how a given test file was produced and have definitely regretted *not* doing so :)

cmyr: oh definitely, I was just confused by the wording. :)

PR #547, opened 2023-11-07
Correct family for WghtVar_NoExport.glyphs so it exports to expected .designspace

rsheeter:
Correct mistake in #546 where family name for NoExport was wrong so the .designspace export was wrong. It's all about family.


cmyr (approved):
lgtm

Issue #532, opened 2023-10-31
We should skip glyphs marked as non-export


anthrotype:
Glyphs.app allows to mark selected glyphs as non-exportable, i.e. they are not meant to be included in the final font's glyph set.
Designspace+UFO also provides a similar setting called "public.skipExportGlyphs" (a global list of glyph names stored in lib.plist as opposed to a per-glyph bool flag as in the former).
This is how glyphsLib translates between the two:

https://github.com/googlefonts/glyphsLib/blob/cb8a4a914b0a33431f0a77f474bf57eec2f19bcc/Lib/glyphsLib/builder/glyph.py#L62-L67

see:
https://unifiedfontobject.org/versions/ufo3/lib.plist/#publicskipexportglyphs

anthrotype:
Actually, the "public.skipExportGlyphs" should be looked up in the designspace's "<lib>" element. It could also be defined inside a single UFO's lib.plist however fontc only supports building off designspace documents currently, not from single UFOs, so we don't need to check the latter's lib.plist, as the one in the designspace is more general and applies to all UFO sources referenced by the designspace.
https://fonttools.readthedocs.io/en/latest/designspaceLib/index.html#public-skipexportglyphs

Other important things to also bear in mind:
- If the non-export glyphs are used as components in other composite glyphs, [the components will get decomposed](https://github.com/googlefonts/ufo2ft/blob/efb4658890ba645292a9c6bc055c91afe225c4a1/Lib/ufo2ft/util.py#L65-L87) as the base glyphs they point to get pruned.
- kerning groups and pairs will be pruned of the non-export glyphs
- non-export glyphs are skipped when collecting anchors for generating mark/mkmk feature
- the GDEF glyph class definitions (as built from ["public.openTypeCategories"](https://unifiedfontobject.org/versions/ufo3/lib.plist/#publicopentypecategories)) and the ligature carets (built from entry/exit anchors in ufo2ft's gdefFeatureWriter) will not include the non-export glyphs either
- the FEA feature file is left untouched and is not affected by skipExportGlyphs setting; that mostly because subsetting FEA is not trivial. So basically if users use this feature they should be careful not to flag as non-export glyphs that may be referenced from hand-written FEA otherwise feature compiler will be angry.

anthrotype:
I think we support this now, hence closing

PR #548, opened 2023-11-07
Make the export status of a glyph available

rsheeter:
Don't do anything useful with it yet. Step toward #532.


cmyr (approved):
I've not a single nit to pick.

PR #549, opened 2023-11-07
Do not retain glyphs we should not retain

rsheeter:
Helps with #532. We are a few tests short of saying fixed, for example we don't verify the effects on layout noted in the issue.

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -4,7 +4,7 @@
 //! the contours and one updated glyph with no contours that references the new gyph as a component.
 
 use std::{
-    collections::{HashSet, VecDeque},
+    collections::{HashMap, HashSet, VecDeque},
     sync::Arc,
 };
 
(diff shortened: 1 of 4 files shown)


fontir/src/glyph.rs
```diff
@@ -407,6 +420,19 @@ impl Work<Context, WorkId, WorkError> for GlyphOrderWork {
             }
         }
 
+        // Resolve component references to glyphs that are not retained by conversion to contours
+        // Glyphs have to have consistent components at this point so it's safe to just check the default
+        // See https://github.com/googlefonts/fontc/issues/532
+        for glyph_name in new_glyph_order.clone().iter() {
```
cmyr, line 426: hm, I'd normally write this as `iter().cloned()`, which lazily clones each item as needed, instead of cloning the whole collection before iteration? probably slightly better.

cmyr (approved):
one non-blocker inline, otherwise looks good!

fontir/src/glyph.rs
rsheeter: Amusingly the clone is completely unnecessary; I think I initially anticipated modifying new glyph order in the loop then didn't.

Issue #530, opened 2023-10-31
We don't match fontmake for the test file WghtVar_Anchors.glyphs


rsheeter:
```shell
$ python resources/scripts/ttx_diff.py --compare default resources/testdata/glyphs3/WghtVar_Anchors.glyphs
COMPARISON
  Only fontmake produced 'MVAR'
  Identical 'GDEF'
  Identical 'GPOS'
  Identical 'GlyphOrder'
  Identical 'HVAR'
  DIFF 'OS_2', build/default/fontc.OS_2.ttx build/default/fontmake.OS_2.ttx
  Identical 'STAT'
  Identical 'cmap'
  Identical 'fvar'
  DIFF 'glyf', build/default/fontc.glyf.ttx build/default/fontmake.glyf.ttx
  DIFF 'gvar', build/default/fontc.gvar.ttx build/default/fontmake.gvar.ttx
  DIFF 'head', build/default/fontc.head.ttx build/default/fontmake.head.ttx
  DIFF 'hhea', build/default/fontc.hhea.ttx build/default/fontmake.hhea.ttx
  DIFF 'hmtx', build/default/fontc.hmtx.ttx build/default/fontmake.hmtx.ttx
  Identical 'loca'
  Identical 'maxp'
  Identical 'name'
  Identical 'post'
```

Rather more diffs than one might have expected.

anthrotype:
I don't get the same output as yours...
I checked out the PR branch #492 "back_when_mark_was_marky_mark" at the latest commit 01ad09c82b89c1eadad46ea604bc422d1919e8fe, then run the following command:

```
$ rm -rf build && python resources/scripts/ttx_diff.py resources/testdata/glyphs3/WghtVar_Anchors.glyphs
Compare default in build/default
  (cd build/default && cargo run -p fontc -- --no-production-names --source /usr/local/google/home/clupo/oss/fontc/resources/testdata/glyphs3/WghtVar_Anchors.glyphs --build-dir . > fontc.log 2>&1)
  (cd build/default && ttx -o font.ttx font.ttf > ttx.log 2>&1)
  (cd build/default && fontmake -o variable --output-dir WghtVar_Anchors.glyphs --drop-implied-oncurves --no-production-names --debug-feature-file debug.fea /usr/local/google/home/clupo/oss/fontc/resources/testdata/glyphs3/WghtVar_Anchors.glyphs > fontmake.log 2>&1)
[...]

PR #553, opened 2023-11-08
Match fontmake defaults for fsType

rsheeter:
Turns out the default is format-specific. Helps with #530.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -2432,4 +2432,27 @@ mod tests {
     fn compile_obeys_no_export_designspace() {
         assert_noexport("designspace_from_glyphs/WghtVar_NoExport.designspace");
     }
+
+    fn assert_fs_type(source: &str, expected_fs_type: u16) {
+        let temp_dir = tempdir().unwrap();
+        let build_dir = temp_dir.path();
+        compile(Args::for_test(build_dir, source));
+
+        let font_file = build_dir.join("font.ttf");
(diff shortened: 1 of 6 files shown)


dfrg (approved):

fontc/src/lib.rs
```diff
@@ -2432,4 +2432,27 @@ mod tests {
...
+        let temp_dir = tempdir().unwrap();
+        let build_dir = temp_dir.path();
+        compile(Args::for_test(build_dir, source));
+
+        let font_file = build_dir.join("font.ttf");
+        let buf = fs::read(font_file).unwrap();
+        let font = FontRef::new(&buf).unwrap();
+        let os2 = font.os2().unwrap();
+
+        assert_eq!(expected_fs_type, os2.fs_type());
```
cmyr, line 2446: one thing to be aware of with this pattern of having the assert outside of the test case is that, if you were calling this multiple times from the same test it can end up being slightly annoying to figure out which of the calls is failing. In that case I find it's nicer to have your helper function just return a bool or something, and then you can have the `assert!` be in the test case instead.


This doesn't matter here, since it's only called once per test, but just wanted to point it out as a general tip.

cmyr (approved):
🚢

fontc/src/lib.rs
rsheeter: I agree that if we called it many times from the same test it would need more context or a return. That would be a very odd usage here.

Issue #555, opened 2023-11-09
MissingGlyphId error while compiling GS after #549


anthrotype:
this is possibly a regression from https://github.com/googlefonts/fontc/pull/549

"K.alt" is among the glyphs in the public.skipExportGlyphs set. It doesn't look like "K.alt" is mentioned in the hand-written features.fea (I could not grep it), so we must be incorrectly introducing it when autogenerating either kerning or mark/mkmk

```
clupo@clupo-macbookpro3 ~/oss/googlesans (main)
$ RUST_BACKTRACE=1 fontc -s source/GoogleSans/GoogleSans.designspace
thread '<unnamed>' panicked at fontbe/src/features.rs:525:33:
called `Result::unwrap()` on an `Err` value: MissingGlyphId(K.alt)
stack backtrace:
   0: rust_begin_unwind
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:595:5
   1: core::panicking::panic_fmt
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:67:14
   2: core::result::unwrap_failed
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/result.rs:1652:5
   3: <fontbe::features::FeatureWriter as fea_rs::compile::feature_writer::FeatureProvider>::add_features
   4: fea_rs::compile::compile_ctx::CompilationCtx::compile
   5: fea_rs::compile::compiler::Compiler::compile
   6: <fontbe::features::FeatureWork as fontdrasil::orchestration::Work<fontbe::orchestration::Context,fontbe::orchestration::AnyWorkId,fontbe::error::Error>>::exec
   7: fontc::work::AnyWork::exec
   8: std::panicking::try
   9: <rayon_core::job::HeapJob<BODY> as rayon_core::job::Job>::execute
  10: rayon_core::registry::WorkerThread::wait_until_cold
  11: rayon_core::registry::ThreadBuilder::run
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
[2023-11-09T09:47:49.742882Z main fontc::workload ERROR] Be(Features) failed Job panicked: 'called `Result::unwrap()` on an `Err` value: MissingGlyphId(K.alt)'
At least one work item failed
```

anthrotype:
[...]

PR #557, opened 2023-11-09
Fix MissingGlyphId while building mark feature with a non-export glyph with anchors

anthrotype:
the first commit adds a bunch of anchors to a no-export glyph "hyphen" in WghtVar_NoExport.glyphs test file to  trigger issue https://github.com/googlefonts/fontc/issues/555

Will follow up with a fix

### DIFF
--- a/resources/testdata/glyphs3/WghtVar_NoExport.glyphs
+++ b/resources/testdata/glyphs3/WghtVar_NoExport.glyphs
@@ -148,9 +148,19 @@ unicode = 33;
 {
 export = 0;
 glyphname = hyphen;
-lastChange = "2023-11-07 18:00:55 +0000";
+lastChange = "2023-11-09 15:45:52 +0000";
 layers = (
 {
+anchors = (
+{
+name = _top;
+pos = (279,186);
+},
+{
+name = top;
+pos = (273,411);
+}
+);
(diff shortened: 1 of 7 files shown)


anthrotype:
as expected, ci fails with
https://github.com/googlefonts/fontc/actions/runs/6814230043/job/18530625328?pr=557#step:4:283 
```
called `Result::unwrap()` on an `Err` value: MissingGlyphId(hyphen)
```

resources/testdata/glyphs3/WghtVar_NoExport.glyphs
```diff
@@ -401,6 +421,9 @@ key = versionString;
 value = "New Value";
 }
 );
+settings = {
+disablesAutomaticAlignment = 1;
```
anthrotype, line 425: note that I had to disable automatic component alignment otherwise the "manual-component" composite glyph which uses "hypen" as a component would be automatically modified by Glyphs.app. This feature uses anchors definined in the component (in this case, hyphen) to automatically align the components inside a composite glyph. I didn't want that so I explicitly turned it off, in case you were wondering what this is...

Screenshot 2023-11-09 at 16 02 17

rsheeter (approved):
LGTM, ty.

PR #558, opened 2023-11-09
ensure non-export glyphs get pruned from kerning groups

anthrotype:
to trigger the other issue noted at https://github.com/googlefonts/fontc/issues/555#issuecomment-1804163641

> when collecting the glyph groups, we let non-export glyphs in the kerning groups IR, but the the fontbe kern feature generator arbitrarily assigns gid 0 (.notdef?!) to missing glyphs when construcing the coverages which is wrong. Since we are pruning pairs from the kerning IR when single glyphs are not in the glyph order, we should similarly also prune missing/non-export glyphs when they occur inside a kerning group as we construct the kerning IR.

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -244,14 +244,12 @@ impl<'a> FeatureWriter<'a> {
             .kerning
             .groups
             .iter()
-            .map(|(class_name, glyph_set)| {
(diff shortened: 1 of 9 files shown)


rsheeter (approved):

anthrotype:
going to rebase so I get rid of the merge commit

Issue #559, opened 2023-11-09
Compile error with GS project


chrissimpkins:
@MariannaPaszkowska identified the following compiler error today

fontc 484450e970dbdf36a2229ed7198976b95dd576be
rustc 1.72.1 (d5c2e9c34 2023-09-13)
macOS Monterey v12.6.7
[GS main branch commit](https://github.com/googlefonts/googlesans/commit/06c36f526f318302e3c1b34eec722f78e0573a68)

```
cargo run --package fontc -- --source /Users/…/GitHub/googlesans/source/GoogleSans/GoogleSans.designspace 
    Finished dev [unoptimized + debuginfo] target(s) in 1.70s
     Running `target/debug/fontc --source /Users/…./GitHub/googlesans/source/GoogleSans/GoogleSans.designspace`
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `Space(0)`,
 right: `Space(1)`', /Users/…/.cargo/registry/src/index.crates.io-6f17d22bba15001f/write-fonts-0.18.0/src/graph.rs:797:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[2023-11-09T13:21:38.379561Z main fontc::workload ERROR] Be(Features) failed Job panicked: 'assertion failed: `(left == right)`
  left: `Space(0)`,
 right: `Space(1)`'
At least one work item failed
```

rsheeter:
Ty for the detailed report; I believe this is fixed by #560

PR #560, opened 2023-11-10
Push building a lookup up one level

rsheeter:
We used to produce a lookup per base (ty @cmyr for noticing!) which isn't quite right.


cmyr (approved):
looks good. Also afaik this was the cause of #559, although I'm confused about why that was the case and am going to dig a bit more into that and try to identify the root cause.

PR #561, opened 2023-11-10
[fontc] Ensure feature code is deterministic

cmyr:
This replaces our use of HashMap with BTreeMap in various places where we later iterate over a collection, which ensures that we are constructing our features in the same order (ensuring that, for instance, the same item has the same delta set index between builds)

JMM


anthrotype:
alternatively we could have sorted the collections right before using them in places where their order matters

PR #562, opened 2023-11-10
Directly interfere to start kern sooner

rsheeter:
Simplified version of #466. Impact is most readily visible in the timing output:

Before:

image

kern may start as far right as the edge of the blue (be glyph) jobs.

After:

image

kern will consistently start at the left edge (earlier) of the blue jobs. Trying to detect a measurable performance boost with hyperfine was inconclusive, which is better flips run to run. Amusingly Oswald is now too fast for hyperfine to get useful #s at all on Mac. I tried the following:

```shell
# LINUX
$ cargo build --release && hyperfine --warmup 10 --runs 50 --prepare 'rm -rf build/' 'target/release/fontc --source ../OswaldFont/sources/Oswald.glyphs --emit-ir false'
# ~140ms

$ cargo build --release && hyperfine --warmup 3 --prepare 'rm -rf build/' 'target/release/fontc --source ../googlesans/source/GoogleSans/GoogleSans.designspace --emit-ir false'
# ~5.7s

# M1 Mac
$ cargo build --release && hyperfine --warmup 10 --runs 50 --prepare 'rm -rf build/' 'target/release/fontc --source ../OswaldFont/sources/Oswald.glyphs --emit-ir false'
#  Too fast to measure reliably

$ cargo build --release && hyperfine --warmup 3 --prepare 'rm -rf build/' 'target/release/fontc --source ../googlesans/source/GoogleSans/GoogleSans.designspace --emit-ir false'
# ~3.7s, apparently if you work on GS you should buy a mac

```

### DIFF
--- a/fontc/src/workload.rs
+++ b/fontc/src/workload.rs
@@ -232,11 +232,15 @@ impl<'a> Workload<'a> {
             .queued()
             .run();
 
-        let launchable: Vec<_> = self
+        let mut has_kern = false;
+        let mut launchable: Vec<_> = self
             .jobs_pending
             .iter()
             .filter_map(|(id, job)| {
                 if !job.running && self.can_run(job) {
+                    if matches!(id, AnyWorkId::Fe(FeWorkIdentifier::Kerning)) {
+                        has_kern = true;
+                    }
                     Some(id.clone())
                 } else {
                     None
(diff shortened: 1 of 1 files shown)


cmyr (approved):
nice.

fontc/src/workload.rs
```diff
@@ -232,11 +232,15 @@ impl<'a> Workload<'a> {
             .queued()
             .run();
 
-        let launchable: Vec<_> = self
+        let mut has_kern = false;
```
anthrotype, line 235: to single out kerning in this context feels to me a bit out of place/too specific, ideally it'd be preferable to have a more generic way to prioritise tasks that does not involve special-casing

Issue #456, opened 2023-09-25
backend feature compile can be a long pole


rsheeter:
#443 reveals features can be the long pole. There are two non-exclusive options to improve this:

1. Make it faster
   * We write fea for content (and soon marks) for fea-rs to parse; we could build some other form, e.g. the AST, directly
1. Make it parallel
   * The more cores you have the more fea-rs pops out as long pole (ref #443)
   * Could we break up the compile, for example process GSUB and GPOS separately?
      * Requires splitting up things like `feature foo { pos a b -5; sub x by y; }` (courtesy of @cmyr)
      * Some parts, e.g. classes, are common

anthrotype:
may I also suggest:

3. Start it sooner rather than later

in https://github.com/googlefonts/fontc/pull/443 you wrote:

> We can see any speedup to glyph order production is a big win as it's delaying everything.

since auto-FEA generation and FEA compilation are blocked on glyph order, we might also try to speed the latter so that they get an early start?

rsheeter:
> since auto-FEA generation and FEA compilation are blocked on glyph order, we might also try to speed the latter so that they get an early start?

+1, but it looks to me like on a machine with lots of cores fea will still be a problem. https://github.com/googlefonts/fontc/pull/443#issuecomment-1734065103 has a plot on a 72-way system that seesm to show this.

rsheeter:
https://github.com/googlefonts/fontc/issues/458 for glyph order

cmyr:
I've dipped into this a little bit, and have some useful preliminary findings, based on some profiling of the compilation of Google Sans (running fea-rs directly, using as input the fea we generate internally, dumped with `--emit-debug`):

- ~20-30% of our time is spent parsing.
  - There may be some room for small optimizations here, by reducing the number of nodes (in which we group tokens) in certain cases     (specifically I'm thinking about parsing GPOS ValueRecords)
  - In cases where features are split across multiple files, we may be able to parse multiple files in parallel, which might offer significant improvements. A simple sketch of how I think this might work:
    - use a queue to store 'sources to parse'
    - when a source is loaded, in addition to going onto the work queue we also do a `strcmp` for `include` statements
    - parse the files separately, and then stitch them together at the end.
  - There are other options; for instance it would be simpler/faster if the client could just provide a list of relevant feature files ahead of time?
  - If the bottleneck is parsing a single large file we could also look into breaking the input into chunks and parsing those in parallel.
[...]

PR #567, opened 2023-11-11
Move construction of kerning and mark structures to feed to fea-rs into separate jobs that can run in parallel

rsheeter:
Build a structure that converts easily into a fea-rs structure because we are not able to build multiple partial delta sets and combine so we can't get var indices.

https://github.com/googlefonts/fontc/issues/571 may help with letting us build fea-rs structures without the intermediary.

hyperfine does not report a clear speedup. Emit timing suggests we create the potential for it but fail to exploit; lots of whitespace (idle) and seemingly delayed starts for the critical path. I need to take a detour to emojitown and then I can poke at that.

image

Pleasingly between not generating text to parse and splitting up the work fea is now not so dramatically the single long pole for Oswald (remember #456). Also nice to see fe kern, be marks, and hvar all run at once.


fontbe/src/kern.rs
```diff
@@ -0,0 +1,138 @@
...
+        // now for each kerning entry, directly add a rule to the builder:
+        for ((left, right), values) in &kerning.kerns {
+            let (default_value, deltas) = resolve_variable_metric(&static_metadata, values)?;
+
+            let mut x_adv_record = ValueRecord::new().with_x_advance(default_value);
+            let empty_record = ValueRecord::default();
+
+            if !deltas.is_empty() {
+                // COLIN HOW DO I DO THIS WITHOUT A BUILDER?
+                let var_idx = builder.add_deltas(deltas);
```
rsheeter, line 76: @cmyr I'm stuck here. IIUC it's all based on temporary ids until the actual compile so maybe I can fabricate one?

cmyr: hmmm you do need the real builder to get an ID, currently, but I think I could make it so that you can have two separate builders and merge them? So maybe for now you should just create your own builder and then I'll figure out the merging after.

rsheeter: Per IM & VC, might need to produce non-fea-rs structures until such time as we have the ability to accumulate deltas without var indices. That's https://github.com/googlefonts/fontc/issues/571.

fontbe/src/marks.rs
```diff
@@ -0,0 +1,208 @@
...
+            matches!(
+                id,
+                AnyWorkId::Fe(FeWorkId::StaticMetadata)
+                    | AnyWorkId::Fe(FeWorkId::GlyphOrder)
+                    | AnyWorkId::Fe(FeWorkId::Anchor(..))
+            )
+        }))
+    }
+
+    /// Generate kerning data structures.
```
anthrotype, line 114:
```suggestion
    /// Generate mark data structures.
```

fontc/src/change_detector.rs
```diff
@@ -246,6 +246,22 @@ impl ChangeDetector {
                 .is_file()
     }
 
+    pub fn mark_be_change(&self) -> bool {
+        self.glyph_order_ir_change()
```
anthrotype, line 250: I think the be marks task also depends on the anchors IR, not only on the glyph order IR, should you not capture this in here?

rsheeter: I should!

rsheeter: Fixed.

anthrotype (approved):

PR #569, opened 2023-11-13
Add support for 'Virtual Master' Glyphs.app's custom parameter

anthrotype:
I modified the IntermediateLayer.glyphs test file by adding a new `CPHT` axis (for dynamically adjusting the cap-height) and virtual masters that define the min/max extents of this new axis; then added intermediate ('brace') layers to the capital letter glyphs only ("H" and "I" in this specific test file) at each end of the CPHT poles.
It makes sense to use the same test file to test both strictly-defined "intermediate" layers (i.e. in-between two full masters) and layers associated to virtual masters because they use the same layer attributes (or brace naming convention in g2).

It should look like this when the feature is fully implemented:

https://github.com/googlefonts/fontc/assets/6939968/7571ca08-c3e9-4fbb-8b65-cfe41191f63c

The first commits update the test file and the expected test result. I will follow up with the actual implementation that makes the test pass.


rsheeter (approved):
I'm not confident I completely understand the Virtual Master mechanism but the changes all look sane to me

PR #573, opened 2023-11-15
Remove Access::Custom; it forces table scans to figure out what can run

rsheeter:
Access::Custom took a closure and required a scan over pending jobs to determine if it matched any. Instead count the number of pending jobs of each type and resolve "blocks on all of type" (e.g. all IR Glyphs) by checking if the count of those pending is 0.

Key outcome is deletion of `can_run_scan` from `fontc/src/workload.rs` . hyperfine reports no significant change in performance:

```shell
$ cargo build --release && hyperfine --warmup 25 --runs 100 --prepare 'rm -rf build/' 'target/release/fontc ../OswaldFont/sources/Oswald.glyphs'
main   130.6 ms ±  10.9 ms
branch 127.9 ms ±   8.0 ms
```

A mildly amusing outcome is that you can potentially detect that all-of(something) is done before you can detect a specific id is done because the former is a decrement to an atomic and the latter requires a message to pass through a channel.

I feel like there should be a nicer way than how I've rigged AllOrOne but I haven't pinned it down.

### DIFF
--- a/fontbe/src/cmap.rs
+++ b/fontbe/src/cmap.rs
@@ -1,8 +1,9 @@
 //! Generates a [cmap](https://learn.microsoft.com/en-us/typography/opentype/spec/cmap) table.
 
-use std::sync::Arc;
-
-use fontdrasil::orchestration::{Access, Work};
+use fontdrasil::{
+    orchestration::{Access, AccessBuilder, Work},
+    types::GlyphName,
+};
 use fontir::orchestration::WorkId as FeWorkId;
 
 use write_fonts::{tables::cmap::Cmap, types::GlyphId};
@@ -25,12 +26,10 @@ impl Work<Context, AnyWorkId, Error> for CmapWork {
     }
 
     fn read_access(&self) -> Access<AnyWorkId> {
-        Access::Custom(Arc::new(|id| {
-            matches!(
-                id,
-                AnyWorkId::Fe(FeWorkId::GlyphOrder) | AnyWorkId::Fe(FeWorkId::Glyph(..))
-            )
-        }))
+        AccessBuilder::new()
+            .variant(FeWorkId::GlyphOrder)
+            .variant(FeWorkId::Glyph(GlyphName::NOTDEF))
+            .build()
     }
 
     /// Generate [cmap](https://learn.microsoft.com/en-us/typography/opentype/spec/cmap)
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -3,14 +3,11 @@
 //! Each glyph is built in isolation and then the fragments are collected
 //! and glued together to form a final table.
 
-use std::{
-    collections::{BTreeSet, HashMap, HashSet},
-    sync::Arc,
-};
+use std::collections::{BTreeSet, HashMap, HashSet};
 
 use fontdrasil::{
     coords::{Location, NormalizedCoord, NormalizedLocation},
-    orchestration::{Access, Work},
+    orchestration::{Access, AccessBuilder, Work},
     types::GlyphName,
 };
 use fontir::{
@@ -324,10 +321,10 @@ impl Work<Context, AnyWorkId, Error> for GlyphWork {
     }
 
     fn write_access(&self) -> Access<AnyWorkId> {
-        Access::Set(HashSet::from([
-            WorkId::GlyfFragment(self.glyph_name.clone()).into(),
-            WorkId::GvarFragment(self.glyph_name.clone()).into(),
-        ]))
+        AccessBuilder::new()
+            .specific_instance(WorkId::GlyfFragment(self.glyph_name.clone()))
+            .specific_instance(WorkId::GvarFragment(self.glyph_name.clone()))
+            .build()
     }
 
     fn also_completes(&self) -> Vec<AnyWorkId> {
@@ -853,26 +850,20 @@ impl Work<Context, AnyWorkId, Error> for GlyfLocaWork {
     }
 
     fn read_access(&self) -> Access<AnyWorkId> {
-        Access::Custom(Arc::new(|id| {
-            matches!(
-                id,
-                AnyWorkId::Fe(FeWorkId::StaticMetadata)
-                    | AnyWorkId::Fe(FeWorkId::GlyphOrder)
-                    | AnyWorkId::Be(WorkId::GlyfFragment(..))
-            )
-        }))
+        AccessBuilder::new()
+            .variant(FeWorkId::StaticMetadata)
+            .variant(FeWorkId::GlyphOrder)
+            .variant(WorkId::GlyfFragment(GlyphName::NOTDEF))
+            .build()
     }
 
     fn write_access(&self) -> Access<AnyWorkId> {
-        Access::Custom(Arc::new(|id| {
-            matches!(
-                id,
-                AnyWorkId::Be(WorkId::Glyf)
-                    | AnyWorkId::Be(WorkId::Loca)
-                    | AnyWorkId::Be(WorkId::LocaFormat)
-                    | AnyWorkId::Be(WorkId::GlyfFragment(..))
-            )
-        }))
+        AccessBuilder::new()
+            .variant(AnyWorkId::Be(WorkId::Glyf))
+            .variant(AnyWorkId::Be(WorkId::Loca))
+            .variant(AnyWorkId::Be(WorkId::LocaFormat))
+            .variant(WorkId::GlyfFragment(GlyphName::NOTDEF))
+            .build()
     }
 
     fn also_completes(&self) -> Vec<AnyWorkId> {
--- a/fontc/src/workload.rs
+++ b/fontc/src/workload.rs
@@ -4,14 +4,17 @@ use std::{
     collections::{HashMap, HashSet},
     panic::AssertUnwindSafe,
     sync::{
-        atomic::{AtomicBool, Ordering},
+        atomic::{AtomicBool, AtomicUsize, Ordering},
         Arc, Mutex,
     },
 };
 
 use crossbeam_channel::{Receiver, TryRecvError};
 use fontbe::orchestration::{AnyWorkId, Context as BeContext, WorkId as BeWorkIdentifier};
-use fontdrasil::{orchestration::Access, types::GlyphName};
+use fontdrasil::{
+    orchestration::{Access, AccessBuilder, AccessType, Identifier, IdentifierDiscriminant},
+    types::GlyphName,
+};
 use fontir::{
     orchestration::{Context as FeContext, WorkId as FeWorkIdentifier},
     source::Input,
@@ -35,6 +38,7 @@ pub struct Workload<'a> {
     // When K completes also mark all entries in V complete
     also_completes: HashMap<AnyWorkId, Vec<AnyWorkId>>,
     pub(crate) jobs_pending: HashMap<AnyWorkId, Job>,
+    pub(crate) count_pending: HashMap<IdentifierDiscriminant, Arc<AtomicUsize>>,
 
     pub(crate) timer: JobTimer,
 }
@@ -62,6 +66,26 @@ enum RecvType {
     NonBlocking,
 }
 
+/// Higher is better
+///
+/// We basically want things that block the glyph order => kern => fea sequence to go asap
+fn priority(id: &AnyWorkId) -> u32 {
+    match id {
+        AnyWorkId::Fe(FeWorkIdentifier::Features) => 99,
+        AnyWorkId::Fe(FeWorkIdentifier::Kerning) => 99,
+        AnyWorkId::Fe(FeWorkIdentifier::GlyphOrder) => 99,
+        AnyWorkId::Fe(FeWorkIdentifier::PreliminaryGlyphOrder) => 99,
+        AnyWorkId::Fe(FeWorkIdentifier::StaticMetadata) => 99,
+        AnyWorkId::Fe(FeWorkIdentifier::GlobalMetrics) => 99,
+        AnyWorkId::Be(BeWorkIdentifier::Kerning) => 99,
+        AnyWorkId::Be(BeWorkIdentifier::Features) => 99,
+        AnyWorkId::Be(BeWorkIdentifier::GlyfFragment(..)) => 0,
+        AnyWorkId::Be(BeWorkIdentifier::GvarFragment(..)) => 0,
+        AnyWorkId::Fe(FeWorkIdentifier::Glyph(..)) => 1,
+        _ => 32,
+    }
+}
+
 impl<'a> Workload<'a> {
     pub fn new(change_detector: &'a ChangeDetector, timer: JobTimer) -> Workload {
         Workload {
@@ -71,11 +95,13 @@ impl<'a> Workload<'a> {
             error: Default::default(),
             also_completes: Default::default(),
             jobs_pending: Default::default(),
+            count_pending: Default::default(),
             timer,
         }
     }
 
     /// True if job might read what other produces
+    #[cfg(test)]
     fn might_read(&self, job: &Job, other: &Job) -> bool {
         let result = job.read_access.check(&other.id)
             || self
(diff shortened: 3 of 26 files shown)


fontbe/src/glyphs.rs
```diff
@@ -853,26 +850,20 @@ impl Work<Context, AnyWorkId, Error> for GlyfLocaWork {
...
-                id,
-                AnyWorkId::Fe(FeWorkId::StaticMetadata)
-                    | AnyWorkId::Fe(FeWorkId::GlyphOrder)
-                    | AnyWorkId::Be(WorkId::GlyfFragment(..))
-            )
-        }))
+        Access::Set(HashSet::from([
+            AnyWorkId::Fe(FeWorkId::StaticMetadata).into(),
+            AnyWorkId::Fe(FeWorkId::GlyphOrder).into(),
+            AllOrOne::All(AnyWorkId::AllOfBe(WorkId::GlyfFragment(GlyphName::NOTDEF))), // specific name doesn't matter
```
anthrotype, line 856: hm, what if the glyph name did matter some whatever reason?
honestly, it read much clearer before this changes..

fontc/src/workload.rs
```diff
@@ -211,35 +281,49 @@ impl<'a> Workload<'a> {
...
     }
 
-    fn can_run_scan(&self, job: &Job) -> bool {
-        // Job can only run if *no* incomplete jobs exist whose output it could read
-        !self
-            .jobs_pending
-            .iter()
-            .filter(|(other_id, _)| !self.success.contains(other_id))
-            .any(|(_, other_job)| self.might_read(job, other_job))
+    fn can_run_all_or_one(&self, all_or_one: &AllOrOne<AnyWorkId>) -> bool {
```
anthrotype, line 284: a little comment in plain English similar to the one in the previous `car_run_scan` would help the reader understand what this quite dense `can_run_all_or_one` is doing

fontdrasil/src/orchestration.rs
```diff
@@ -160,9 +158,29 @@ impl<I: Identifier> Access<I> {
...
+            Access::One(all) => all == id,
+            Access::Set(ids) => ids.iter().any(|allow| allow.check(id)),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum AllOrOne<I: Identifier> {
+    All(I),
+    One(I),
```
anthrotype, line 170: I noticed that in all the read/write_access method where you used this AllOrOne enum, you only ever used the AllOrOne:All variant, never the AllOrOne:One variant (except when matching all the enum variants).. I wonder what the latter is for. Is it just to logically counterpose the All variant, or is there a more concrete use for it that I am not seeing?

anthrotype: Oh is it used implicitly when you do e.g. `AnyWorkId::Fe(FeWorkId::GlyphOrder).into()` because of the `impl From for AllOrOne` that returns an AllOrOne::One variant?

fontbe/src/cmap.rs
```diff
@@ -25,12 +28,10 @@ impl Work<Context, AnyWorkId, Error> for CmapWork {
...
 
     fn read_access(&self) -> Access<AnyWorkId> {
-        Access::Custom(Arc::new(|id| {
-            matches!(
-                id,
-                AnyWorkId::Fe(FeWorkId::GlyphOrder) | AnyWorkId::Fe(FeWorkId::Glyph(..))
-            )
-        }))
+        Access::Set(HashSet::from([
+            AllOrOne::All(AnyWorkId::Fe(FeWorkId::GlyphOrder)),
```
anthrotype, line 32: now this is a bit confusing..
Shouldn't this be an AllOrOne::One or `AnyWorkId::Fe(FeWorkId::GlyphOrder).into()`? There is only one hob with FeWorkId::GlyphOrder, not many that you may need to block on _all_?

fontdrasil/src/orchestration.rs
cmyr: Would it be possible to just have an 'AllGlyphs' or 'AllGvarFragments' work item, instead of trying to generalize over the whole idea of one/many?

fontdrasil/src/orchestration.rs
```diff
@@ -4,15 +4,22 @@ use std::{
...
     hash::Hash,
-    sync::Arc,
 };
 
 pub const MISSING_DATA: &str = "Missing data, dependency management failed us?";
 
+/// Identifies all items of the same group, typically enum variant, of an identifer
+///
+/// String because it's enormously easier to understand debug output that way
+pub type IdentifierDiscriminant = &'static str;
```
cmyr, line 14: This could just as easily be a type with a `Display` impl?

cmyr (commented):
It feels like there is an awful lot of complexity here, and it is not clear to me how much of that complexity is inherent versus accumulated. Are you happy with the state of things right now? Are there any particular bits that you find annoying?

I have a hunch that there is a better way to represent this concept of a 'work id' in the type system, and to do so in such a way that lets us delete a lot of code.

I also understand that the glyphs/gvar ids are tricky, because unlike all of the other work tasks these can't be known at compile time, as the set of glyphs obviously depends on the source format.

fontdrasil/src/orchestration.rs
```diff
@@ -4,15 +4,22 @@ use std::{
...
+
 /// A type that affords identity.
 ///
 /// Frequently copied, used in hashmap/sets and printed to logs, hence the trait list.
-pub trait Identifier: Debug + Clone + Eq + Hash {}
+pub trait Identifier: Debug + Clone + Eq + Hash {
+    /// Return a value that is consistent across all instances in the same group, e.g. enum variant.
+    fn discriminant(&self) -> IdentifierDiscriminant;
```
cmyr, line 21: also: what is this used for? You might be able to just use [`type_id`](https://doc.rust-lang.org/1.74.0/std/any/trait.Any.html#tymethod.type_id) or [`type_name`](https://doc.rust-lang.org/1.74.0/std/any/fn.type_name.html)?

fontbe/src/glyphs.rs
rsheeter: It cannot matter for the All case as it specifically means all of that variant. As far as I can tell there isn't a good way to express that without an exemplar.

fontdrasil/src/orchestration.rs
rsheeter: The only time you would want One is when you want something narrower than "all of this variant." Concretely, you want to block on a specific glyph not all glyphs. For example, create a One for every component to block on your components but not on every glyph in the world. I can add comments.

fontdrasil/src/orchestration.rs
rsheeter: The intent is identifiers are usually enums; you want to express all of a variant so there isn't a type.

rsheeter:
> Are there any particular bits that you find annoying?

Due to lack of a way to get a discriminant other than from an instance you end up with this `AllOrOne::All(AnyWorkId::AllOfBe(WorkId::GlyfFragment(GlyphName::NOTDEF)))` which misleadingly has to specify a specific glyph name. It's also more nested than I'd like but I'm not sure how to avoid that while retaining separate enums for FE vs BE work.

I would also like to make One only applicable to variants that do not contain data or to have a way to identify that and treat One(variant without data) as equivalent to All(variant without data) because All can resolve completion by simply checking a counter.

fontdrasil/src/orchestration.rs
rsheeter: It could be but it would make use of the value more awkward for no obvious benefit?

fontdrasil/src/orchestration.rs
rsheeter: I think AccessBuilder makes this clearer as now callsites are explict about matching a variant vs a specific instance.

fontdrasil/src/orchestration.rs
rsheeter: Given the lack of a type for a variant I think this is as intended

fontbe/src/glyphs.rs
anthrotype: yes, that's better thanks

fontdrasil/src/orchestration.rs
rsheeter: I played with this locally and it seemed to make my life worse so I'm inclined not to do it.

rsheeter:
Rebased to resolve conflict with main

fontc/src/workload.rs
rsheeter: Oh good catch, added such a comment including explaining why the counter is preferred

anthrotype (approved):

rsheeter:
> I'm not sure that complexity is well spent

Do you have a specific suggestion? - it seems at least somewhat required to me. You need an identity for tasks to be able to say A depends on B. You could dump access checks but that opens the door for inconsistently reproducible ordering problems.

rsheeter:
The majority of what you are describing is already the case.

> dependencies should be described in terms of the actual resources/artifacts

The work identifier specifies an asset or resource (naming is the hardest part) more than actual work given than one work can both consume and produce many things so this is true other than the name not being resource or artifact or asset or w/e. Maybe we should rename things.

> I also think that existence of separate id types and context types for fe/be is unnecessary, and these could be unified

It's possible but I'm of the opinion it would be harmful. The phases are deliberately separated and some errors, such as a FE task trying to use a BE output, won't compile.

> "this task depends on the existence of a particular glyph outline", or "this task requires all glyph outlines", where there would be a separate resource id for each individual glyph, as well as a resource id for 'all of the outlines'.

This is true already

> a struct that would store a type id and an optional 'instance identifier'

The set of possible ids is known in advance. Which ones have an instance identifier is known in advance. To me that's describing an enum. What would the advantage be of this change?

rsheeter:
The code example helps, ty. I was imagining our current contexts glued together which implies everything can see everything. The sketch neatly avoids that.

I agree it feels like there might be something nicer than what we do currently here (assuming no blocker exists going from sketch to real implementation). I'm a bit concerned that the ROI would be low. It would be a lot of work and the benefits I see are:

1. A specific type of test is easier
1. Adding fields to Context is easier, we are spared some repetitive manual changes

Are there other benefits?

> an end state where we have a Fontc type which we configure with some input arguments, and then we are able to (in one or two lines) say 'now give me the ir::Kerning for this input', and the compiler can figure out how to resolve all of its dependencies internally

This part seems strongly informed by the problem you encountered in writing a test where we create and run specific work rather awkwardly. Those tests are poorly written, we should probably go fix them. However, I'm not immediately certian that a non-test users needs the "run to an arbitrary point" capability?

Thinking "aloud," I think we could make the test that seems to inform the end state easier through a less invasive change. For example, to run until a specific thing is produced you would, roughly:

1. Make a method to process [an input] until you have produced [a work id], or perhaps a set of ids so I can write a test that says "run until you produce {glyph ir for these glyphs}". That method would:
   1. Call `fontc::create_workload` to generate a full graph
   1. Call a new method to prune the workload down to only things you need to produce let's say fontir::WorkId::Kerning
      * this would start at the work for kerning, look at it's read access, and close over work reachable by read access to build the set of ids retained
      * then dump all the ids that aren't retained
   1. Execute the workload
   1. Return context
1. You read your item out of Context

Not as elegant as `let kern: Kern = context.get()?` but readily achievable in a few hours.

> without needing us to manually create any work or otherwise do any per-execution setup.

You need something to create a graph of work, and to update it at runtime because the graph is not a pure function of source but changes in response to operations on that source, such as to generate a new glyph based on what we found in the glyphs of the source. You've still got "work," you still need to know what work depends on, and you've still got something responsible for executing a workload.

Hm. In the sketched model whose job is it to create the work to build the things you need to provide the ir::Kerning? You need to know what it reads, transitively, and to assemble a graph. Getting that graph right is arguably the interesting part.

cmyr:
> The code example helps, ty. I was imagining our current contexts glued together which implies everything can see everything. The sketch neatly avoids that.
> 
> I agree it feels like there might be something nicer than what we do currently here (assuming no blocker exists going from sketch to real implementation). I'm a bit concerned that the ROI would be low. It would be a lot of work and the benefits I see are:
> 
> 1. A specific type of test is easier
> 2. Adding fields to Context is easier, we are spared some repetitive manual changes
> 
> Are there other benefits?

The main benefits as I see them (just for this context sketch):

- easier to write tests: I actually think this is a really valuable property; reducing the friction around writing up a new simple test case can go a long way to improving our robustness long term.
- better encapsulation: we can have a context type that doesn't need to know everything that it's storing, which is a nice API boundary, and the public API surface will be very small. Adding new resources is simple, and there's no confusion about what context I'm using etc.
- better maintainability: less code, fewer types, one clear idea. The implementation will need to be documented, but it should be easy enough for a new contributor to go do a quick skim of the types involved and understand what's going on.
- the API itself enforces our constraints: by this I mean that you can only get resources out of the context that you can name. It would be easy enough to also layer an access control mechanism on top of this, where you are only allowed to access items that were listed in your read_access/dependency list, but at an even more fundamental level you don't see all the fields in the context, so you couldn't accidentally call the wrong one.

> > an end state where we have a Fontc type which we configure with some input arguments, and then we are able to (in one or two lines) say 'now give me the ir::Kerning for this input', and the compiler can figure out how to resolve all of its dependencies internally
> 
> This part seems strongly informed by the problem you encountered in writing a test where we create and run specific work rather awkwardly. Those tests are poorly written, we should probably go fix them. However, I'm not immediately certain that a non-test users needs the "run to an arbitrary point" capability?

Less than "run to an arbitrary point" I've been thinking of this as "generate this arbitrary artifact". In the test example the artifact might be some IR, but an artifact might also be an instance or a subset or something else. The idea is that there is a single mechanism/abstraction through which resources are resolved, regardless of what point in the program graph you find yourself. I think this has a very nice conceptual simplicity.

> Thinking "aloud," I think we could make the test that seems to inform the end state easier through a less invasive change. For example, to run until a specific thing is produced you would, roughly:

I think that any changes we might make should be done in stages, and I think that there are reasonable ways to break it up. For instance, I think that we could probably unify the contexts without changing too much else, and iterate from there. I definitely agree that anything we can do in the immediate term that improves testing would be great.

> Not as elegant as `let kern: Kern = context.get()?` but readily achievable in a few hours.
> 
> > without needing us to manually create any work or otherwise do any per-execution setup.
> 
> You need something to create a graph of work, and to update it at runtime because the graph is not a pure function of source but changes in response to operations on that source, such as to generate a new glyph based on what we found in the glyphs of the source. You've still got "work," you still need to know what work depends on, and you've still got something responsible for executing a workload.
> 
> Hm. In the sketched model whose job is it to create the work to build the things you need to provide the ir::Kerning? You need to know what it reads, transitively, and to assemble a graph. Getting that graph right is arguably the interesting part.


Right, I think you're definitely identifying the right questions here.

I have thought about this a reasonable amount, and I have some ideas. I'll try to sketch out at a high level how I've imagined this whole system fitting together:

- as mentioned, we have a clear distinction between 'resources' and 'work'.
- we have a trait that looks a lot like the current `Work` trait.
- (as an aside, I picture all of the types and core functionality related to the plumbing and logic of the compiler living somewhere like fontdrasil, e.g. somewhere outside of the be/ir crates, that is a dependency of everything)
- we have a *new* trait called, say, `ResourceProvider`. The various frontends, as well as the backend, would provide types that implement this trait. It will have an API like,
  ```rust
  /// A type that is capable of creating tasks that produce resources.
  trait ResourceProvider {
      fn provide_work(&self, for_resource: &ResourceIdentifier) -> Option<AnyWork>;
  }
  ```
- There will be some sort of compilation context that owns the graph and manages execution.
- Compilation starts when some root resource is requested, and the compiler is provided some input arguments (sources e.g.)
- based on the inputs to the compiler, it configures some (small) set of resource providers. For the sake of simplicity let's just imagine there are always two, one for the frontend and one for the backend.
- The process of building the graph is recursive:
  - the compiler looks at the requested resource, and checks if it exists.
  - if the resource exists, we're done.
  - if the resource does not exist, then the compiler asks each `ResourceProvider` if they can provide a `Work` that will generate it.
  - (it is an invariant that only one provider can exist for any resource, and there should always be a provider for any resource, although we will error nicely if this is violated)
  - we inspect the provided `Work` items dependencies, recursing for each one.
  - we use this to build a graph of required work, which we execute.

So that's a hand-wavy sketch. The devil is in the details; for instance in some cases (e.g. glyphs) we do not know all of the resources we need (assuming one per glyph) until we have done some preliminary work, and so some `Work` items cannot actually name their dependencies until some other work has completed (getting the glyph order, e.g.) and so I'd also want some mechanism for `Work` to basically be polled for dependencies: the `Work` item could signal from its dependencies/access method that it wasn't ready yet, and should be called back after some preliminary dependency had been resolved, or something.

And then all of this would be rolled up in a nice `Compiler` struct of some kind that would own the runloop and handle things like tracing execution and updating the graph as work completed.


And again, this is all just ideas, and although I have played around with implementing some of this there are lots of unanswered questions; but I do think that figuring out a scalable & ergonomic architecture would offer some significant long term benefits.

I also think that we can get here iteratively. Ultimately most of the important logic in `fontc` is the stuff in the `exec` methods, and none of that logic really needs to change (outside of the stuff that directly implicates the current architecture, like getting/setting fields on the contexts.)

Indeed, in a hypothetical world where we wanted to explore these changes, the route I would propose would be to first refactor `exec` so that all the business logic is in independent fns that know nothing about `fontc`, and are just passed arguments and return a result, with the `exec` method just being used to get the arguments and set the results on the context; that way we could have the core of this new compiler exist alongside the current one, and it could reuse all the current logic.

PR #574, opened 2023-11-16
Add GS main branch build CI testing

chrissimpkins:
Adds a GH Action workflow to build the GS project on pull request commits with a debug build of the fontc compiler.

### DIFF
--- a/.github/workflows/gsbuild.yml
+++ b/.github/workflows/gsbuild.yml
@@ -11,9 +11,7 @@ jobs:
       - name: Check out fontc source repository
         uses: actions/checkout@v4
       - name: Install the latest stable Rust toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
+        uses: dtolnay/rust-toolchain@stable
       - name: Build and install fontc
         run: cd fontc && cargo install --path . --debug
       - name: Check out font project source repository
--- a/.github/workflows/gsbuild.yml
+++ b/.github/workflows/gsbuild.yml
@@ -1,0 +1,27 @@
+name: GS Test Build
+
+on:
+  pull_request:
+
+jobs:
+  build-gs:
+    runs-on: ubuntu-latest
+    name: Build GS variable fonts
+    steps:
+      - name: Check out fontc source repository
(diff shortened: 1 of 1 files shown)


rsheeter (approved):
Fantastic, please merge once it passes :)

.github/workflows/gsbuild.yml
```diff
@@ -0,0 +1,43 @@
...
+          token: ${{ secrets.GS_READ_FONTC }}
+      - name: Install the latest stable Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - name: Install fontc and compile fonts
+        uses: googlefonts/fontc-action@v1
+        with:
+          source-path: source/GoogleSans/GoogleSans.designspace
+          debug: true
```
rsheeter, line 24: You could just assume we are in the root of our repo and run cargo commands akin to what https://github.com/googlefonts/fontc/blob/main/.github/workflows/rust.yml does?

.github/workflows/gsbuild.yml
```diff
@@ -0,0 +1,43 @@
...
+
+  build-italic:
+    runs-on: ubuntu-latest
+    name: Build Italic variable font
```
rsheeter, line 28: If compiling the compiler is by far the slowest step  it's possible that building Romand and Italic in one job would make sense?

.github/workflows/gsbuild.yml
chrissimpkins: yeah, I guess that was kind of silly, eh?

chrissimpkins: 7db4cb1f058d0a01e687312099dd784390dd904f

.github/workflows/gsbuild.yml
chrissimpkins: 7db4cb1f058d0a01e687312099dd784390dd904f

chrissimpkins: With the current approach the fontc compile takes about 50s and each variable font compile takes 1.5 mins.  OK to keep this sequential approach or should we run the builds in parallel again?

PR #575, opened 2023-11-16
don't emit IR by default and rename option to --incremental

anthrotype:
As discussed with Rod in chat, we prefer to have incremental builds be opt-in rather than opt-out. Not only it's faster to build without, but also it reduces the risk that the casual user might incur in potential bugs on subsequent builds related to incomplete inter-task dependencies (I know, we should test this code path more to be more confident about it).

Also I propose to rename the option from `--emit-ir` to `--incremental` because the latter communicates the intent better to the user. The old name is kept for backward compatibility but it won't be shown in the --help.

The flag takes an optional bool argument, which can thus be omitted altogether for simplicity.
E.g. --incremental is equivalent to --emit-ir=true; omitting the whole flag (default behavior) is equivalent to --emit-ir=false or --incremental=false.

### DIFF
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -14,8 +14,17 @@ pub struct Args {
     pub source: PathBuf,
 
     /// Whether to write IR to disk. Must be true if you want incremental compilation.
-    #[arg(short, long, default_value = "true", action = ArgAction::Set)]
-    pub emit_ir: bool,
+    #[arg(
+        short,
+        long,
+        num_args = 0..=1,
+        default_value = "false",
+        default_missing_value = "true",
+        // pre-existing long name kept as alias for backwards compatibility
+        alias = "emit-ir",
+        action = ArgAction::Set,
+    )]
+    pub incremental: bool,
 
     /// Whether to write additional debug files to disk.
     #[arg(long, default_value = "false")]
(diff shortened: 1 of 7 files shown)


fontc/src/args.rs
```diff
@@ -101,7 +110,7 @@ impl Args {
         Args {
             glyph_name_filter: None,
             source,
-            emit_ir: true,
+            incremental: true,
```
anthrotype, line 113: Note that even if it's off by default, I decided to keep incremental=true when initialising Args::for_test, since many tests expect to be able to look inside the build/ directory for IR files.

anthrotype:
The clippy lint failure is probably not my fault, or is it?

https://github.com/googlefonts/fontc/actions/runs/6892002555/job/18748318625?pr=575

fontc/src/args.rs
```diff
@@ -14,8 +14,17 @@ pub struct Args {
...
+    #[arg(
+        short,
+        long,
+        num_args = 0..=1,
+        default_value = "false",
+        default_missing_value = "true",
+        // pre-existing long name kept as alias for backwards compatibility
+        alias = "emit-ir",
+        action = ArgAction::Set,
+    )]
```
cmyr, line 26: this seems very complicated for what should be a very common use-case? Like a flag that sets a bool? Is this all about backwards compat? I'm not sure I would bother, or if that's our concern I would maybe just leave the `emit_ir` item and just exit with a helpful message if it gets set?

cmyr (approved):
Looks reasonable!

fontc/src/args.rs
anthrotype: yes, totally about backward compat

@rsheeter said he'd prefer to have the old --emit-ir continue working becuase "I have lots of copy/paste examples with the old name"

PR #576, opened 2023-11-16
make input arg positional, add -o/--output-file option

anthrotype:
This adds a new `-o`/`--output-file` CLI flag to tell fontc where to save the output file. It keeps defaulting to the existing `build/font.ttf` like before (we may want to change that to use, say, the font's postscript name e.g. Oswald-Regular.ttf, so one could build more than one font in the same default build directory without overwriting them? but that's a separate issue). 

Also, I changed so that the input source is a positional argument, but I didn't want to break existing usages of `-s`/`--source` option so I kept the latter and made them two methods mutually exclusive, and require at least one of the two is not None before the final Args can be constructed.

So the old way of calling fontc continues to work (e.g. `fontc --source Oswald.glyphs` or `fontc -s Oswald.glyphs`), but now you can simply do `fontc Oswald.glyphs -o Oswald.ttf` which looks nicer. Or even `fontc -o Oswald.{ttf,glyphs}` :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Install the latest version of Rust, https://www.rust-lang.org/tools/install.
 ### Build a simple test font
 
 ```shell
-$ cargo run -p fontc -- --source resources/testdata/wght_var.designspace
+$ cargo run -p fontc -- resources/testdata/wght_var.designspace
 ```
 
 ### Emit IR to enable incremental builds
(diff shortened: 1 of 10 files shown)


anthrotype:
another thing to note is the build directory is only created if we are doing incremental builds (to save the IR files) or if the user didn't provide an explicit --output-file, because we want to continue saving our `build/font.ttf` in the same old location so we need the `build/` directory.
Otherwise, if one doesn't do incremental (which is now disabled by default) _and_ has provided an explicit output file, then no build directory is required and all is done in memory.

cmyr (approved):
Looks good, I think these small improvements are really valuable :)

PR #580, opened 2023-11-17
[fontc/args] don't take true|false argument in --incremental

anthrotype:
It creates problems when you do `fontc -i font.glyphs` (as it attempts to parse the positional input source as the bool argument for --incremental) so it forces one to have to put the filename in front followed by the optional parameter, e.g. `fontc font.glyphs -i`..
I think this is an unnecessary complication and would rather drop support for the old --emit-ir=true|false flag alias... I think it's fine to simplify and only have a single bool flag `-i/--incremental` that takes no parameters.


cmyr (approved):
A nice simplification in any case!

PR #583, opened 2023-11-17
[ttx_diff] Run black formatter

cmyr:
Totally unnecessary, but I want to have this file up to date so that an upcoming patch only includes my actual changes.

JMM


dfrg (approved):
Dunno what this is, but looks good?

PR #584, opened 2023-11-17
[fontc] More code reuse in fontc tests

cmyr:
In the process of looking into how to add mark-mark tests I got a bit distracted trying to figure out a way to reduce the boilerplate in the existing tests.

This mostly does that, although it also slightly tweaks how some asserts are structured.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -433,7 +433,7 @@ mod tests {
             variations::{DeltaSetIndexMap, ItemVariationData},
         },
         types::NameId,
-        TableRef,
+        ReadError, TableRef,
     };
     use skrifa::{
         charmap::Charmap,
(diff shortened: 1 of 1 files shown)


cmyr:
Also as a general observation (not having really looked at this in much detail before) a bunch(/most/all?) of the tests in `lib.rs` really feel like integration tests, which can potentially live outside of the crate in the `tests` directory, assuming that they don't need to access much private internal API. I will put this on my 'one day' todo list...

fontc/src/lib.rs
```diff
@@ -678,6 +678,47 @@ mod tests {
         assert_eq!(expected, completed);
     }
 
+    /// A convenience type for the common case of "just compile this source with
+    /// default arguments".
+    ///
+    /// When you need more control, just the `compile` fn/`TestCompile` type
+    /// directly
+    struct SimpleCompile {
```
rsheeter, line 686: I like the elimination of boilerplate but I'm not sure why you need a new type, just improve TestCompile?

rsheeter (approved):
LGTM with one nit

cmyr:
I looked into that, and it isn't obvious how to present this sort of simple interface there, while preserving the flexibility of the existing type. This solution doesn't feel _great_, but it does feel like a small improvement.

fontc/src/lib.rs
```diff
@@ -678,6 +678,47 @@ mod tests {
...
+        /// the font bytes
+        buf: Vec<u8>,
+        result: TestCompile,
+    }
+
+    impl SimpleCompile {
+        fn new(source: &str) -> std::io::Result<Self> {
+            let temp_dir = tempdir()?;
+            let build_dir = temp_dir.path();
+            let result = compile(Args::for_test(build_dir, source));
```
anthrotype, line 699: I don't particularly like that the heavy-lifting part of the code where things get compiled is hidden in a default new() constructor.. Why not use a free function that takes a str and returns a struct and is named with something like "compile" to make it clear this is, well, compiling?

rsheeter: +1, I like the current `result = compile()` pattern. I also like how this PR removes some of the boilerplate. Seems possible to combine.

cmyr: fair, how about `SimpleCompile::compile`?

fontc/src/lib.rs
rsheeter: Filed as https://github.com/googlefonts/fontc/issues/594 to unblock merging

cmyr:
okay I'm inclined to get this in, it isn't a silver bullet but it marginally improves the status quo

PR #586, opened 2023-11-20
[fontbe/avar] don't skip no-op segment maps

### LINKED ISSUE #582: GS build with fontc raises OTS error + warning
[fontc source commit](https://github.com/googlefonts/fontc/commit/796d5bdd978ba1877b7f8b738c99ede0603a0a04)
[GS source commit](https://github.com/googlefonts/googlesans/commit/2ca11933d045352af2ada4d4feeacb3e191bb245)

Fonts compiled in the default GH Actions Linux runner environment with the current stable Rust toolchain

```
ERROR: avar: Axis count mismatch
ERROR: Table discarded
WARNING: STAT: Unexpected non-zero offsetToAxisValueOffsets
File sanitized successfully!
```

anthrotype:
GS contains a `GRAD` axis whose avar segment maps does not have any interesting mappings.
It happens to be the last one in the axis order. We were filtering out axis segment maps that contain only identity maps, leading to an avar table with fewer segment maps than there are fvar axes, which is invalid as OTS correctly points out.

So we now still emit these non-interesting segment maps when bulding avar and there's at least one of them that is not an identity map.

Also note that, despite the OT spec allows for empty segment maps for axis that don't distort the default normalization, we want to match fontmake/fonttools and emit the required {-1:1, 0:0, 1:1} mappings even when no other interesting mappings are present for an axis, because apparently some old implementations need that (see https://github.com/fonttools/fonttools/pull/1026).

Fixes #582 avar axis mismatch OTS error

I can confirm that after this, the GS avar table is the same when built with fontc vs fontmake.

### DIFF
--- a/fontbe/src/avar.rs
+++ b/fontbe/src/avar.rs
@@ -1,5 +1,6 @@
 //! Generates a [avar](https://learn.microsoft.com/en-us/typography/opentype/spec/avar) table.
 
+use font_types::F2Dot14;
 use fontdrasil::{
     coords::{CoordConverter, DesignCoord, NormalizedCoord},
     orchestration::{Access, Work},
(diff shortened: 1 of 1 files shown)


fontbe/src/avar.rs
```diff
@@ -21,7 +22,30 @@ pub fn create_avar_work() -> Box<BeWork> {
...
 }
 
-fn to_segment_map(axis: &Axis) -> Option<SegmentMaps> {
+/// Return true if the SegmentMaps is not a boring identity map
+fn is_interesting(segment_map: &SegmentMaps) -> bool {
+    segment_map
+        .axis_value_maps
+        .iter()
+        .any(|av| av.from_coordinate != av.to_coordinate)
+}
```
cmyr, line 31: I think this makes sense as a method on the `SegmentMaps` type in write-fonts? I'd probably flip the logic and call it `is_identity`.


One option would be to add the method there, merge this as-is, and open an issue to fix this when we next publish write-fonts.

fontbe/src/avar.rs
```diff
@@ -21,7 +22,30 @@ pub fn create_avar_work() -> Box<BeWork> {
...
+
+/// Return a default avar SegmentMaps containing the required {-1:-1, 0:0, 1:1} maps
+fn default_segment_map() -> SegmentMaps {
```
cmyr, line 34: and this feels like it should probably be the impl of `Default` for `SegmentMaps`; however we currently autogenerate `Default` impls for all write types, and there's no mechanism to override this, so this is probably fine.

cmyr (approved):
Looks good, I appreciate all the explanation inline.

fontbe/src/avar.rs
rsheeter: +1 making the method is_identity

rsheeter:
> I appreciate all the explanation inline

Vigorous +1, it's helpful to us now and incredibly helpful to future editors

PR #589, opened 2023-11-20
[ci] Add OTS tests on GS builds

chrissimpkins:
Addresses https://github.com/googlefonts/fontc/issues/582#issuecomment-1816811638 and https://github.com/googlefonts/fontc/issues/582#issuecomment-1817191570

### DIFF
--- a/.github/workflows/gsbuild.yml
+++ b/.github/workflows/gsbuild.yml
@@ -19,7 +19,15 @@ jobs:
         with:
           repository: googlefonts/googlesans
           token: ${{ secrets.GS_READ_FONTC }}
+      - name: Download and unzip OTS v9.1.0
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/gsbuild.yml
```diff
@@ -19,7 +19,15 @@ jobs:
         with:
           repository: googlefonts/googlesans
           token: ${{ secrets.GS_READ_FONTC }}
+      - name: Download and unzip OTS v9.1.0
+        run: |
+            curl -OL "https://github.com/khaledhosny/ots/releases/download/v9.1.0/ots-9.1.0-Linux.zip"
```
rsheeter, line 24: Optional, we could confirm the expected checksum of the zip here

rsheeter (approved):

PR #592, opened 2023-11-21
Build builder with 'build' instead of TryInto

cmyr:
This is a little bit of cleanup to be more consistent in our (non)use of TryInto.

No functional change, JMM

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -79,7 +79,7 @@ fn split_glyph(glyph_order: &GlyphOrder, original: &Glyph) -> Result<(Glyph, Gly
         });
     });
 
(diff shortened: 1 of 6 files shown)


rsheeter:
This seems like a fairly reasonable use of TryInto, I'm not entirely sure why it needs to be fixed?

rsheeter (approved):
I don't find it obvious this is an improvement but I don't particularly object either.

cmyr:
Yea, this is mostly about semantics and Rust idiom. I think this has come up before, but `TryFrom` is specifically about conversions between types that may fail in specific, controlled circumstances. The docs do a better job than I of explaining: http://doc.rust-lang.org/1.72.1/std/convert/trait.TryFrom.html

PR #596, opened 2023-11-22
[chore] Rationalize imports

cmyr:
This enforces a new convention: crate should only ever depend on *one* of write-fonts, read-fonts, or font-types.

- If you use read-fonts, you can access font-types as read_fonts::types
- If you use write-fonts, read_fonts is at write_fonts::read, and types is at write_fonts::types

In the process of making this change, I also did a bit of general cleanup of imports.


rsheeter:
If you want to pursue this model I suggest always choosing write-fonts for fontc. I think that would mean the entire workspace would have {write-fonts, skrifa} and skrifa is currently only for tests.

rsheeter (approved):
LGTM with one suggestion in a separate comment

Issue #474, opened 2023-09-29
default cu2qu tolerance should be relative to UPEM


anthrotype:
fontTools.cu2qu.ufo.fonts_to_quadratic (used by fontmake via ufo2ft.preProcessor) defaults to a relative maximum error of 0.001 * font.info.unitsPerEm, which works out as 1.0 absolute font units for fonts with UPEM=1000 as it's more frequently the case in font editors.

fontc cubics_to_quadratics method currently defaults to 1.0 as the absolute tolerace, which only matches fontmake/fonttools when the input source has UPEM 1000; if UPEM is larger, fontc' default tolerance is too fine grained (leading to more quad control points), if UPEM is smaller, the default tolerance is too coarse, not precise enough.

So we want to match fontmake default tolerance by making it 1/1000th of the UPEM.

fontmake also has an option to control the cu2qu --conversion-error which I believe is rarely used (note, unlike the default error, this optional one is actually measured in absolute font units), we may want to add a similar flag to fontc.

rsheeter:
+1 to making it computed. I hope we can avoid adding a flag if the default is good.

PR #599, opened 2023-11-24
[fontbe] make cu2qu tolerance relative to UPEM

anthrotype:
Instead of hard-coding tolerance=1.0, we now use the same default tolerance as fontTools.cu2qu, which is relative to the font's UPEM, i.e. UPEM/1000

Fixes https://github.com/googlefonts/fontc/issues/474


anthrotype:
Fair enough. But I think it's good to be able to parametrize fontc tests more generally, we also use it over at fontations btw

Issue #581, opened 2023-11-17
--incremental panics "Spawned more jobs than items available to run" when run 2nd time


anthrotype:
If you build GS twice with the --incremental flag using the latest fontc main, the second time it panics with the following error:

```
clupo@clupo-macbookpro3 ~/oss/fontc (main)
$ rm -rf build
clupo@clupo-macbookpro3 ~/oss/fontc (main)
$ cargo run -- ../googlesans/source/GoogleSans/GoogleSans.designspace --incremental
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/fontc ../googlesans/source/GoogleSans/GoogleSans.designspace --incremental`
clupo@clupo-macbookpro3 ~/oss/fontc (main)
$ cargo run -- ../googlesans/source/GoogleSans/GoogleSans.designspace --incremental
    Finished dev [unoptimized + debuginfo] target(s) in 0.23s
     Running `target/debug/fontc ../googlesans/source/GoogleSans/GoogleSans.designspace --incremental`
thread '<unnamed>thread '' panicked at <unnamed>fontc/src/workload.rs' panicked at :fontc/src/workload.rs372::37229::
29Spawned more jobs than items available to run:

Spawned more jobs than items available to runnote: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread '<unnamed>' panicked at thread 'fontc/src/workload.rs<unnamed>:' panicked at 372fontc/src/workload.rs::372:29:
thread '29:
Spawned more jobs than items available to runSpawned more jobs than items available to run

thread '<unnamed>' panicked at thread '<unnamed>thread '' panicked at fontc/src/workload.rs<unnamed>fontc/src/workload.rs<unnamed>' panicked at ' panicked at :fontc/src/workload.rsfontc/src/workload.rs:372:::29:
Spawned more jobs than items available to run372
372:372::2929:
:
Spawned more jobs than items available to run29
:
Spawned more jobs than items available to run
Spawned more jobs than items available to run
thread 'thread 'thread 'thread '<unnamed><unnamed><unnamed>' panicked at <unnamed>' panicked at fontc/src/workload.rs:372:29:
fontc/src/workload.rsSpawned more jobs than items available to runthread 'thread '
<unnamed>' panicked at <unnamed>' panicked at thread '' panicked at fontc/src/workload.rs<unnamed>fontc/src/workload.rs::372:29thread ':
[...]

PR #600, opened 2023-11-24
[workload] exclude non-runnable jobs when popping items from the run_queue

anthrotype:
With this --incremental works again

Fixes https://github.com/googlefonts/fontc/issues/581

### DIFF
--- a/fontc/src/workload.rs
+++ b/fontc/src/workload.rs
@@ -303,6 +303,8 @@ impl<'a> Workload<'a> {
                     nth_wave += 1;
                 }
 
+                // count of launchable jobs that are runnable, i.e. excluding !run jobs
+                let mut runnable_count = launchable.len();
                 {
                     let mut run_queue = run_queue.lock().unwrap();
                     for id in launchable.iter() {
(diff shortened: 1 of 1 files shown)


fontc/src/workload.rs
```diff
@@ -303,6 +303,7 @@ impl<'a> Workload<'a> {
                     nth_wave += 1;
                 }
 
+                let mut job_count = launchable.len();
```
rsheeter, line 306: Can you add a comment this is the count of runnable jobs? - at a first glance I thought this was Very Wrong and then I realized it was effectively filtering the !run ones

rsheeter (approved):

fontc/src/workload.rs
anthrotype: or I can just rename the variable to runnable_count

anthrotype: Done in https://github.com/googlefonts/fontc/pull/600/commits/8fff172087baba11d4c8c56d0f97f5bf73f26a12 (squashed)

PR #602, opened 2023-11-28
Note what each crate is for in its readme

### LINKED ISSUE #493: What is it you DO here?
Add some documentation of what each crate does and what its name means would be grand. Some things are self explanatory - ufo2fontir for example - but others aren't, fontdrasil is still a bit of a mystery, and this then leads to being uncertain as to whether or not it's okay to use types from that crate in other ones, that sort of thing. A tiny README in each crate with three sentences this is what it does and this is where you should/shouldn't use it would make the world of difference

rsheeter:
Fixes #493. JMM if happy.

### DIFF
--- a/fontc/README.md
+++ b/fontc/README.md
@@ -1,0 +1,11 @@
+# fontc
+
+[![MIT/Apache 2.0](https://img.shields.io/badge/license-MIT%2FApache-blue.svg)](#license)
+
+fontc: Font Compiler
+
+This crate provides the command line entrypoint into our font compiler.
+
+It's primary task is to accept arguments from the user, create a task graph, and execute it.
+
+This crate can reference all other crates in the workspace.
--- a/fontdrasil/README.md
+++ b/fontdrasil/README.md
@@ -1,0 +1,11 @@
+# fontdrasil
+
+[![MIT/Apache 2.0](https://img.shields.io/badge/license-MIT%2FApache-blue.svg)](#license)
+
+fontdrasil: Common types and functionality shared between all layers of fontc.
+
+The name is a reference to Yggdrasil, the world tree.
+
+This crate provides types that are usable at any layer of fontc. For example,
+support for the three major [units](https://github.com/googlefonts/fontc/blob/main/resources/text/units.md)
+used in font compilation.
--- a/fontir/README.md
+++ b/fontir/README.md
@@ -1,0 +1,10 @@
+# fontir
+
+[![MIT/Apache 2.0](https://img.shields.io/badge/license-MIT%2FApache-blue.svg)](#license)
+
+fontir: Font Intermediate Representation
+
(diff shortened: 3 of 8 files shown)


fontir/README.md
```diff
@@ -0,0 +1,10 @@
+# fontir
+
+[![MIT/Apache 2.0](https://img.shields.io/badge/license-MIT%2FApache-blue.svg)](#license)
+
+fontir: Font Internal Representation
+
+This crate defines the internal representation (IR) of a font used by fontc and common operations
+on that IR.
+
+It should be referenced by fontc and any frontend (_format_2ir) crates.
```
anthrotype, line 10:
```suggestion
It should be referenced by fontc and any frontend (_format_2fontir) crates.
```

fontc/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+[![MIT/Apache 2.0](https://img.shields.io/badge/license-MIT%2FApache-blue.svg)](#license)
+
+fontc: Font Compiler
+
+This crate provides the command line entrypoint into our font compiler.
+
+It's primary task is to accept arguments from the user, create a task graph, and execute it.
+
+This crate can reference all other crates in the workspace.
```
anthrotype, line 11: how about a DOT graph that visualizes all the interdependencies?

anthrotype (approved):

fontdrasil/README.md
```diff
@@ -0,0 +1,9 @@
+# fontdrasil
+
+[![MIT/Apache 2.0](https://img.shields.io/badge/license-MIT%2FApache-blue.svg)](#license)
+
+fontdrasil: Connects all layers of fontc. A reference to Yggdrasil, the world tree.
```
cmyr, line 5:
```suggestion
fontdrasil: Common types and functionality shared between all layers of fontc.
```

fontir/README.md
```diff
@@ -0,0 +1,10 @@
+# fontir
+
+[![MIT/Apache 2.0](https://img.shields.io/badge/license-MIT%2FApache-blue.svg)](#license)
+
+fontir: Font Internal Representation
```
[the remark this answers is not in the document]
anthrotype: me too

fontdrasil/README.md
rsheeter: Less Yggdrasil is clearly worse. I'll take your preamble but leave the Yggdrasil.

fontir/README.md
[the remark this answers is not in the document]
rsheeter: it is, editing sleepy ftw. Will fix.

fontc/README.md
rsheeter: I think perhaps that would be good in the root readme as a follow-on?

rsheeter: https://github.com/googlefonts/fontc/issues/607

PR #604, opened 2023-11-28
Delete SimpleCompile

### LINKED ISSUE #594: Get rid of SimpleCompile
Make it work as suggested in https://github.com/googlefonts/fontc/pull/584/files#r1397981719

rsheeter:
Fixes #594 and removes a few lines of noise from tests

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -814,12 +814,9 @@ mod tests {
 
     fn assert_compiles_with_gpos_and_gsub(
         source: &str,
-        adjust_args: impl Fn(&mut Args),
+        adjust_args: impl Fn(Args) -> Args,
     ) -> TestCompile {
-        let result = TestCompile::compile(source, |mut args| {
-            adjust_args(&mut args);
-            args
-        });
+        let result = TestCompile::compile(source, adjust_args);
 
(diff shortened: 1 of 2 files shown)


fontc/src/lib.rs
```diff
@@ -859,16 +816,12 @@ mod tests {
...
     ) -> TestCompile {
-        let temp_dir = tempdir().unwrap();
-        let build_dir = temp_dir.path();
-        let mut args = Args::for_test(build_dir, source);
-        adjust_args(&mut args);
-        let result = compile(args);
+        let result = TestCompile::compile(source, |mut args| {
+            adjust_args(&mut args);
+            args
+        });
```
cmyr, line 822: if this fn took the same closure type as the one on `TestCompile::compile` (this one mutates args, the other one takes ownership and returns them) then you could just pass the closure straight through here, like, `TestCompile::compile(source, adjust_args);` which is a bit cleaner.

cmyr (approved):
This is much nicer!

fontc/src/lib.rs
rsheeter: I don't know why I made it different! Good catch.

PR #609, opened 2023-11-28
Construct kerning/mark lookup builders direclty

cmyr:
This is followup to #598; since we don't need access to the feature builder in order to construct anchors or value records, we can now construct the various builders directly in the relevant tasks.

### DIFF
--- a/fontbe/src/error.rs
+++ b/fontbe/src/error.rs
@@ -72,7 +72,7 @@ pub enum Error {
     #[error("No glyph class '{0}'")]
     MissingGlyphClass(GlyphName),
     #[error("Multiple assignments for class: {0:?}")]
-    PreviouslyAssignedClass(PreviouslyAssignedClass),
+    PreviouslyAssignedClass(#[from] PreviouslyAssignedClass),
 }
 
 #[derive(Debug)]
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -18,7 +18,7 @@ use ordered_float::OrderedFloat;
 use fea_rs::{
     compile::{
         Compilation, FeatureBuilder, FeatureProvider, MarkToBaseBuilder, MarkToMarkBuilder,
-        PairPosBuilder, VariationInfo,
+        VariationInfo,
     },
     parse::{SourceLoadError, SourceResolver},
     Compiler,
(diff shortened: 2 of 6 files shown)


fontbe/src/error.rs
```diff
@@ -72,7 +72,7 @@ pub enum Error {
     #[error("No glyph class '{0}'")]
     MissingGlyphClass(GlyphName),
     #[error("Multiple assignments for class: {0:?}")]
-    PreviouslyAssignedClass(PreviouslyAssignedClass),
+    PreviouslyAssignedClass(#[from] PreviouslyAssignedClass),
```
rsheeter, line 75: Optional, to me naming closer to the error string (MultipleAssignmentsForClass) would be more intuitive

fontbe/src/features.rs
```diff
@@ -242,58 +235,21 @@ impl<'a> FeatureWriter<'a> {
...
-                    base.gid,
-                    &mark_mark.class,
-                    base.create_anchor_table(),
-                );
-            }
             mark_mark_lookups.push(builder.add_lookup(
                 LookupFlag::default(),
-                Some(mark_mark.filter_set.clone().into()),
-                vec![mark_mark_builder],
+                //TODO: figure out where the mark filter set gets computed & stored?
```
rsheeter, line 250: We have an issue for this, I don't know that the comment is necessary/helpful?

rsheeter (approved):
Much nicer!

rsheeter:
Btw does hyperfine show any improvement on this vs main?

```shell
$ cargo build --release && hyperfine --warmup 25 --runs 100 --prepare 'rm -rf build/' 'target/release/fontc ../OswaldFont/sources/Oswald.glyphs'

$ cargo build --release && hyperfine --warmup 5 --prepare 'rm -rf build/' 'target/release/fontc ../googlesans/source/GoogleSans/GoogleSans.designspace'
````

**Comment by cmyr** [comment]
before:

```
Benchmark 1: target/release/fontc ../fontfiles/googlesans/source/GoogleSans/GoogleSans.designspace
  Time (mean ± σ):      4.349 s ±  0.244 s    [User: 7.100 s, System: 3.082 s]
  Range (min … max):    4.152 s …  4.919 s    10 runs
```

after:

```
Benchmark 1: target/release/fontc ../fontfiles/googlesans/source/GoogleSans/GoogleSans.designspace
  Time (mean ± σ):      4.260 s ±  0.052 s    [User: 7.196 s, System: 3.029 s]
  Range (min … max):    4.188 s …  4.368 s    10 runs
```

so nothing super significant.

This shouldn't really surprise us, since we're not saving ourselves a ton of work here, just removing one extra iteration/allocation pass over all the mark/kern rules.

fontbe/src/error.rs
cmyr: sure, can do that as followup.

rsheeter:
Probably slightly faster, seems about right. Ty for checking.

Issue #495, opened 2023-10-19
Did we pointlessly overcomplicate our serde?


rsheeter:
Why for serde do you have near-identical copies of structs (e.g. StaticMetadata(SerdeRepr)), instead of using custom (de)serialization logic on specific fields you want to handle differently? It gives a first impression of a fair bit of redundancy and boilerplate, how did it come about and can a more elegant solution be used.

rsheeter:
How it came about is that we wanted to write both text and binary versions of things and some of our types were imploding at least one of the serde implementations. I think we flipped at least one of the serde libraries we use from toml => yaml so it is possible some of our serde impls are no longer required.

cmyr:
This has now been significantly simplified.

rsheeter:
Report and fix most appreciated :)

PR #614, opened 2023-11-29
[fontir] Remove custom serde impl for Kerning

cmyr:
The problem was that serde_yaml doesn't like it when BTreeMap is used as the key to a collection. Since this is only the case in `Location`, it's much simpler to just have a cusom serde impl for that type.

I expect this will lead to further simplification, although I will leave that for additional PRs.

progress towards #495

### DIFF
--- a/fontdrasil/src/serde.rs
+++ b/fontdrasil/src/serde.rs
@@ -1,11 +1,15 @@
-use crate::coords::{CoordConverter, DesignCoord, UserCoord};
+use crate::coords::{CoordConverter, DesignCoord, Location, UserCoord};
 use serde::{Deserialize, Serialize};
+use write_fonts::types::Tag;
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub(crate) struct CoordConverterSerdeRepr {
     default_idx: usize,
     user_to_design: Vec<(f32, f32)>,
 }
+
+#[derive(Serialize, Deserialize, Debug)]
+pub(crate) struct LocationSerdeRepr<T>(Vec<(Tag, T)>);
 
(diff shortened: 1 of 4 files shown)


fontdrasil/src/serde.rs
```diff
@@ -1,12 +1,18 @@
...
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub(crate) struct CoordConverterSerdeRepr {
     default_idx: usize,
     user_to_design: Vec<(f32, f32)>,
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub(crate) struct LocationSerdeRepr<T> {
+    parts: Vec<(Tag, T)>,
```
rsheeter, line 13: nit: would be nicer as a tuple struct if that doesn't break serde?

fontir/src/ir.rs
```diff
@@ -206,7 +206,7 @@ type KernValues = BTreeMap<NormalizedLocation, OrderedFloat<f32>>;
 /// In UFO terms, roughly [groups.plist](https://unifiedfontobject.org/versions/ufo3/groups.plist/)
 /// and [kerning.plist](https://unifiedfontobject.org/versions/ufo3/kerning.plist/) combined.
 #[derive(Serialize, Deserialize, Debug, Default, Clone, PartialEq, Eq)]
-#[serde(from = "KerningSerdeRepr", into = "KerningSerdeRepr")]
+//#[serde(from = "KerningSerdeRepr", into = "KerningSerdeRepr")]
```
rsheeter, line 209: actually delete

rsheeter (approved):
Nice!

PR #616, opened 2023-11-29
[fontir] Simplify StateSet serialization

cmyr:
The problem here is that serde really does not want to use an enum as a
key. Fortunately for us, we don't really need the key to be an enum,
since the value in the map encodes whether or not something is a path or
not. This means we can simplify things and just use PathBuf as the key
type, and treat this purely as an implementation detail.


fontir/src/stateset.rs
```diff
@@ -251,7 +251,7 @@ mod tests {
     use filetime::set_file_mtime;
     use tempfile::{tempdir, TempDir};
 
-    use super::{StateDiff, StateIdentifier, StateSet};
+    use super::*;
```
rsheeter, line 254: nit: not a fan of this

fontir/src/stateset.rs
```diff
@@ -13,32 +12,28 @@ use std::{
...
-// The key for a stateful thing of some specific type
-#[derive(Debug, Clone, Hash, PartialEq, Eq)]
-pub enum StateIdentifier {
-    File(PathBuf),
-    Memory(String),
+    // our entries can either be files on disk or items in memory, but
+    // as a simplification just use PathBuf as the key type; we know
+    // if it's an actual file or not based on the value type.
+    pub(crate) entries: HashMap<PathBuf, State>,
 }
```
rsheeter, line 20: Giving up the clear indication of state type seems unfortunate to me, I'd rather we not do that and if it means a bit more serde conversion that's ok

rsheeter:
> we don't really need the key to be an enum

I don't love this change, can we simplify serde w/o doing that?

fontir/src/stateset.rs
cmyr: This is a standard convention for the test module, you'll see it in lots of rust projects.

The basic idea is: assume the test module has access to everything in the parent.

fontir/src/stateset.rs
cmyr: Yea, I tried a bunch of other things first, but this is the simplest, and its worth remembering that the type is preserved in the `State` enum.

fontir/src/stateset.rs
rsheeter: I get the idea but I still don't like it. Where is this standard documented? - it seems suspiciously like something that is largely up to project owners preference.

PR #617, opened 2023-12-01
[ttx_diff] Get ttx_diff working again

cmyr:
An update in fontations had made us not always write the AxisValueArray field, which led to the select_one function triggering an assert.


resources/scripts/ttx_diff.py
```diff
@@ -234,9 +234,10 @@ def stat_like_fontmake(ttx):
     ver.attrib["value"] = "0x00010001"
 
     # fontc reporting a blank axis value array isn't a very interesting diff
-    axis_values = select_one(el, "AxisValueArray")
-    if len(axis_values) == 0:
-        axis_values.getparent().remove(axis_values)
+    if len(el.xpath("AxisValueArray")) > 0:
```
anthrotype: or you could do

```suggestion
    if el.find("AxisValueArray")):
```

anthrotype:
I think you should be able to remove the whole block after `# fontc reporting a blank axis value array` because we now no longer emit that empty array, just like fontmake does

anthrotype:
```diff
diff --git a/resources/scripts/ttx_diff.py b/resources/scripts/ttx_diff.py
index 38546d3..440da4e 100755
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -233,11 +233,6 @@ def stat_like_fontmake(ttx):
     # So until such time as we start writing format 4 axis value tables it doesn't matter
     ver.attrib["value"] = "0x00010001"

-    # fontc reporting a blank axis value array isn't a very interesting diff
-    axis_values = select_one(el, "AxisValueArray")
-    if len(axis_values) == 0:
-        axis_values.getparent().remove(axis_values)
-

 def reduce_diff_noise(fontc, fontmake):
     for ttx in (fontc, fontmake):
```

anthrotype (approved):

PR #622, opened 2023-12-05
move italic_angle to StaticMetadata; set in post table

anthrotype:
the post.italicAngle is not a variable global metric. MVAR does not have a tag for it, but it is implicitly set to the 'slnt' axis value of a variable font (if any). Therefore it's better to move it to the StaticMetadata.

Note how in Glyphs.app italicAngle is clockwise, whereas in UFO and OpenType it is counter-clockwise angle so we need to flip the sign when doing glyphs2fontir.

The SlantedFont.glyphs test font has one 'slnt' axis, with two masters; the default master has italicAngle = -12, the other master has italigAngle = 0. The post table is expected to have italicAngle = -12.


dfrg (approved):
This all looks good to me. Thanks!

PR #624, opened 2023-12-05
[fontc] Validate glyph fliter regex at parse time

cmyr:
This removes the generally unhelpful 'BadRegex' error variant and ensures that if a glyph filter is passed at the command line, we ensure that it is valid regex, and let clap handle displaying a helpful error if not.


fontc/src/change_detector.rs
```diff
@@ -72,7 +72,7 @@ impl ChangeDetector {
             .glyph_name_filter
             .map(|raw_filter| Regex::new(&raw_filter))
             .transpose()
-            .map_err(Error::BadRegex)?;
+            .expect("validated during arg parsing");
```
rsheeter, line 75: If we ever launch other than by CLI this can still happen so I suggest leaving the error despite it being much less likely now

rsheeter (approved):

fontc/src/change_detector.rs
cmyr: do we ever or might we ever do that, outside of a test?

cmyr:
Okay I've updated this so that there is no way to construct an `Args` without ensuring that the regex is valid.

fontc/src/change_detector.rs
rsheeter: We might in future

PR #625, opened 2023-12-05
Improve error messages in fontc/fontir

cmyr:
There were a number of places where we had access to additional information that we were not presenting in our error messages.


anthrotype (approved):
ah thank you!

rsheeter (approved):

PR #626, opened 2023-12-05
[glypsh-reader] fix ParseIntError when custom param has unquoted string

anthrotype:
If you create a New Font from Glyphs.app and then add a 'Variable Font Origin' custom parameter (which we currently do support) and set this to the first and only master that the newly created font comes with (which Glyphs.app assigns the id of `m01` using an unquoted plist string), and then attempt to compile, you will get a ParseIntError when parsing the custom parameters, because glyphs-reader thinks that it is an integer.

I add a reproducer first, and follow up with the actual fix.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2402,6 +2402,18 @@ mod tests {
     }
 
     #[test]
+    fn vf_origin_unquoted_string() {
+        // the 'Variable Font Origin' custom parameter has the value `m01`,
+        // un un-quoted plist string, which happens to be the default master.id
+        // that Glyphs.app assigns to the predefined 'Regular' master that any
+        // "New Font" comes with when it is first crated.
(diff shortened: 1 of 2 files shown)


anthrotype:
https://github.com/googlefonts/fontc/actions/runs/7104626540/job/19340026073?pr=626
```
failures:

---- font::tests::vf_origin_unquoted_string stdout ----
thread 'font::tests::vf_origin_unquoted_string' panicked at glyphs-reader/src/font.rs:411:84:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }


failures:
    font::tests::vf_origin_unquoted_string

test result: FAILED. 50 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.08s
```

rsheeter (approved):

PR #627, opened 2023-12-05
move underline_thickness/position to GlobalMetrics as they are MVAR-iable

anthrotype:
they are not _static_ metadata, as they can vary across masters and MVAR defines tags for them.

Note I'm using the Glyphs.app default underlinePosition for fontir::GlobalMetrics' default `UnderlinePosition` (ie. -0.1 * units_per_em), however in ufo2fontir I override that with ufo2ft's default underlinePosition.

This way, we match the current fontmake output (in absence of explicit underline settings) for both .glyphs and .designspace inputs, even though... well they don't match one another for historical reasons which nobody can care less

### DIFF
--- a/fontbe/src/post.rs
+++ b/fontbe/src/post.rs
@@ -7,6 +7,7 @@ use fontir::orchestration::WorkId as FeWorkId;
 use write_fonts::{
     tables::post::Post,
     types::{FWord, Fixed},
+    OtRound,
 };
 
 use crate::{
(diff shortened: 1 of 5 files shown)


anthrotype:
this is just confirming that the default underlines are the same as they were before, when they were set in the StaticMetadata.

in follow up PRs when I will add support for MVAR, I will add test files with explicit `underlineThickness` and `underlinePosition` per-master custom parameters. So I think this is good to go as is

rsheeter (approved):

Issue #483, opened 2023-10-03
A handful of points are off by one in Oswald


rsheeter:
As noted in https://github.com/googlefonts/fontc/pull/480#issuecomment-1744620883 we're *very* close to matching Oswald. A couple of points are off by 1. @anthrotype notes this might be fine, in which case we need to allow for it in ttx_diff. Ideally we'd fix it and have a perfect match but it is possible it's OK if 1% of points are off by 1, maybe it depends which is closer to the source value, ... bit of digging required.

anthrotype:
the off-by-one diffs seems all related to off-curve points which is why I thought it might have to do with cu2qu, or the different order in which decompose and cu2qu operations are applied in fontc vs fontmake

anthrotype:
also, we don't have any diffs in glyf/gvar in the default profile where we match vanilla fontmake, only in the gftools profile where we add the flatten-components and decompose-transfored-components filter, so it might have to do with the interaction between decomposing and cu2qu'ing

rsheeter:
Reminder to self: tried telling the filters to run before but didn't help. In ttx_diff.py

```python
       cmd += [
            "--filter",
            "FlattenComponentsFilter(pre=True)",
            "--filter",
            "DecomposeTransformedComponentsFilter(pre=True)",
        ]
```

rsheeter:
If the glyphs are decomposed in Glyphs then gvar is off by one for two points. This is probably #235 per @anthrotype.

rsheeter:
I'm increasingly convinced we should allow a non-zero number of off-by-one in both glyf and gvar. Oswald has 3 of 10,000 points off by one ... it's fine. We should still look at #235 because that sounds like it could cause NLI to go haywire.

rsheeter:
(we did ^)

Issue #235, opened 2023-04-04
Add rounding to deltas


rsheeter:
Add the test and fix from https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Tests/varLib/models_test.py#L167

If we do not have rounding within the delta computation we subtract delta(s) that will be applied to avoid double-application, but when applied they are rounded so we can accumulate rounding error from each delta.

anthrotype:
also, note that when rounding deltas (e.g. see below for gvar) in fontTools we use the Python3 built-in `round` function, which uses Banker's rounding, whereby ties are rounded up or down towards the nearest even number:

https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/varLib/__init__.py#L286

This way of rounding is statistically better, and we tend to use in place of the otRound whenever we deal with abstract numbers that aren't visual absolute coordinates.
Also see @behdad's comment in https://github.com/fonttools/fonttools/pull/2214#issuecomment-790742479

I'm not sure if Rust has a direct equivalent for this, but we should probably match that for consistency with the python build pipeline.

anthrotype:
the rust_decimal crate has https://docs.rs/rust_decimal/latest/rust_decimal/enum.RoundingStrategy.html#variant.MidpointNearestEven

anthrotype:
and looks like the rust_decimal::Decimal::round method defaults to theBanker's method:

https://docs.rs/rust_decimal/latest/rust_decimal/struct.Decimal.html#method.round

rsheeter:
https://github.com/rust-lang/rust/issues/96710 appears to be coming. We could likely do it ourselves and flip once that lands.

anthrotype:
Fixed by #1070

PR #629, opened 2023-12-06
Allow 0.1% of coordinates to differ from fontmake by 1

rsheeter:
Arguably having a few points off by one after cu2qu, decomposition, and/or transformation isn't worth the effort to fix. Exploring what allowing a small but non-zero number of x/y values to differ from fontmakes by 1 might look like.

See #483. We should still address rounding (#235) because for NLI that could accumulate a significant error.

@anthrotype wdyt?

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -44,6 +44,11 @@ flags.DEFINE_enum(
     ["both", "fontc", "fontmake", "none"],
     "Which compilers to rebuild with if the output appears to already exist. None is handy when playing with ttx_diff.py itself.",
 )
+flags.DEFINE_float(
+    "off_by_one_budget",
+    0.1,
+    "The percentage of point (glyf) or delta (gvar) values allowed to differ by one without counting as a diff",
+)
 
 
 def run(cmd: MutableSequence, working_dir: Path, log_file: str, **kwargs):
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -233,7 +240,53 @@ def stat_like_fontmake(ttx):
...
+
+    coord_tag = coord_holder.rpartition("/")[-1]
+
+    for fontmake_container in fontmake.xpath(f"//{container}"):
+        name = fontmake_container.attrib[name_attr]
+        fontc_container = select_one(fontc, f"//{container}[@{name_attr}='{name}']")
+
+        fontc_it = fontc_container.iter()
+        for fontmake_el in fontmake_container.iter():
+            fontc_el = next(fontc_it)
```
anthrotype, line 261: you could `zip` the two iterators. Also I think you'd get StopIteration error here if next() fails to yield because fontmake_container is longer than fontc_container

anthrotype (approved):
I think that's fine, it lets us move on to higher priority diffs. Athough I would like at some point to reduce that to 0...

resources/scripts/ttx_diff.py
rsheeter: Oops, good catch.

PR #631, opened 2023-12-06
support strikeout/subscripts/superscripts pos/size settings in ir::GlobalMetrics

anthrotype:


anthrotype:
these are the remaining variable global metrics (that fontmake also supports) that we need to read from the masters to be able to eventually produce MVAR

rsheeter (approved):

PR #635, opened 2023-12-07
Add an error type to the VariationInfo trait

### LINKED ISSUE #535: Add an Error type to VariationInfo
https://github.com/googlefonts/fontc/pull/492#discussion_r1377769316 notes we should add

> an associated error on the trait, which lets the client decide on the error type, since we will generally be returning it back to them anyway

rsheeter:
Drop a dyn while we're at it as any given use of the compiler seems likely to use a specific type. Maybe we should do that to source resolver too?

Fixes #535.

### DIFF
--- a/fea-rs/src/compile/compiler.rs
+++ b/fea-rs/src/compile/compiler.rs
@@ -23,27 +23,27 @@ const DEFAULT_N_MESSAGES_TO_PRINT: usize = 100;
 /// This is intended as the principal public API for this crate.
 ///
 /// ```no_run
-/// # use fea_rs::Compiler;
+/// # use fea_rs::{Compiler, compile::{NopFeatureProvider, NopVariationInfo}};
 /// # fn make_glyph_map() -> fea_rs::GlyphMap { todo!() }
 /// let glyph_map = make_glyph_map();
-/// let my_font_bytes = Compiler::new("path/to/features.fea", &glyph_map)
+/// let my_font_bytes = Compiler::<'_, NopFeatureProvider, NopVariationInfo>::new("path/to/features.fea", &glyph_map)
 ///     .verbose(true)
 ///     .compile_binary().unwrap();
 /// ```
-pub struct Compiler<'a> {
+pub struct Compiler<'a, F: FeatureProvider, V: VariationInfo> {
     root_path: OsString,
     project_root: Option<PathBuf>,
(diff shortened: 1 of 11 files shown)


fea-rs/src/compile/variations.rs
```diff
@@ -14,6 +19,9 @@ use write_fonts::{tables::variations::VariationRegion, types::Tag};
 ///
 /// This trait abstracts over that info.
 pub trait VariationInfo {
+    /// The error type
+    type Error: Debug + Display;
```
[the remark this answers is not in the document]
rsheeter: I needed Display because we assume that in some of our printing. Good call re Error.

fea-rs/src/compile/compiler.rs
```diff
@@ -23,27 +23,27 @@ const DEFAULT_N_MESSAGES_TO_PRINT: usize = 100;
 /// This is intended as the principal public API for this crate.
 ///
 /// ```no_run
-/// # use fea_rs::Compiler;
+/// # use fea_rs::{Compiler, compile::{NopFeatureProvider, NopVariationInfo}};
 /// # fn make_glyph_map() -> fea_rs::GlyphMap { todo!() }
 /// let glyph_map = make_glyph_map();
-/// let my_font_bytes = Compiler::new("path/to/features.fea", &glyph_map)
+/// let my_font_bytes = Compiler::<'_, NopFeatureProvider, NopVariationInfo>::new("path/to/features.fea", &glyph_map)
```
anthrotype, line 29: I don't exactly know what the consequences of this changes are (or why is it important to not have the `dyn` and such), but merely looking at the API, this appears to make constructing a new fea-rs Compiler a lot more verbose..

cmyr: I agree that this feels like a worse API, I'll play around a bit and see if there's a middle ground.

rsheeter: dyn is an extra layer of indirection to make calls. One of Rusts strengths is that you don't have to do that very often. Every execution of a compiler will do ever so slightly less work.

Ideally we'd have a default but iiuc from https://rust-lang.github.io/rfcs/0213-defaulted-type-params.html that is not currently possible.

rsheeter:
> this should probably have been two PRs

Fair point.

PR #637, opened 2023-12-07
use units_per_em for default caret_slope_rise

anthrotype:
I just made a similar change in ufo2ft, see https://github.com/googlefonts/ufo2ft/issues/804

JMM


cmyr (approved):
looks good!

PR #638, opened 2023-12-07
match ufo2ft's default ascender/descender settings

anthrotype:
if the source file does provide generic ascender/descender values, we need to use those to derive the OS/2 and hhea values. We want to match ufo2ft defaults.

Surprisingly the code already contained urls to ufo2ft's source code but it didn't actually match the code it linked to..

### DIFF
--- a/fontbe/src/os2.rs
+++ b/fontbe/src/os2.rs
@@ -864,7 +864,8 @@ mod tests {
     #[test]
     fn build_basic_os2() {
         let default_location = NormalizedLocation::new();
-        let mut global_metrics = GlobalMetrics::new(default_location.clone(), 1000, None, 0.0);
+        let mut global_metrics =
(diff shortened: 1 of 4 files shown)


anthrotype:
running ttx_diff.py againt WghtVar.glyphs now produces the same OS/2 and hhea vertical metrics as those produced by fontmake

anthrotype:
also, note the matter here is not about what's the correct vertical metrics settings (a can of worms) but rather "what does ufo2ft does by default"

rsheeter:
> Surprisingly the code already contained urls to ufo2ft's source code but it didn't actually match the code it linked to..

IIRC the intent for some was merely to match the default. For example, if ascender wasn't wired through so the best you can do is 0.8 * upem to match the python default. Os2TypoAscender looks to me like it was transcribed from the wrong python function as I do see one with similar name (openTypeOS2WinAscentFallback) that does add typo line gap.

rsheeter (approved):

Issue #610, opened 2023-11-29
Generating GPOS takes a long time


rsheeter:
If I print the time for `bytes_for` in `fontbe/src/font.rs` GPOS takes ~10x GSUB which takes ~50x the next table in line:

```
$ cargo build --release && rm -rf build/ && time target/release/fontc ../googlesans/source/GoogleSans/GoogleSans.designspace
...
bytes_for Avar took 0.0ms
bytes_for Cmap took 0.1ms
bytes_for Fvar took 0.0ms
bytes_for Head took 0.0ms
bytes_for Hhea took 0.0ms
bytes_for Hmtx took 0.0ms
bytes_for Glyf took 0.3ms
bytes_for Gvar took 0.6ms
bytes_for Loca took 0.0ms
bytes_for Maxp took 0.0ms
bytes_for Name took 0.1ms
bytes_for Os2 took 0.0ms
bytes_for Post took 1.7ms
bytes_for Stat took 0.0ms
bytes_for Hvar took 0.1ms
bytes_for Gpos took 1025.9ms
bytes_for Gsub took 85.5ms
bytes_for Gdef took 0.3ms

real	0m6.334s
```

GPOS is about 16% of the total build time (real). We know we generate over-large GPOS tables (https://github.com/googlefonts/fontc/issues/593), that probably doesn't help.

belluzj:
Random thought (I've told @anthrotype on the side I think): you may benefit from translating to Rust the GPOS compaction from here: https://github.com/fonttools/fonttools/pull/2326. It's not slow even though it's Python, so it might be fast enough in Rust, and it doesn't need to know about the glyph's Unicode properties (so no need for GSUB closure) in order to decide how to split the subtables, it does so only by looking at the filled-in/empty areas of the matrices. Let me know if you have any questions.

rsheeter:
Kerning is the majority of the critical path when I build GS locally. Hopefully we can speed it up and/or parallelize it!

![image](https://github.com/googlefonts/fontc/assets/6466432/db1c9949-a729-48b1-9baa-44407a6c27cc)

cmyr:
why is kerning there twice?

rsheeter:
The first is FE the second is BE. Apparently I didn't see the need to distinguish them very much :D

PR #642, opened 2023-12-08
Try spreading FE kerning across threads per location

### LINKED ISSUE #639: Consider splitting FE kerning/marks work into per-master tasks
Currently we have one task that (at least for ufo) sequentially loads each master and collects the kerning rules from each into one big structure. Since we will block on the IO, it might make sense to split this operation up and generate structures on a per-master basis, merging them at the end?

Whether or not this would represent a real improvement likely depends in large part on how efficiently we are spawning new work tasks.

rsheeter:
Step toward #639. Doesn't produce a speedup yet but I like the better threading. Basically breaks up the leftmost kern from https://github.com/googlefonts/fontc/issues/610#issuecomment-1833125480. 

image

Things to poke in subsequent PRs:

1. Read file sooner, that has no need to block on glyph order etc
1. Looks like kern_at could have started sooner than it did
1. BE kerning spends a lot of time in a big loop, might be worth splitting up
   * GS has about 60k pairs to loop over, we could just break into N groups, say num CPUs or just a fixed value

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -495,7 +495,7 @@ mod tests {
 
     use fea_rs::compile::VariationInfo;
     use fontdrasil::{
-        coords::{CoordConverter, DesignCoord, NormalizedCoord, UserCoord},
+        coords::{Coord, CoordConverter, DesignCoord, NormalizedCoord, UserCoord},
         types::Axis,
     };
     use fontir::ir::StaticMetadata;
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -2,7 +2,7 @@
 
 use std::{
     cell::RefCell,
-    collections::{BTreeMap, HashMap},
+    collections::HashMap,
     ffi::{OsStr, OsString},
     fmt::Display,
     fs,
@@ -123,12 +123,12 @@ impl<'a> FeaVariationInfo<'a> {
 //except they have slightly different inputs?
 pub(crate) fn resolve_variable_metric(
     static_metadata: &StaticMetadata,
-    values: &BTreeMap<NormalizedLocation, OrderedFloat<f32>>,
+    values: impl IntoIterator<Item = (NormalizedLocation, OrderedFloat<f32>)>,
 ) -> Result<(i16, Vec<(VariationRegion, i16)>), Error> {
     let var_model = &static_metadata.variation_model;
 
     let point_seqs = values
-        .iter()
+        .into_iter()
         .map(|(pos, value)| (pos.to_owned(), vec![value.0 as f64]))
         .collect();
     let raw_deltas: Vec<_> = var_model
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -121,9 +121,9 @@ impl<'a> FeaVariationInfo<'a> {
 
 //NOTE: this is basically identical to the same method on FeaVariationInfo,
 //except they have slightly different inputs?
-pub(crate) fn resolve_variable_metric(
+pub(crate) fn resolve_variable_metric<'a>(
     static_metadata: &StaticMetadata,
-    values: impl IntoIterator<Item = (NormalizedLocation, OrderedFloat<f32>)>,
+    values: impl Iterator<Item = &'a (NormalizedLocation, OrderedFloat<f32>)>,
 ) -> Result<(i16, Vec<(VariationRegion, i16)>), Error> {
     let var_model = &static_metadata.variation_model;
 
--- a/fontbe/src/kern.rs
+++ b/fontbe/src/kern.rs
@@ -1,13 +1,20 @@
 //! Generates a [Kerning] datastructure to be fed to fea-rs
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 
 use fea_rs::{
     compile::{PairPosBuilder, ValueRecord as ValueRecordBuilder},
     GlyphSet,
 };
-use fontdrasil::orchestration::{Access, AccessBuilder, Work};
-use fontir::{ir::KernParticipant, orchestration::WorkId as FeWorkId};
+use fontdrasil::{
+    coords::NormalizedLocation,
+    orchestration::{Access, AccessBuilder, Work},
+};
+use fontir::{
+    ir::{KernPair, KernParticipant},
+    orchestration::WorkId as FeWorkId,
+};
+use ordered_float::OrderedFloat;
 use write_fonts::types::GlyphId;
 
 use crate::{
@@ -31,8 +38,9 @@ impl Work<Context, AnyWorkId, Error> for KerningWork {
     fn read_access(&self) -> Access<AnyWorkId> {
         AccessBuilder::new()
             .variant(FeWorkId::StaticMetadata)
-            .variant(FeWorkId::Kerning)
             .variant(FeWorkId::GlyphOrder)
+            .variant(FeWorkId::KerningGroups)
+            .variant(FeWorkId::KerningAtLocation(NormalizedLocation::default()))
             .build()
     }
 
@@ -40,7 +48,9 @@ impl Work<Context, AnyWorkId, Error> for KerningWork {
     fn exec(&self, context: &Context) -> Result<(), Error> {
         let static_metadata = context.ir.static_metadata.get();
         let glyph_order = context.ir.glyph_order.get();
-        let ir_kerning = context.ir.kerning.get();
+        let ir_groups = context.ir.kerning_groups.get();
+        let ir_kerns = context.ir.kerning_at.all();
+
         let gid = |name| {
             glyph_order
                 .glyph_id(name)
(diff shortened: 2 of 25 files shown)


fontbe/src/features.rs
```diff
@@ -123,12 +123,12 @@ impl<'a> FeaVariationInfo<'a> {
 //except they have slightly different inputs?
 pub(crate) fn resolve_variable_metric(
     static_metadata: &StaticMetadata,
-    values: &BTreeMap<NormalizedLocation, OrderedFloat<f32>>,
+    values: impl IntoIterator<Item = (NormalizedLocation, OrderedFloat<f32>)>,
```
cmyr, line 126: this reminds me that another potential speedup could be memoizing the results of this fn call, I think we call it with a lot of redundant values and it's expensive.

Also possibly a better signature would be just `impl Iterator`, since this signature requires ownership which means an additional clone at the call site?

fontbe/src/kern.rs
```diff
@@ -64,8 +74,25 @@ impl Work<Context, AnyWorkId, Error> for KerningWork {
 
         let mut builder = PairPosBuilder::default();
 
+        // Add IR kerns to builder. IR kerns are split by location so put them back together again.
+        // We want to iterate over (left, right), map<location: adjustment>
+        let mut kerns: HashMap<&KernPair, Vec<(NormalizedLocation, OrderedFloat<f32>)>> =
```
cmyr, line 79: I'm also typing here right now, and something I'm wondering is whether it makes sense to group everything by glyph in the IR? my particular observation is that building this data structure is slow, and we could build the full data structure up after we do alignment.

fontc/src/change_detector.rs
```diff
@@ -249,28 +289,37 @@ impl ChangeDetector {
...
             || self.glyph_order_ir_change()
             || self.current_inputs.features != self.prev_inputs.features
             || !self
                 .ir_paths
-                .target_file(&FeWorkIdentifier::Kerning)
+                .target_file(&FeWorkIdentifier::KerningGroups)
+                .is_file()
+    }
+
+    pub fn kerning_at_ir_change(&self, at: NormalizedLocation) -> bool {
```
cmyr, line 317: In a world where we had a more concrete graph representation ideally we would be to use that to determine what gets invalidated, and then we could simplify this API to something like,

```rust
fn is_up_to_date(&self, work_id: &WorkId) -> bool { ..}
```

or something?

fontc/src/lib.rs
```diff
@@ -1760,25 +1789,36 @@ mod tests {
...
-                    values
-                        .iter()
-                        .map(|(loc, val)| {
-                            assert_eq!(loc.axis_tags().count(), 1, "Should be only wght");
-                            let (axis, pos) = loc.iter().next().unwrap();
-                            (format!("{axis} {}", pos.to_f32()), val.0)
-                        })
-                        .collect::<Vec<_>>(),
-                )
+        let wght = Tag::from_be_bytes(*b"wght");
```
cmyr, line 1792: this works, but you should just use `Tag::new`.

fontc/src/workload.rs
```diff
@@ -531,20 +548,33 @@ impl<'a> Workload<'a> {
...
+                    self.timer.add(timing.complete());
                 }
 
-                // Block for things to phone home to say they are done
-                // Then complete everything that has reported since our last check
+                // Complete everything that has reported since our last check
                 if successes.is_empty() {
+                    let timing = create_timer(AnyWorkId::InternalTiming("rc"), nth_wave)
+                        .queued()
+                        .run();
```
cmyr, line 564: I think tracing is going to become very important for us and we should investigate current good practices. back in the xi days we wrote [`xi_trace`](https://docs.rs/xi-trace/latest/xi_trace/) which was very useful, but there are probably better alternatives now.

fontdrasil/src/coords.rs
```diff
@@ -363,6 +363,21 @@ impl Debug for NormalizedLocation {
     }
 }
 
+/// For testing only
+#[doc(hidden)]
+pub fn norm_loc(positions: &[(&str, f32)]) -> NormalizedLocation {
```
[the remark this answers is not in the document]
cmyr: I think even if it's just for testing it makes sense for this to be a type method on `NormalizedLocation`.

glyphs2fontir/src/source.rs
```diff
@@ -610,8 +625,14 @@ fn is_kerning_class(name: &str) -> bool {
 }
 
 #[derive(Debug)]
-struct KerningWork {
+struct KerningGroupWork {
+    font_info: Arc<FontInfo>,
+}
+
+#[derive(Debug)]
+struct KerningAtWork {
```
cmyr, line 633: Maybe `KerningFragment`?

ufo2fontir/src/source.rs
```diff
@@ -1236,74 +1259,118 @@ impl Work<Context, WorkId, WorkError> for KerningWork {
...
+            master_locations(&static_metadata.all_source_axes, &self.designspace.sources);
+
+        // We know all the groups, read all the kerning and update group naming
+        let source = self
+            .designspace
+            .sources
+            .iter()
+            .find(|s| {
+                !is_glyph_only(s)
+                    && *master_locations.get(s.name.as_ref().unwrap()).unwrap() == self.at
```
cmyr, line 1315: unrelated, but do we validate at any point that the name always exists? Or would it make sense to store the validated set of master locations in the static metadata?

cmyr (approved):
A bunch of comments inline, most of which are just general questions prompted by this PR.

fontir/src/ir.rs
```diff
@@ -193,29 +193,39 @@ impl IntoIterator for GlyphOrder {
...
+    ///
+    /// The rhs should be the name used in the groups map.
+    pub old_to_new_group_names: BTreeMap<GroupName, GroupName>,
+}
+
+/// IR representation of kerning for a location.
+///
+/// In UFO terms, roughly [kerning.plist](https://unifiedfontobject.org/versions/ufo3/kerning.plist/).
+#[derive(Serialize, Deserialize, Debug, Default, Clone, PartialEq, Eq)]
+pub struct KerningAtLocation {
```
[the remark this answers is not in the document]
rsheeter: I like that, *much* better name. Will update.

fontdrasil/src/coords.rs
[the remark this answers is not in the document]
rsheeter: Seems obvious in retrospect, will move to NormalizedLocation

fontc/src/workload.rs
rsheeter: Flamegraph is awesome but at times over-broad. What I'd really like right now is the ability to drill from `--emit-timing` => flamegraph of just that task.

fontbe/src/features.rs
rsheeter: +1 to impl Iterator, will change

glyphs2fontir/src/source.rs
rsheeter: I like Cosimo's KerningInstance, consistant with other fontc naming.

fontbe/src/kern.rs
rsheeter: https://github.com/googlefonts/fontc/pull/644 suggests once sharded kerning isn't too bad, lets keep that in our back pocket until it pops out in a profile. Does the fea compilation contain any opportunity for ||ism? - it ends up being quite long, as does packing tables (https://github.com/googlefonts/fontc/issues/610).

fontc/src/change_detector.rs
rsheeter: +1 but not in this PR :D

fontbe/src/kern.rs
rsheeter: I suggest we resolve here and pursue in #610

fontc/src/workload.rs
rsheeter: https://github.com/googlefonts/fontc/issues/646

ufo2fontir/src/source.rs
rsheeter: I believe at this point all the named sources have to have locations so this could be made nicer

fontir/src/ir.rs
```diff
@@ -193,29 +193,39 @@ impl IntoIterator for GlyphOrder {
...
-/// and [kerning.plist](https://unifiedfontobject.org/versions/ufo3/kerning.plist/) combined.
+/// plus the set of locations that can have kerning.
 #[derive(Serialize, Deserialize, Debug, Default, Clone, PartialEq, Eq)]
-pub struct Kerning {
+pub struct KerningGroups {
     pub groups: BTreeMap<GroupName, BTreeSet<GlyphName>>,
+    /// The locations that have kerning defined.
+    ///
+    /// Must be a subset of the master locations.
+    pub locations: BTreeSet<NormalizedLocation>,
```
cmyr, line 208: now that i'm actually working in this file again, what's the rationale for having the locations here? Is it related to groups in particular in any way?

rsheeter: Those are the locations you should spawn kern instance jobs for

PR #644, opened 2023-12-12
Spread BE kerning across threads

rsheeter:
Builds on #642.

Numbers bounce around a bit run to run but mean + σ on branch is consistantly better than mean - σ on main.
```shell
$ cargo build --release && hyperfine --warmup 5 --prepare 'rm -rf build/' 'target/release/fontc ../googlesans/source/GoogleSans/GoogleSans.designspace'

# Linux
main   6.925 s ±  0.248 s
branch 6.356 s ±  0.217 s

# Mac
main  3.904 s ±  0.370 s
branch 3.828 s ±  0.037 s
```

Linux  `$  cargo build --release &&  rm -rf build/ && target/release/fontc ../googlesans/source/GoogleSans/GoogleSans.designspace --emit-timing`:

image

Observations:

* fea is 15%
* font, which per #610 is mostly packing gpos, is 20%
* kernat should probably start sooner, such as right after glyph order
   * I need to take another pass at making sure we refer to kern instance not kern at :)


fontbe/src/orchestration.rs
```diff
@@ -377,6 +401,48 @@ impl Persistable for Kerning {
...
+            PairPosEntry::Class(set0, rec0, set1, rec1) => {
+                builder.insert_classes(set0, rec0, set1, rec1)
+            }
+        }
+    }
+}
+
+/// A chunk of kerning that needs to be fed into a [PairPosBuilder]
+#[derive(Default, Clone, Serialize, Deserialize, PartialEq)]
+pub struct KernSegment {
```
anthrotype, line 425: for glyf/gvar (and also for the KerningSegment's timing label) you used the word "fragment" instead of "segment", might make sense to be consistent

fontbe/src/paths.rs
```diff
@@ -57,6 +57,11 @@ impl Paths {
         self.glyph_dir.join(safe_filename(name, ".gvar"))
     }
 
+    fn kern_slice_file(&self, segment: usize) -> PathBuf {
```
anthrotype, line 60: segment, slice, fragment?... so many synonyms

fontbe/src/paths.rs
```diff
@@ -76,7 +81,9 @@ impl Paths {
             WorkId::Hhea => self.build_dir.join("hhea.table"),
             WorkId::Hmtx => self.build_dir.join("hmtx.table"),
             WorkId::Hvar => self.build_dir.join("hvar.table"),
-            WorkId::Kerning => self.build_dir.join("kerning.bin"),
+            WorkId::KernPairs => self.build_dir.join("kern_scatter.bin"),
```
anthrotype, line 84: KernPairs=>"kern_scatter"? why not make the names of the structs and paths (and timing labels) be consistent with one another?

fontbe/src/orchestration.rs
```diff
@@ -120,7 +127,9 @@ impl Identifier for WorkId {
             WorkId::Hhea => "BeHhea",
             WorkId::Hmtx => "BeHmtx",
             WorkId::Hvar => "BeHvar",
-            WorkId::Kerning => "BeKerning",
+            WorkId::KernPairs => "BeKerning",
+            WorkId::KernSegment(..) => "BeKern",
+            WorkId::Kerns => "BeKernGather",
```
anthrotype, line 132: the non-matching between WorkId variant name and identifiers makes me very confused

fontbe/src/orchestration.rs
```diff
@@ -352,21 +361,36 @@ impl Persistable for Marks {
     }
 }
 
-/// Precomputed kerning, to the extent possible given that we cannot create temporary var indices in advance.
-///
-/// TODO: update once <https://github.com/googlefonts/fontc/issues/571> is fixed. Then we can build a
-/// [`fea_rs::compile::PairPosBuilder`] in advance.
 #[derive(Default, Clone, Serialize, Deserialize, PartialEq)]
-pub struct Kerning {
+pub struct Kerns {
```
anthrotype, line 365: hard to figure out what a "Kerns" struct is/do, how it differs from similarly named "Kerning"..
Also "kern" is more frequently used as a verb; when used as a noun, it's "a part of a typeset letter that projects beyond its side bearings" like the tails of a cursive f.  See https://www.merriam-webster.com/dictionary/kern

Admittedly I don't have an alternative to suggest at the moment.

fontbe/src/kern.rs
```diff
@@ -124,29 +210,72 @@ impl Work<Context, AnyWorkId, Error> for KerningWork {
...
+        let segments = context.kern_segments.all();
+        let mut builder = PairPosBuilder::default();
+
+        for ppe in segments
+            .iter()
+            .flat_map(|(_, segment)| segment.kerns.iter())
+            .cloned()
+        {
+            ppe.add_to(&mut builder);
+        }
```
anthrotype, line 272: (how) do we ensure that the pairs are added to the builder in a deterministic order? IIUC, the kern_segments are produced in different threads in parallel so we might not guarantee their ordering is the same as if they were built serially

anthrotype: Or perhaps the order does not matter because the builder will sort them somehow?

anthrotype: I tried to compile GS multiple times from this branch, and then diff the resulting ttx files, and can see that GDEF and GPOS tables are different each time I run fontc.. Maybe we should have a CI check that ensures that building the same source file multiple times produces identical results.

anthrotype: actually this happens from the current main branch too.. Something very weird is going on, the size of the GPOS table seems to fluctuate, and ttx dumps contain more or less data every time I run fontc on the same GoogleSans.designspace...

anthrotype: it seems like the indeterminism of GDEF/GPOS has been introduced after https://github.com/googlefonts/fontc/pull/642 got merged to main, if I checkout immediately before that it seems produce identical GDEF and GPOS tables each time I re-run fontc.

anthrotype: besides GDEF/GPOS, the files are still not 100% identical. I see diffs in GSUB with stylistic set features using FeatureParams, the nameID seems to change every time I rebuild.. Ok time to file a separate issue

```diff
--- /dev/fd/63  2023-12-13 13:24:16.601441808 +0000
+++ /dev/fd/62  2023-12-13 13:24:16.601441808 +0000
@@ -7635,7 +7635,7 @@
     <flags value="00000000 00000011"/>
     <unitsPerEm value="1000"/>
     <created value="Thu Jul  6 09:41:44 2017"/>
-    <modified value="Wed Dec 13 13:18:42 2023"/>
+    <modified value="Wed Dec 13 13:20:01 2023"/>
     <xMin value="-978"/>
     <yMin value="-989"/>
     <xMax value="3284"/>
@@ -4137695,7 +4137695,7 @@
         <Feature>
           <FeatureParamsStylisticSet>
             <Version value="0"/>
-            <UINameID value="258"/>  <!-- Grade -->
+            <UINameID value="263"/>  <!-- Text Medium -->
           </FeatureParamsStylisticSet>
           <!-- LookupCount=1 -->
           <LookupListIndex index="0" value="1462"/>
@@ -4137706,7 +4137706,7 @@
         <Feature>
           <FeatureParamsStylisticSet>
             <Version value="0"/>
-            <UINameID value="257"/>  <!-- Weight -->
+            <UINameID value="261"/>  <!-- Bold -->
           </FeatureParamsStylisticSet>
           <!-- LookupCount=1 -->
           <LookupListIndex index="0" value="1463"/>
@@ -4137728,7 +4137728,7 @@
         <Feature>
           <FeatureParamsStylisticSet>
             <Version value="0"/>
-            <UINameID value="260"/>  <!-- Medium -->
+            <UINameID value="256"/>  <!-- Size -->
           </FeatureParamsStylisticSet>
           <!-- LookupCount=1 -->
           <LookupListIndex index="0" value="1464"/>
```

anthrotype: https://github.com/googlefonts/fontc/issues/647

fontbe/src/kern.rs
```diff
@@ -9,55 +9,91 @@ use fea_rs::{
...
     orchestration::{Access, AccessBuilder, Work},
+    types::GlyphName,
 };
 use fontir::{
-    ir::{KernPair, KernParticipant},
+    ir::{GlyphOrder, KernPair, KernParticipant},
     orchestration::WorkId as FeWorkId,
 };
-use ordered_float::OrderedFloat;
+use log::debug;
```
cmyr, line 18: Just as a general observation but I prefer not to import these macros by name, and instead just call them as `log::debug!("my message")`. The main rationale for this is simply that it makes it easier to grep for all the places where you're logging. A secondary rationale is that there's a (loose) convention in Rust to avoid importing functions directly, and instead import their parent module.. I can't find a great source for this, but it's [mentioned here by klabnik](https://github.com/rust-lang/style-team/issues/24#issuecomment-252288601). I like this convention but am not zealous about it.

fontbe/src/kern.rs
```diff
@@ -9,55 +9,91 @@ use fea_rs::{
...
+pub fn create_kerns_work() -> Box<BeWork> {
+    Box::new(KerningGatherWork {})
+}
+
+fn gid(glyph_order: &GlyphOrder, name: &GlyphName) -> Result<GlyphId, Error> {
+    glyph_order
+        .glyph_id(name)
+        .map(|gid| GlyphId::new(gid as u16))
+        .ok_or_else(|| Error::MissingGlyphId(name.clone()))
+}
```
cmyr, line 72: Maybe we just make this a method on `GlyphOrder`? Not sure why I didn't do that initially, probably no good reason.

fontbe/src/kern.rs
cmyr: it should be easy enough to track the start position of each segment, and then sort them here at the end?

fontbe/src/orchestration.rs
cmyr: yea maybe something like `FinalizedKerning`, if I'm understanding the rationale correctly

cmyr (approved):
So the main thing missing from all of this, as best I can tell, is some high-level overview of how it works: basically what are the different stages of the work and how to they fit together, what are the invariants at each stage, etc; I think that would be very useful to have somewhere.

fontbe/src/paths.rs
rsheeter: I'll try to land on fragment :)

fontbe/src/orchestration.rs
rsheeter: Updated to GatherIrKerning, KernFragment, and GatherKernFragments

fontbe/src/kern.rs
rsheeter: I think we do this relatively consistently; maybe file an issue and we'll fix it eveyrwhere plus add a presubmit?

fontbe/src/kern.rs
rsheeter: Ha, indeed. Will do.

fontbe/src/kern.rs
rsheeter: Fantastic catch, we're clearly missing a test :( Will address in a subsequent PR.

PR #648, opened 2023-12-13
build MVAR table from variable GlobalMetrics 

### LINKED ISSUE #531: Emit MVAR
fontmake / fonttools produce an MVAR table when the masters have different global metrics (ascender/descender etc.).

https://learn.microsoft.com/en-us/typography/opentype/spec/mvar

https://github.com/fonttools/fonttools/blob/49570c359966897386d9a71c90cb9dd2f4da30f9/Lib/fontTools/varLib/__init__.py#L660-L735

We currently don't and we should match that.

anthrotype:
Fixes #531 

This matches fontmake's and fonttools' MVAR tables for a bunch of fonts I tested locally, will add more thorough unit tests soon.

Note how in ufo2fontir and glyphs2fontir we make sure that non-default global metrics masters are not sparse (the default metrics all populated based on basic metrics like UPEM or italic angle), that's because when varLib merges them into MVAR, it starts with _whole_ OS/2 or post etc. tables, so it is never the case that, e.g., only the ascender or the descender provides masters. 

Our fontir GlobalMetrics allow for sparse-ness but we currently want to match the output of fontmake in this regard, until maybe some new source format is designed for sparse variable-first use cases.


fontbe/src/mvar.rs
```diff
@@ -0,0 +1,329 @@
...
+pub fn create_mvar_work() -> Box<BeWork> {
+    Box::new(MvarWork {})
+}
+
+/// Helper to build MVAR table from global metrics sources.
+struct MvarBuilder {
+    /// Variation axes
+    axes: Vec<Axis>,
+    /// Sparse variation models, keyed by the set of locations they define
+    models: HashMap<BTreeSet<NormalizedLocation>, VariationModel>,
```
rsheeter, line 42: I assume it's a hashmap (unstable order) of btreesets (stable order) for a good reason but the comment doesn't tell me what that reason is, perhaps it could?

fontbe/src/mvar.rs
```diff
@@ -0,0 +1,329 @@
...
+
+    fn read_access(&self) -> Access<AnyWorkId> {
+        AccessBuilder::new()
+            .variant(FeWorkId::StaticMetadata)
+            .variant(FeWorkId::GlobalMetrics)
+            .build()
+    }
+
+    /// Generate [MVAR](https://learn.microsoft.com/en-us/typography/opentype/spec/MVAR)
+    fn exec(&self, context: &Context) -> Result<(), Error> {
```
rsheeter, line 138: Optional, is there corresponding python code worth referencing?

fontbe/src/mvar.rs
anthrotype: well, because it must be hashable otherwise it can't be used as a key in a hashmap? we do similar things all over fontc

fontbe/src/mvar.rs
```diff
@@ -0,0 +1,329 @@
...
+        let static_metadata = context.ir.static_metadata.get();
+        let metrics = context.ir.global_metrics.get();
+        let var_model = &static_metadata.variation_model;
+
+        let mut mvar_builder = MvarBuilder::new(var_model.clone());
+        for (metric, values) in metrics.iter() {
+            // maybe we should get rid of GlobalMetric variants that aren't MVAR-relevant?
```
rsheeter, line 145: Suggest making this an issue rather than a comment

fontbe/src/mvar.rs
```diff
@@ -0,0 +1,329 @@
...
+        assert_eq!(mvar.value_records().len(), 1);
+
+        let rec = &mvar.value_records()[0];
+        assert_eq!(rec.value_tag(), Tag::new(b"xhgt"));
+        assert_eq!(rec.delta_set_outer_index(), 0);
+        assert_eq!(rec.delta_set_inner_index(), 0);
+
+        let Some(Ok(varstore)) = mvar.item_variation_store() else {
+            panic!("MVAR has no ItemVariationStore?!");
+        };
```
rsheeter, line 246: Was there a particular reason mvar.item_variation_store.unwrap() wouldn't do?

fontbe/src/mvar.rs
```diff
@@ -0,0 +1,329 @@
...
+        let bytes = dump_table(&mvar).unwrap();
+        let mvar = read_mvar::Mvar::read(FontData::new(&bytes)).unwrap();
+
+        assert_eq!(mvar.version(), MajorMinor::VERSION_1_0);
+        assert_eq!(mvar.value_records().len(), 1);
+
+        let rec = &mvar.value_records()[0];
+        assert_eq!(rec.value_tag(), Tag::new(b"xhgt"));
+        assert_eq!(rec.delta_set_outer_index(), 0);
+        assert_eq!(rec.delta_set_inner_index(), 0);
```
rsheeter, line 242: Optional, I continue to greatly prefer the test output of this sort of thing done as one assert, e.g.

```rust
assert_eq!(vec![(Tag::new(b"xhgt"), 0, 0),], mvar.value_records().iter().map(...)]
```

I've always found it much easier to troubleshoot and harder to forget to assert the length

fontbe/src/mvar.rs
anthrotype: Done

fontbe/src/mvar.rs
```diff
@@ -0,0 +1,329 @@
...
+        let Some(mvar) = builder.build() else {
+            panic!("no MVAR?!");
+        };
+
+        let bytes = dump_table(&mvar).unwrap();
+        let mvar = read_mvar::Mvar::read(FontData::new(&bytes)).unwrap();
+
+        assert_eq!(mvar.value_records().len(), 2);
+        assert_eq!(mvar.value_records()[0].value_tag(), Tag::new(b"strs"));
+        assert_eq!(mvar.value_records()[1].value_tag(), Tag::new(b"xhgt"));
```
rsheeter, line 320: Optional, why not one assert?

Do you care which index strs and xhgt end up in or just that there is one value record for each of them? If the order doesn't matter perhaps make the test insensitive to it?

anthrotype: well, I do care, because they get sorted alphabetically by tag

fontir/src/ir.rs
```diff
@@ -362,38 +362,117 @@ pub enum GlobalMetric {
...
+        // We support the same subset of the metrics defined in the spec
+        // as fonttools does (except vertical ones which we don't support yet):
+        // https://github.com/fonttools/fonttools/blob/0c5cb3b/Lib/fontTools/varLib/mvar.py
+        match self {
+            GlobalMetric::Os2TypoAscender => Some(Tag::new(b"hasc")),
+            GlobalMetric::Os2TypoDescender => Some(Tag::new(b"hdsc")),
+            GlobalMetric::Os2TypoLineGap => Some(Tag::new(b"hlgp")),
+            GlobalMetric::Os2WinAscent => Some(Tag::new(b"hcla")),
+            GlobalMetric::Os2WinDescent => Some(Tag::new(b"hcld")),
+            // TODO: support vertical global metrics
```
rsheeter, line 382: Suggest filing an issue and citing it, e.g. // TODO(#xyz): support vertical global metrics

glyphs-reader/src/font.rs
```diff
@@ -764,24 +764,24 @@ impl FontMaster {
...
-    pub fn cap_height(&self) -> Option<OrderedFloat<f64>> {
-        self.read_metric("cap height")
+    pub fn cap_height(&self) -> Option<f64> {
+        self.read_metric("cap height").map(|x| x.into_inner())
     }
 
-    pub fn italic_angle(&self) -> Option<OrderedFloat<f64>> {
-        self.read_metric("italic angle")
+    pub fn italic_angle(&self) -> Option<f64> {
+        self.read_metric("italic angle").map(|x| x.into_inner())
```
rsheeter, line 784: Optional, this repeats enough I'd be tempted to add a helper. self.read_metric_f64("italic angle") or some such.

Or just update the read_metric fn to return f64 if that wouldn't break the world?

rsheeter (approved):

fontbe/src/mvar.rs
anthrotype: there are some global metrics that do not have an MVAR tag, notably hhea ascender/descender/lineGap (these are irrelevant and supposed to match the respective OS/2 metrics in a VF which implies USE_TYPO_METRICS); and the non-qualified (neither OS/2 nor hhea) ascender/descender, which both source formats have and is mostly used to encode the "real" height of the ascender/descender, used to define hinting zones (as opposed to the one that gets written in the font and used for line spacing purposes with all the baggage of legacy and weird recommendations like "linegap must be 0 otherwise BAD THINGS happen, etc."); we actually use the real ascender/descender to draw the default .notdef glyph, it's better to keep doing that than trust whatever gets into OS/2 or hhea.
So I decided that I'm not going to force GlobalMetrcis to map to what MVAR supports but allow the mvar_tag() to keep returning an Option<Tag>, and skip whatever metric is not associated with an MVAR tag. No need to file issues.

fontir/src/ir.rs
anthrotype: #668

Issue #643, opened 2023-12-11
Honor lookups during layout normalization for at least marks and kerning


rsheeter:
Per f2f and https://github.com/googlefonts/fontc/issues/641#issuecomment-1847847586 we are not correctly handling lookups. We should for _at least_ the layout types we generate for marks and kerning. Multiple lookups can be active and have overlapping rules. Resolution is:

* Within a lookup the first matching subtable wins
  * For example (potential test?), in a lookup with two subtables that kern AW the first value is taken
* For marks, the value from the last lookup wins.
   * That is, marks accumulate by assignment. Each new value overwrites any prior value.
* For paripos, the value accumulates by addition
   * That is, if many lookups kern AW the final value is the sum of the adjustments

cmyr:
a fun detail of this: we can't just combine the lookups as we process them, because different language systems may reference different sets of lookups. This means that it is possible for one lang/script to (for instance) include two different pairpos lookups with adjustments for the same pair (in which case they are summed) but another lang/script to include only one of those lookups.

Basically: for interactions across lookup boundaries we can't do the work ahead of time, and instead need to do it for each language system.

anthrotype:
> Within a lookup the first matching subtable wins
> For example (potential test?), in a lookup with two subtables that kern AW the first value is taken

I think it's more complicated than that at least for PairPos. The first glyph "A" could be in the coverage of multiple subtables, say one PairPosFormat1 contains pair "AW" and then another PairPosFormat1 subtable has the pair "AV"; if we only stop at the first occurence of "A" we will never evaluate the "AV" pair in the subsequent subtable.
Only when we have a match for _both_ glyphs in the pair, not just the first, we can skip any duplicate pairs in following subtables.
I noticed this issue while attempting to use the updated layout-normalizer (following the #650 merge which was supposed to handle subtables) on the ufo2ft's compile-variable-features PR branch https://github.com/googlefonts/ufo2ft/pull/635#issuecomment-1855778142

anthrotype:
As @simoncozens noted in chat, "you should stop at the first matching rule, not the first matching subtable. The coverage tables are just there to help you filter out subtables you don't need to process at all."

Quoting https://learn.microsoft.com/en-us/typography/opentype/spec/chapter2#coverage-table (emphasis mine)

> If a glyph does _not_ appear in a Coverage table, the client can _skip_ that subtable and move immediately to the next subtable.

anthrotype:
so doing this is actually incorrect:

[...]

PR #650, opened 2023-12-13
[ttx_diff] Fix incorrect subtable handling

cmyr:
This now ensures that rules from all subtables in a lookup are correctly combined, and also adds the 'first hit wins' logic, ignoring any item (in the main coverage table) in a subtable that has appeared in a previous subtable for the same lookup.


rsheeter (approved):
LGTM, one tiny nit: I suggest citing https://github.com/googlefonts/fontc/issues/643. The rationale there is important.

cmyr:
Yep good call this is progress on that, although just the subtables bit.

PR #654, opened 2023-12-13
[ttx_diff] Actually handle extension lookups

### LINKED ISSUE #641: Layout-normalizer: does not handle extension lookups
```
% wget https://notofonts.github.io/latin-greek-cyrillic/fonts/NotoSans/unhinted/ttf/NotoSans-Regular.ttf
% layout-normalizer -t GPOS NotoSans-Regular.ttf | grep Mark
# 8686 MarkToBase rules
# 8686 MarkToBase rules
# 8686 MarkToBase rules
# 8686 MarkToBase rules
# 8686 MarkToBase rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
```

```
% wget "https://notofonts.github.io/latin-greek-cyrillic/fonts/NotoSans/unhinted/variable/NotoSans[wdth,wght].ttf"
%  layout-normalizer -t GPOS NotoSans\[wdth,wght\].ttf | grep Mark
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
# 415 MarkToMark rules
```

cmyr:
Verified manually, we are now printing something much more reasonable for the (very useful!) examples in #641.

- fixes #641

### DIFF
--- a/layout-normalizer/src/gpos.rs
+++ b/layout-normalizer/src/gpos.rs
@@ -9,8 +9,8 @@ use indexmap::IndexMap;
 use write_fonts::read::{
     tables::{
         gpos::{
-            AnchorTable, MarkBasePosFormat1, MarkMarkPosFormat1, PairPos, PairPosFormat1,
-            PairPosFormat2, PositionLookup, PositionLookupList, ValueRecord,
+            AnchorTable, ExtensionSubtable, MarkBasePosFormat1, MarkMarkPosFormat1, PairPos,
+            PairPosFormat1, PairPosFormat2, PositionLookup, PositionLookupList, ValueRecord,
         },
         layout::{DeviceOrVariationIndex, LookupFlag},
     },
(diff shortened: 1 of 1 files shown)


rsheeter (approved):
LGTM. One concern, we plan to rely on this, I think perhaps it should have tests?

anthrotype (approved):

cmyr:
I'm trying to think of how to write tests in a way that is meaningful. It would be relatively easy to write something that took a known font, loaded it, and then asserted that we had the correct number of lookups of the relevant type, for instance, but that doesn't seem _super_ useful. In the general case we want to check that our transformations make sense (e.g. that we are skipping rules in subtables correctly, or not) and this is tricky. The most reasonable & useful approach I can think of is that we could use write-fonts and compile specific small examples, and then verify our behaviour against these.

I'm going to open a separate issue to figure out how best to start testing here, since it feels like this tool is sort of developing a life of its own. More generally the way a lot of this code has been written so far was with the intention of just doing a single simple representation, and I think we've sort of generally agreed that it would be useful if it could work in different modes, which means that I need to use a more flexible representation (currently I'm basically transforming everything into a final representation in the first pass)

Issue #647, opened 2023-12-13
repeated builds are not deterministic


anthrotype:
ok the head.modified changes every time (and checksums would also change accoridngly), but I would not expect other things to differ when I compare ttx dumps of font files that were freshly built from the same source file using the same copy of fontc.

It turns out there are some differences in things like GDEF and GPOS (as noted in https://github.com/googlefonts/fontc/pull/644#discussion_r1425331084, this one possible regression introduced with #642), as well as in GSUB (in the FeatureParamsStylisticSet nameIDs).

Also when doing `ttx -l` to show the table directory, the gvar table seems to have different checksums each time round, even though the length is the same.. But diffing its ttx repesentations gives no differences, so it might be the bytes themselves are serialized differently each time, for some reasons.

anthrotype:
forgot to mention - I was able to notice these diffs on repeated builds of GoogleSans.designspace, I didn't try other fonts.

cmyr:
In the past I have seen things like this arise when using collections that provide a non-deterministic iteration order (e.g. hashmaps) so that would be one thing I would look into first. I have done some work in the past in this regard (for instance #561) and it's very possible there are other places where we've missed this.

anthrotype:
I think here it's also to do with the fact that we split some tasks in parallel threads that may complete at different times and then we gather their result in some other collection where the order does matter

cmyr:
yes that is totally possible as well.

rsheeter:
Good catch. Can we reproduce in a test with our test files? - that should make it easy to fix and keep fixed.

anthrotype:
I don't think our test files contain enough kerning pairs to trigger this..

rsheeter:
Let's fix this after we land https://github.com/googlefonts/fontc/pull/644. We can add a test file with lots of kerns or a knob to help it trigger with fewer kerns if necessary.

rsheeter:
In IM @anthrotype noted "the GPOS vary in size between 106684 bytes and 4384362, an order of magnitude different.."; that's mildly alarming

rsheeter:
Reproduction:

```shell
$ for i in {0..9}; do rm -rf build/ && cargo run ../googlesans/source/GoogleSans/GoogleSans.designspace && ttx -l build/font.ttf > run$i.ttxl; done
$ grep -H GPOS run?.ttxl
run0.ttxl:    GPOS  0xE96ECCD8   4384230     18876
[...]

PR #656, opened 2023-12-14
CI to expose build instability

rsheeter:
Expose #647. This check will fail until we fix that. I also moved build GS to be part of CI because in my mind it is.


.github/workflows/rust.yml
```diff
@@ -160,4 +160,69 @@ jobs:
...
+        run: fontc source/GoogleSans/GoogleSans.designspace
+      - name: OTS tests, Roman
+        run: ots-9.1.0-Linux/ots-sanitize build/font.ttf
+      - name: Compile Italic variable font
+        run: fontc source/GoogleSans/GoogleSans-Italic.designspace
+      - name: OTS tests, Italic
+        run: ots-9.1.0-Linux/ots-sanitize build/font.ttf
+
+  # Check if GS build is repeatable, https://github.com/googlefonts/fontc/issues/647
+  build-gs-repeatably:
```
anthrotype, line 193: do we actually need a separate task for this? can't you add the extra builds at the end of `build-gs`?

.github/workflows/rust.yml
```diff
@@ -160,4 +160,69 @@ jobs:
...
+          cp build/font.ttf ./first.ttf
+
+      - name: Compile me twice, shame on me
+        run: |
+          rm -rf build
+          fontc source/GoogleSans/GoogleSans.designspace
+          cp build/font.ttf ./second.ttf
+
+      - name: Same output?
+        run: cmp first.ttf second.ttf
```
anthrotype, line 228: this will always fails (even after we fix other bugs) at least until the head.modified timestamp changes each time one builds the same font.
I think you should be able to set `SOURCE_DATE_EPOCH` environment variable to make sure that doesn't change

.github/workflows/rust.yml
anthrotype: rebuilding GS should take a few seconds, bootstrapping again a whole container + rust toolchain + rebuilding fontc + checking out GS repo etc. will take much more

.github/workflows/rust.yml
```diff
@@ -160,4 +160,69 @@ jobs:
...
+          pwd
+          ls
+          fontc source/GoogleSans/GoogleSans.designspace
+          cp build/font.ttf ./first.ttf
+
+      - name: Compile me twice, shame on me
+        run: |
+          rm -rf build
+          fontc source/GoogleSans/GoogleSans.designspace
+          cp build/font.ttf ./second.ttf
```
anthrotype, line 225: I suggest you also list the table directory with `ttx -l` for both fonts, or even diff the two directories, 
helpful for inspecting diffs in table sizes/checksums

.github/workflows/rust.yml
rsheeter: We don't _need_ to separate, I just like getting separate checkmarks for builds GS at all vs builds it repeatably.

.github/workflows/rust.yml
rsheeter: Oh good point. I'd like a flag to do that rather than, or as well as, an environment variable. How awful an idea is it for repeatable to be the default?

anthrotype: > How awful an idea is it for repeatable to be the default?

well, so far we tried to stick with matching fontmake's default behavior, it would be a bit surprising that one has to opt-in to have the head.modified be updated to current time. I'd prefer if you instead add an option for repeatable. What's wrong with the SOURCE_DATE_EPOCH? it's standard and we do support it

.github/workflows/rust.yml
rsheeter: The timing is interesting, looking at the new job and the old one:

* Build for debug takes :50
* Build for release takes 1:50
* Building GS with a debug compiler takes :50
* Building GS with a release compiler takes :06
* Checkout of source takes :14

All in all they both land around 2:40. I think will combine and use the release compiler, it should be basically always passing as soon as those bugs are fixed.

.github/workflows/rust.yml
rsheeter: I keep forgetting that's a quasi-standard not a hack. Feels like a hack to me still lol.

.github/workflows/rust.yml
rsheeter: So updated.

.github/workflows/rust.yml
rsheeter: Good idea. I guess that means I have to install python. Hopefully that's fast.

.github/workflows/rust.yml
anthrotype: yeah but it's meant for this very purpose and we implemented it already so we may well use it :)

.github/workflows/rust.yml
anthrotype: I think the ubuntu container may already come with one.. try

anthrotype: yeah looks like it already has python https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md

anthrotype (approved):

.github/workflows/rust.yml
```diff
@@ -160,4 +160,62 @@ jobs:
...
+      - name: Compile me twice, shame on me
+        run: |
+          rm -rf build
+          fontc source/GoogleSans/GoogleSans.designspace
+          cp build/font.ttf ./second-roman.ttf
+          rm -rf build
+          fontc source/GoogleSans/GoogleSans-Italic.designspace
+          cp build/font.ttf ./second-italic.ttf
+
+
```
anthrotype, line 216: maybe remove the extra newline here

rsheeter: Done

.github/workflows/rust.yml
rsheeter: > I think the ubuntu container may already come with one.. try

shiny. Will rig.

rsheeter:
Rebased on main

anthrotype:
I believe it's still going to fail because of varying GSUB FeatureParams nameIDs

anthrotype:
how about instead of just calling ttx -l you actually diff the output for two ttx -l calls? So it only shows what changed which is what we are interested in

```sh
$ diff -u <(pipx run fonttools ttx first-roman.ttf) <(pipx run fonttools ttx second-roman.ttf)
$ diff -u <(pipx run fonttools ttx first-italic.ttf) <(pipx run fonttools ttx second-italic.ttf)
```

rsheeter:
> I believe it's still going to fail because of varying GSUB FeatureParams nameIDs

That's right. I don't think we can merge it quite yet.

anthrotype:
just rebased on main hoping that things just work

anthrotype:
looks like gvar and head are still reporting same lengths but different checksums

```
--- /dev/fd/63	2023-12-18 17:41:27.394633835 +0000
+++ /dev/fd/62	2023-12-18 17:41:27.394633835 +0000
@@ -10,8 +10,8 @@
     cmap  0xA7FB7457     10444   4523220
     fvar  0x47F5205B       172   4533664
     glyf  0xDB038597   1080400   4533836
-    gvar  0xB3F94B2D   2789905   56[14](https://github.com/googlefonts/fontc/actions/runs/7251757185/job/19754763563?pr=656#step:12:15)236
-    head  0x1F4E2CFF        54   8404144
+    gvar  0xD666863A   2789905   5614236
+    head  0x1F4E2D0D        54   8404144
     hhea  0x0CCB29[20](https://github.com/googlefonts/fontc/actions/runs/7251757185/job/19754763563?pr=656#step:12:21)        36   8404200
     hmtx  0x0C2B634B     30364   8404[23](https://github.com/googlefonts/fontc/actions/runs/7251757185/job/19754763563?pr=656#step:12:24)6
     loca  0xA6461EE2     30372   8434600
```

rsheeter:
> looks like gvar and head are still reporting same lengths but different checksums

Guess we can't declare victory yet. Even so, I'm quite pleased GSUB is now fixed :)

anthrotype:
After https://github.com/googlefonts/fontations/pull/750 we can temporarily declare victory ✌️ 
We need to bump write-fonts first to see the fix but I can confirm that it now works locally (after exporting SOURCE_DATE_EPOCH of course).

anthrotype:
with latest write-fonts 0.21.1, CI is now green!

anthrotype:
shall I revert this until we figure out https://github.com/googlefonts/fontc/issues/647#issuecomment-1888865488? otherwise all new PRs are going to fail because of it..

rsheeter:
I'd vote not, if CI is failing we shouldn't be merging much of anything except a fix and we still have the power to merge despite this failure if we really really want to

PR #658, opened 2023-12-14
[ci] Fail CI if docs contain broken links

cmyr:
This also fixes some existing broken links.

locally I'm getting warnings printed everytime I run the pre-push checks, so we should be catching this in CI.

JMM

### DIFF
--- a/fontbe/src/kern.rs
+++ b/fontbe/src/kern.rs
@@ -1,4 +1,4 @@
-//! Generates a [Kerning] datastructure to be fed to fea-rs
+//! Generates an [FeaRsKerns] datastructure to be fed to fea-rs
 
 use std::collections::{BTreeMap, HashMap};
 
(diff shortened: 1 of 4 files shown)


fontbe/src/marks.rs
```diff
@@ -1,4 +1,4 @@
-//! Generates a [Marks] datastructure to be fed to fea-rs
+//! Generating mark GPOS features
```
rsheeter, line 1: I prefer the prior comment, can we just update it to point to FeaRsMarks?

fontbe/src/kern.rs
```diff
@@ -1,4 +1,4 @@
-//! Generates a [Kerning] datastructure to be fed to fea-rs
+//! Generating kerning rules
```
rsheeter, line 1: I prefer the prior comment, could we update to FeaRsKerns?

rsheeter (approved):

PR #660, opened 2023-12-14
[WIP] What if all the collections were stable

rsheeter:
In hopes of helping with #647


rsheeter:
This did not help; further investigation points at fontatins packing, see https://github.com/googlefonts/fontc/issues/647#issuecomment-1858251211.

PR #665, opened 2023-12-16
[chore] Update write-fonts to 0.21.0

cmyr:
This involves a breaking change to the VariationStoreBuilder, which in turn requires us to tweak the FeatureProvider trait in fea-rs, but overall fairly painless.

### DIFF
--- a/layout-normalizer/src/gpos/pairpos.rs
+++ b/layout-normalizer/src/gpos/pairpos.rs
@@ -233,7 +233,7 @@ mod tests {
         }
 
         fn build_exactly_one_subtable(self) -> PairPos {
-            let mut varstore = VariationStoreBuilder::new();
+            let mut varstore = VariationStoreBuilder::new(0);
             let subs = self.build(&mut varstore);
             assert_eq!(subs.len(), 1);
             subs.into_iter().next().unwrap()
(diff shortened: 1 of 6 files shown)


layout-normalizer/src/gpos/pairpos.rs
```diff
@@ -233,7 +233,7 @@ mod tests {
         }
 
         fn build_exactly_one_subtable(self) -> PairPos {
-            let mut varstore = VariationStoreBuilder::new();
+            let mut varstore = VariationStoreBuilder::new(0);
```
anthrotype, line 236: how about we add a `VariationStoreBuilder::nop()` alternative constructor that would set axis_count = 0? It doesn't make much sense for the _actual_ `new()` default constructor to take an axis_count that is equal to 0, because there's nothing to "build" in that case, the varstore won't contain any regions or deltas if there are no axes...

anthrotype (approved):

layout-normalizer/src/gpos/pairpos.rs
cmyr: given that this should only ever be needed in tests I'm not too fussed either way? That we need this at all is mostly just indicates there's something smelly in our API (although there's a rationale for that)

PR #670, opened 2023-12-22
[ttx_diff]: Normalize kern rules

cmyr:
This adds the proper normalization logic for pairpos rules: basically we will now sum the values in subsequent lookups that apply to the same pairs.

this PR also involves moving some test helpers to a common `test_helpers` module, since they're needed by different tests that live in different places.

There's one question inline: basically I'm not sure how we should think about merging of device tables, or if we should care. Also I'm assuming that if deltas are provided there should be the same number of deltas in each master, and we can just sum them individually?

### DIFF
--- a/layout-normalizer/src/gpos.rs
+++ b/layout-normalizer/src/gpos.rs
@@ -277,13 +277,13 @@ impl LookupRules {
             .filter(|lookup| lookups.contains(&lookup.lookup_id))
             .flat_map(|lookup| lookup.iter())
         {
-            match pairmap.entry((rule.rule.first, rule.rule.second.clone())) {
+            match pairmap.entry((rule.rule().first, rule.rule().second.clone())) {
                 hash_map::Entry::Vacant(entry) => {
                     entry.insert(rule);
                 }
                 hash_map::Entry::Occupied(mut entry) => {
-                    let prev_rule = entry.get_mut().rule.to_mut();
-                    prev_rule.merge(&rule.rule);
+                    let prev_rule = entry.get_mut().rule_mut();
+                    prev_rule.merge(rule.rule());
                 }
             };
         }
@@ -293,9 +293,9 @@ impl LookupRules {
         let mut seen = HashMap::<_, _>::new();
         for rule in pairmap.into_values() {
             match seen.entry((
-                rule.rule.first,
-                rule.rule.record1.clone(),
-                rule.rule.record2.clone(),
+                rule.rule().first,
+                rule.rule().record1.clone(),
+                rule.rule().record2.clone(),
             )) {
                 hash_map::Entry::Vacant(entry) => {
                     entry.insert(rule);
@@ -303,10 +303,9 @@ impl LookupRules {
                 hash_map::Entry::Occupied(mut entry) => {
                     entry
                         .get_mut()
-                        .rule
-                        .to_mut()
+                        .rule_mut()
                         .second
-                        .combine(&rule.rule.second);
+                        .combine(&rule.rule().second);
                 }
             }
         }
@@ -555,7 +554,7 @@ mod tests {
         let our_rules = rules
             .pairpos_rules(&[0, 1])
             .into_iter()
-            .map(|r| r.rule.as_ref().to_owned())
+            .map(|r| r.rule().to_owned())
             .collect::<Vec<_>>();
 
         // (left gid, [right gids], x advance)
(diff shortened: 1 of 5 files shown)


layout-normalizer/src/gpos.rs
```diff
@@ -228,6 +239,22 @@ impl ResolvedValue {
     fn is_zero(&self) -> bool {
         self.default == 0 && self.device_or_deltas.is_none()
     }
+
+    fn add_in_place(&mut self, other: &ResolvedValue) {
+        self.default += other.default;
+        //CODE REVIEW:
+        // - if we have device tables we're just dropping the second, which
+        // is fine for our purposes? And if we have mixed deltas & devices than
+        // god help us
```
anthrotype, line 248: this is fine, I've actually never seen device tables in the wild. how about we at least warn?

layout-normalizer/src/gpos.rs
```diff
@@ -228,6 +239,22 @@ impl ResolvedValue {
...
+        if let (Some(DeviceOrDeltas::Deltas(d1)), Some(DeviceOrDeltas::Deltas(d2))) = (
+            self.device_or_deltas.as_mut(),
+            other.device_or_deltas.as_ref(),
+        ) {
+            // I think this is expected/required?
+            assert_eq!(d1.len(), d2.len());
```
anthrotype, line 254: well, in theory the number of deltas might be different in different lookups.. We can't merge them if they aren't.

cmyr: hmm. In my code we are always using the same item variation store and the same set of master locations when resolving the deltas. Are you thinking about sparseness? In that case I think we're fine, and should be inserting zeros, although maybe I'm misunderstanding?

layout-normalizer/src/gpos.rs
cmyr: we can warn, but when we run with `ttx_diff` (which is the most common case) we suppress stderr so we might never see it. 🤔

layout-normalizer/src/gpos.rs
anthrotype: yes i'm thinking about sparse masters, even if right now fontc doesn't emit them for the generated GPOS features (https://github.com/googlefonts/fontc/issues/407) it doesn't mean pos rules have to always be "dense" and contain all the same number of masters.
It's not as simple as inserting zeros, I'm afraid. It is conceivable that an item variation store has multiple vardata tables each specifying different number of regions and then various pos rules referencing one or the other.

cmyr: hmm okay I'm worried.


This is how, in `layout-normalizer`, we compute the deltas (they aren't actually deltas; they're absolute value at each master location?) https://github.com/googlefonts/fontc/blob/dab7473fe4ae791deaf5fb934b1fb78493829de5/layout-normalizer/src/variations.rs#L50

For a given set of master locations, this will always return the same number of values. Is my logic wrong somewhere?

anthrotype: Yes, one delta for each master location, but each given coordinate could potentially be defined along a different number of master locations

cmyr: I'm sorry but I'm still confused. We only use a single set of master locations, which we retrieve from the IVS itself; then for each location we call `ItemVariationStore::compute_delta`, which adds up the deltas for all the regions that are active for that coordinate; so there are potentially a different number of deltas but there are always the same number of total master locations?

anthrotype: > We only use a single set of master locations, which we retrieve from the IVS itself

then we are wrong, because the IVS has a global VarRegionList and then each VariationData table can reference different sets of regions, and the various variable things that can have a VarIndex are free to reference one or the other VarData, effectively selecting only some from the IVS global list of regions

cmyr: maybe the problem is just my naming? As in this is called 'deltas' but it's really the resolved value for each defined master/location.

anthrotype: no, it's not just naming, it's wrong to assume a single set of master locations (aka. region peaks) for the entire IVS like our DeltaComputer is doing. There are potentially many sets of master locations, one per VariationData subtable

anthrotype (approved):

layout-normalizer/src/gpos.rs
cmyr: This is an internal tool, and we don't use device tables, so I'm just going to let this slide.

PR #673, opened 2024-01-12
bump write-fonts to 0.21.1

anthrotype:
required for https://github.com/googlefonts/fontc/issues/647


anthrotype:
the clippy warnings are new and completely unrelated. I'm going to merge this as is and tackle those in a separate PR

PR #674, opened 2024-01-12
make clippy happy again

anthrotype:
JMM


anthrotype:
uh-oh, looks like https://github.com/googlefonts/fontc/issues/647#issuecomment-1888865488 is back

anthrotype:
I'm pretty sure this specific, relatively trivial changes have nothing to do with #647 so I'm going to merge this PR so at least the clippy fail goes away and we can focus on the other issue

PR #676, opened 2024-01-12
Add repeatable-builds.sh and test Oswald as well

anthrotype:
This PR adds a new shell script to help debugging repeatable builds locally, and adds Oswald to the CI testing repeatable builds (the gvar randomness about order of shared tuples would only trigger when building Oswald, not GS somehow).
Finally, I added a step to upload the built ttfs as CI artifacts if any failure occurs, also useful for remote debugging.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -202,9 +202,6 @@ jobs:
           repository: googlefonts/OswaldFont
           path: oswald
 
-      - name: Pin SOURCE_DATE_EPOCH
-        run: echo "SOURCE_DATE_EPOCH=$(date +%s)" >> "$GITHUB_ENV"
-
       - name: Compile GS Roman twice
         run: ./resources/scripts/repeatable-builds.sh googlesans/source/GoogleSans/GoogleSans.designspace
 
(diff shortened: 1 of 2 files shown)


anthrotype:
finally, repeatable-builds.sh for GoogleSans just failed on CI (it's never failed locally for me):

https://github.com/googlefonts/fontc/actions/runs/7504914976/job/20433196909#step:9:12

anthrotype:
re-run the jobs and now GS passes but Oswald failed:

https://github.com/googlefonts/fontc/actions/runs/7504914976/job/20433507498#step:11:12

but _not_ with the gvar diff that I get sometimes locally on my machine (that I mentioned in https://github.com/googlefonts/fontc/issues/647#issuecomment-1889582400)!
It seems to fail with GDEF/GPOS diffs similar to those seen with GS (but only when run on CI).

anthrotype:
after a few trials, I managed to also reproduce the Oswald gvar diff:

https://github.com/googlefonts/fontc/actions/runs/7505218418/job/20434131713#step:11:25

rsheeter (approved):

PR #680, opened 2024-01-23
[fea-rs] Tweak ligature sort order

cmyr:
Discussed in chat; this adopts a simple common sorting method for ligatures.

The sorting of ligature substitutions for a given glyph is surprisingly contentious. The spec says that the implementation should sort the input such that longer ligatures are applied first, but this means there is no way for the user to indicate that a given ligature should be prioritized over another.

This tweaks our implementation to try and address both concerns: we will now ensure that longer ligatures are first, but for ligatures of equal length we will respect the order provided by the user.

This tweaks a few of the 'expected_diff' files for the fonttools tests, as our behaviour is now different from fonttools in a *new way*; but the expectation is that fonttools will adopt this new sorting behaviour as well, and then everyone will be happy forever.

see https://github.com/fonttools/fonttools/issues/3428 & https://github.com/adobe-type-tools/afdko/issues/1727#issuecomment-1906703943


rsheeter (approved):

anthrotype:
I can confirm that after applying this and https://github.com/fonttools/fonttools/pull/3429, I no longer see differently ordered ligatures when ttx_diff'ing Oswald built with fontc vs fontmake

PR #685, opened 2024-01-30
Add fontra2fontir without an actual implementation

rsheeter:
Fontra source can describe varc so having an IR source for it seems like it will be handy.
Per discussion with Just we will want to ingress their json format, not .rcjk.

Fontra json examples can be found in https://github.com/notofonts/noto-cjk-varco action runs, e.g. https://github.com/notofonts/noto-cjk-varco/actions/runs/7714671850. In subsequent PRs I will add sample sources to resources.

### DIFF
--- a/fontra2fontir/README.md
+++ b/fontra2fontir/README.md
@@ -1,0 +1,6 @@
+# fontra2fontir
+
+This crate is a frontend for compilation of Fontra json sources. It converts from .fontra to the fontc
+internal representation.
+
+It should be referenced only by fontc.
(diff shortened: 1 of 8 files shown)


fontra2fontir/README.md
```diff
@@ -0,0 +1,6 @@
+# fontra2fontir
+
+This crate is a frontend for compilation of Fontra json sources. It converts from .fontra to the fontc
```
cmyr, line 3: can we link to fontra?

cmyr (approved):
one note inline :)

fontra2fontir/README.md
rsheeter: Will do

PR #686, opened 2024-01-30
Find the glyphs of a .fontra 

rsheeter:
Read glyph-info.csv and find the corresponding glyph files. The next set of reads (future PR) should be possible w/serde_json I believe.


fontra2fontir/src/source.rs
```diff
@@ -1,22 +1,146 @@
...
+            fontdata_file,
+            glyphinfo_file,
+            glyph_dir,
+            glyph_info: Default::default(),
+        })
+    }
+
+    fn glyph_file(&self, glyph: GlyphName) -> PathBuf {
+        // TODO: this probably over-encodes. Should be sufficient to get started.
+        let filename = utf8_percent_encode(glyph.as_str(), NON_ALPHANUMERIC);
```
anthrotype, line 47: I think fontra's glyph->filename algorithm is implemented here:
https://github.com/googlefonts/fontra/blob/main/src/fontra/backends/filenames.py

fontra2fontir/src/source.rs
```diff
@@ -1,22 +1,146 @@
...
+            let line = line.map_err(Error::IoError)?;
+            let parts: Vec<_> = line.split(';').collect();
+            if parts.len() != 2 {
+                return Err(Error::ParseError(
+                    self.glyphinfo_file.clone(),
+                    format!("Expected two parts in line {i} separated by ;"),
+                ));
+            }
+            let glyph_name = GlyphName::new(parts[0].trim());
+            // TODO: support multiple codepoints
```
anthrotype, line 75: the mutliple codepoints are simply comma separated, see https://github.com/googlefonts/fontra/blob/a4edd06837118e583804fd963c22ed806a315b04/src/fontra/backends/fontra.py#L171-L181

anthrotype (approved):
that's a good start!

fontra2fontir/src/source.rs
```diff
@@ -1,22 +1,146 @@
...
 }
 
 impl Source for FontraIrSource {
     fn inputs(&mut self) -> Result<fontir::source::Input, fontir::error::Error> {
-        todo!()
+        let static_metadata = self.static_metadata_state()?;
+        self.load_glyphinfo()?;
+        let glyphs = self.glyph_state()?;
+
+        // Not sure how fontra features work
```
anthrotype, line 133: I think fontra doesn't support OpenType Features yet

rsheeter: I suppose it's coming? @justvanrossum do you know?

justvanrossum: We have no concrete timeline for that.

PR #690, opened 2024-01-31
Serde type for the fontra font-data.json file

rsheeter:
Trying to sneak things between meetings :)

JMM

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -1047,7 +1047,11 @@ mod tests {
         assert_eq!(1, glyph.contours().len());
         assert_eq!(
             4_usize,
-            glyph.contours().iter().map(|c| c.iter().count()).sum::<usize>()
+            glyph
+                .contours()
+                .iter()
+                .map(|c| c.iter().count())
+                .sum::<usize>()
         );
         assert_eq!(
             Bbox {
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -1047,7 +1047,7 @@ mod tests {
         assert_eq!(1, glyph.contours().len());
         assert_eq!(
             4_usize,
-            glyph.contours().iter().map(|c| c.iter().count()).sum()
+            glyph.contours().iter().map(|c| c.iter().count()).sum::<usize>()
         );
         assert_eq!(
             Bbox {
--- a/fontra2fontir/src/source.rs
+++ b/fontra2fontir/src/source.rs
@@ -1,8 +1,8 @@
 use std::{
     collections::HashMap,
-    fs::File,
+    fs::{self, File},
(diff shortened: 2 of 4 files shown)


fontra2fontir/src/source.rs
```diff
@@ -188,6 +191,36 @@ impl Source for FontraIrSource {
...
+impl FontData {
+    fn from_file(p: &Path) -> Result<Self, Error> {
+        let raw = fs::read_to_string(p).map_err(Error::IoError)?;
+        serde_json::from_str(&raw).map_err(|e| Error::ParseError(p.to_path_buf(), format!("{e}")))
+    }
+}
+
+#[derive(Debug, Clone, Deserialize)]
+struct FontraAxis {
+    _name: String,
```
cmyr, line 212: why all the underbars? Is this just to avoid dead code warnings? In general I'd prefer to just mark items as `#[allow(dead_code)]` while things are in progress, but 🤷

fontc/src/lib.rs
```diff
@@ -1047,7 +1047,11 @@ mod tests {
         assert_eq!(1, glyph.contours().len());
         assert_eq!(
             4_usize,
-            glyph.contours().iter().map(|c| c.iter().count()).sum()
+            glyph
+                .contours()
+                .iter()
+                .map(|c| c.iter().count())
+                .sum::<usize>()
```
cmyr, line 1054: weird that this had to change?

cmyr (approved):
lgtm, two notes inline

fontc/src/lib.rs
rsheeter: Yeah, I'm fuzzy why but clippy was quite insistent

fontra2fontir/src/source.rs
rsheeter: For dead code. It becomes live in the follow-on so for this PR I'm inclined to leave the _'s

fontra2fontir/src/source.rs
```diff
@@ -224,4 +273,38 @@ mod tests {
...
         );
     }
+
+    #[test]
+    fn fontdata_of_minimal() {
+        let font_data =
+            FontData::from_file(&testdata_dir().join("minimal.fontra/font-data.json")).unwrap();
+        assert_eq!(1000, font_data._units_per_em);
+        assert_eq!(
+            vec![("Weight", Tag::from_be_bytes(*b"wght"), 200.0, 200.0, 900.0),],
```
anthrotype, line 283: doesn't `Tag::new(b"wght")` do the same with less characters?

cmyr: oh yes, good call, `new` is the blessed constructor.

PR #691, opened 2024-01-31
Make a rudimentary static metadata for .fontra

rsheeter:
Builds on #690


fontra2fontir/src/source.rs
```diff
@@ -59,11 +67,16 @@ impl FontraIrSource {
     }
 
     fn load_glyphinfo(&mut self) -> Result<(), Error> {
+        if !self.glyph_info.is_empty() {
```
cmyr, line 70: is this expecting to be called multiple times, and we're just trying to avoid doing duplicate work?

fontra2fontir/src/source.rs
```diff
@@ -195,9 +212,9 @@ impl Source for FontraIrSource {
 #[derive(Debug, Clone, Deserialize)]
 struct FontData {
     #[serde(rename = "unitsPerEm")]
-    _units_per_em: u16,
+    units_per_em: u16,
```
cmyr, line 215: aha.

fontra2fontir/src/source.rs
```diff
@@ -209,16 +226,114 @@ impl FontData {
...
+                    })
+                    .collect();
+                let default_idx = examples
+                    .iter()
+                    .position(|(u, _)| *u == default)
+                    .ok_or_else(|| WorkError::AxisMustMapDefault(a.tag))?;
+                examples
+                    .iter()
+                    .position(|(u, _)| *u == min)
+                    .ok_or_else(|| WorkError::AxisMustMapMin(a.tag))?;
```
cmyr, line 284: this might be easier with `Vec::contains`?

cmyr (approved):
👍

fontra2fontir/src/source.rs
rsheeter: Victory :)

fontra2fontir/src/source.rs
rsheeter: Examples has (User, Design) and I want to search by only User.

fontra2fontir/src/source.rs
rsheeter: It's not really supposed to get called over and over but doing so felt achievable enough the check felt worth it. You could argue it shouldn't be there.

PR #692, opened 2024-02-01
Parse Fontra glyphs as far as contours

rsheeter:
https://codepen.io/rs42/pen/qBvoaLy shows the first outlines ever loaded from fontra by this code.

We have yet to reach the interesting (varc) bits. Baby steps.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -1028,7 +1028,7 @@ pub struct Glyph {
     pub emit_to_binary: bool,
     pub codepoints: HashSet<u32>, // single unicodes that each point to this glyph. Typically 0 or 1.
     default_location: NormalizedLocation,
-    instances: HashMap<NormalizedLocation, GlyphInstance>,
+    sources: HashMap<NormalizedLocation, GlyphInstance>,
     has_consistent_2x2_transforms: bool,
 }
 
@@ -1094,27 +1094,27 @@ impl Glyph {
             emit_to_binary,
             codepoints,
             default_location,
-            instances,
+            sources: instances,
             has_consistent_2x2_transforms,
         })
     }
 
     pub fn default_instance(&self) -> &GlyphInstance {
-        self.instances.get(&self.default_location).unwrap()
+        self.sources.get(&self.default_location).unwrap()
     }
 
     pub fn sources(&self) -> &HashMap<NormalizedLocation, GlyphInstance> {
-        &self.instances
+        &self.sources
     }
 
     pub fn sources_mut(
         &mut self,
     ) -> impl Iterator<Item = (&NormalizedLocation, &mut GlyphInstance)> {
-        self.instances.iter_mut()
+        self.sources.iter_mut()
     }
 
     pub fn source_mut(&mut self, loc: &NormalizedLocation) -> Option<&mut GlyphInstance> {
-        self.instances.get_mut(loc)
+        self.sources.get_mut(loc)
     }
 
     /// Does the Glyph use the same components, (name, 2x2 transform), for all instances?
@@ -1131,7 +1131,7 @@ impl Glyph {
     ///
     /// See <https://github.com/googlefonts/fontc/issues/291#issuecomment-1557358538>
     pub(crate) fn has_nonidentity_2x2(&self) -> bool {
-        self.instances
+        self.sources
             .values()
             .flat_map(|inst| inst.components.iter())
             .any(|c| c.transform.as_coeffs()[..4] != [1.0, 0.0, 0.0, 1.0])
@@ -1343,7 +1343,7 @@ impl From<Glyph> for GlyphBuilder {
             name: value.name,
             emit_to_binary: value.emit_to_binary,
             codepoints: value.codepoints,
-            sources: value.instances,
+            sources: value.sources,
         }
     }
 }
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -1028,7 +1028,7 @@ pub struct Glyph {
     pub emit_to_binary: bool,
     pub codepoints: HashSet<u32>, // single unicodes that each point to this glyph. Typically 0 or 1.
     default_location: NormalizedLocation,
-    sources: HashMap<NormalizedLocation, GlyphInstance>,
+    instances: HashMap<NormalizedLocation, GlyphInstance>,
     has_consistent_2x2_transforms: bool,
 }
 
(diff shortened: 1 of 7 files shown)


fontir/src/ir.rs
```diff
@@ -1028,7 +1028,7 @@ pub struct Glyph {
     pub emit_to_binary: bool,
     pub codepoints: HashSet<u32>, // single unicodes that each point to this glyph. Typically 0 or 1.
     default_location: NormalizedLocation,
-    sources: HashMap<NormalizedLocation, GlyphInstance>,
+    instances: HashMap<NormalizedLocation, GlyphInstance>,
```
anthrotype, line 1031: why the attribute name change? These are called sources in DesignSpace jargon, instances usually mean the obtain from interpolating.. the sources

fontra2fontir/src/toir.rs
```diff
@@ -0,0 +1,265 @@
...
+        }
+        path_builder.move_to((first.x, first.y))?;
+        add_to_path(
+            glyph_name.clone(),
+            &mut path_builder,
+            contour.points[1..].iter(),
+        )?;
+    } else {
+        // I think Fontra does the start at the end bit
+        // Rotate right by 1 by way of chaining iterators
```
anthrotype, line 194: where did you get that impression? I think it just follows ufo, where the open path is signalled by the first point having a "move" type, whereas closed path can start with anything (doesn't really start or ends, it's a cyclical list of points) 

See how a fontra PackedPath is drawn onto a (UFO) PointPen:

https://github.com/googlefonts/fontra/blob/a4edd06837118e583804fd963c22ed806a315b04/src/fontra/core/path.py#L137-L179

and this is the pen that constructs a fontra PackedPath from a set of PointPen commands:

https://github.com/googlefonts/fontra/blob/a4edd06837118e583804fd963c22ed806a315b04/src/fontra/core/path.py#L301-L382

I don't see any rotation occurring there, but maybe I'm wrong and you're right. /cc @justvanrossum

justvanrossum: I have no idea what this code is doing, but: Fontra contours have an explicit `isClosed` attribute.

fontir/src/ir.rs
rsheeter: From the IRs point of view they are instances. The source terminology is only applicable to UFO+ds and thus struck me as not belonging.

anthrotype: to me at least it's confusing to call the sources "instances", and I'm sure a lot of font developers would feel the same.

fontra2fontir/src/toir.rs
justvanrossum: Fontra does not rotate. At least as long as we stick to the a-contour-is-a-list-of-points view.

anthrotype: thanks for confirming. But at least for open paths it's invalid to have leading/trailing off curves, right?

rsheeter: Fixing, ty for confirming

fontir/src/ir.rs
rsheeter: FWIW I thought I'd find the instances in something with instance in the name, "sources" actually delayed me. It's relatively normal to name collections after what's in them in my experience. A bunch of GlyphInstances is instances, "sources" requires domain knowledge to grok.

I also thought, perhaps mistakenly, that was leaking a specific ... source (UFO+ds)... into our abstraction over many source formats. 

If none of that is convincing I can change it back :) Actually probably what I should do is back it out here and submit as a different PR (if you are convinced) since it's relatively unrelated to the rest of this.

anthrotype: I see what you mean about the names of collections but I am not convinced. If anything I'd propose to rename GlyphInstance struct. In all font formats and font editors, a master.. ehm sorry a "source" is the input that the designer draws, a whole value that is associated with some location in the variation space, and is used for... interpolating instances, which are the output or the result of the interpolation. This is unversally the case.

rsheeter: > instances usually mean the thing you obtain from interpolating.. the sources

Thinking "aloud" I really like the name GlyphInstance, I think it captures what it is very accurately. If I pursue this line of reasoning I think i would conclude it's the wrong name and I don't like that conclusion. That would lead me to argue IR simply doesn't use "instance" in this way.

anthrotype: just try and ask in the font engs chat and see what they mean by "instance"

rsheeter: I have undone the naming change. I would be quite strongly opposed to renaming GlyphInstance. I don't think the IR aims to model what an editor interacts with. It's meant to support a SWE like a Google Fonts member and to that group "something instance" generally has meaning, the correct meaing, whereas "source" is a rather broad term that quite likely means the wrong thing.

I read Fontra's VariableGlyph as aligning with our "Glyph" and StaticGlyph as aligning with GlyphInstance. .glyphs just has "glyphs". Needing to lookup a layer to find a location appears to me to be a product of how a user-facing editor works: you create a layer with a location then draw on it. I don't think that concept really applies to IR, we wouldn't gain anything that I can currently identify by modelling layers.

I propose we decide to rename nothing. GlyphInstance is a good name and my world won't end if I have to find the glyph instances in "sources"

anthrotype: > I read Fontra's VariableGlyph as aligning with our "Glyph" and StaticGlyph as aligning with GlyphInstance

right, as I said Fontra's VariableGlyph has "sources", not "instances".

> meant to support a SWE like a Google Fonts member and to that group "something instance" generally has meaning, the correct meaing

whatever :)

> I propose we decide to rename nothing.

agreed!

anthrotype: > "source" is a rather broad term that quite likely means the wrong thing

it highlights the _function_. It's "example or specific occurrence" (instance) of a glyph that is designed (not generated) and is positioned along variation axes in order to obtain other such instances through interpolation. If I go to fonttools or glyphsLib codebase and find/replace all occurrences of the word "master" (a banned word that we replaced with "source") with "instance", it surely will raise a few eyebrows.

fontra2fontir/src/toir.rs
justvanrossum: > But at least for open paths it's invalid to have leading/trailing off curves, right?

Sorry, no, leading/trailing off-curve points are perfectly legal in Fontra... For rendering you can just strip them, though. This is only important for a work-in-progress contour.

fontir/src/ir.rs
justvanrossum: It's indeed about the function. A "source" is a designed source of (glyph/kerning/metrics/whatever) data at a location in a designspace, aka a "master value". An "instance" is an interpolation at any location in a designspace. Since they can both be represented with the same type I came up with `StaticGlyph` as the object type, to contrast with `VariableGlyph`, and then it is less weird to use the same object type for both functions. I agree that `GlyphInstance` is a bit weird for a "master value", but `GlyphSource` would definitely be weird for an instance.

justvanrossum: It may take some getting used to, but I think ultimately "source" is a pretty great synonym for "master". (I think this meaning was introduced by Erik van Blokland in the .designspace file format.)

rsheeter: I would prefer the entire concept of "master" get dropped as we push towards sparse definitions. This glyph has definition at {some set of locations}, this other one has them at {some other set}.

I don't see a need for IR to distinguish whether an example/occurance/instance is designed source or not, it's just the definition at a location. IR is a bridge between source formats and binary, not an editors representation.

For context, the IR currently uses XInstance consistently (GlyphInstance, KerningInstance, GlobalMetricsInstance, NamedInstance) to mean a thing defined at a specific location in designspace. That makes "sources" as a collection of them seem weird, something doesn't line up right. If we wanted to "fix" XInstance naming - I don't think we should, but let's pursue it just to see where it takes us - then perhaps what would make sense would be to strip Instance suffixes and instead have a naming convention for things that are variable. For example, instead of Glyph/GlyphInstance you'd have VariableGlyph/Glyph. 

I still think I land at don't actually change anything :)

fontra2fontir/src/fontra.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub(crate) max_value: f64,
+    #[serde(default)]
+    pub(crate) mapping: Vec<[f64; 2]>,
+}
+
+/// serde type used to load .fontra/glyphs/namelike.json files
+///
+/// <https://github.com/googlefonts/fontra/blob/a4edd06837118e583804fd963c22ed806a315b04/src/fontra/core/classes.py#L104-L116>
+#[derive(Debug, Clone, Deserialize)]
+#[allow(dead_code)] // TEMPORARY
```
dfrg, line 66: I have no real preference from a review perspective other than the noise, but I think it's easier to just add `#![allow(dead_code)]` at the top of the module when pushing incremental PRs.

fontra2fontir/src/toir.rs
```diff
@@ -0,0 +1,298 @@
...
+            .iter()
+            .map(|e| match e {
+                PathEl::MoveTo(..) => "M",
+                PathEl::LineTo(..) => "L",
+                PathEl::QuadTo(..) => "Q",
+                PathEl::CurveTo(..) => "C",
+                PathEl::ClosePath => "Z",
+            })
+            .collect::<Vec<_>>()
+            .join("")
```
dfrg, line 247:
```suggestion
            .map(|e| match e {
                PathEl::MoveTo(..) => 'M',
                PathEl::LineTo(..) => 'L',
                PathEl::QuadTo(..) => 'Q',
                PathEl::CurveTo(..) => 'C',
                PathEl::ClosePath => 'Z',
            })
            .collect()
```
Tiny little trick: you can collect an iterator of `char` directly into a `String`.

fontra2fontir/src/toir.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    #[test]
+    fn static_metadata_of_2glyphs() {
+        let fontdata_file = testdata_dir().join("2glyphs.fontra/font-data.json");
+        let font_data = FontraFontData::from_file(&fontdata_file).unwrap();
+        let static_metadata = to_ir_static_metadata(&font_data).unwrap();
+        assert_eq!(1000, static_metadata.units_per_em);
+        assert_eq!(
+            vec![
+                ("Weight", Tag::from_be_bytes(*b"wght"), 200.0, 200.0, 900.0),
```
dfrg, line 274: `Tag::new(b"wght")` is a bit more concise and doesn't require the dereference.

dfrg (approved):
I can't speak directly to the fontra specific bits, but this code LGTM. I do see lots of TODOs without links to issues although some relevant issues seem to exist :)

Nice to have a new front end for a very promising font editor!

fontra2fontir/src/toir.rs
rsheeter: Oops, will fix

fontra2fontir/src/fontra.rs
rsheeter: The noise is annoying indeed. I actually prefer to just prefix fields with _ and not so annotate at all.

rsheeter:
> TODOs without links to issues

Added two links

PR #694, opened 2024-02-01
[fontir] Rename Features -> FeaturesSrc

cmyr:
To differentiate it from the FeaturesAst we will be storing later.


anthrotype (approved):
spelled out FeaturesSource is not so verbose either, but LGTM

PR #695, opened 2024-02-01
Split feature work into separate parsing & compilation steps

cmyr:
This is my first PR that actually adds a new `Work` implementation. This was relatively painless, but slightly confusing in a few places:

- how work creation works is not very well documented or obvious.
- there are a few places where I think a trait could simplify things significantly. Mainly It feels like all of the `add_x_job` methods want to be generic, so that you just have something like `Workflow::add_job<T: WorkOrWorkMakerOrWhaterver>`, and this calls a trait method like `Work::create`.
- the `add_x_job` implementations are kind of opaque and it's not clear what you're supposed to do here. Another benefit of a trait is that it gives you somewhere to document what is expected.
- the need for everything that gets stored in a context to serialize itself is a papercut. In the cast of the AST it is cheap to recreate and inefficient to serialize, and it would be nice if this were expressible.

Overall though not too bad, and I got this working without much struggle.


rsheeter:
Want to file an issue for context items that don't serialize? - I agree that would be a good change.

+1 to the add job bit; I've been gradually making those more self-similar with roughly that intent.

fontbe/src/paths.rs
```diff
@@ -65,6 +65,7 @@ impl Paths {
     pub fn target_file(&self, id: &WorkId) -> PathBuf {
         match id {
             WorkId::Features => self.build_dir.join("features.marker"),
+            WorkId::FeaturesAst => self.build_dir.join("features_ast.marker"),
```
rsheeter, line 68: .marker should mean the file is near-empty. If you are actually writing out the ast in binary it should probably be .bin. Admittedly this is entirely undocumented but you can see it how other files are named.

rsheeter (approved):

cmyr:
opened #697 and #698.

PR #701, opened 2024-02-03
Start to parse Fontra components

### LINKED ISSUE #688: Implement Fontra's glyph->filename algorithm
https://github.com/googlefonts/fontc/pull/686#discussion_r1472648418 notes that

> fontra's glyph->filename algorithm is implemented here: https://github.com/googlefonts/fontra/blob/main/src/fontra/backends/filenames.py

`fontra2fontir/src/source.rs` dumbs this down too far.

rsheeter:
Not yet turned into IR

Components require a better version of string to filename so that is added, fixes #688.


fontra2fontir/src/fontra.rs
```diff
@@ -3,21 +3,105 @@
...
+const SEPARATOR_CHAR: char = '^';
+
+fn is_reserved_char(c: char) -> bool {
+    let c = c as u32;
+    match c {
+        _ if c < 32 => true,
+        _ if c == 0x7F => true,
+        _ if c == SEPARATOR_CHAR as u32 => true,
+        _ => false,
+    }
```
cmyr, line 25: A bit more idiomatic to write this as,

```suggestion
        match c {
        '\0' ..= '\x1F' => true,
        SEPARATOR_CHAR => true,
        '\x7F' => true,
        _ => false,
    }
```

and you could condense it down to a once-line `matches!` statement by combining the matches using `|`, like,


```rust
    matches!(c, '\0'..='\x1F' | SEPARATOR_CHAR | '\x7F')
```

fontra2fontir/src/fontra.rs
```diff
@@ -3,21 +3,105 @@
...
+                bit <<= 1;
+            }
+            digit
+        })
+        .collect();
+    while let Some(0) = code_digits.last() {
+        code_digits.pop();
+    }
+
+    let mut filename = Vec::new();
```
cmyr, line 71: It might be easier to just declare this as a `String`? You still get `push` which takes a `char`, but you also get `push_str` to concatenate, which would simplify the code below.

fontra2fontir/src/fontra.rs
```diff
@@ -3,21 +3,105 @@
...
+    for (i, c) in string.chars().enumerate() {
+        if i == 0 && c == '.' {
+            filename.extend("%2E".chars());
+        } else if !is_reserved_char(c) {
+            filename.push(c);
+        } else {
+            let c = c as u32;
+            for c in format!("{c:02X}").chars() {
+                filename.push(c);
+            }
```
cmyr, line 81: `extend` like above works here too, or `push_str` if you use a string instead of a vec.

fontra2fontir/src/fontra.rs
```diff
@@ -78,7 +180,7 @@ pub(crate) struct FontraSource {
     #[serde(rename = "layerName")]
     pub(crate) layer_name: String,
     #[serde(default)]
-    pub(crate) location: HashMap<Tag, f64>,
+    pub(crate) location: HashMap<String, f64>,
```
cmyr, line 183: if this `String` has any specific meaning (is it supposed to be a tag, or a human readable name, or what?) a comment would be useful.

cmyr (approved):
A few rust idiom/api notes inline, but nothing blocking.

fontra2fontir/src/fontra.rs
rsheeter: Much nicer, will update

fontra2fontir/src/fontra.rs
rsheeter: I think it'll help if I newtype some of the name fields that are used as identifiers. `HashMap<AxisName, f64>`, much like the `<Tag, f64>` version. Aside: I wish it was the tag.

fontra2fontir/src/fontra.rs
```diff
@@ -3,21 +3,105 @@
...
+            let mut bit = 1;
+            for c in string[i..(i + 5).min(string.len())].chars() {
+                if c.is_ascii_uppercase() {
+                    digit |= bit
+                }
+                bit <<= 1;
+            }
+            digit
+        })
+        .collect();
```
dfrg, line 66: Slicing a string like this will panic if the range doesn't align with a character boundary. 

If we can make the assumption that the input is ASCII, I'd just write the map over `string.as_bytes().chunks(4)` (you can call is_ascii_uppercase() on u8).

dfrg (approved):
Already had some comments attached to this so I dropped the ones Colin already mentioned.

fontra2fontir/src/fontra.rs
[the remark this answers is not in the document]
rsheeter: Good catch, will fix.

fontra2fontir/src/fontra.rs
rsheeter: Python behavior here is mildly fascinating.

```
>>> len("❤️‍🔥")
4
>>> s = "Hi ❤️‍🔥 hru"
>>> len(s)
11
>>> list(range(0, len(s), 5)
... )
[0, 5, 10]
>>> s[0:5]
'Hi ❤️'
>>> s[5:10]
'\u200d🔥 hr'
>>> stringToFileName("Hi ❤️‍🔥 hru")
'Hi ❤️\u200d🔥 hru^1'
```

I will add a test and fix. Good catch.

rsheeter: Confirmed panic which is definitely not desired.

fontra2fontir/src/fontra.rs
rsheeter: I will do this in a followon

resources/testdata/fontra/README.md
Sersery0606, line 1: @Sersery0606

fontra2fontir/Cargo.toml
Sersery0606, line 1: #701

Sersery0606, line 1: / #724 #701

PR #703, opened 2024-02-06
fea-rs: conditional compilation of GPOS & GSUB

cmyr:
This passes these options through to the compile step, which will skip processing the relevant statements if the corresponding table is not desired.


fea-rs/src/compile/opts.rs
```diff
@@ -22,4 +27,82 @@ impl Opts {
...
     }
+
+    /// Specify the number of errors to print when printing a [`DiagnosticSet`].
+    ///
+    /// Default is some arbitrary 'reasonable' number (currently 100.) To
+    /// suppress errors, pass `0`. To print all errors, pass a number as large
+    /// as the number of errors you expect to encounter.
+    ///
+    /// [`DiagnosticSet`]: crate::DiagnosticSet
+    pub fn max_error_messages(mut self, max_n_errors: usize) -> Self {
```
rsheeter, line 38: Optional, I think I might have expected an option (-1?) to print all of them

rsheeter (approved):

fea-rs/src/compile/opts.rs
cmyr: this is my fault for trying to be funny in my docs. Can update to just say `usize::MAX`, the joke isn't good anyway 😅

PR #705, opened 2024-02-06
[fontbe] Write out glyph order when --emit-debug is set

cmyr:
This is useful for writing certain tests for fea-rs, since I require a glyph order in order to compile features.


fontbe/src/features.rs
```diff
@@ -393,22 +393,27 @@ impl FeatureCompilationWork {
     }
 }
 
+fn write_glyph_order(context: &Context, glyphs: &GlyphOrder) {
```
rsheeter, line 396: Nit: write_debug_glyph_order would be more consistent

rsheeter (approved):

rsheeter (approved):

PR #706, opened 2024-02-06
[fontbe] Validate FEA AST at parse time

cmyr:
This makes more sense, instead of requiring every future user of the AST to perform a separate validation pass.


dfrg (approved):
Looks good!

dfrg:
... if it compiled :)

PR #711, opened 2024-02-06
Try to make field matching in fontra more obvious with newtypes. Don't reinvent chunks.

rsheeter:
Followon to #701

### DIFF
--- a/fontra2fontir/src/fontra.rs
+++ b/fontra2fontir/src/fontra.rs
@@ -13,6 +13,9 @@ use fontir::error::Error;
 use serde::Deserialize;
 use write_fonts::types::Tag;
 
+pub(crate) type AxisName = String;
+pub(crate) type LayerName = String;
+
 const SEPARATOR_CHAR: char = '^';
 
 fn is_reserved_char(c: char) -> bool {
@@ -63,13 +66,12 @@ const BASE_32_CHARS: [char; 32] = [
 /// Matches <https://github.com/googlefonts/fontra/blob/15bc0b8401054390484cfb86d509d633d29657a1/src/fontra/backends/filenames.py#L40-L64>
 fn string_to_filename(string: &str, suffix: &str) -> String {
     let string_bytes = string.as_bytes();
-    let mut code_digits: Vec<_> = (0..string_bytes.len())
-        .step_by(5)
-        .map(|i| {
+    let mut code_digits: Vec<_> = string_bytes
+        .chunks(5)
+        .map(|chunk| {
             let mut digit = 0;
             let mut bit = 1;
-            let string = string.as_bytes();
-            for byte in string[i..(i + 5).min(string.len())].iter() {
+            for byte in chunk {
                 if byte.is_ascii_uppercase() {
                     digit |= bit
                 }
@@ -140,7 +142,7 @@ impl FontraFontData {
 
 #[derive(Debug, Clone, Deserialize)]
 pub(crate) struct FontraAxis {
-    pub(crate) name: String,
+    pub(crate) name: AxisName,
     pub(crate) tag: Tag,
     #[serde(default)]
     pub(crate) hidden: bool,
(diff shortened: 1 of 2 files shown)


dfrg (approved):
LGTM

anthrotype (commented):
i'm not entirely sure what exactly this buys us besides making the use of these strings slightly more verbose. Do AxisName or LayerName differ in behavior from each other or from a regular string?

fontra2fontir/src/fontra.rs
```diff
@@ -13,6 +14,24 @@ use fontir::error::Error;
 use serde::Deserialize;
 use write_fonts::types::Tag;
 
+#[derive(Debug, Clone, Deserialize, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct AxisName(pub(crate) String);
```
cmyr, line 18: I'm really not sure in this particular case that this is an improvement over `type AxisName = String;`. 


I like new types in many cases, but generally when there is either some invariant that we want to enforce by construction (like we want to ensure that all axis names are a given length or include only characters in some allowed range) or when there is some type-specific API that we would like to provide; here it feels like the only thing we are achieving here is essentially documentation, and I think that a typedef is fine for that.


In general I also think that "newtype with a public inner type" has a bit of code smell, since it adds the additional complexity of a new type without any guarantees.

fontra2fontir/src/fontra.rs
```diff
@@ -13,6 +14,24 @@ use fontir::error::Error;
...
+
+impl Display for AxisName {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.0.as_str())
+    }
+}
```
cmyr, line 24: if we *were* going to go this route, I think it would also be useful to implement the conversion traits (`From`) as well as probably `AsRef<str>`.

fontra2fontir/src/fontra.rs
```diff
@@ -418,7 +436,10 @@ mod tests {
         let glyph = read_test_glyph("minimal.fontra", ".notdef");
         assert_eq!(GlyphName::new(".notdef"), glyph.name, "{glyph:#?}");
         assert_eq!(
-            1000.0, glyph.layers["foreground"].glyph.x_advance,
+            1000.0,
+            glyph.layers[&LayerName("foreground".into())]
```
cmyr, line 440: this is the bit that feels particularly bad: where previously we could use a `&str` as the key here (because `String` implements `Borrow<str>` now we need to allocate and be more verbose, in a situation where confusion seems unlikely.

cmyr (commented):
I'm with cosimo here, I'm not sure this is buying us anything.

fontra2fontir/src/fontra.rs
rsheeter: Updating to alias.

rsheeter:
I feel strongly that leaving it as naked String is misleading and error prone. There is at least one spot in fontra already where pythong declares str, no comment, and I had to look at actual data to see if it was a tag an axis name. I'm not especially tied to using a new type so I will update to type aliases.

anthrotype:
>  There is at least one spot in fontra already where pythong declares str, no comment, and I had to look at actual data to see if it was a tag an axis name.

`String` is certainly not a `Tag`. And both a layer name and an axis name, unlike a tag, don't have any particular requirements beyond being a string. So it would be correct to type both as a String in my opinion. You can argue that the python code you were looking at was ambiguous in that it seemed to be using `str` type for an axis _tag_ (which it wasn't in the end).

Also, by the way, the two type aliases are just that, alternative names to the _same_ type, which is still String. One could still use a LayerName or an AxisName (or String) interchangeably and nothing would happen -- and that's fine.

rsheeter:
The goal is merely it's a little bit clearer what matches what. You can still do the wrong thing but perhaps it's a little bit less likely. 

For example, in python I see `Location = dict[str, float]` without a comment. What's the key? - maybe you know from convention it's likely to be a name but if you aren't aware of that convention it's not at all obvious. If it said `Location = dict[AxisName, float]` or had a comment it would be more clear. I hope using the type aliases - I agree they don't need to be newtypes - is similarly helpful. If not ... well, it probably doesn't do any harm as far as I can tell.

PR #718, opened 2024-02-13
[fea-rs] Support inline 'by Null' in contextual sub

cmyr:
- closes https://github.com/cmyr/fea-rs/issues/246

cc @simoncozens


simoncozens:
Nice, thank you!

PR #719, opened 2024-02-13
Correct case of subfamily name

simoncozens:
Currently with a UFO like this:

```
    <key>familyName</key>
    <string>Noto Sans Balinese</string>
    <key>styleMapFamilyName</key>
    <string>Noto Sans Balinese</string>
    <key>styleMapStyleName</key>
    <string>bold</string>
    <key>styleName</key>
    <string>Bold</string>
    <key>trademark</key>
```

fontc produces a font like this:

```
    <namerecord nameID="1" platformID="3" platEncID="1" langID="0x409">
      Noto Sans Balinese
    </namerecord>
    <namerecord nameID="2" platformID="3" platEncID="1" langID="0x409">
      bold
    </namerecord>
    <namerecord nameID="3" platformID="3" platEncID="1" langID="0x409">
      2.005;GOOG;NotoSansBalinese-bold
    </namerecord>
    <namerecord nameID="4" platformID="3" platEncID="1" langID="0x409">
      Noto Sans Balinese bold
    </namerecord>
    <namerecord nameID="5" platformID="3" platEncID="1" langID="0x409">
      Version 2.005
    </namerecord>
    <namerecord nameID="6" platformID="3" platEncID="1" langID="0x409">
      NotoSansBalinese-bold
    </namerecord>
```

This is wrong (and worse, stops us comparing the files with diffenator2!). The stylemap names should be capitalised when used to create the Name ID 2 subfamily name.

PR #721, opened 2024-02-13
Correct ordering of transform coefficients

simoncozens:
Components with a xyScale and yxScale are currently broken. Noto Sans Balinese before this patch:

shape

After:

shape

Comment explains why.

### DIFF
--- a/ufo2fontir/src/toir.rs
+++ b/ufo2fontir/src/toir.rs
@@ -60,6 +60,12 @@ fn to_ir_component(component: &norad::Component) -> ir::Component {
         base: component.base.as_str().into(),
         transform: Affine::new([
             component.transform.x_scale,
+            // For the 2nd and 3rd field of its 2x3 affine transformation, norad uses
+            // the same labels as fonttools' Transform, respectively `xy` and `yx`.
+            // Elsewhere (e.g. FreeType, Cairo or read-fonts Transform) these labels are
+            // inverted, but their meaning is still the same: i.e. the "y component of
+            // the î basis vector", and "x component of the ĵ basis vector".
+            // See https://github.com/googlefonts/fontc/pull/721
             component.transform.xy_scale,
             component.transform.yx_scale,
             component.transform.y_scale,
--- a/ufo2fontir/src/toir.rs
+++ b/ufo2fontir/src/toir.rs
@@ -292,19 +292,19 @@ mod tests {
             Affine::new([1.0, 0.0, 0.0, 1.0, 10.0, 10.0])
         );
 
-        // <component base="a" xScale="0.4366" xyScale="-0.4366" yScale="0.4415" yxScale="0.4415" xOffset="282" yOffset="5" identifier="5402E799"/>
+        // <component base="a" xScale="0.4366" xyScale="-0.4366" yScale="0.4425" yxScale="0.4415" xOffset="282" yOffset="5" identifier="5402E799"/>
         c.transform = AffineTransform {
             x_scale: 0.4366,
             xy_scale: -0.4366,
-            y_scale: 0.4415,
             yx_scale: 0.4415,
+            y_scale: 0.4425,
             x_offset: 282.0,
             y_offset: 5.0,
         };
         // Switchy switchy!
         assert_eq!(
             to_ir_component(&c).transform,
-            Affine::new([0.4366, -0.4366, 0.4415, 0.4415, 282.0, 5.0])
+            Affine::new([0.4366, -0.4366, 0.4415, 0.4425, 282.0, 5.0])
         );
     }
 }
--- a/ufo2fontir/src/toir.rs
+++ b/ufo2fontir/src/toir.rs
@@ -187,7 +187,7 @@ mod tests {
 
     use fontdrasil::coords::{NormalizedCoord, NormalizedLocation};
     use fontir::ir::AnchorBuilder;
-    use norad::ContourPoint;
+    use norad::{AffineTransform, Component, ContourPoint, Name};
 
     use super::*;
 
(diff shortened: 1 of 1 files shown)


simoncozens:
Tests are failing but how much do you trust your tests? ;-)

anthrotype:
The eternal fight between `xy` vs `yx` in 2D graphics...
 
Can you please update the tests as well?

this reminds me of a discussion we had when drafting the COLR spec:

https://github.com/googlefonts/colr-gradients-spec/pull/85

anthrotype:
After re-reading the old long thread at https://github.com/googlefonts/colr-gradients-spec/pull/85 and I think the flip you're doing here is not at the right place, it might be that it's the ufo2fontir that is setting the kurbo Transform incorrectly.
The labels xy and yx or b or c are always confusing which is why one needs to look at the _meaning_ of those values, i.e. interpreting them as x or y components of the î or ĵ basis vectors, and also take into account whether the values of the matrix are stored in column-first or row-first order.
Please let me double check

anthrotype:
@simoncozens can you please help me to reproduce the issue you've shown in the screenshots above?

You compiled https://github.com/notofonts/balinese using fontc? Which glyph/character is that?

I think the test results that you have modified to make the test pass following your change, were actually correct. And in fact if I build the same test font resources/testdata/glyphs2/Component.glyphs using fontmake 3.8.1 I get exactly the same glyf table.
If I inspect the composite glyphs that contain some 2x2, like 'comma' or 'non_uniform_transform' (encoded as digit 0), they render as expected.

simoncozens:
> You compiled https://github.com/notofonts/balinese using fontc? Which glyph/character is that?

I did `fontmake -o ufo -g sources/NotoSansBalinese.glyphs; fontc master_ufo/NotoSansBalinese-Bold.ufo; shape build/font.ttf -u 25cc`.

anthrotype:
After applying your patch the "non_uniform_scale" and "comma" glyphs look like this when dumped with ttx and diff'ed. fontc is using your PR branch, fontmake is v3.8.1:

```diff
$ diff -u build/default/fontc.glyf.ttx build/default/fontmake.glyf.ttx | colordiff | less
--- build/default/fontc.glyf.ttx        2024-02-14 13:09:27
+++ build/default/fontmake.glyf.ttx     2024-02-14 13:09:27
@@ -19,12 +19,12 @@
       <instructions/>
     </TTGlyph>

-    <TTGlyph name="comma" xMin="241" yMin="-481" xMax="508" yMax="-253">
-      <component glyphName="period" x="-233" y="-129" scalex="1.6655" scale01="-1.1611" scale10="1.1611" scaley="1.6655" flags="0x4"/>
+    <TTGlyph name="comma" xMin="67" yMin="245" xMax="334" yMax="473">
+      <component glyphName="period" x="-233" y="-129" scalex="1.6655" scale01="1.1611" scale10="-1.1611" scaley="1.6655" flags="0x4"/>
     </TTGlyph>

-    <TTGlyph name="non_uniform_scale" xMin="8" yMin="-481" xMax="143" yMax="-253">
-      <component glyphName="period" x="-233" y="-129" scalex="0.8452" scale01="-1.1611" scale10="0.58923" scaley="1.6655" flags="0x4"/>
+    <TTGlyph name="non_uniform_scale" xMin="-138" yMin="102" xMax="26" yMax="259">
+      <component glyphName="period" x="-233" y="-129" scalex="0.8452" scale01="0.58923" scale10="-1.1611" scaley="1.6655" flags="0x4"/>
     </TTGlyph>

     <TTGlyph name="period" xMin="250" yMin="50" xMax="375" yMax="100">
\ No newline at end of file
```

Basicaly you swapped "scale01" and "scale10" component attributes. But the glyphs look wrong with this patch.

anthrotype:
ah so you compiled from a UFO, not directly from the NotoSansBalinese.glyphs source. Ok so I need to look somewhere else for the cuplrit. But it definitely this is not the right fix, because both kurbo::Affine and the fontations Transform structs use a column-first order for representing the matrix values

simoncozens:
Ah, OK, so it's probably ufo2fontir

anthrotype:
can you please try with:

```
diff --git a/ufo2fontir/src/toir.rs b/ufo2fontir/src/toir.rs
index 566b67ff..5fdfdef2 100644
--- a/ufo2fontir/src/toir.rs
+++ b/ufo2fontir/src/toir.rs
@@ -60,8 +60,8 @@ fn to_ir_component(component: &norad::Component) -> ir::Component {
         base: component.base.as_str().into(),
         transform: Affine::new([
             component.transform.x_scale,
-            component.transform.yx_scale,
             component.transform.xy_scale,
+            component.transform.yx_scale,
             component.transform.y_scale,
             component.transform.x_offset,
             component.transform.y_offset,
```

I'm off to lunch and can take a closer look later

simoncozens:
Yes, that works on my fonts.

rsheeter:
Ahaha I still wish transforms more directly spoke to basis vectors. 

Good catch. Sounds like a hole in our tests, quite likely by me :)

Please add a new test to prove it was broken (e.g. test fails on the current code) and is now fixed.

rsheeter (approved):
LGTM conditional on addition of test(s) that prove it was broken and is now fixed. Unit test(s) for a handful of to_ir_component scenarios that checks the Affine on the ir::Component I assume.

simoncozens:
> Sounds like a hole in our tests, quite likely by me :)

After #716 and #719, we're now in a position where we can use diffenator2 to compare fontc/fontmake outputs of UFO builds, which will make isolated regression testing a lot easier. I'm hoping to be able to oxidise *some* of the Noto production build (just the instance statics), using this to find differences and failures, and hopefully iterate fontc closer to correctness.

> Please add a new test to prove it was broken 

Yep, will do.

rsheeter:
> After https://github.com/googlefonts/fontc/pull/716 and https://github.com/googlefonts/fontc/pull/719, we're now in a position where we can use diffenator2 to compare fontc/fontmake outputs of UFO builds, which will make isolated regression testing a lot easier. I'm hoping to be able to oxidise some of the Noto production build (just the instance statics), using this to find differences and failures, and hopefully iterate fontc closer to correctness.

Great news!

anthrotype:
So all the confusion was due to norad's AffineTransform using the same terminology as fonttools' Transform where it refers to the "y component of the î basis vector" as _xy_ and to the "x component of the ĵ basis vector" as _yx_.

https://github.com/linebender/norad/blob/e58e32009786e767066b9af2cba17bcff23c9ae3/src/glyph/mod.rs#L443-L456

On the other hand, read-fonts' 2x2 Transform (also used by write-fonts) names the same values with the opposite labels: _yx_ is "y component of the î basis vector" and _xy_ is "x component of the ĵ basis vector". This follows a naming convention also used in FreeType and Cairo.

Note however that the values of the matrix are all listed in _column-first_ order, the same that kurbo Affine uses, or that the OpenType glyf Component's "scale01" and "scale10" values are (the first index is the column, the second is the row).

So it's not like we are swapping places, we're simply calling the same things differently.

ufo2fontir/src/toir.rs
```diff
@@ -252,4 +252,53 @@ mod tests {
...
+            to_ir_component(&c).transform,
+            Affine::new([1.0, 0.0, 0.0, 1.0, 10.0, 10.0])
+        );
+
+        // <component base="a" xScale="0.4366" xyScale="-0.4366" yScale="0.4415" yxScale="0.4415" xOffset="282" yOffset="5" identifier="5402E799"/>
+        c.transform = AffineTransform {
+            x_scale: 0.4366,
+            xy_scale: -0.4366,
+            y_scale: 0.4415,
+            yx_scale: 0.4415,
```
anthrotype, line 294: I think it might help avoid any further confusions if you keep the struct fields here ordered the same way they are defined in the norad::AffineTransform struct, so 
```suggestion
            x_scale: 0.4366,
            xy_scale: -0.4366,
            yx_scale: 0.4415,
            y_scale: 0.4415,
```

anthrotype (approved):

Lorp:
It would help enormously if documentation shows the matrix itself using the local terms. (Until today I thought glyf’s _scale10_ and _scale01_ were the other way around!)

For example, in [glyf](https://learn.microsoft.com/en-us/typography/opentype/spec/glyf):

```math
\begin{bmatrix}
    xscale & scale10 \\
    scale01 & yscale
\end{bmatrix}
```

Thus for a 90º CCW rotation…

```math
\begin{bmatrix}
    0 & -1 \\
    1 & 0
\end{bmatrix}
```

… we have:

```
xscale = 0
scale01 = 1
scale10 = -1
yscale = 0
```

PR #723, opened 2024-02-29
[fea-rs] Remove ClassId, add gdef classes to output

cmyr:
- ClassId was a pre-write-fonts enum standing in for the GlyphClassDef enum, which we can now juse use directly
- passing any explicit GDEF classes through to the output feels like the simplest way to expose these to manual feature generation code, since that code needs to compile anyway to get the GSUB table.

The motivation here was that I want the explicit classes in a more user-friendly format, instead of having to go muck about in the compiled GDEF ClassDef table/recheck the AST to figure out if it was inferred or not, etc.

### DIFF
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -13,7 +13,7 @@ use smol_str::SmolStr;
 use write_fonts::{
     tables::{
         self,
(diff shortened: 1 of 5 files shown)


dfrg (approved):
I don’t quite understand how this all fits together but the rationale seems sound and the code LGTM.

PR #725, opened 2024-03-06
Add KernGroup type

cmyr:
This encodes in the type system the idea that kerning groups have an associated side, instead of having this be encoded as a prefix on a string.

This removes the need to hae separate `KernSide` enums, and removes possible complications around using the same type for glyph names and group names.

----

I was expecting this to be a smaller diff, but overall I think it's at least a modest improvement.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -1,7 +1,7 @@
 //! Font IR types.
 use std::{
     collections::{hash_map::RandomState, BTreeMap, BTreeSet, HashMap, HashSet},
-    fmt::Debug,
+    fmt::{Debug, Display},
     io::Read,
     path::PathBuf,
 };
@@ -11,7 +11,8 @@ use indexmap::IndexSet;
 use kurbo::{Affine, BezPath, PathEl, Point};
 use log::{log_enabled, trace, warn};
 use ordered_float::OrderedFloat;
-use serde::{Deserialize, Serialize};
+use serde::{de::Error, Deserialize, Serialize};
+use smol_str::SmolStr;
 use write_fonts::{
     tables::os2::SelectionFlags,
     types::{GlyphId, NameId, Tag},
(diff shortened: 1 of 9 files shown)


fontir/src/ir.rs
```diff
@@ -239,7 +250,42 @@ pub struct KerningInstance {
...
+impl<'de> Deserialize<'de> for KernGroup {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+        s.strip_prefix("side1.")
+            .map(|s| KernGroup::Side1(s.into()))
+            .or_else(|| s.strip_prefix("side2.").map(|s| KernGroup::Side2(s.into())))
+            .ok_or_else(|| D::Error::custom(format!("missing side1/side prefix: {s}")))
```
rsheeter, line 287: Missing the 2

rsheeter (approved):

glyphs2fontir/src/source.rs
```diff
@@ -608,31 +608,17 @@ impl Work<Context, WorkId, WorkError> for FeatureWork {
...
-        }
-    }
-
-    fn group_prefix(&self) -> &'static str {
-        match self {
-            KernSide::Side1 => "public.kern1.",
-            KernSide::Side2 => "public.kern2.",
-        }
-    }
+fn parse_kern_group(name: &str) -> Option<KernGroup> {
```
anthrotype, line 611: could this be a method on KernGroup maybe?

anthrotype (approved):
LGTM

glyphs2fontir/src/source.rs
cmyr: the issue is that the parsing behaviour depends on the backend, so it doesn't make sense to hang it off the common type.

PR #727, opened 2024-03-06
Kern alignment

cmyr:
This was way more confusing than it should've been, but ultimately I
believe this is correct. It brings the oswald diff from 8.6k/3.3k to
7.5k/2.2k, which is very promising.

I do now see one additional difference in kerning, which is that in some cases we have identical values at each position in the designspace; this means we don't need to store deltas at all. We're also sometimes storing explicit 0 values, which is also unnecessary.

CI is currently broken, pending #726.
this is based on #725, which will go in when CI is fixed.

### DIFF
--- a/fontbe/src/kern.rs
+++ b/fontbe/src/kern.rs
@@ -1,17 +1,22 @@
 //! Generates an [FeaRsKerns] datastructure to be fed to fea-rs
 
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, HashMap, HashSet};
 
 use fea_rs::{
     compile::{PairPosBuilder, ValueRecord as ValueRecordBuilder},
     GlyphSet,
 };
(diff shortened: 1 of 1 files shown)


fontbe/src/kern.rs
```diff
@@ -140,6 +147,96 @@ impl Work<Context, AnyWorkId, Error> for GatherIrKerningWork {
...
+            .collect::<Vec<_>>();
+
+        for pair in missing_pairs {
+            let value = lookup_kerning_value(
+                pair,
+                instance,
+                &side1_glyph_to_group_map,
+                &side2_glyph_to_group_map,
+            );
+            log::info!("got aligned value {value} for kern pair {pair:?}");
```
anthrotype, line 191: maybe log::info level would be too noisy for this, debug might work better

cmyr: oh good call, I just added this for my personal edification so I could see if what was happening looked sensible, I think we can just remove it.

PR #728, opened 2024-03-07
[kerning] Prune nop deltas

cmyr:
This skips writing sets of all-zero deltas.

This removes the last funny numerical diffs I was seeing in the generated kerning, which hopefully means the only remaining differences are related to script splitting (and so will make debugging that work much simpler)


dfrg (approved):
Whittling away at the diffs…

cmyr:
this will be unblocked when https://github.com/googlefonts/fontations/pull/819 lands

Issue #619, opened 2023-12-01
GPOS: feature writers should split lookups based on language system


cmyr:
This is a significant project, and includes splitting based on writing direction. A good place to start will be [`kernFeatureWriter2.py`](https://github.com/googlefonts/ufo2ft/blob/cea60d71dfcf0b1c0fa4e133ec4231ba06fe0da0/Lib/ufo2ft/featureWriters/kernFeatureWriter2.py#L315).

belluzj:
Hello, here I think you're first talking about splitting by direction (LTR vs RTL) which is necessary.

As a second step, or while you're doing this, you might be tempted to also split into one lookup per script,[ as we're doing in recent ufo2ft](https://github.com/googlefonts/ufo2ft/blob/cea60d71dfcf0b1c0fa4e133ec4231ba06fe0da0/Lib/ufo2ft/featureWriters/kernFeatureWriter.py). After discussing this issue several times with Cosimo, here are my thoughts on the topic: basically, splitting scripts into different lookups was a compile-time performance and file size optimization, and it's not the best one, so I don't think it should be adopted here. 

Pros of splitting into one lookup per script:
* it happens early in the pipeline (as opposed to GPOS compaction near the end) so it saves processing time early by keeping small lookup sizes and avoiding overflow resolution
* in terms of file size savings, it's a very good approximation of the probable best splitting (we observed that splitting into one lookup per script is as good for file size as doing GPOS compaction which doesn't know about scripts, and splits based on clustering the numeric data in the table)

Con: It's not functionally equivalent to the old way of having just one big lookup for all LTR kerning, and another big lookup for all RTL kerning. The functional differences are minor but they are regressions:
  * it [prevents cross-script kerning](https://github.com/googlefonts/ufo2ft/issues/808): Khaled [implemented a fix](https://github.com/googlefonts/ufo2ft/pull/811) here, but the fix is to go back to grouping scripts together, so I believe it loses the performance and file size benefits, in favour of correctness
  * it causes regressions in Adobe InDesign ([one](https://github.com/googlefonts/ufo2ft/pull/787) and [two](https://github.com/googlefonts/ufo2ft/pull/788)) which arguably are Adobe's fault, but unfortunately will matter to customers. I tried to fix these by registering the lookups everywhere, and while that preserves the compile-time performance and small file size, it creates a runtime performance issue, as profiled by Behdad, so it's not good either.
  
[...]

PR #731, opened 2024-03-12
Split kerning by script

cmyr:
This patch is imperfect, but I think it is worth getting something committed that we can iterate from.

- modulo a very small number of differences, this matches the output of fontmake on Oswald.
- this does not correctly set the lookup flags (or mark filtering set); I'll do that as a followup
- this does not handle the 'dist' feature
- the testing here is anemic. I have at least figured out *how* to write tests against this, and I see that there are some ideas at https://github.com/googlefonts/ufo2ft/blob/cea60d71dfcf0b1c0fa4e133ec4231ba06fe0da0/tests/featureWriters/kernFeatureWriter_test.py that look worth porting, but I think it makes sense to focus on this as I continue to refine.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,6 @@ temp-env = "0.3.3"
 rstest = "0.18.2"
 
 [workspace]
-
 resolver = "2"
 
 members = [
@@ -52,3 +51,6 @@ members = [
     "layout-normalizer",
 ]
 
+[patch.crates-io]
+icu_properties = { version = "1.4", git = "https://github.com/unicode-org/icu4x.git", rev = "728eb44" }
+tinystr = { version = "0.7.5", git = "https://github.com/unicode-org/icu4x.git", rev = "728eb44" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,8 @@ members = [
     "layout-normalizer",
 ]
 
+# remove this when icu_properties 1.5 is released:
+# https://github.com/unicode-org/icu4x/issues?q=is%3Aopen+is%3Aissue+milestone%3A%221.5+Blocking+⟨P1⟩%22
 [patch.crates-io]
 icu_properties = { version = "1.4", git = "https://github.com/unicode-org/icu4x.git", rev = "728eb44" }
 tinystr = { version = "0.7.5", git = "https://github.com/unicode-org/icu4x.git", rev = "728eb44" }
--- a/fontbe/src/features/kern.rs
+++ b/fontbe/src/features/kern.rs
@@ -443,8 +443,6 @@ impl KerningGatherWork {
 
         let gdef = compilation.gdef_classes;
 
-        // serves as a standin for cmap in the fonttools impl
-
         let pairs = fragments
             .iter()
             .flat_map(|frag| frag.kerns.iter())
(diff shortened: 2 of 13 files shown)


anthrotype (commented):
thanks Colin this is big! I've only started looking at this today, will continue review in the next few days

Cargo.toml
```diff
@@ -52,3 +51,6 @@ members = [
     "layout-normalizer",
 ]
 
+[patch.crates-io]
+icu_properties = { version = "1.4", git = "https://github.com/unicode-org/icu4x.git", rev = "728eb44" }
```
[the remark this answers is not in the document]
cmyr: they do, but I had to fix a few things to get this to work, and those things aren't published yet. See https://github.com/unicode-org/icu4x/issues/4681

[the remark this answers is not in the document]
rsheeter: https://docs.rs/icu_properties/latest/icu_properties/, maybe there's an unreleased feature we need? If so we should get them to publish.

madig:
Before looking at this, does this PR incorporate the ideas in https://github.com/googlefonts/fontc/issues/619#issuecomment-1909750996?

anthrotype:
> Before looking at this, does this PR incorporate the ideas in https://github.com/googlefonts/fontc/issues/619#issuecomment-1909750996?

no, I don't think so. This is generating multiple per-script lookups, not subtables, like ufo2ft currently does. I think Colin wants to just match fontmake output for now.

anthrotype:
@madig also note that Colin's PR does implements the feature which Khaled added recently to ufo2ft whereby kerning lookups get merged into one when cross-script kerning is present, which I think addresses your concerns about Indesign dumb composer.

Cargo.toml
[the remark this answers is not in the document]
cmyr: see my comment above: I had to PR a few things to icu_properties. They have a release planned, but it's a ways away.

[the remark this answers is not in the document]
rsheeter: Let's add a comment pointing at https://github.com/unicode-org/icu4x/issues/4681 and noting that we should be able to drop this as of their 1.5 release.

cmyr:
okay so there's one remaining issue here afaik, which rod posted in #733:

https://github.com/googlefonts/fontc/pull/733#discussion_r1524973948

fontbe/src/features/properties.rs
```diff
@@ -0,0 +1,329 @@
...
+pub const COMMON_SCRIPT: UnicodeShortName =
+    unsafe { UnicodeShortName::from_bytes_unchecked(*b"Zyyy") };
+pub const INHERITED_SCRIPT: UnicodeShortName =
+    unsafe { UnicodeShortName::from_bytes_unchecked(*b"Zinh") };
+
+static SCRIPT_DATA: ScriptWithExtensionsBorrowed<'static> =
+    icu_properties::script::script_with_extensions();
+
+/// The type used by icu4x for script names
+pub type UnicodeShortName = tinystr::TinyAsciiStr<4>;
```
dfrg, line 31: I think we should just define a `ScriptTag([u8; 4])` type (or just reuse `Tag`?) and bypass `TinyAsciiStr`. At least for shaping, we definitely don't want to deal with this intermediate type and I don't think it adds any value.

I'd add a const `new([u8; 4]) -> Self` constructor for this and a `new_checked([u8; 4]) -> Option<Self>` constructor that tries to round trip through the `name_to_enum_mapper`.

This also lets us avoid the unsafe code.

dfrg: thinking on this a bit more, I don't think the checked constructor is even necessary but `fn icu_script(self) -> Option<icu_properties::Script>` might be useful and provides the same validation

cmyr: the reason to prefer this type is because it is used by `icu_propeties`, so that we can do things like query the properties of a unicode value, which returns [`Script`](https://docs.rs/icu_properties/1.4.0/icu_properties/struct.Script.html)s, and then we can convert those to the unicode names via that crate.

dfrg: oof, I just wrote something similar a couple weeks ago: https://github.com/dfrg/fount/blob/731e374eb963d30abb4d0f938bb515cc3de0a823/fontique/src/script.rs 🤷

cmyr: looking at yours, is the `unicode-script` crate necessary? I think that information is available via `icu_properties`?

cmyr: in any case my vote right now is that we punt on this and can figure out a more holistic approach when we get to shaping?.

dfrg: yeah, happy to punt on this for now

I added optional unicode-script support because there are potential users that already depend on that library for scripts.

fontbe/src/features/kern.rs
```diff
@@ -359,31 +382,719 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+                                )));
+                            }
+                            // these various combos all go in the marks group
+                            for (side1, side2) in [
+                                (&side1_bases, &side2_marks),
+                                (&side1_marks, &side2_bases),
+                                (&side1_marks, &side2_marks),
+                            ] {
+                                if !side1.is_empty() && !side2.is_empty() {
+                                    base_pairs.push(Cow::Owned(PairPosEntry::Class(
```
anthrotype, line 916: @cmyr If I followed the logic correctly, shouldn't this `base_pairs` actually be `mark_pairs`?

anthrotype (commented):
belated LGTM! you did an excellent job porting over to Rust the complicated logic of ufo2ft kernFeatureWriter 💯 

Left just a little comment above

fontbe/src/features/kern.rs
cmyr: good catch, I have tests for this in #733 but apparently not handling the class case, will fix that now :)

PR #733, opened 2024-03-13
set lookupflag and mark filtering set for kerning lookups

cmyr:
This adds a `PendingLookup` type so that we can compute and store the lookup flag and the mark filtering sets when we generate the kerning lookups.

based on #731, which should go in first.

### DIFF
--- a/fontbe/src/features/kern.rs
+++ b/fontbe/src/features/kern.rs
@@ -27,7 +27,7 @@ use log::debug;
 use ordered_float::OrderedFloat;
 use write_fonts::{
     read::{tables::gsub::Gsub, FontRead, ReadError},
-    tables::gdef::GlyphClassDef,
+    tables::{gdef::GlyphClassDef, layout::LookupFlag},
     types::{GlyphId, Tag},
 };
 
@@ -43,7 +43,7 @@ use crate::{
     },
 };
 
-use super::properties::CharMap;
+use super::{properties::CharMap, PendingLookup};
 
 /// On Linux it took ~0.01 ms per loop, try to get enough to make fan out worthwhile
 /// based on empirical testing
@@ -463,11 +463,14 @@ impl KerningGatherWork {
     /// <https://github.com/googlefonts/ufo2ft/blob/f6b4f42460b/Lib/ufo2ft/featureWriters/kernFeatureWriter.py#L772>
     fn assign_lookups_to_scripts(
         &self,
-        lookups: BTreeMap<BTreeSet<UnicodeShortName>, Vec<PairPosBuilder>>,
+        lookups: BTreeMap<BTreeSet<UnicodeShortName>, Vec<PendingLookup<PairPosBuilder>>>,
         ast: &ParseTree,
         // one of 'kern' or 'dist'
         current_feature: Tag,
-    ) -> (Vec<Vec<PairPosBuilder>>, BTreeMap<FeatureKey, Vec<usize>>) {
+    ) -> (
+        Vec<PendingLookup<PairPosBuilder>>,
+        BTreeMap<FeatureKey, Vec<usize>>,
+    ) {
         let dflt_langs = vec![DFLT_LANG];
 
         let is_kern_feature = current_feature == KERN;
@@ -482,13 +485,15 @@ impl KerningGatherWork {
 
         // in python this part happens earlier, as part of splitKerning.
         for (scripts, lookups) in lookups {
-            let idx = ordered_lookups.len();
-            ordered_lookups.push(lookups);
-            for script in scripts {
-                lookups_by_script
-                    .entry(script)
-                    .or_insert(Vec::new())
-                    .push(idx);
+            for lookup in lookups {
+                let idx = ordered_lookups.len();
+                ordered_lookups.push(lookup);
+                for script in &scripts {
+                    lookups_by_script
+                        .entry(script.to_owned())
+                        .or_insert(Vec::new())
+                        .push(idx);
+                }
             }
         }
 
@@ -568,10 +573,10 @@ impl KerningGatherWork {
 
 fn debug_ordered_lookups(
     features: &BTreeMap<FeatureKey, Vec<usize>>,
-    lookups: &[Vec<PairPosBuilder>],
+    lookups: &[PendingLookup<PairPosBuilder>],
 ) {
-    for (i, subtables) in lookups.iter().enumerate() {
-        let total_rules = subtables.iter().map(|x| x.len()).sum::<usize>();
+    for (i, lookup) in lookups.iter().enumerate() {
+        let total_rules = lookup.subtables.iter().map(|x| x.len()).sum::<usize>();
         log::trace!("lookup {i}, {total_rules} rules");
     }
 
@@ -641,7 +646,7 @@ impl KernSplitContext {
     fn make_lookups(
         &self,
         pairs: &[&PairPosEntry],
-    ) -> BTreeMap<BTreeSet<UnicodeShortName>, Vec<PairPosBuilder>> {
+    ) -> BTreeMap<BTreeSet<UnicodeShortName>, Vec<PendingLookup<PairPosBuilder>>> {
         if !self.opts.ignore_marks {
             let pairs = pairs.iter().map(|x| Cow::Borrowed(*x)).collect::<Vec<_>>();
             return self.make_split_script_kern_lookups(&pairs, false);
@@ -665,21 +670,18 @@ impl KernSplitContext {
     /// Returns a map of scripts: `[lookup]`, where 'scripts' is a set of scripts
     /// referencing the lookups.
     ///
-    /// Although this method always/only returns a single lookup per unique set
-    /// of scripts, we wrap it in a vec because that is what we need at the
-    /// call site.
     // <https://github.com/googlefonts/ufo2ft/blob/f6b4f42460b/Lib/ufo2ft/featureWriters/kernFeatureWriter.py#L694>
     fn make_split_script_kern_lookups(
         &self,
         pairs: &[Cow<PairPosEntry>],
         //TODO: handle marks
         _are_marks: bool,
-    ) -> BTreeMap<BTreeSet<UnicodeShortName>, Vec<PairPosBuilder>> {
+    ) -> BTreeMap<BTreeSet<UnicodeShortName>, Vec<PendingLookup<PairPosBuilder>>> {
         let mut lookups_by_script = BTreeMap::new();
         let kerning_per_script = self.split_kerns(pairs);
         let mut bidi_buf = HashSet::new(); // we can reuse this for each pair
         for (scripts, pairs) in kerning_per_script {
-            let mut lookup = PairPosBuilder::default();
+            let mut builder = PairPosBuilder::default();
             for mut pair in pairs {
                 bidi_buf.clear();
                 for (direction, glyphs) in &self.bidi_glyphs {
@@ -706,8 +708,9 @@ impl KernSplitContext {
                 if pair_is_rtl {
                     pair.make_rtl_compatible();
                 }
-                pair.add_to(&mut lookup);
+                pair.add_to(&mut builder);
             }
+            let lookup = PendingLookup::new(vec![builder], LookupFlag::empty(), None);
             lookups_by_script.insert(scripts, vec![lookup]);
         }
         lookups_by_script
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -52,3 +51,6 @@ members = [
     "layout-normalizer",
 ]
 
+[patch.crates-io]
```
rsheeter, line 54: Presumably we should move to unpatched deps before merge.

fea-rs/src/compile/lookups.rs
```diff
@@ -182,7 +183,8 @@ pub(crate) struct LookupFlagInfo {
 }
 
 /// A feature associated with a particular script and language.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct FeatureKey {
```
rsheeter, line 188: Worth considering types for feature vs language vs script? - it appears easy to confuse them under the current definition.

fontbe/src/features/kern.rs
```diff
@@ -359,31 +390,843 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+        .map_err(|err| {
+            Error::FeaCompileError(fea_rs::compile::error::CompilerError::CompilationFail(err))
+        })?;
+
+        let gsub = compilation
+            .gsub
+            .as_ref()
+            .map(write_fonts::dump_table)
+            .transpose()
+            .expect("if this doesn't compile we will already panic when we try to add it to the context");
```
rsheeter, line 446: Not quite sure I follow. This seems like it should return an Error and if that causes a downstream panic that's a bug?

fontbe/src/features/kern.rs
```diff
@@ -359,31 +390,843 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+        let gsub = gsub
+            .as_ref()
+            .map(|data| write_fonts::read::tables::gsub::Gsub::read(data.as_slice().into()))
+            .transpose()?;
+
+        let gdef = compilation.gdef_classes;
+
+        // serves as a standin for cmap in the fonttools impl
```
rsheeter, line 454: permalink to fonttools code?

fontbe/src/features/kern.rs
```diff
@@ -359,31 +390,843 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+            .collect();
+        let split_ctx = KernSplitContext::new(&glyphs, &known_scripts, gsub, mark_glyphs)?;
+
+        let lookups = split_ctx.make_lookups(&pairs);
+        let (lookups, features) = self.assign_lookups_to_scripts(lookups, ast, KERN);
+        Ok(FeaRsKerns { lookups, features })
+    }
+
+    /// returns a vec of lookups (as a vec of subtables), along with a map of features -> lookups
+    /// (by order in the first vec)
```
rsheeter, line 491: This is mildly confusing. I suggest a struct with a comment explaining it's layout for the return value. I _think_ it's saying the values in the map parallel the vector of pending lookups?

fontbe/src/features/kern.rs
```diff
@@ -359,31 +390,843 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+    fn assign_lookups_to_scripts(
```
rsheeter, line 492: it would be nice for the function to permalink to the python function or block(s) it mirrors.

fontbe/src/features/kern.rs
```diff
@@ -359,31 +390,843 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+        Vec<PendingLookup<PairPosBuilder>>,
+        HashMap<FeatureKey, Vec<usize>>,
+    ) {
+        let dflt_langs = vec![DFLT_LANG];
 
-        let mut entries = 0;
-        for ppe in fragments
+        let is_kern_feature = current_feature == KERN;
+        assert!(is_kern_feature || current_feature == DIST);
+        // this is based on the _registerLookups fn, and this is where we should start typing
```
rsheeter, line 506: Stranded thought?

fontbe/src/features/kern.rs
```diff
@@ -359,31 +390,843 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+        let mut default_lookups = Vec::new();
+        if let Some(common_lookups) = lookups_by_script
+            .get(&COMMON_SCRIPT)
+            .filter(|_| is_kern_feature)
+        {
+            log::debug!("found {} default lookups", common_lookups.len());
+            default_lookups.extend(common_lookups.iter().copied());
+        }
+
+        //inDesign bugfix:
```
rsheeter, line 540: I assume copied from python, suggest stating it explicitly=

fontbe/src/features/kern.rs
```diff
@@ -359,31 +390,843 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+            default_lookups.extend(rtl_lookups);
         }
-        context.fea_rs_kerns.set(kerns);
+        default_lookups.sort_unstable();
 
-        Ok(())
+        let mut features = HashMap::new();
+        if !default_lookups.is_empty() {
+            let languages = fea_langs_by_script.get(&DFLT_SCRIPT).unwrap_or(&dflt_langs);
+            log::debug!("DFLT languages: {languages:?}");
```
rsheeter, line 565: Suggest trace

fontbe/src/features/kern.rs
```diff
@@ -359,31 +390,843 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+            (None, None) => Vec::new(),
+        };
+
+        for (script, mut lookups) in lookups_by_script {
+            lookups.extend(dflt_lookups.iter().copied());
+
+            for tag in super::properties::script_to_ot_tags(&script) {
+                let languages = fea_langs_by_script.get(&tag).unwrap_or(&dflt_langs);
+                for lang in languages {
+                    log::info!("added {} lookups for {tag}/{lang}", lookups.len());
```
rsheeter, line 593: Suggest debug

fontbe/src/features/kern.rs
```diff
@@ -359,31 +390,843 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+                }
+            }
+        }
+
+        debug_ordered_lookups(&features, &ordered_lookups);
+        (ordered_lookups, features)
+    }
+}
+
+fn debug_ordered_lookups(
```
rsheeter, line 604: I think maybe trace?

fontbe/src/features/kern.rs
```diff
@@ -359,31 +390,843 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+    /// map of all mark glyphs; bool indicates if mark is spacing
+    mark_glyphs: HashMap<GlyphId, MarkSpacing>,
+    glyph_scripts: HashMap<GlyphId, HashSet<UnicodeShortName>>,
+    bidi_glyphs: HashMap<BidiClass, HashSet<GlyphId>>,
+    opts: KernSplitOptions,
+    dflt_scripts: HashSet<UnicodeShortName>,
+    common_scripts: HashSet<UnicodeShortName>,
+}
+
+// unused for now, but included so that our impl can more closely follow fonttools
```
rsheeter, line 632: feels like a link to fonttools might want to go here?

fontbe/src/features/kern.rs
```diff
@@ -359,31 +390,843 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+                kerning_per_script
+                    .entry(scripts)
+                    .or_insert(Vec::new())
+                    .push(pair);
+            }
+        }
+
+        kerning_per_script = merge_scripts(kerning_per_script);
+        for scripts in kerning_per_script.keys().filter(|x| x.len() > 1) {
+            log::info!("merging kerning lookups for {scripts:?}");
```
rsheeter, line 781: debug

fontbe/src/features/kern.rs
```diff
@@ -359,31 +390,843 @@ impl Work<Context, AnyWorkId, Error> for KerningGatherWork {
...
+            bases.subtables.iter().map(|sub| sub.len()).sum::<usize>(),
+            2
+        );
+
+        let marks = &lookups[1];
+        assert_eq!(marks.flags, LookupFlag::empty());
+        assert_eq!(
+            marks.subtables.iter().map(|sub| sub.len()).sum::<usize>(),
+            1
+        );
```
rsheeter, line 1202: This looks like it could be one assert and thus give more context when it fails. 

```rust
assert_eq!(
  vec![(ignore_marks, 2), (LookupFlag::empty())], lookups.iter()...etc);
```

mark_to_base_only below seems similar, so much so you could hoist out a helper to check you got one lookup with these flags and sizes.

fontbe/src/features/properties.rs
```diff
@@ -11,12 +12,75 @@ use write_fonts::{
     types::{GlyphId, Tag},
 };
 
+// SAFETY: we can easily verify that neither of these strings contains a nul byte.
+// (this is the only way we can declare these as constants)
+pub const COMMON_SCRIPT: UnicodeShortName =
+    unsafe { UnicodeShortName::from_bytes_unchecked(*b"Zyyy") };
+pub const INHERITED_SCRIPT: UnicodeShortName =
+    unsafe { UnicodeShortName::from_bytes_unchecked(*b"Zinh") };
```
rsheeter, line 20: Having just gone through considerable delay as a result of unsafe I would very much like to avoid the unsafe.

Perhaps a function with a static OnceCell hidden inside would suffice?

fontbe/src/features/properties.rs
```diff
@@ -11,12 +12,75 @@ use write_fonts::{
...
+impl CharMap for Vec<(Arc<Glyph>, GlyphId)> {
+    fn iter_glyphs(&self) -> impl Iterator<Item = (GlyphId, u32)> {
+        self.iter()
+            .flat_map(|(glyph, gid)| glyph.codepoints.iter().map(|uv| (*gid, *uv)))
+    }
+}
+
+impl ScriptDirection {
+    /// Returns the writing direction for the provided script
+    pub(crate) fn for_script(script: &UnicodeShortName) -> Self {
```
rsheeter, line 59: Is there a reference for where the RTL list comes from?

fontbe/src/features/properties.rs
```diff
@@ -103,22 +190,37 @@ pub(crate) fn scripts_by_glyph(
...
+        |codepoint| unicode_bidi_type(codepoint).into_iter(),
         gsub,
     )
 }
+
+const DFLT_SCRIPT: Tag = Tag::new(b"DFLT");
+
 static SCRIPT_ALIASES: &[(Tag, Tag)] = &[(Tag::new(b"jamo"), Tag::new(b"hang"))];
 
+static SCRIPT_EXCEPTIONS: &[(&str, Tag)] = &[
```
rsheeter, line 207: Reference to data source?

fontbe/src/features/properties.rs
```diff
@@ -134,6 +236,122 @@ static NEW_SCRIPTS: &[(Tag, Script)] = &[
     (Tag::new(b"tml2"), Script::Tamil),
 ];
 
+// I don't know what's going on here, just copying OTTags.py
```
rsheeter, line 239: Plz permalink so we can see the exact version of the file this was taken from.

fontbe/src/features/properties.rs
```diff
@@ -194,6 +412,21 @@ fn ot_tag_to_unicode_short_name(tag: Tag) -> UnicodeShortName {
     UnicodeShortName::try_from_raw(out).expect("cannot fail, as tag cannot have leading nul byte")
 }
 
+/// a script can correspond to one or two tags, because
```
rsheeter, line 415: missing words?

rsheeter (commented):
The Rust largely LGTM. I think @anthrotype better have a look before we merge as he has greater understanding of the nuances of the Python.

fontbe/src/features/properties.rs
cmyr: this isn't in fontations though, do we have similar concerns with fontc?

fontbe/src/features/kern.rs
cmyr: what this comment is saying is that if we panic here (because GSUB fails to pack) it's a case where we are already going to panic, because the way contexts work we unwrap this when we save it to the context.

fontbe/src/features/kern.rs
cmyr: this was stale.

fea-rs/src/compile/lookups.rs
cmyr: my feeling is that at this point the types would be more noise than help, but if it ends up being a problem we can always go that route.

fontbe/src/features/properties.rs
rsheeter: If this is temporary it's fine to file an issue for it and submit. I wouldn't expect to put fontc in google3 *soon*.

fontbe/src/features/kern.rs
```diff
@@ -585,7 +619,8 @@ fn debug_ordered_lookups(
 
 /// All the state needed for splitting kern pairs by script & writing direction
 struct KernSplitContext {
-    gdef: Option<HashMap<GlyphId, GlyphClassDef>>,
+    /// map of all mark glyphs; bool indicates if mark is spacing
```
anthrotype, line 622: the MarkSpacing enum is not actually a "bool" though, is it? Should it be? maybe you changed it to enum and the comment is stale

anthrotype (approved):

fontbe/src/features/kern.rs
cmyr: your deduction is correct 🕵️

PR #734, opened 2024-03-14
[fontc] update skrifa test to new outline API

dfrg:
We recently removed the old scale module. Updating here now so this doesn't cause unnecessary breakage when we publish.

JMM


cmyr (approved):
👍

PR #736, opened 2024-03-18
[kern] Use LookupFlag constants

cmyr:
These were not available when this code was first written.

based on #735


anthrotype (approved):
much nicer, lgtm

PR #738, opened 2024-03-19
[ttx_diff] Make generated fonts easier to find

cmyr:
Previously the fontc output would live in `build/{compare}/font.ttf`, and the fontmake output would live in `build/{compare}/{source}`, which was unintuitive.

Now fontc write `build/{compare}/fontc.ttf` and fontmake writes `build/{compare}/fontmake.ttf`.

- this requires #737

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -136,12 +136,14 @@ def build_fontc(source: Path, build_dir: Path, compare: str):
         "--no-production-names",
         "--build-dir",
         ".",
+        "-o",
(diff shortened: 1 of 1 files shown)


dfrg (approved):
Not sure if this breaks any testing scripts that may live out of repo but the change makes sense to me.

cmyr:
> Not sure if this breaks any testing scripts that may live out of repo but the change makes sense to me.

it shouldn't, this change only changes the behaviour of `ttx_diff.py`, not the compiler itself.

Issue #457, opened 2023-09-25
should raise an error when multiple glyphs are mapped to the same unicode codepoint


anthrotype:
fontmake (via ufo2ft) raises an error if the source file if the same unicode codepoint maps to more than one glyphs, as it becomes ambiguous/undefined which character-to-glyph mapping should prevail when building a cmap table:

https://github.com/googlefonts/ufo2ft/blob/6787e37e63530ecf46ea11fb5f5bddc00224d3ec/Lib/ufo2ft/util.py#L221-L243

the issue arises from the fact that, in source formats, unicodes list is a property of the glyphs, so one could have, e.g. both "foo": [0x1234] and "bar": [0x1234], whereas in OpenType cmap it's the reverse and you only have 0x1234: "foo" (or "bar").

we could either raise it in fontbe where we create the mappings that we feed to write-fonts:

https://github.com/googlefonts/fontc/blob/13307caf219550b001b9d2e5a5a355be44bee636/fontbe/src/cmap.rs#L42-L57

or we could make write-fonts' Cmap::from_mappings fallible and return a Result; right now it never fails (it may panic until #451 gets fixed).

rictic:
In updating fontbe to use the fallible Cmap::from_mappings I'm seeing a failure in the test `resolve_contour_and_composite_glyph_in_non_legacy_mode` which is processing `glyphs2/MixedContourComponent.glyphs`. It looks like it maps the codepoint for comma (44) to both gid 2 and gid 3.

Is this an invalid .glyphs file or is the new validity check over-eager?

<details>
  <summary>Expand for debug output of the decoded glyphs as the fontbe CmapWork sees them</summary>

```
[fontbe/src/cmap.rs:45:17] gid = 0
[fontbe/src/cmap.rs:45:17] &glyph = Glyph {
    name: .notdef,
    emit_to_binary: true,
    codepoints: {},
    default_location: Normalized {XXXX: 0.00, wdth: 0.00, wght: 0.00},
    sources: {
        Normalized {XXXX: 0.00, wdth: 0.00, wght: 0.00}: GlyphInstance {
            width: 500.0,
            height: None,
            contours: [
                BezPath(
                    [
                        MoveTo(
                            (50.0, -200.0),
                        ),
                        LineTo(
                            (450.0, -200.0),
                        ),
                        LineTo(
                            (450.0, 800.0),
                        ),
                        LineTo(
                            (50.0, 800.0),
                        ),
                        LineTo(
[...]

PR #742, opened 2024-03-20
Update to write-fonts 0.25.0, handling the new cmap conflict

rictic:
Fixes https://github.com/googlefonts/fontc/issues/457

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -62,6 +62,7 @@ fn split_glyph(glyph_order: &GlyphOrder, original: &Glyph) -> Result<(Glyph, Gly
     simple_glyph.sources.iter_mut().for_each(|(_, inst)| {
         inst.components.clear();
     });
+    simple_glyph.codepoints.clear();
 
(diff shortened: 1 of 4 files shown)


fontir/src/glyph.rs
```diff
@@ -62,6 +62,7 @@ fn split_glyph(glyph_order: &GlyphOrder, original: &Glyph) -> Result<(Glyph, Gly
     simple_glyph.sources.iter_mut().for_each(|(_, inst)| {
         inst.components.clear();
     });
+    simple_glyph.codepoints.clear();
```
rictic, line 65: See https://github.com/googlefonts/fontc/issues/457#issuecomment-2008651486

rsheeter (approved):
LGTM

fontir/src/glyph.rs
anthrotype: thanks for spotting this! turns out not raising an error in case of cmap conflicts was actually hiding another bug :)

PR #745, opened 2024-03-20
[fontbe] Fixup errors for variable metrics

cmyr:
There were a few giant FIXMEs in the resolve_variable_metric fn; if memory serves I think I'd initially written that patch as "just a sketch" and then it was merged to unblock other work.

In any case, this rationalizes things a bit, adding error variants for the case where we can't compute deltas for anchors & kerns.

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -23,6 +23,7 @@ use fea_rs::{
 use fontir::{
     ir::{FeaturesSource, GlyphOrder, StaticMetadata},
     orchestration::{Flags, WorkId as FeWorkId},
+    variations::DeltaError,
 };
 
 use fontdrasil::{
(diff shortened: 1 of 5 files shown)


fontbe/src/features.rs
```diff
@@ -182,10 +182,15 @@ pub(crate) fn resolve_variable_metric<'a>(
         // Array of region axis coordinates records, in the order of axes given in the 'fvar' table.
         let mut region_axes = Vec::with_capacity(static_metadata.axes.len());
         for axis in static_metadata.axes.iter() {
-            let Some(tent) = region.get(&axis.tag) else {
-                todo!("FIXME: add this error conversion!")
+            let coords = match region.get(&axis.tag) {
+                Some(tent) => tent.to_region_axis_coords(),
+                None => {
+                    // it would be nice if this was all validated before now
+                    log::warn!("unknown axis '{}' in region", axis.tag);
```
rsheeter, line 189: why isn't this an error?

rsheeter (approved):

fontbe/src/features.rs
cmyr: This felt like an expedient compromise:

- adding a variant to the `DeltaError` that we're returning doesn't seem right, since this is technically unrelated
- adding a new error type just for this function feels like overkill, especially because:
- how would this error condition ever occur? The location here is the location for a given master, which was retrieved (afaict) from the static metadata. Presumably when we are generating the IR we ensure that the locations for various masters are defined in a common axis-space. The locations are all reused for all sets of deltas. At no point do we conjure up new locations.

realistically, it feels like this should maybe be a fatal error, since it probably represents a major inconsistency in our inputs?

anthrotype: I think you can unwrap here

rsheeter: +1, unwrap (if it's definitely meant to be impossible) or error, a warning log for a condition that should never occur doesn't seem right.

PR #748, opened 2024-03-21
Handle multiple codepoints in .fontra files, as well as axes with a values array instead of min and max

### LINKED ISSUE #696: Support multiple codepoints in .fontra inputs
They are comma separated, see https://github.com/googlefonts/fontra/blob/a4edd06837118e583804fd963c22ed806a315b04/src/fontra/backends/fontra.py#L171-L181. 

https://github.com/googlefonts/fontc/blob/main/resources/testdata/fontra/2glyphs.fontra/glyph-info.csv shows a single-codepoint example.

rictic:
Fixes #696

Structured for easy review if you go commit by commit

To test this I pulled in the MutatorSans test font from the fontra repo, and as part of parsing it I handled another case in our fontra parsing. A `font-data.json` file might define an axis without a minValue or maxValue field, but instead a vector of values. e.g. MutatorSans defines its italic axis as:

```json
{
  "name": "italic",
  "label": "italic",
  "tag": "ital",
  "values": [0, 1],
  "defaultValue": 0
}
```


anthrotype:
>  A font-data.json file might define an axis without a minValue or maxValue field, but instead a vector of values.

I think you stumbled on a distinction between a regular variable axis, on the one hand, and a so-called "discrete" axis, see:

https://github.com/googlefonts/fontra/blob/1c330c3e4243611191d9999dd6b4af37cca9daff/src/fontra/core/classes.py#L104-L126

The two are not interchangeable, i'm afraid. I think fontra borrowed the concept from DesignSpace format v5:

https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#axis-element

We don't have those in our fontir IR yet. We will need to support them at some point when we add full suport for DSv5 and, now, fontra.

anthrotype:
e.g. for the MutatorSans in your example above, the `"values": [0, 1],` means that italic axis can only either be 0 or 1, but not something in between.

anthrotype:
I think for the time being we could raise an error that discrete axes are not implemented in fontc.
We should file a separate issue for those.

One major complication that these discrete axes add is that one can no longer assume there is a 1:1 mapping between 1 input source (e.g. a .fontra or .designspace or .glyphs file) on the one hand, and only _one_ variable font. Once you have a discrete axes, you immeditaley have a 1:N situation, i.e. more than one VFs that can be generated from a given source.

A VF can define continuously variable axes (the fvar kind), discrete axes was a way to allow grouping multiple, related VFs in the same source file (e.g. to allow defining shared attributes like STAT)

fontra2fontir/src/source.rs
```diff
@@ -244,24 +247,127 @@ mod tests {
...
+        source.inputs().unwrap();
         assert_eq!(
-            vec![GlyphName::new(".notdef"), GlyphName::new("u20089")],
-            glyph_names
+            Arc::new(
+                [
+                    (
+                        GlyphName::new(".notdef"),
+                        (
+                            "../resources/testdata/fontra/2glyphs.fontra/glyphs/%2Enotdef.json"
```
rsheeter, line 281: Nit: testdata_dir().join(...) ? - here and below

fontra2fontir/src/source.rs
```diff
@@ -244,24 +247,127 @@ mod tests {
...
+                ]
+                .into_iter()
+                .collect::<BTreeMap<_, _>>()
+            ),
+            source.glyph_info
         );
     }
+
+    #[test]
+    fn glyph_info_of_mutator_sans() {
```
rsheeter, line 302: The name of the test isn't tremendously informative wrt what it validates. Based on the PR title the goal is to confirm support for 0..N codepoints mapped to a glyph. I suggest validating that a bit more minimally.

For example, rather than adding the entire MutatorSans and checking tons of glyphs add a new test font that has the absolute minimum to verify what you wish to test. Perhaps a new manually created font that has say .notdef (no codepoints), A (1 codepoint), Aaacute (2 codepoints) and some emoji that has lots (say https://emojipedia.org/handshake-medium-light-skin-tone-medium-skin-tone#technical).

fontra2fontir/src/fontra.rs
```diff
@@ -351,6 +415,10 @@ mod tests {
         font_data
             .axes
             .iter()
+            .map(|a| match a {
+                FontraAxis::Continuous(a) => a,
+                FontraAxis::Discrete(a) => panic!("Unexpected discrete axis: {a:#?}"),
```
rsheeter, line 420: I don't think it's worth fixing here but I think the Continuous/Discrete and multiple codepoints could have been two separate PRs.

fontra2fontir/src/fontra.rs
```diff
@@ -141,19 +141,83 @@ impl FontraFontData {
...
+    }
+
+    pub(crate) fn hidden(&self) -> bool {
+        match self {
+            FontraAxis::Continuous(a) => a.hidden,
+            FontraAxis::Discrete(a) => a.hidden,
+        }
+    }
+
+    pub(crate) fn mapping(&self) -> &Vec<[f64; 2]> {
```
rsheeter, line 180: This I think warrants a comment on how to interpret each pair, e.g. interpret as [user, design] forming a piecewise linear map

fontra2fontir/src/toir.rs
```diff
@@ -22,7 +22,14 @@ pub(crate) fn to_ir_static_metadata(
     let axes = font_data
         .axes
         .iter()
+        .map(|a| match a {
+            crate::fontra::FontraAxis::Discrete(_) => Err(WorkError::UnsupportedConstruct(
```
rsheeter, line 26: Suggest adding a minimal example of a fontra font with discrete axes to testdata and a test to prove we reject it.

rsheeter (approved):
LGTM but request minimization of testdata or an explanation of why that would be undesireable. Edit: minimized.

fontra2fontir/src/source.rs
rsheeter: https://github.com/googlefonts/fontc/pull/754 (PR to this branch) aspires to so update

fontra2fontir/src/source.rs
rsheeter: Fixed by #754

PR #749, opened 2024-03-21
[kerning] Delay decomposing kern rules

cmyr:
This reworks how we do kerning in order to preserve information about whether rules included a glyph class until we are ready to do the final compilation.

Preserving this information lets us match the behaviour of FEA in ordering these rules, where we put (glyph, glyph) first, followed by (glyph, class), (class, glyph) and finally (class, class).

---

With this patch `ttx_diff` is producing identical output to oswald for generated kerning.

### DIFF
--- a/fea-rs/src/common/glyph_class.rs
+++ b/fea-rs/src/common/glyph_class.rs
@@ -36,13 +36,11 @@ pub(crate) struct GlyphClass(Vec<GlyphId>);
 pub struct GlyphSet(Vec<GlyphId>);
 
 impl GlyphClass {
+    /// An empty glyph class
+    pub const EMPTY: Self = GlyphClass(Vec::new());
+
     pub(crate) fn items(&self) -> &[GlyphId] {
         &self.0
-    }
-
-    /// Return a new, empty glyph class
-    pub fn empty() -> Self {
-        Self(Default::default())
     }
 
     /// Return a `GlyphSet` containing the unique glyphs in this class.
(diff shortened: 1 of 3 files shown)


fea-rs/src/common/glyph_class.rs
```diff
@@ -60,6 +60,9 @@ impl GlyphClass {
 }
 
 impl GlyphSet {
+    /// The empty glyph set
+    pub const EMPTY: Self = GlyphSet(Vec::new());
```
rsheeter, line 64: GlyphClass has empty a few lines above. It's a bit weird they are inconsistent about this. Perhaps we should standardize on this style since it provides for const?

fontbe/src/features/kern.rs
```diff
@@ -156,7 +158,7 @@ impl Work<Context, AnyWorkId, Error> for GatherIrKerningWork {
 
         align_kerning(&ir_groups, &mut kern_by_pos);
         // Use a BTreeMap  because it seems the order we process pairs matters. Maybe we should sort instead...?
```
rsheeter, line 160: Can this comment be updated now?

fontbe/src/features/kern.rs
```diff
@@ -156,7 +158,7 @@ impl Work<Context, AnyWorkId, Error> for GatherIrKerningWork {
...
-        let mut adjustments: BTreeMap<KernPair, KernAdjustments> = Default::default();
+        let mut adjustments: BTreeMap<IrKernPair, KernAdjustments> = Default::default();
```
rsheeter, line 161: Optional IIRC we've chosen `ir::KernPair` over aliasing in other places.

rsheeter (approved):

PR #758, opened 2024-03-28
remove the GlyphName defined in fea-rs, and use the one in fontdrasil

qxliu76:
for #757

### DIFF
--- a/fea-rs/src/common/glyph_map.rs
+++ b/fea-rs/src/common/glyph_map.rs
@@ -1,6 +1,7 @@
 use write_fonts::tables::post::Post;
 
-use super::{GlyphId, GlyphIdent, GlyphName};
+use super::{GlyphId, GlyphIdent};
+use fontdrasil::types::GlyphName;
 use std::{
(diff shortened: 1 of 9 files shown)


rsheeter (approved):

fea-rs/src/common/glyph_map.rs
```diff
@@ -158,9 +161,15 @@ mod sealed {
         }
     }
 
+    impl AsGlyphIdent for SmolStr {
+        fn named(&self) -> Option<&str> {
+            Some(self.as_str())
+        }
+    }
+
```
cmyr, line 169: I'm not sure why this is necessary?

cmyr (commented):
one Q inline, overall looks good!

fea-rs/src/common/glyph_map.rs
cmyr: ok I see what breaks when I remove it, LGTM :)

qxliu76:
Fixes #757

PR #760, opened 2024-03-28
[marks] Flesh out mark anchor name parsing

cmyr:
This tries to match the more complicated behaviour of ufo2ft, which performs additional validation and also handles the convention of numbering the anchors for ligature components.

As this parsing can now fail, this PR includes a bunch of additonal error handling code. This begs the question of whether we could be doing this parsing earlier, during the conversion to IR; that would mean we only need to deal with this error in one place, which would let us simplify subsequent code.

Ir might also be a nice place for this since there are possibly differences in naming conventions between ufo and glyphs.app.


anthrotype:
any idea why the ./resources/scripts/repeatable-builds.sh is failing on CI?

anthrotype:
not sure if related to this PR or not, but I see the GPOS table length changing from 1300286 (first build) to 1300324 (second build) of repeatable-builds.sh

cmyr:
this is superseded by #762

PR #762, opened 2024-04-02
Move anchor name parsing into fontir

cmyr:
This is shared between frontends, and doing this in ir simplifies backend code.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -21,11 +21,11 @@ use write_fonts::{
 
 use fontdrasil::{
     coords::{NormalizedCoord, NormalizedLocation, UserLocation},
-    types::{AnchorName, Axis, GlyphName},
+    types::{Axis, GlyphName},
 };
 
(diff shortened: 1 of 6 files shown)


fontir/src/ir.rs
```diff
@@ -991,14 +991,82 @@ impl GlyphAnchors {
...
+    }
 }
 
 /// A variable definition of an anchor.
 ///
 /// Must have at least one definition, at the default location.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct Anchor {
-    pub name: AnchorName,
+    pub type_: AnchorType,
```
anthrotype, line 1069: how about `kind` to avoid the reserved `type` name and the need for `_`?

anthrotype (approved):

fontir/src/ir.rs
cmyr: good call, done :)

PR #764, opened 2024-04-02
[marks] Move marks logic into a context struct

cmyr:
This patch just matches the existing behaviour, but it will be improved to match the behaviour of fontmake.

no functional change, diffs are all the same etc; just thought this would be a good checkpoint that will reduce the size of subsequent diffs.


dfrg (approved):
Love the changes here. The new code is much easier to read and almost trivial to follow step by step with the comments.

PR #765, opened 2024-04-03
[g2f] Use smolstr when parsing glyphs.app data

cmyr:
Particularly for types that end up being SmolStr later on, like glyph names; it's inconvenient and inefficient to parse to String and then later convert to SmolStr, so this is a minor simplification.

Arguably the whole plist parsing process should be using smolstr, which should offer a reasonable speedup, but we can worry about that when we're thinking more generally about optimization.

----------

...looking at this again, maybe it would make sense to use `GlyphName` whenever we're dealing with a glyph name? that's the type we're going to want later anyway, and it would save us from needing to convert things later...

### DIFF
--- a/glyphs-reader/Cargo.toml
+++ b/glyphs-reader/Cargo.toml
@@ -20,6 +20,7 @@ env_logger.workspace = true
 regex.workspace = true
 
 chrono.workspace = true
+smol_str.workspace = true
 
 [dev-dependencies]
 pretty_assertions.workspace = true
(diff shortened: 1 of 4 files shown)


rsheeter:
Is the impact detectable (https://github.com/googlefonts/fontc?tab=readme-ov-file#comparing-branch-performance-with-hyperfine)? - happy to land either regardless, just curious.

rsheeter (approved):

cmyr:
no it's not... if there's any benefit I think you'd need to have a parsing-specific benchmark for it to show up. Mostly about ergonomics in any case. :)

PR #768, opened 2024-04-05
Add fontdrasil::glyph_data module and codegen

cmyr:
This is based on the non_spacing_marks.py script that is being used in glyphs2fontir.

In order to properly generate mark positioning lookups, I need access to the categories and subcategories that glyphs.app assigns to glyphs. This patch adds a new module to fontdrasil that exposes this data. Currently this is limited to mapping from glyph names to category information, but it would be trivial to update this to also provide conversion from unicode values to glyph names (and thus from unicode values to categories).

conclusions:

- not using build.rs was helpful in this case, because it's nice to have `glyphsLib` be the source of truth,
- *but* this means we end up having 90k lines of pretty gross generated code in the repo, which I don't love.
- I couldn't benchmark this against a version using a naive match statement, since that version... wouldn't compile.


cmyr:
So this is now working but I am seriously reconsidering whether or not it should live in fontdrasil; I was thinking that it would be used in the shared anchor propagation implementation, but I think it will continue to only be used in glyphs2fontir.

PR #771, opened 2024-04-09
[g2fir] Handle magic '*origin' anchor

cmyr:
I think this is the most principaled approach; we apply this delta in the conversion to IR, and drop the anchor then.

cc @anthrotype, does this seem reasonable to you? If yes let's try and figure out what a test case looks like?

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -12,7 +12,7 @@ use std::str::FromStr;
 use std::{fs, path};
 
 use crate::plist::FromPlist;
-use kurbo::{Affine, Point};
+use kurbo::{Affine, Point, Vec2};
 use log::{debug, warn};
 use ordered_float::OrderedFloat;
 use plist_derive::FromPlist;
@@ -721,6 +721,28 @@ pub struct RawAnchor {
 pub struct Anchor {
     pub name: SmolStr,
     pub pos: Point,
+}
+
+impl Anchor {
+    // an anchor with this name on a mark glyph is used to adjust positions of other anchors
+    const SPECIAL_ORIGIN_NAME: &'static str = "*origin";
+
+    /// If this anchor is the special '*origin' anchor, return its value as a `Vec2`
+    pub fn special_origin_adjustment(&self) -> Option<Vec2> {
+        if self.name == Self::SPECIAL_ORIGIN_NAME {
+            Some(self.pos.to_vec2())
+        } else {
+            None
+        }
+    }
+
+    /// Returns `true` if this anchor has special semantic meaning
+    ///
+    /// Anchors with not be included in IR.
+    pub fn is_special(&self) -> bool {
+        //TODO: figure out what other names we should be ignoring/handling?
+        self.name == Self::SPECIAL_ORIGIN_NAME
+    }
 }
 
 impl Hash for Anchor {
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -844,8 +844,7 @@ impl Work<Context, WorkId, WorkError> for GlyphIrWork {
             });
         }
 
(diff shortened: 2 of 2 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -845,7 +845,7 @@ impl Work<Context, WorkId, WorkError> for GlyphIrWork {
         }
 
         // https://github.com/googlefonts/fontmake-rs/issues/285 glyphs non-spacing marks are 0-width
```
rsheeter, line 847: I think you should move the comment to the `if !is_nonspacing` if you rename it so

glyphs-reader/src/font.rs
```diff
@@ -723,6 +723,28 @@ pub struct Anchor {
     pub pos: Point,
 }
 
+impl Anchor {
+    // an anchor with this name on a mark glyph is used to adjust positions of other anchors
+    const MAGIC_ORIGIN_NAME: &'static str = "*origin";
+
+    /// If this anchor is the magic '*origin' anchor, return its value as a `Vec2`
+    pub fn magic_origin_adjustment(&self) -> Option<Vec2> {
```
rsheeter, line 731: Nit: "magic" isn't really adding anything, to me it reads a bit disparaging

glyphs-reader/src/font.rs
```diff
@@ -723,6 +723,28 @@ pub struct Anchor {
...
+        }
+    }
+
+    /// Returns `true` if this anchor has special semantic meaning
+    ///
+    /// Anchors with not be included in IR.
+    pub fn is_special(&self) -> bool {
+        //TODO: figure out what other names we should be ignoring/handling?
+        self.name == Self::MAGIC_ORIGIN_NAME
+    }
```
rsheeter, line 745: I think we could better encapsulate this. For example, 

```rust
/// If Some, this isn't an anchor itself, rather an adjustment to other anchors
pub fn anchor_adjustment(&self) -> Option<Vec2> {
  // if named origin return our pos as vec2
}
```

source.rs would then just reduce anchors => a single adjustment (if any) and apply it.

rsheeter:
> If yes let's try and figure out what a test case looks like?

Naively this seems readily unit testable in `glyphs2fontir/src/source.rs`, is there a barrier to testing I'm missing?

glyphs-reader/src/font.rs
cmyr: This is separate because I believe there are going to be additional special anchors that we will need to skip.

cmyr:
> > If yes let's try and figure out what a test case looks like?
> 
> Naively this seems readily unit testable in `glyphs2fontir/src/source.rs`, is there a barrier to testing I'm missing?

I want an end-to-end test that goes from 'source including *origin anchor' to 'mark compiled at correct position', and I want to be able to validate this against what glyphs is compiling for some sample source. Last we spoke cosimo was working on putting something like this together.

glyphs-reader/src/font.rs
rsheeter: I think this might feel better if Anchor was an enum then. We already have two distinct cases, Origin and whatever you want to call the kind that actually flow into the output as anchors. We also already have RawAnchor => Anchor so it fits to classify them in that process.

rsheeter:
> I want an end-to-end test

Agreed, but I don't think we need to block this PR on it. We could unit test the funcitonality here and then update it if the full compile test reveals problems.

cmyr:
So after conversation with @anthrotype, I'm going to close this; I don't think it's the correct approach.


- the presence of this anchor likely means we need to adjust other aspects of the glyph, like potentially shifting all points.
- these anchors are not used in any of our fonts, and are not supported in glyphsLib or fontmake, and we don't totally know what they're supposed to do
- so it feels premature to try and handle them here; we agreed to defer this for the time being and focus on matching the current behaviour of the python toolchain

PR #772, opened 2024-04-09
[fontc] Simplify timing color code

cmyr:
I was in here to add a new IR type and thought this could be cleaned up.

In addition to the cleanup, a tiny additional benefit of this change is that gray is now the default, and you don't need to add an explicit case anytime a new workid is added.

JMM


fontc/src/timing.rs
```diff
@@ -167,46 +167,13 @@ fn short_name(id: &AnyWorkId) -> &'static str {
...
-        AnyWorkId::Be(BeWorkIdentifier::Gdef) => "gray",
-        AnyWorkId::Be(BeWorkIdentifier::Glyf) => "gray",
         AnyWorkId::Be(BeWorkIdentifier::GlyfFragment(..)) => "#00c1c9",
-        AnyWorkId::Be(BeWorkIdentifier::Gpos) => "gray",
-        AnyWorkId::Be(BeWorkIdentifier::Gsub) => "gray",
-        AnyWorkId::Be(BeWorkIdentifier::Gvar) => "gray",
         AnyWorkId::Be(BeWorkIdentifier::GvarFragment(..)) => "#008786",
-        AnyWorkId::Be(BeWorkIdentifier::Head) => "gray",
-        AnyWorkId::Be(BeWorkIdentifier::Hhea) => "gray",
         AnyWorkId::Be(BeWorkIdentifier::Hmtx) => "v",
```
rsheeter, line 174: = "v" seems like it might be a mistake, maybe we should delete that one

rsheeter (approved):

Issue #766, opened 2024-04-04
External GlyphData.xml files


simoncozens:
Some fonts in the GF catalogue expect to use external GlyphData.xml files to determine categories and production names for glyphs with well-known names. Is this in scope for fontc? (Alternately, we can require that babelfont preprocesses such Glyphs files to bake the data into the sources.)

rsheeter:
Which specific fonts use this and what would the impact of not having it be?

simoncozens:
I know at least the Noto Arabic fonts use it. I can't tell which other fonts use it. Not having it means glyph categories are wrong and shaping breaks.

PR #773, opened 2024-04-12
The continued adventures of GlyphData

cmyr:
This adopts the approach to GlyphData that was finally agreed on in chat:

- the data lives in `glyph-reader`, in the `glyphdata` module.
- the source XML files are checked in, in `glyph-reader/data`.
- there is an update script that gets these XML files from the python `glyphsLib`.
- we use a build.rs script to parse this XML and generate bincode, which we bake into our binary
- we have the ability (not currently exercised, but tested) to load user GlyphData files and merge them into the bundled ones (see #766)
- we use `OnceLock` (the thread-safe `OnceCell`) to initialize a global data structure for querying this data.

None of this is perfectly efficient, but we can make iterative improvements as needed.

- the approach to querying the data is very naive: we have a big vec of per-glyph `GlyphInfo` structs, and then maintain hashmaps of key -> index, where 'key' is glyph name or codepoint. In particular this means that we're not particularly memory efficient, and all the names are duplicated. We could improve on this, but it would require being much fancier.
- we're committing the xml files, and these are not very efficient. do we want to gzip these? they're very compressible (`GlyphData.xml` goes from 4.8M to 460k with a naive gzip)

This PR removes the previous `glyphs2fontir::glyphdata` module and python script, and implements the same functionality using the new data. 

The new version is 2-3x slower than the previous version, but we're still talking about low tens of ns per lookup.


glyphs2fontir/src/source.rs
```diff
@@ -935,6 +934,23 @@ impl Work<Context, WorkId, WorkError> for GlyphIrWork {
...
 
+// this is intended as a drop-in replacement for a previous function of this
+// name. This will eventually need to be replaced with something that can handle
+// custom GlyphData.xml files, as well as handle overrides that are part of the
+// glyph source.
+fn is_nonspacing_mark(codepoints: &HashSet<u32>, name: &GlyphName) -> bool {
+    fn glyph_data() -> &'static GlyphData {
+        static GLYPH_DATA: OnceLock<GlyphData> = OnceLock::new();
+        GLYPH_DATA.get_or_init(|| GlyphData::new(None).unwrap())
+    }
```
rsheeter, line 945: Why do you need a wapper function? - just set data directly to GLYPH_DATA.get_or_init

glyphs2fontir/src/source.rs
```diff
@@ -935,6 +934,23 @@ impl Work<Context, WorkId, WorkError> for GlyphIrWork {
     }
 }
 
+// this is intended as a drop-in replacement for a previous function of this
```
rsheeter, line 937: The first sentence is not going to be very helpful to anyone who didn't read the prior version. Suggest delete or change it to explain what it does.

rsheeter:
This generally seems much better. In IM I think we said we might be able to avoid support for custom glyphdata. If that's true I would be enthused about rewriting the xml in update.py to make it smaller and faster to load.

* definitely not xml
* potentially compressed
* serde-compatible yaml or json come to mind; build.rs would simply decompress and deserialize the text format then serialize bincode

rsheeter (approved):

cmyr:
given that the glyphdata sources are only used to generate the bincode I don't think 'faster to load' is important to us; we're only parsing it once, at build time, so if we want to add an additional processing step now I think it would need to be justified principally by reducing the size of the files we commit, which doesn't feel _critical_.

rsheeter:
> which doesn't feel critical.

I agree that isn't critical and approved the PR as-is on that basis. Smaller files and faster builds is enough I see it as worth the couple hours to implement, but it needn't block, it can go to an issue and be done in a followup.

PR #775, opened 2024-04-15
[glyphs] Store codepoints in glyph struct

cmyr:
Previously we were generating a separate cmap-like thing in the glyphs module, which was then used to set the codepoints in the ir::Glyph. This just keeps the codepoints in the glyphs Glyph, which will be nice because it means that all the things that we use to determine the glyph category will be in one place.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -51,7 +51,6 @@ pub struct Font {
     pub default_master_idx: usize,
     pub glyphs: BTreeMap<SmolStr, Glyph>,
     pub glyph_order: Vec<SmolStr>,
-    pub glyph_to_codepoints: BTreeMap<SmolStr, BTreeSet<u32>>,
     // tag => (user:design) tuples
     pub axis_mappings: RawUserToDesignMapping,
     pub virtual_masters: Vec<BTreeMap<String, OrderedFloat<f64>>>,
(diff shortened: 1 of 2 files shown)


glyphs-reader/src/font.rs
```diff
@@ -1630,23 +1620,26 @@ impl TryFrom<RawLayer> for Layer {
     }
 }
 
-impl TryFrom<RawGlyph> for Glyph {
-    type Error = Error;
-
-    fn try_from(from: RawGlyph) -> Result<Self, Self::Error> {
+impl RawGlyph {
+    // we pass in the radix because it depends on the version, stored in the font struct
+    fn cook(self, codepoint_radix: u32) -> Result<Glyph, Error> {
```
rsheeter, line 1625: I don't find "cook" overly intuitive, `to_glyph` perhaps ?

rsheeter (approved):

glyphs-reader/src/font.rs
cmyr: for non-copy types, clippy expects `to_` methods to take self by reference. Can go with 'build', which is maybe clearer?

fwiw the idea of a 'cook' or 'bake' method to convert a 'raw' type into its non-raw equivalent is something I've seen elsewhere. Given that this is an internal method that's only called in one place I don't think we need to stress too much..

PR #776, opened 2024-04-16
Add test files to reproduce issue #407

anthrotype:
I modified an existing test file that contained intermediate glyph layers. These only contribute sources for the glyph outlines and are expected to be ignored for other global font metrics or kerning.
I added a kerning pair between 'i' and 'i', defined at Regular and Black masters.

The expectaction is that these intermediate locations don't participate in kerning variations; however because we currently use a single VariationModel initialized from the union of all the _glyphs_ locations for generating deltas for the OpenType Layout features (see #407), the regions' min or max for variable kerning are computed incorrectly: in this particular example, instead of expected 0.0 (normalized Regular) and 1.0 (normalzied Black), the support's min is somewhere in between (0.6), which is where the unrelated intermediate layer happens to be, so the kerning will not vary at all between 0.0 and 0.6.

In the recording below, the top font is the incorrect one that fontc builds, the one below is the correct expected one:

https://github.com/googlefonts/fontc/assets/6939968/a26653fb-b2fb-4038-b09a-649d906cbb78

### DIFF
--- a/resources/testdata/designspace_from_glyphs/IntermediateLayer-Black.ufo/kerning.plist
+++ b/resources/testdata/designspace_from_glyphs/IntermediateLayer-Black.ufo/kerning.plist
@@ -1,0 +1,11 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <key>i</key>
+    <dict>
+      <key>i</key>
+      <integer>-400</integer>
+    </dict>
+  </dict>
+</plist>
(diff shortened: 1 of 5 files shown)


rsheeter (approved):

anthrotype:
Note that this issue can only be reproduced when building from .designspace, not when building from .glyphs.
This is an unintented consequence of the fact that in ufo2fontir the StaticMetadata constructor's `glyph_locations` parameter (used to initialise the StaticMetadata.variation_model) is fed with all the source's locations (both the full UFO masters' and the sparse layers' locations); whereas in the case of glyphs2fontir, `glyph_locations` only includes the global font masters' locations, and not also the intermediate (brace) layers' locations. 

https://github.com/googlefonts/fontc/blob/d14fa12dc1a69af1e1b593628f1b1be0ceffba3c/glyphs2fontir/src/source.rs#L351-L355

So it accidentally works in the latter case, however the intention of a parameter named "glyph_locations" doesn't actually match the way it's used

PR #777, opened 2024-04-17
[glyphs] track category info in glyphs

cmyr:
This means parsing the overrides in the source file, and then falling
    back to the glyphdata when needed.

    This also implements additional logic (ported from glyphsLib) for
    determining the category for glyph names that might be alternates or
    ligatures.

This is doing _most_ of what fontools, is, with two notable exceptions:

- fonttools does lookup based on production names, and also spends a bunch of time computing production names. I'm not sure if this is important for our purposes?
- fonttools has a fallback branch in the code that computes categories when they aren't found, and this fallback branch queries the [Adobe Glyph List](https://github.com/adobe-type-tools/agl-aglfn). We currently bundle the shorter version of this in layout-normalizer, and so maybe it's worth just putting this in fontdrasil?


glyphs-reader/src/font.rs
```diff
@@ -1630,20 +1648,65 @@ impl RawGlyph {
...
+            category,
+            sub_category,
         })
     }
 }
 
+// This will eventually need to be replaced with something that can handle
+// custom GlyphData.xml files, as well as handle overrides that are part of the
+// glyph source.
+fn get_glyph_category(name: &str, codepoints: &BTreeSet<u32>) -> Option<(Category, Subcategory)> {
```
rsheeter, line 1703: or maybe link to the py equivalent here?

glyphs-reader/src/glyphdata.rs
```diff
@@ -73,10 +76,48 @@ impl GlyphData {
...
+            codepoints
+                .into_iter()
+                .flat_map(|cps| cps.iter())
+                .find_map(|cp| self.get_by_codepoint(*cp))
+        }) {
+            return Some(Cow::Borrowed(info));
+        }
+
+        // we don't have info for this glyph: can we synthesize it?
+        // TODO: python does production name here. do we care?
```
rsheeter, line 101: suggest TODO(#xxx) referencing the prod names issue

glyphs-reader/src/glyphdata.rs
```diff
@@ -85,11 +126,170 @@ impl GlyphData {
...
-    pub fn get_by_codepoint(&self, codepoint: u32) -> Option<&GlyphInfo> {
+    fn get_by_codepoint(&self, codepoint: u32) -> Option<&GlyphInfo> {
         self.unicode_map
             .get(&codepoint)
             .and_then(|idx| self.data.get(*idx as usize))
     }
+
+    // https://github.com/googlefonts/glyphsLib/blob/e2ebf5b517d/Lib/glyphsLib/glyphdata.py#L199
+    fn construct_category(&self, name: &str) -> Option<(Category, Subcategory)> {
+        if name.starts_with('_') {
```
rsheeter, line 137: a few comments would help. I know I can read the python but ... ideally I wouldn't have to. E.g. what is the signiicance of starting with _? Maybe we could have a function, if is_something(name) would be clearer.

glyphs-reader/src/glyphdata.rs
```diff
@@ -85,11 +126,170 @@ impl GlyphData {
...
+        }
+        let base_name = self
+            .split_glyph_name(name)
+            .map(|(base, _)| base)
+            .unwrap_or(name);
+        if let Some(info) = self.get_by_name(base_name) {
+            return Some((info.category, info.subcategory));
+        }
+
+        if let Some(base_names) = self.split_ligature_glyph_name(base_name) {
```
rsheeter, line 148: This block is a touch hard to follow at a glance, a wee comment explaining what's happening would be welcome.

Seems almost worthy of it's own function.

Indeed ... I wonder if several of these blocks could be functions returning Option and this whole function becomes an unwrap_or_else chain?

rsheeter (approved):

glyphs-reader/src/glyphdata.rs
anthrotype: glyph names that start with an underscore are usually used as "parts" to be combined in other composite glyphs, and are set as non-export by default, see https://glyphsapp.com/learn/smart-components, section "Setting up a Smart Glyph":

> enter a name that starts with _part, because it is a part we can reuse in other glyphs. Also, glyphs created with an underscore at the beginning will be set to non-exporting by default, which makes sense for us.

glyphs-reader/src/glyphdata.rs
cmyr: a lot of this is about me wanting to try and fairly closely follow the source I'm working from.

I tried refactoring this block into a separate function but then the overall control flow feels harder to read, because you need to jump down to read that block before continuing. The bit below can go into a separate function, though!

glyphs-reader/src/font.rs
```diff
@@ -1630,20 +1648,65 @@ impl RawGlyph {
             }
             instances.push(layer.try_into()?);
         }
+        // if category/subcategory were set in the source, we keep them;
+        // otherwise we look them up based on the bundled GlyphData.
+        // (we use this info later to determine GDEF categories, zero the width
+        // on non-spacing marks, etc)
+        let mut category = self
```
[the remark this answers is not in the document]
cmyr: python doesn't set this and just looks it up in the glyphdata whenever it's needed, I'm trying to just do the lookup once and then use that everywhere else.

glyphs-reader/src/font.rs
cmyr: this is just a helper/wrapper around `get_glyph`, and doesn't have a python equivalent. `get_glyph` does, though, and that's documented below.

PR #778, opened 2024-04-17
[fontdrasil] Add agl module to fontdrasil

cmyr:
This provides an API for querying the Adobe Glyph List dataset.

This module is a nearly direct port of the 'agl' module in fonttools, modulo the fact that we are not including the special handling for Zapf Dingbats.


fontdrasil/build.rs
```diff
@@ -0,0 +1,82 @@
+//! Generate static arrays with data from the Adobe Glyph Lists
+
+use std::{env, fs::File, io::BufWriter, path::Path, str::FromStr};
+
+const OUT_FILE: &str = "agl_codegen.rs";
+
+fn main() {
+    println!("cargo::rerun-if-changed=data");
```
rsheeter, line 8: Not data/ ?

rsheeter (approved):
LGTM

fontdrasil/build.rs
cmyr: seems fine: https://doc.rust-lang.org/1.77.2/cargo/reference/build-scripts.html#rerun-if-changed

Issue #769, opened 2024-04-09
handle magic '*origin' anchor in glyphs2ir


cmyr:
The anchor propagation code we've gotten from glyphs.app includes a new special case for handling mark anchors. From @anthrotype:

```py
   if layer.anchors and glyph.category == "Mark":
       new_anchors = {a.name: a for a in layer.anchors}
       if "*origin" in new_anchors:
           origin = new_anchors.pop("*origin").position
           for anchor_name, anchor in new_anchors.items():
               new_anchors[anchor_name] = GSAnchor(
                   anchor_name,
                   NSPoint(anchor.position.x - origin.x, anchor.position.y - origin.y),
               )
       return new_anchors
```

this looks like something that should definitely be in the glyphs2ir phase, although it would be nice to know exactly what the rationale/use case is, for this logic.

Additionally it would be really nice to see a clear description of what the glyphs.app anchor naming scheme is; for instance I know that `#` is a magic character, and apparently [`&`  is as well](https://forum.glyphsapp.com/t/multiple-entry-anchors-possible/28825/2) (and I think I recall seeing something that non-alphanumeric prefixes generally have some special behaviour?)

cmyr:
@anthrotype I think this is also something where it would be useful to ping georg; in particular I'd like to get a test case for this, but that would require a better understanding of what we're actually trying to accomplish.

anthrotype:
> it would be nice to know exactly what the rationale/use case is, for this logic

the use case is that.. combining mark glyphs' advance widths are meant to be zero-ed at compile time, however font developers like to keep some non-zero spacing in their mark glyphs otherwise they'd disappear or clash with preceding glyphs when laid out next to other glyphs and they are being edited in Glyphs.app.
In Glyphs.app, zeroing the advance width at compile time does not simply set their width=0, but also shifts the mark glyphs so that their origin is moved all the way to the right (xMax + rightSideBearing): this way they appear to hang somewhere on top of the preceding glyph even if the latter doesn't have mark pos rules. The user can control this shifting behavior using this special `*origin` anchor.

E.g. see https://forum.glyphsapp.com/t/custom-lsb-rsb-for-zero-width-mark-attachment/15943/7

> I'd like to get a test case for this

I'm working on making those right now

rsheeter:
@cmyr is this done? - I feel like I saw code related to handling origin but I'm not immediatley sure if it's complete

PR #782, opened 2024-04-24
Anchor propagation

cmyr:
This implements basic anchor propagation in `glyphs-reader`. Propagation happens before the conversion to IR. Handling this here lets us more closely match the behaviour of glyphs.app.

This work is based on a python rewrite of the original objc; that work (by @anthrotype) is available in [a google doc](https://docs.google.com/document/d/1bXF8usWtsInYFUYIMvq3ds4jjPdbf2uDykewYyxQ-wU/edit?resourcekey=0-EqPApxblj5YTtDOG-1bZfA).

This code is now fairly well covered by tests, and I am reasonably confident that it replicates the behaviour of the original code in at least all the obvious cases.

~The original code is tricky and handles a bunch of edge cases. Some of that is included in this, and some of that is not. In particular, this does not handle the "*origin" anchor (see #769) and it does not handle the possible presence of the "anchor" field on a component, which we do not currently parse from the source.~

~This includes some tests for the most common cases, as well as some nice infra for adding additional tests in the future. This code is branchy and complex, and there are doubtless a bunch we aren't hitting. In particular there is special logic to swap top/bottom, left/right, and entry/exit in cases where a component has a rotation applied. I'm not even really sure what the use case is here, and thus what a good test case would be, but perhaps it's for something like using a 'u' as an 'n' by just rotating a component?~

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -9,7 +9,6 @@ use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::ffi::OsStr;
 use std::hash::Hash;
 use std::str::FromStr;
-use std::sync::OnceLock;
 use std::{fs, path};
 
 use crate::glyphdata::{Category, GlyphData, Subcategory};
@@ -170,7 +169,7 @@ impl FeatureSnippet {
     }
 }
 
-#[derive(Debug, PartialEq, Hash)]
+#[derive(Clone, Default, Debug, PartialEq, Hash)]
 pub struct Glyph {
     pub name: SmolStr,
     pub export: bool,
@@ -204,7 +203,7 @@ impl Glyph {
     }
 }
 
-#[derive(Debug, PartialEq, Hash)]
+#[derive(Debug, Default, Clone, PartialEq, Hash)]
 pub struct Layer {
     pub layer_id: String,
     pub associated_master_id: Option<String>,
@@ -268,7 +267,7 @@ impl FromPlist for LayerAttributes {
     }
 }
 
-#[derive(Debug, PartialEq, Hash)]
+#[derive(Debug, Clone, PartialEq, Hash)]
 pub enum Shape {
     Path(Path),
     Component(Component),
@@ -1725,9 +1724,8 @@ impl RawGlyph {
 // custom GlyphData.xml files, as well as handle overrides that are part of the
 // glyph source.
 fn get_glyph_category(name: &str, codepoints: &BTreeSet<u32>) -> Option<(Category, Subcategory)> {
-    static GLYPH_DATA: OnceLock<GlyphData> = OnceLock::new();
-    let data = GLYPH_DATA.get_or_init(|| GlyphData::new(None).unwrap());
-    data.get_glyph(name, Some(codepoints))
+    GlyphData::bundled()
+        .get_glyph(name, Some(codepoints))
         .map(|info| (info.category, info.subcategory))
 }
 
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -172,7 +172,7 @@ impl FeatureSnippet {
 
 #[derive(Debug, PartialEq, Hash)]
 pub struct Glyph {
-    pub glyphname: SmolStr,
+    pub name: SmolStr,
     pub export: bool,
     pub layers: Vec<Layer>,
     pub unicode: BTreeSet<u32>,
@@ -194,6 +194,14 @@ impl Glyph {
             )
         )
     }
+
+    pub(crate) fn has_components(&self) -> bool {
+        self.layers
+            .iter()
+            .flat_map(Layer::components)
+            .next()
+            .is_some()
+    }
 }
 
 #[derive(Debug, PartialEq, Hash)]
@@ -213,6 +221,13 @@ impl Layer {
 
     pub fn is_intermediate(&self) -> bool {
         self.associated_master_id.is_some() && !self.attributes.coordinates.is_empty()
+    }
+
+    pub(crate) fn components(&self) -> impl Iterator<Item = &Component> + '_ {
+        self.shapes.iter().filter_map(|shape| match shape {
+            Shape::Path(_) => None,
+            Shape::Component(comp) => Some(comp),
+        })
     }
 
     // TODO add is_alternate, is_color, etc.
@@ -1694,7 +1709,7 @@ impl RawGlyph {
         }
 
         Ok(Glyph {
-            glyphname: self.glyphname,
+            name: self.glyphname,
             export: self.export.unwrap_or(true),
             layers: instances,
             left_kern: self.kern_left,
@@ -2016,7 +2031,7 @@ impl TryFrom<RawFont> for Font {
             })
             .collect();
 
-        Ok(Font {
+        let mut font = Font {
             units_per_em,
             fs_type,
             use_typo_metrics,
@@ -2026,7 +2041,6 @@ impl TryFrom<RawFont> for Font {
             default_master_idx,
             glyphs,
             glyph_order,
-            //glyph_to_codepoints,
             axis_mappings,
             virtual_masters,
             features,
@@ -2036,7 +2050,10 @@ impl TryFrom<RawFont> for Font {
             version_minor: from.versionMinor.unwrap_or_default() as u32,
             date: from.date,
             kerning_ltr: from.kerning_LTR,
-        })
+        };
+
+        font.propagate_all_anchors();
+        Ok(font)
     }
 }
 
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -13,7 +13,7 @@ use std::{fs, path};
 
 use crate::glyphdata::{Category, GlyphData, Subcategory};
 use crate::plist::FromPlist;
-use kurbo::{Affine, Point};
+use kurbo::{Affine, Point, Vec2};
 use log::{debug, warn};
 use ordered_float::OrderedFloat;
 use plist_derive::FromPlist;
@@ -761,6 +761,16 @@ struct RawAnchor {
 pub struct Anchor {
     pub name: SmolStr,
     pub pos: Point,
+}
+
+impl Anchor {
+    pub(crate) fn is_origin(&self) -> bool {
+        self.name == "*origin"
+    }
+
+    pub(crate) fn origin_delta(&self) -> Option<Vec2> {
+        self.is_origin().then_some(self.pos.to_vec2())
+    }
 }
 
 impl Hash for Anchor {
@@ -2029,7 +2039,7 @@ impl TryFrom<RawFont> for Font {
             })
             .collect();
 
-        let mut font = Font {
+        Ok(Font {
             units_per_em,
             fs_type,
             use_typo_metrics,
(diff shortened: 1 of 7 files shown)


glyphs-reader/src/propagate_anchors.rs
```diff
@@ -0,0 +1,883 @@
...
+    // to make this work, we write the anchors to a separate data structure, and
+    // then only update the actual glyphs after we've done all the work.
+    let mut all_anchors = HashMap::new();
+    for name in todo {
+        // temporarily remove the glyph so we can mess with it without borrowing all glyphs
+        let glyph = glyphs.get(&name).unwrap();
+        // we work with indices here to also get around borrowck, because
+        // if we borrow the layers in the loop we can't set the anchors in the loop body
+        for layer in &glyph.layers {
+            let anchors = anchors_traversing_components(
```
rsheeter, line 46: Optional, might just be me, I'd have grasped something like `reachable_anchors` faster

glyphs-reader/src/propagate_anchors.rs
```diff
@@ -0,0 +1,883 @@
...
+    );
+}
+
+/// Return the anchors for this glyph, including anchors from components
+fn anchors_traversing_components<'a>(
+    glyph: &'a Glyph,
+    layer: &'a Layer,
+    glyphs: &BTreeMap<SmolStr, Glyph>,
+    // map of glyph -> anchors, per-layer, updated as we do each glyph
+    visited_anchors: &HashMap<SmolStr, Vec<Vec<Anchor>>>,
```
rsheeter, line 90: `visited` in a graph traversal led me to expect a fairly specific pattern of use (if already visited, return that answer, otherwise compute and record) that this doesn't follow. I wonder if it's worth renaming to just `anchors`

glyphs-reader/src/propagate_anchors.rs
```diff
@@ -0,0 +1,883 @@
...
+    let prev_names: Vec<_> = layer.anchors.iter().map(|a| &a.name).collect();
+    let new_names: Vec<_> = anchors.iter().map(|a| &a.name).collect();
+    log::info!(
+        "propagated anchors for ('{}': {prev_names:?} -> {new_names:?}",
+        glyph.name,
+    );
+}
+
+/// Return the anchors for this glyph, including anchors from components
+fn anchors_traversing_components<'a>(
```
rsheeter, line 85: This seems to do more than the name or comment suggests, for example I see application of adjustments like origin. Suggest updating at least the comment.

glyphs-reader/src/propagate_anchors.rs
```diff
@@ -0,0 +1,883 @@
...
+    base_glyph_counts: &mut HashMap<(SmolStr, &'a str), usize>,
+) -> Vec<Anchor> {
+    if layer.anchors.is_empty() && layer.components().count() == 0 {
+        return Vec::new();
+    }
+
+    // if this is a mark and it has anchors, just return them.
+    if !layer.anchors.is_empty() && glyph.category == Some(Category::Mark) {
+        let mut anchors = layer.anchors.clone();
+        // if there is an '*origin' anchor, other positions are adjusted by its pos:
```
rsheeter, line 101: Where do we assure ourselves that there aren't multiple origin anchors? - either comment or return an error

glyphs-reader/src/propagate_anchors.rs
```diff
@@ -0,0 +1,883 @@
...
+    // map of glyph -> anchors, per-layer, updated as we do each glyph
+    visited_anchors: &HashMap<SmolStr, Vec<Vec<Anchor>>>,
+    // each (glyph, layer) writes its number of base glyphs into this map during traversal
+    base_glyph_counts: &mut HashMap<(SmolStr, &'a str), usize>,
+) -> Vec<Anchor> {
+    if layer.anchors.is_empty() && layer.components().count() == 0 {
+        return Vec::new();
+    }
+
+    // if this is a mark and it has anchors, just return them.
```
rsheeter, line 98: "just return them" is a bit misleading given that we are adjusting them.

glyphs-reader/src/propagate_anchors.rs
```diff
@@ -0,0 +1,883 @@
...
+        if let Some(origin) = anchors
+            .iter()
+            .position(Anchor::is_origin)
+            .map(|idx| anchors.remove(idx))
+        {
+            anchors
+                .iter_mut()
+                .for_each(|a| a.pos -= origin.pos.to_vec2())
+        }
+        return anchors;
```
rsheeter, line 111: Suggest extracting this into it's own function

glyphs-reader/src/propagate_anchors.rs
```diff
@@ -0,0 +1,883 @@
...
+    }
+
+    let is_ligature = glyph.sub_category == Some(Subcategory::Ligature);
+    let mut has_underscore = layer
```
rsheeter, line 115: A name that indicates what has_underscore means would help.

There seems to be quite a lot of anchor name comparison with strings, many of which are repeated ("_", "entry", "exit", etc appear several times). I wonder if we might either make helper functions like `is_exit` or `is_whatever_starts_with_underscore_means` or even something like an AnchorType enum that can be build from the name.

glyphs-reader/src/propagate_anchors.rs
```diff
@@ -0,0 +1,883 @@
...
+        .anchors
+        .iter()
+        .any(|anchor| anchor.name.starts_with('_'));
+
+    let mut number_of_base_glyphs = 0usize;
+    // we use an index map so we get the same ordering behaviour as python
```
rsheeter, line 121: point to the python source here and wherever else applicable?

glyphs-reader/src/propagate_anchors.rs
rsheeter: It's also quite long and hard to follow, I wonder if it could be broken down into smaller functions such that this and they end up simpler?

glyphs-reader/src/propagate_anchors.rs
```diff
@@ -0,0 +1,883 @@
...
+    const ROUND_TO: f64 = 1e6;
+    let mut pos = (transform * anchor.pos).to_vec2();
+    pos *= ROUND_TO;
+    pos = pos.round();
+    pos /= ROUND_TO;
+    anchor.pos = pos.to_point();
+}
+
+fn maybe_rename_component_anchor(comp_name: SmolStr, anchors: &mut [Anchor]) {
+    // e.g, go from 'top_1' to 'top'
```
rsheeter, line 276: Just to confirm, we definitely want the _first_ `_` or assume there is only one of them?

glyphs-reader/src/propagate_anchors.rs
```diff
@@ -0,0 +1,883 @@
...
+
+        let new_glyph = glyphs.get("commaturnedabovecomb").unwrap();
+        assert_eq!(
+            new_glyph.layers[0].anchors,
+            [("_top", (300., 502.)), ("mytop", (312., 810.)),]
+        );
+    }
+
+    #[test]
+    fn affine_scale() {
```
rsheeter, line 829: This looks like a bunch of independent tests that don't need to be one big test?

rsheeter (approved):

glyphs-reader/src/propagate_anchors.rs
```diff
@@ -0,0 +1,883 @@
...
+            assert_eq!(layers.len(), glyph.layers.len());
+            for (i, layer_anchors) in layers.into_iter().enumerate() {
+                glyph.layers[i].anchors = layer_anchors;
+            }
+        }
+    }
+}
+
+fn debug_new_anchors(anchors: &[Anchor], glyph: &Glyph, layer: &Layer) {
+    if !log::log_enabled!(log::Level::Info) {
```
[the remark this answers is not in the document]
cmyr: oh yea i just confused trace & info

glyphs-reader/src/propagate_anchors.rs
cmyr: This name was chosen to match the source impl, happy to change...

glyphs-reader/src/propagate_anchors.rs
cmyr: anchor names are unique.

glyphs-reader/src/propagate_anchors.rs
cmyr: nice we can also reuse that in the other place we handle "*origin", below.

glyphs-reader/src/propagate_anchors.rs
cmyr: that code isn't public, and there also isn't much to point to in this case except for the fact that python dictionaries track insertion order.

glyphs-reader/src/propagate_anchors.rs
cmyr: the original code is assuming only one.

glyphs-reader/src/propagate_anchors.rs
cmyr: just me sanity checking the behaviour of this fn, doesn't feel like much of an improvement to break this into five two line tests..

glyphs-reader/src/propagate_anchors.rs
cmyr: - this is the original name, and the original logic, and the original code doesn't really document its intention.
- I tried the helpers, but unfortunately half the places where we're checking these things we're just working with strings, so we can't really add a method without getting fancy.
- the enum thing is nice and we do that once we get to IR, but I think down here we just have to do our best to match the glyphs code without trying to be more systematic than it is.

glyphs-reader/src/propagate_anchors.rs
cmyr: I did a lot of that, but most of the rest of it is pretty entwined, and the more refactoring we do the harder it is for us to compare back to the original (particularly in the future, where the reference impl might change)

glyphs-reader/src/propagate_anchors.rs
rsheeter: Even if we just made a function it would be an improvement over repeating the same string comparison everywhere.

If we have questions about the intent we can ask Georg, perhaps we should assemble a list of questions?

cmyr: okay, added 'is_exit' and 'is_entry' helpers.


Ultimately here it feels like the implementation is the spec; there are a lot of heuristics and squishy bits here that seem like they're probably the result of trying to get some specific pattern to work. It would be nice if there was something in the glyphs docs that detailed the naming convention for anchors, but for this code in particular I don't have questions, exactly; I'm just taking it at face value and trying to match the behaviour. This is a very literal port.

glyphs-reader/src/font.rs
```diff
@@ -2051,6 +2073,14 @@ fn preprocess_unparsed_plist(s: &str) -> Cow<str> {
 
 impl Font {
     pub fn load(glyphs_file: &path::Path) -> Result<Font, Error> {
+        let mut font = Self::load_impl(glyphs_file)?;
+        font.propagate_all_anchors();
```
anthrotype, line 2077: I feel like this `propagate_all_anchors ` should be optional somehow, or at least there should be a method to _only_ load the .glyphs file without the anchor propagation.. It's not our current use case in fontc, but it's conceivable that one may wish to read a .glyphs file, change some bit that's unrelated to anchors, and write it back to a file to do something else.

anthrotype:
Thanks Colin for implementing this. I know it must not have been fun 👍

glyphs-reader/src/propagate_anchors.rs
rsheeter: If we don't understand the naming conventions why don't we ask the Glyphs folks? - decent chance they'd be up for explaining.

PR #786, opened 2024-04-30
[marks] Store GDEF categories in StaticMetadata

cmyr:
We need these categories in order to determine which glyphs should be considered marks/based/ligatures for the purpose of generating layout tables.


fontc/src/lib.rs
```diff
@@ -1768,6 +1769,32 @@ mod tests {
...
+        let categories = &staticmeta.gdef_categories;
+        assert_eq!(categories.get("a"), Some(&GlyphClassDef::Base));
+        assert_eq!(categories.get("acutecomb"), Some(&GlyphClassDef::Mark));
+        assert_eq!(categories.get("brevecomb"), Some(&GlyphClassDef::Mark));
+        assert_eq!(
+            categories.get("brevecomb_acutecomb"),
+            Some(&GlyphClassDef::Mark)
+        );
+        assert_eq!(categories.get("f_a"), Some(&GlyphClassDef::Ligature));
+    }
```
rsheeter, line 1785: ideally add a case where the category is None?

fontir/src/ir.rs
```diff
@@ -1059,6 +1064,11 @@ impl AnchorKind {
         }
         Ok(AnchorKind::Base(name.into()))
     }
+
+    /// Returns `true` If this is a base or ligature base anchor
```
rsheeter, line 1068: describing what an attaching anchor is might be more helpful, e.g. what attaches to what

rsheeter (approved):

PR #790, opened 2024-05-08
[glyphs] Add test for anchor propagation on digraph

cmyr:
This is just to sanity check that our behaviour is matching glyphs.app in the case where we have a glyph with multiple base components but which is not a ligature.

JMM

### DIFF
--- a/glyphs-reader/src/propagate_anchors.rs
+++ b/glyphs-reader/src/propagate_anchors.rs
@@ -643,6 +643,7 @@ mod tests {
     #[test]
     fn propagate_ligature_anchors() {
         // derived from the observed behaviour of glyphs 3.2.2 (3259)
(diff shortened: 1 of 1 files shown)


moyogo:
This seems wrong whether Glyphs.app is doing it or not.

cmyr:
@moyogo do you mean in the sense that the generated anchors don't really make sense? I agree that the font designer should probably be either adding anchors manually or else opting out of propagation here if possible, but we definitely don't want these glyphs to be treated as ligatures.

In any case we are defining 'correct' here in terms of what glyphs.app does, and this is correct by that measure.

moyogo:
@cmyr I commented on https://github.com/googlefonts/glyphsLib/issues/368#issuecomment-2103445759

PR #791, opened 2024-05-08
[fea] Move PendingLookup into fea-rs

cmyr:
This makes more sense there, and lets us provide a better API.

This was motivated by wanting to start storing the flags and filter glyph set for mark/mark rules.


rsheeter (approved):
LGTM. Note that one of the checks is failing.

PR #792, opened 2024-05-09
[ttx_diff] Add basic support for mark2lig rules

cmyr:
We weren't handling this at all. This implementation does not do some of the fancier stuff that we do for other mark attachment rules, since these lookups are less common and it seems much less likely that there will be overlapping lookups that apply rules to the same glyphs.

### DIFF
--- a/layout-normalizer/src/gpos/marks.rs
+++ b/layout-normalizer/src/gpos/marks.rs
@@ -4,7 +4,7 @@ use std::{
 };
 
 use write_fonts::read::{
-    tables::gpos::{MarkBasePosFormat1, MarkMarkPosFormat1},
+    tables::gpos::{MarkBasePosFormat1, MarkLigPosFormat1, MarkMarkPosFormat1},
     types::GlyphId,
     ReadError,
 };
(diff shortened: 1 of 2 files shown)


layout-normalizer/src/gpos/marks.rs
```diff
@@ -35,6 +42,31 @@ impl PrintNames for MarkAttachmentRule {
...
+        for (i, anchor) in self.base_anchors.iter().enumerate() {
+            if i != 0 {
+                write!(f, ", ")?;
+            }
+            match anchor {
+                Some(a) => write!(f, "{a}"),
+                None => write!(f, "<NULL>"),
+            }?
+        }
+        writeln!(f, "]")?;
```
rsheeter, line 58: this looks suspiciously likely to be a repetitive pattern, if it is perhaps we could share a fn?

rsheeter (approved):

layout-normalizer/src/gpos/marks.rs
cmyr: if we end up repeating it we can always refactor, but this is the last mark attachment rule so I don't anticipate that happening.

PR #794, opened 2024-05-10
Generate mark2lig lookups

cmyr:
*This is based on #793, which should go in first.*

This is the last major remaining piece that we were missing in order to be able to match oswald.

This code *is* generating the same lookups as fontmake for oswald, but it is pretty naive; in particular it's just generating one big lookup and assuming that anchors don't collide (that is, that the same glyph is not part of multiple mark classes) whereas I believe the markFeatureWriter does the same sort of splitting for mark2lig that it does for mark2base, where it generates one lookup for each mark class?

This raises the bigger question of testing: there's currently no good way to test our feature generating code in isolation. I started working on this a bit for kerning but ran into some issues, but I think now it's worth spending some time to figure that problem out, since I think that the "last 10%" of layout is going to be much more test driven (isolate some funny quirk, write a test case, write a fix)

### DIFF
--- a/fea-rs/src/compile/lookups/gpos_builders.rs
+++ b/fea-rs/src/compile/lookups/gpos_builders.rs
@@ -532,6 +532,7 @@ impl Builder for MarkToBaseBuilder {
     }
 }
 
+/// A builder for GPOS Lookup Type 5, Mark-to-Ligature
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct MarkToLigBuilder {
@@ -540,6 +541,10 @@ pub struct MarkToLigBuilder {
 }
 
 impl MarkToLigBuilder {
+    /// Add a new mark glyph.
+    ///
+    /// If this glyph already exists in another mark class, we return the
+    /// previous class; this is likely an error.
     pub fn insert_mark(
         &mut self,
         glyph: GlyphId,
@@ -549,14 +554,21 @@ impl MarkToLigBuilder {
         self.marks.insert(glyph, class, anchor)
     }
 
+    /// Add a ligature base, providing a set of anchors for each component.
+    ///
+    /// There must be an item in the vec for each component in the ligature glyph,
+    /// but the anchors can be sparse; null anchors will be added for any classes
+    /// that are missing.
     pub fn add_lig(&mut self, glyph: GlyphId, components: Vec<BTreeMap<SmolStr, Anchor>>) {
         self.ligatures.insert(glyph, components);
     }
 
+    /// Returns an iterator over all of the mark glyphs
     pub fn mark_glyphs(&self) -> impl Iterator<Item = GlyphId> + Clone + '_ {
         self.marks.glyphs()
     }
 
+    /// Returns an iterator over all of the ligature glyphs
     pub fn lig_glyphs(&self) -> impl Iterator<Item = GlyphId> + Clone + '_ {
         self.ligatures.keys().copied()
     }
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -241,6 +241,7 @@ impl<'a> FeatureWriter<'a> {
                 .borrow_mut()
                 .push(("Start add marks", Instant::now()));
         }
+
         self.marks.add_features(builder);
 
         {
--- a/fontbe/src/features/marks.rs
+++ b/fontbe/src/features/marks.rs
@@ -4,8 +4,8 @@ use std::collections::{BTreeMap, BTreeSet, HashSet};
 
 use fea_rs::{
     compile::{
-        FeatureProvider, MarkToBaseBuilder, MarkToMarkBuilder, NopFeatureProvider,
-        NopVariationInfo, PendingLookup,
+        FeatureProvider, MarkToBaseBuilder, MarkToLigBuilder, MarkToMarkBuilder,
+        NopFeatureProvider, NopVariationInfo, PendingLookup,
     },
     typed::{AstNode, LanguageSystem},
     GlyphSet, Opts, ParseTree,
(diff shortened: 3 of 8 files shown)


rsheeter:
> "last 10%" of layout is going to be much more test driven (isolate some funny quirk, write a test case, write a fix)

+1

rsheeter:
> it's just generating one big lookup and assuming that anchors don't collide (that is, that the same glyph is not part of multiple mark classes) whereas I believe the markFeatureWriter does the same sort of splitting for mark2lig

Sounds worth filing an issue for. Perhaps we can create a test case (in a follow-on) taht deliberately puts a glyph into multiple mark classes and see hwo our results compare?

fea-rs/src/compile/lookups/gpos_builders.rs
```diff
@@ -540,6 +541,10 @@ pub struct MarkToLigBuilder {
 }
 
 impl MarkToLigBuilder {
+    /// Add a new mark glyph.
+    ///
+    /// If this glyph already exists in another mark class, we return the
+    /// previous class; this is likely an error.
```
rsheeter, line 547: Is that what fontmake does? If not suggest filing an issue about this so we don't forget.

fontbe/src/features.rs
```diff
@@ -262,24 +262,27 @@ impl<'a> FeatureWriter<'a> {
 
         // Build the actual mark base and mark mark constructs using fea-rs builders
 
-        let mut mark_base_lookups = Vec::new();
-        let mut mark_mark_lookups = Vec::new();
+        let mut mark_lookups = Vec::new();
+        let mut mkmk_lookups = Vec::new();
 
         for mark_base in marks.mark_base.iter() {
             // each mark to base it's own lookup, whch differs from fontmake
```
rsheeter, line 269: can you specify how it differs and link to the fontmake code?

rsheeter: [oops, you didn't change that]

fontbe/src/features.rs
```diff
@@ -262,24 +262,27 @@ impl<'a> FeatureWriter<'a> {
 
         // Build the actual mark base and mark mark constructs using fea-rs builders
 
-        let mut mark_base_lookups = Vec::new();
-        let mut mark_mark_lookups = Vec::new();
+        let mut mark_lookups = Vec::new();
+        let mut mkmk_lookups = Vec::new();
```
rsheeter, line 266: I find the prior names clearer. mkmk is particularly obtuse.

fontbe/src/features/marks.rs
```diff
@@ -303,6 +307,65 @@ impl<'a> MarkLookupBuilder<'a> {
...
+            let mut component_anchors = Vec::new();
+            for anchor in anchors {
+                if let fontir::ir::AnchorKind::Ligature { group_name, index } = &anchor.kind {
+                    liga_anchor_groups.insert(group_name);
+                    // ensure our component array has an item for this component:
+                    if *index > component_anchors.len() {
+                        component_anchors.resize(*index, BTreeMap::new());
+                    }
+                    let anchor = resolve_anchor(anchor, self.static_metadata, glyph_name)?;
+                    component_anchors[*index - 1].insert(group_name.clone(), anchor);
```
rsheeter, line 333: I wonder if we should just have a fn or trait to insert at with resize ... if we do this elsewhere as well I think we should. 

I also wonder if it's worth just finding the index of the last ligature and resizing once. I guess we can leave that one for the profiler.

fontbe/src/features/marks.rs
```diff
@@ -303,6 +307,65 @@ impl<'a> MarkLookupBuilder<'a> {
...
+                let mark_group = anchor.mark_group_name().unwrap();
+                if !liga_anchor_groups.contains(mark_group) {
+                    continue;
+                }
+
+                let gid = self.glyph_order.glyph_id(mark_name).unwrap();
+                let anchor = resolve_anchor(anchor, self.static_metadata, mark_name)?;
+                if let Err(prev_class) = builder.insert_mark(gid, mark_group.to_owned(), anchor) {
+                    // this can't happen with markbase or markmark because we have one lookup per
+                    // class. Should we be doing something similar for marklig?
```
rsheeter, line 354: Suggest filing an issue and consulting w/@anthrotype

rsheeter (approved):
LGTM subject to adding testing, even if it's just "compile a (new?) test source and confirm we produced mark2lig as expected." fontc has lots of that sort of test.

anthrotype:
> I believe the markFeatureWriter does the same sort of splitting for mark2lig that it does for mark2base, where it generates one lookup for each mark class?

yes, groupMarkClasses=False is the default behavior now, so similarly to markToBase lookup, the mark2Liga ones will only reference one mark class at a time, and the lookups themselves are sorted alphabetically by the mark class name so the more specific ('top.alt' instead 'top') would be applied last and win

see https://github.com/googlefonts/ufo2ft/blob/779bbad84aa2bba9f8e31774e2f27fe84c8c231e/Lib/ufo2ft/featureWriters/markFeatureWriter.py#L280-L290

test case: 
https://github.com/googlefonts/ufo2ft/blob/779bbad84aa2bba9f8e31774e2f27fe84c8c231e/tests/featureWriters/markFeatureWriter_test.py#L1506-L1538

fontbe/src/features.rs
cmyr: unfortunately the previous names are inaccurate; these contain lists of lookups for a particular feature (one of `mark` or `mkmk`, but not necessarily only lookups of one type (the `mark` feature contains both mark2base and mark2lig rules)

fontbe/src/features/marks.rs
cmyr: on review I agree it's simpler to just figure out the length up front, which also gives us a convenient place to skip a glyph if it doesn't actually have any ligature anchors, despite being classified as a ligature 🤷

cmyr:
I have basically completely redone this implementation, so I'm going to close this and reopen as a separate PR.

Issue #684, opened 2024-01-26
GlyphsApp support


rsheeter:
fontmake has a plugin @ https://github.com/schriftgestalt/FontMakeExportPlugin. Let's add fontc, for the time being marked as experimental.

rsheeter:
I wonder if it makes sense to cdn compiler binaries such that the plugin can download them with a priori (baked in) knowledge of the expected size & checksum. That would require a bit of dilligence wrt releasing fontc but it might be worthwhile.

rsheeter:
https://rust-lang.github.io/rustup/cross-compilation.html suggests we can compile e.g. a Mac binary on Linux which is very helpful. While GlyphsApp is Mac-only we should still publish at least Linux binaries to things like Google Sans CI can use them.

PR #795, opened 2024-05-11
Add release workflow to build and publish source and binary artifacts to Github and PyPI

anthrotype:
For the Glyphs.app fontc export plugin (#684) we need to make precompiled binaries of fontc available to download and run without requiring to build from source. Glyphs.app already comes with Python, which in turn comes with an established and secure way to distribute and install binary packages via the Python Package Index (PyPI) and the pip installer.
With the help of a tool called maturin (see [source](https://github.com/PyO3/maturin) and [docs](https://www.maturin.rs/)) it's super easy to make Python wheels out of any Rust crate, including binary-only ones like fontc. Maturin itself is written in Rust, uses cargo under the hood, and it's primarily used to package up hybrid Python/Rust projects -- which fontc technically is not (but who knows, it may in the future get a python API as well); but that's ok, ours would not be the first project on PyPI that doesn't contain any Python itself but is simply an executable that gets unpacked to the environment's bin/ or Scripts/ directories (e.g. ther's also stuff like ninja or cmake). And we have used maturin in the past to package stuff like resvg or pngquant for nanoemoji, for example.
Maturin is great because I don't have to deal with the complications of making sure the binaries works on, e.g., both intel or Apple Silicon macs, or some random linux distro. I'm sure other, more expert Rust devs in here would know how do deal with all that, and may sniff at any snakes creeping in.. but I hope they will accept this as a pragmatic, first-time attempt :)

PS:  While this is more of a side effect than the primary goal, allowing one to `pip install fontc` would increase its reach among font editors (Glyphs.app, Robofont, fontra either expose Python API/Macros or are themselves writtein in it) as well as many font developers who are already accustomed to Python (think fontTools, fontmake, afdko, etc.).

This PR adds a new release.yml Github Actions workflow that is triggered automatically on git tags (actually tags conventionally start with 'v' e.g. `v1.0.0`) or manually from the Actions interface.
It builds the wheels for Mac, Linux and Windows and pushes them to PyPI, where they will be available at https://pypi.org/project/fontc/
It also compresses and publish the same `fontc` binaries that maturin has built for the Python wheels as simple .tar.gz archives and uploads them to a new Github Release, as downloadable assets. Besides allowing one to just curl, unzip and run, this also makes them ready to be installed with the nifty `cargo binstall` (https://github.com/cargo-bins/cargo-binstall), which can automatically finds the Github release associated with a given crate on crates.io and download precompiled binaries from it.
Finally, I added a block that automatically publishes fontc to crates.io but have not enabled that yet, because I think we first need to check we have all the required metatada and maybe we need to make the first cargo publish manually to reserve our spot and make sure everying works.

EDIT: I removed the automatic upload to crates.io, since we decided to publish source crates manually instead like we do in fontations.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -217,6 +217,22 @@ jobs:
           pattern: binaries-*
           merge-multiple: true
 
+      - uses: actions/download-artifact@v4
+        with:
+          pattern: wheels-*
+          merge-multiple: true
+          path: wheels
+
+      # this requirements.txt can be used to pip install fontc more securely
+      # https://pip.pypa.io/en/stable/topics/secure-installs/
+      - name: Generate requirements.txt with SHA256 hashes
+        run: |
+          echo fontc | pipx run --spec pip-tools pip-compile - --no-index --find-links wheels/ --no-emit-find-links --generate-hashes --pip-args '--only-binary=:all:' --no-annotate --no-header --output-file requirements.txt
+
+      - name: Compute checksums of release assets
+        run: |
+          sha256sum *.tar.gz *.zip requirements.txt > checksums.txt
+
       - name: Detect if tag is a pre-release
         id: set_prerelease
         env:
@@ -239,6 +255,8 @@ jobs:
           files: |
             *.tar.gz
             *.zip
+            checksums.txt
+            requirements.txt
           prerelease: ${{ steps.set_prerelease.outputs.is_prerelease }}
           generate_release_notes: true
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -223,3 +223,19 @@ jobs:
             *.zip
           prerelease: ${{ steps.set_prerelease.outputs.is_prerelease }}
           generate_release_notes: true
+
+  # TODO: Make sure fontc's Cargo.toml metadata are ok for publishing to crates.io,
+  # maybe publish it the first time manually to reserve a spot. Also store a token
+  # named CRATES_IO_TOKEN in the repository's secrets before un-commenting this:
+  # release-crates-io:
+  #   name: Publish source to crates.io
+  #   runs-on: ubuntu-latest
+  #   if: "startsWith(github.ref, 'refs/tags/v')"
+  #   needs: [release-github]
+  #   steps:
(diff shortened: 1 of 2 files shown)


anthrotype:
testing this workflow from my anthrotype/fontc fork, the Github Releases produced by it look like the following:

https://github.com/anthrotype/fontc/releases

anthrotype:
Note that the "Build GS" CI failure is because this PR branch sits on my anthrotype fork and as such hasn't access to the token required to download GS from the private repo

rsheeter (approved):
Suggestion (could be a follow-on or ignored): also generate a text file with sha256 and file lengths of the compiler binaries to attach to the release. That would make updating the plugin's download specification easy.

rsheeter:
> we need to make the first cargo publish manually to reserve our spot and make sure everying works.

I believe you are right about first manual. We should do that so our names aren't subject to unexpected changes. IIRC we might also have to manually add everyone as owners in the crates web interface.

anthrotype:
> also generate a text file with sha256 and file lengths of the compiler binaries to attach to the release

77677e3 uploads two additional files to the Github release assets:

- `checksums.txt`: the output of `sha256sum` (or `shasum -a 256`) for all the uploaded assets; I did not add any file lengths for simplicity's sake (checking can be done with simple `sha256sum -c checksums.txt`) and because sha256 ought to be cryptographically robust enough to be sensitive to any changes to file lengths.
- `requirements.txt`: contains the currently pinned, versioned fontc requirement that a release is associated with, alongside precomputed SHA256 hashes of the binary wheel packages that we uploaded to PyPI; we can include this in our Glyphs.app plugin to ensure pip checks the integrity of the downloaded wheel cf. https://pip.pypa.io/en/stable/topics/secure-installs

They will look like this:
https://github.com/anthrotype/fontc/releases/tag/v0.0.2a6

anthrotype:
I think this is ready. Copying the release shell scripts from fontations could be done in a separate PR. After we merged this, could one of you publish fontc to crates.io and push a git tag in the form `fontc-vX.X.X`? the worflow should do the rest (build the wheels, upload them PyPI, create GH release and upload binary assets there)

PR #797, opened 2024-05-14
Mark testing harness

cmyr:
This adds a nice mechanism for testing our mark generation, based on `layout-normalizer`.

Basically this provides a builder where you can set up some set of glyphs + anchors + GDEF categories + user FEA, and then what feature code gets generated, in a normalized textual representation.

Getting this to work required a few tweaks in various places; I made it so that `layout-normalizer` can work only with the specific tables (instead of requiring a full font) and I also tweaked the `MarkLookupBuilder` so that our testing code more closely matches what we do during regular compilation.

### DIFF
--- a/layout-normalizer/src/main.rs
+++ b/layout-normalizer/src/main.rs
@@ -7,7 +7,7 @@ use std::{
 
 use clap::Parser;
 use layout_normalizer::{args, Error, NameMap};
-use write_fonts::read::{FileRef, FontRef, ReadError};
+use write_fonts::read::{FileRef, FontRef, ReadError, TableProvider};
 
 fn main() -> Result<(), Error> {
     let args = args::Args::parse();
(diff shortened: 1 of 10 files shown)


layout-normalizer/src/main.rs
```diff
@@ -29,12 +29,20 @@ fn main() -> Result<(), Error> {
...
+        if let Ok(gpos) = font.gpos() {
+            writeln!(&mut write_target, "# GPOS #")?;
+            layout_normalizer::print_gpos(&mut write_target, &gpos, gdef.as_ref(), &name_map)?;
+        }
     }
 
     if matches!(to_print, args::Table::All | args::Table::Gsub) {
-        layout_normalizer::print_gsub(&mut write_target, &font, &name_map)?;
+        if let Ok(_gsub) = font.gsub() {
+            // we don't currently handle GSUB, and it's not clear we want to?
```
rsheeter, line 42: if we ever use this on google style icon fonts and emoji - both eventual targets - we'll want at least some GSUB. Maybe we should file an issue to normalize GSUB?

rsheeter (approved):

PR #798, opened 2024-05-14
[marks] Fix bad boolean logic in find_mark_glyphs

cmyr:
I am regularly humbled by how hard it is to follow even relatively trivial boolean logic in my head.

This is on #797, which should go in first.

### DIFF
--- a/fontbe/src/features/marks.rs
+++ b/fontbe/src/features/marks.rs
@@ -353,8 +353,14 @@ fn find_mark_glyphs(
     anchors
         .iter()
         .filter(|(name, anchors)| {
-            (!gdef_classes.is_empty() && gdef_classes.get(*name) == Some(&GlyphClassDef::Mark))
-                && anchors.iter().any(|a| a.is_mark())
+            // if we have some classes, and this is not in the mark class:
+            // not a mark glyph.
+            if !gdef_classes.is_empty() && gdef_classes.get(*name) != Some(&GlyphClassDef::Mark) {
+                return false;
+            }
+            // if we have no classes, or this is in the mark class,
+            // then we just look for the presence of a mark anchor.
+            anchors.iter().any(|a| a.is_mark())
         })
         .map(|(name, _)| name.to_owned())
         .collect()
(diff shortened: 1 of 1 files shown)


fontbe/src/features/marks.rs
```diff
@@ -353,8 +353,10 @@ fn find_mark_glyphs(
     anchors
         .iter()
         .filter(|(name, anchors)| {
-            (!gdef_classes.is_empty() && gdef_classes.get(*name) == Some(&GlyphClassDef::Mark))
-                && anchors.iter().any(|a| a.is_mark())
+            let not_a_mark =
+                !gdef_classes.is_empty() && gdef_classes.get(*name) != Some(&GlyphClassDef::Mark);
+            let has_anchors = anchors.iter().any(|a| a.is_mark());
+            !not_a_mark && has_anchors
```
rsheeter, line 359: not not a mark is nasty. I confess to not finding either version terribly easy to follow. Quick attempt to simplify:

`!not_a_mark` is `!(!gdef_classes.is_empty() && gdef_classes.get(*name) != Some(&GlyphClassDef::Mark))` 

by De Morgan it becomes `gdef_classes.is_empty || gdef_classes.get(*name) == Some(&GlyphClassDef::Mark)`

is_empty seems unnecessary, just write `gdef_classes.get(*name) == Some(&GlyphClassDef::Mark)`. This matches my intuitive understanding of "not not a mark": you want to know if it IS a mark.

We want "not not a mark and has anchors [that are marks]" so add the anchors bit and we get:

```rust
.filter(|(name, anchors)| {
   gdef_classes.get(*name) == Some(&GlyphClassDef::Mark)
   && anchors.iter().any(|a| a.is_mark())
})
```

^ reads relatively clear to me: I want to know if name refers to a mark and that mark has mark anchors.

rsheeter (approved):

fontbe/src/features/marks.rs
cmyr: So unfortunately the `is_empty()` check is semantically meaningful, because if _no_ classes are defined  than we don't assign any meaning to the item not being in it or not; this is a major part of what makes this complicated.

so basically if we *know* it's not a mark (some classes are defined, and this isn't in the mark class) than we fail; otherwise we check if there are mark anchors.

In any case I've rewritten this again to just treat the two conditions separately, and use an explicit return on the first. This also more closely matches the referenced python.

PR #799, opened 2024-05-14
[marks] Add tests for using GDEF block from user FEA

cmyr:
This is a funny situation, where for glyphs.app we want to ignore this but when using a UFO source we want to look at it.

based on #797.

### DIFF
--- a/fontbe/src/features/marks.rs
+++ b/fontbe/src/features/marks.rs
@@ -498,6 +498,7 @@ mod tests {
     /// A helper for testing our mark generation code
     #[derive(Clone, Debug)]
     struct MarksInput<const N: usize> {
+        prefer_gdef_categories_in_fea: bool,
         locations: [NormalizedLocation; N],
         anchors: BTreeMap<GlyphName, Vec<Anchor>>,
         categories: BTreeMap<GlyphName, GlyphClassDef>,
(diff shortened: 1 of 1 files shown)


fontbe/src/features/marks.rs
```diff
@@ -568,6 +570,15 @@ mod tests {
             self
         }
 
+        /// Set whether or not to prefer GDEF categories defined in FEA vs in metadata
+        ///
+        /// this is 'true' for UFO sources and 'false' for glyphs sources; default
+        /// here is false.
+        fn set_prefer_fea_gdef_categories(&mut self, flag: bool) -> &mut Self {
+            self.prefer_gdef_categories_in_fea = flag;
```
rsheeter, line 578: Optional, I feel like an enum might be nice? - gdef_preference = GdefPref::Fea or Metadata. The bool leaves it to our imagination [or the comments] to know what the other option is.

rsheeter (approved):

fontbe/src/features/marks.rs
cmyr: If this were public API, I'd be inclined to agree but it doesn't feel like a big simplification here, and is such a funny edge case that you'll still need comments to explain what the heck is going on.


there might be a better name, though, like `gdef_in_fea_should_override_categories` or something... 😒

PR #803, opened 2024-05-16
[ttx_diff] Better normalization of Mark2Lig

cmyr:
This unifies these lookups with the other mark attachment rules, and uses the same normalization process to ensure that when multiple shadow each other we only display the rule that will be applied.

### DIFF
--- a/layout-normalizer/src/gpos.rs
+++ b/layout-normalizer/src/gpos.rs
@@ -29,10 +29,7 @@ mod pairpos;
 #[cfg(test)]
 mod test_helpers;
 
-use self::{
-    marks::{MarkAttachmentRule, MarkLigaRule},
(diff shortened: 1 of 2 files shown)


layout-normalizer/src/gpos.rs
```diff
@@ -440,14 +437,12 @@ fn get_lookup_rules(
                 let subs = subs.iter().flat_map(|sub| sub.ok()).collect::<Vec<_>>();
                 let rules = marks::get_mark_mark_rules(&subs, delta_computer).unwrap();
                 result
-                    .markbase
+                    .markmark
```
anthrotype, line 440: oops

anthrotype (approved):

PR #806, opened 2024-05-16
Generate MarkToLig lookups, take two

cmyr:
This is a rewrite of the initial implementation in #794 to more closely match fontmake (and pass the basic tests there).

- ~Several tests in this PR will fail until it is rebased onto #803; these failures are the result of issues in the normalizer.~
- ~also requires #802.~

### DIFF
--- a/fea-rs/src/compile/lookups/gpos_builders.rs
+++ b/fea-rs/src/compile/lookups/gpos_builders.rs
@@ -532,6 +532,7 @@ impl Builder for MarkToBaseBuilder {
     }
 }
 
+/// A builder for GPOS Lookup Type 5, Mark-to-Ligature
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct MarkToLigBuilder {
@@ -540,6 +541,15 @@ pub struct MarkToLigBuilder {
 }
 
 impl MarkToLigBuilder {
+    /// `true` if this builder contains no rules
+    pub fn is_empty(&self) -> bool {
+        self.ligatures.is_empty()
+    }
+
+    /// Add a new mark glyph.
+    ///
+    /// If this glyph already exists in another mark class, we return the
+    /// previous class; this is likely an error.
     pub fn insert_mark(
         &mut self,
         glyph: GlyphId,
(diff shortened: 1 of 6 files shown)


fea-rs/src/compile/lookups/gpos_builders.rs
```diff
@@ -549,14 +559,53 @@ impl MarkToLigBuilder {
...
+    /// There must be an item in the vec for each component in the ligature glyph,
+    /// but the anchors can be sparse; null anchors will be added for any classes
+    /// that are missing.
+    ///
+    /// NOTE: this API closely mimics how the FEA source represents these rules,
+    /// where you process each component in order, with all the marks defined
+    /// for that component; but this is less useful for public API, where you
+    /// are more often dealing with marks a class at a time. For that reason
+    /// we provide an alternative public method below.
+    pub(crate) fn add_lig_all(
```
rsheeter, line 573: I don't think they bill by the character, `add_ligatures` is probably fine?

rsheeter: (applies to other "lig" methods too, e.g. insert_lig)

fontbe/src/features/marks.rs
```diff
@@ -55,10 +55,17 @@ struct MarkLookupBuilder<'a> {
     mark_glyphs: BTreeSet<GlyphName>,
 }
 
+/// Abstract over the difference in anchor shape between mark2lig and mark2base/mark2mark
+#[derive(Debug, Clone, PartialEq)]
+enum BaseOrLigaAnchors<T> {
```
rsheeter, line 60: Suggest spelling out Ligature somewhere. Enum and variant names seem duplicative. AnchorType::Ligature?

I'd rather we write out Ligature in full but if we must shorten it lets at least pick one shortening. Right now we have lig and liga.

fontbe/src/features/marks.rs
```diff
@@ -76,31 +83,48 @@ impl MarkGroup<'_> {
...
+
+impl MarkAttachmentBuilder for MarkToLigBuilder {
+    fn add_mark(&mut self, gid: GlyphId, group: &GroupName, anchor: FeaAnchor) {
+        let _ = self.insert_mark(gid, group.clone(), anchor);
+    }
+
+    fn add_base(&mut self, gid: GlyphId, group: &GroupName, anchors: BaseOrLigaAnchors<FeaAnchor>) {
+        match anchors {
+            BaseOrLigaAnchors::Liga(anchors) => self.insert_lig(gid, group.clone(), anchors),
+            BaseOrLigaAnchors::Base(_) => unreachable!(),
```
rsheeter, line 126: Provide _at least_ the offending gid and probably the group name as well. I know it's not supposed to happen but if/when it does we'll curse the lack of context provided.

fontbe/src/features/marks.rs
```diff
@@ -305,6 +334,63 @@ impl<'a> MarkLookupBuilder<'a> {
...
     }
+
+    fn make_mark_to_liga_groups(&self) -> BTreeMap<GroupName, MarkGroup<'_>> {
+        let mut groups = BTreeMap::<_, MarkGroup>::new();
+        let mut liga_anchor_groups = HashSet::new();
+
+        // a temporary buffer, reused for each glyph
+        let mut component_groups = HashMap::new();
+        for (glyph_name, anchors) in &self.anchor_lists {
+            let is_not_liga = !self.gdef_classes.is_empty()
```
rsheeter, line 345: negative booleans are a common source of error. Strongly recommend making it positive, e.g. is liga.

This loop is a bit hard to read due to all the conditions before we actually do anything. I wonder if some of the continue cases could be factored into a function so you could say something like `for (name, anchors, max_ligature_index) in self.mark_to_ligature_anchors()` would help?

fontbe/src/features/marks.rs
```diff
@@ -305,6 +334,63 @@ impl<'a> MarkLookupBuilder<'a> {
...
+            for (group_name, anchors) in component_groups.drain() {
+                groups
+                    .entry(group_name)
+                    .or_default()
+                    .bases
+                    .push((glyph_name.clone(), BaseOrLigaAnchors::Liga(anchors)));
+            }
+        }
+
+        // then we do another pass to add the marks:
```
rsheeter, line 374: It would be nice to have an equivalently clear comment explaining what the first pass does

rsheeter (approved):

fontbe/src/features/marks.rs
cmyr: see https://github.com/googlefonts/fontc/blob/945604aa6e0c634165c5d30178ca9a674ade7169/fontbe/src/features/marks.rs#L215-L217 for how this is called;


basically you would need to explicitly use the wrong generic type to hit this branch, in which case it will fail immediately and for all inputs. Added a panic message, though.

fontbe/src/features/marks.rs
cmyr: 👍 on the boolean.

looked at splitting this fn but it feels like a marginal gain, so I've just improved the comments a bit to explain the flow.

PR #819, opened 2024-05-21
Rename plist_derive -> ascii_plist_derive

cmyr:
Also adds a readme and does a bit of additional cleanup.

As we prepare to publish these crates it's a good time to get names right, and the new name clarifies that we are only working with old-style ASCII Plists.


glyphs-reader/Cargo.toml
```diff
@@ -11,7 +11,7 @@ readme = "README.md"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-plist_derive = { path = "plist_derive" }
+ascii_plist_derive = { path = "ascii_plist_derive" }
```
anthrotype, line 14: this is also commonly referred to as OpenStep plist format. Given that our own parser (as well as Glyphs.app's) happily accept UTF-8 literal strings in addition to plain ASCII, I would use that name if we can

cmyr: This is the original PLIST format used by NeXTSTEP (and then OpenStep) but I think it's fair to refer to it as 'ASCII', since this format is commonly referred to as "ascii-style" plists, to differentiate them from the XML format that apple adopted later, but I'll update the README to mention NeXTSTEP/OpenStep..

glyphs-reader/ascii_plist_derive/README.md
```diff
@@ -0,0 +1,9 @@
+# `ascii_plist_derive`
+
+This crate provides a simple derive macro for parsing ASCII-style
+(NeXTSTEP/OpenStep) style [property lists]. It is designed for use by the
```
anthrotype, line 4: one "style" too much?

anthrotype (approved):

PR #821, opened 2024-05-21
Update env-logger

cmyr:
The API changed a bit, so needed to fix that up. Also noticed that our logs were never actually printing a thread name, and this doesn't seem super useful given that the threading isn't something we manage, so I've just removed that bit of the log preamble for the time being.

### DIFF
--- a/fontc/src/main.rs
+++ b/fontc/src/main.rs
@@ -35,14 +35,13 @@ fn run() -> Result<(), Error> {
     env_logger::builder()
         .format(|buf, record| {
             let ts = buf.timestamp_micros();
+            let style = buf.default_level_style(record.level());
             writeln!(
                 buf,
-                "[{ts} {} {} {}] {}",
-                // we manually assign all threads a name
-                std::thread::current().name().unwrap_or("unknown"),
+                "[{ts} {:?} {} {style}{}{style:#}] {}",
+                std::thread::current().id(),
                 record.target(),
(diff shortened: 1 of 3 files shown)


fontc/src/main.rs
```diff
@@ -35,14 +35,12 @@ fn run() -> Result<(), Error> {
     env_logger::builder()
         .format(|buf, record| {
             let ts = buf.timestamp_micros();
+            let style = buf.default_level_style(record.level());
             writeln!(
                 buf,
-                "[{ts} {} {} {}] {}",
-                // we manually assign all threads a name
```
rsheeter, line 41: Knowing what thread things are on *is* helpful, I don't know that we should remove that. If the comment is wrong and we arent' naming them anymore lets just fix the comment?

rsheeter: Hm. I see we get unknown a lot. Let's make it print the thread id if there is no name, unknown is ofc not useful.

rsheeter (approved):
LGTM subject to fixing the thread identifier bit. Unknown is not helpful but thread id would be.

fontc/src/main.rs
cmyr: it's just annoying because the only way to print a thread id is the full debug version where you get `ThreadId(8)` or whatever, and it takes up a lot of space. will make this change though..

PR #825, opened 2024-05-21
[fontbe] Get off patched version of icu_properties

cmyr:
This wasn't too hard, although it does require us to duplicate a bit of code in a few places and otherwise slightly complicate some impls.


rsheeter (approved):

dfrg (approved):
Looks good!

PR #827, opened 2024-05-22
add support for cargo binstall in fontc/Cargo.toml

anthrotype:
there's this nice cargo sub-command called https://github.com/cargo-bins/cargo-binstall which allows one to install pre-compiled binaries for one's own platform/architecture that are hosted on Github Releases.

This is particularly useful e.g. for https://github.com/googlefonts/fontc-action where one would like to install fontc the quickest way possible instead of building from source.

Usually cargo binstall just works and it requires no configuation, but in our particular case, because we push prefixed tags like `fontc-v{version}` instead of bare `v{version}`, then cargo binstall cannot automatically find the github releases download URL, and falls back to the slower `cargo install` (from source).

This little change to fontc/Cargo.toml should make sure the correct URL will be used.

### DIFF
--- a/fontc/Cargo.toml
+++ b/fontc/Cargo.toml
@@ -52,3 +52,20 @@ chrono.workspace = true
 
 [build-dependencies]
 vergen = { version = "8.2.6", features = ["build", "cargo", "git", "gitcl", "rustc"] }
+
+# Support for `cargo binstall fontc`.
+# The default Github download URL template used by cargo-binstall doesn't work
+# because our tag is named `fontc-v{version}` instead of just `v{version}`.
+# https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md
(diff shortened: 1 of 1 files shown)


cmyr:
is this testable in any way?

anthrotype:
well, you could call cargo binstall from the command line overriding the same flags that I have changed in the manifest: 

```
$ cargo binstall --pkg-url="{ repo }/releases/download/{ name }-v{ version }/{ name }-{ target }{ archive-suffix }" --bin-dir="{ bin }{ binary-ext }" --pkg-fmt="tgz" fontc@0.0.1
```

anthrotype:
if we merge these changes and release them to crates.io, then for the new releases one will be able to simply use `cargo install fontc` or `cargo install fontc@0.1.0` etc.

cmyr (approved):
happy to give this a shot

PR #829, opened 2024-05-22
[otl-normalizer] Normalize MultiSub lookups

cmyr:
A very straightforward impl for now.

### DIFF
--- a/otl-normalizer/src/gsub.rs
+++ b/otl-normalizer/src/gsub.rs
@@ -10,8 +10,8 @@ use write_fonts::{
         tables::{
             gdef::Gdef,
             gsub::{
-                Gsub, SingleSubst, SingleSubstFormat1, SingleSubstFormat2, SubstitutionLookupList,
-                SubstitutionSubtables,
+                Gsub, MultipleSubstFormat1, SingleSubst, SingleSubstFormat1, SingleSubstFormat2,
+                SubstitutionLookupList, SubstitutionSubtables,
             },
         },
         ReadError,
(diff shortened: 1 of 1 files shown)


PauloRosarioData (approved):

otl-normalizer/src/gsub.rs
```diff
@@ -148,6 +163,22 @@ impl LookupRules {
         all_rules.dedup();
         all_rules
     }
+
+    fn multisub_rules<'a>(&'a self, lookups: &[u16]) -> Vec<SingleRule<'a, MultiSubRule>> {
+        // normalization here is very simple: we're just going to do first-writer-wins
+        // on targets.
```
rsheeter, line 169: Suggest elaborating on the implications of this. Are there inputs that will cause this approach problems? What are the advantages/disadvantages?

otl-normalizer/src/gsub.rs
```diff
@@ -211,6 +267,21 @@ impl PrintNames for SingleSubRule {
...
+                write!(f, ",")?;
+            }
+            let g2 = names.get(*g2);
+            write!(f, "{g2}")?;
+        }
+        write!(f, "]")
+    }
+}
+
 #[cfg(test)]
```
rsheeter, line 285: Seems testable, add a test?

rsheeter (approved):

cmyr:
I think we're leaning towards not normalizing GSUB, since it's just not clear how to do so without implementing shaping, so I'm going to close this.

Issue #815, opened 2024-05-18
Variable GPOS feature for non-variable glyph? (invalid gvar axisCount)


justvanrossum:
I was trying to find out whether fontc supports variable features, and while I think the error I'm getting proves that it does, I _am_ getting an error, in the following case:

- A .designspace project with a Weight axis
- There is only one source: Regular (at Weight=400)
- The one source contains one glyph A (a square of 500x500)
- The following feature code:
```
feature test {
    pos A <100 0 (wght=400:200 wght=900:400) 0>;
} test;
```

fontmake compiles this into a variable font with a working variable GPOS feature.

fontc errors like this:

```
[2024-05-18T13:38:38.008202Z main fontc::workload ERROR] Be(Features) failed Fea compilation failure Compilation failed with 1 errors
error: failed to compute deltas: 'No variation model for 'Normalized {wght: 1.00}''
in varpos_Regular.ufo/features.fea at 2:17
  | 
2 |     pos A <100 0 (wght=400:200 wght=900:400) 0>;
  |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^

Tasks failed: [(Be(Features), "Fea compilation failure Compilation failed with 1 errors\n\u{1b}[31merror: \u{1b}[0mfailed to compute deltas: 'No variation model for 'Normalized {wght: 1.00}''\n\u{1b}[3;34min\u{1b}[0m varpos_Regular.ufo/features.fea \u{1b}[3;34mat\u{1b}[0m 2:17\n\u{1b}[34m  |\u{1b}[0m \n\u{1b}[34m2 |\u{1b}[0m     pos A <100 0 (wght=400:200 wght=900:400) 0>;\n\u{1b}[34m  |\u{1b}[0m                  \u{1b}[31m^^^^^^^^^^^^^^^^^^^^^^^^^^^\u{1b}[0m\n")]
```

.designspace + .ufo:
[varpos.zip](https://github.com/googlefonts/fontc/files/15365239/varpos.zip)

justvanrossum:
Sparsify all the things!

anthrotype:
Can you try with https://github.com/googlefonts/fontc/pull/779

justvanrossum:
Sorry, not instantly, as I'm running fontc from PyPI, not sources.

anthrotype:
I believe that should fix it, but cannot double check as I’m typing from my mobile. That PR is basically done, only needs a test to ratify it works as intended. I’ll try to complete it next week

[...]

PR #833, opened 2024-05-29
[fontbe] pass explicit axis_count when constructing Gvar

anthrotype:
Will fix https://github.com/googlefonts/fontc/issues/815

CI will be 🔴 until https://github.com/googlefonts/fontations/pull/919 is merged write-fonts dependency bumped

### DIFF
--- a/fontbe/src/gvar.rs
+++ b/fontbe/src/gvar.rs
@@ -52,6 +52,14 @@ impl Work<Context, AnyWorkId, Error> for GvarWork {
         // We built the gvar fragments alongside glyphs, now we need to glue them together into a gvar table
         let static_metadata = context.ir.static_metadata.get();
         let axis_order: Vec<_> = static_metadata.axes.iter().map(|a| a.tag).collect();
+        let axis_count: u16 = axis_order
+            .len()
+            .try_into()
+            // in the unlikely event that we have more than 65535 axes...
+            .map_err(|_| Error::OutOfBounds {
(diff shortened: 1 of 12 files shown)


fontbe/src/gvar.rs
```diff
@@ -52,6 +52,14 @@ impl Work<Context, AnyWorkId, Error> for GvarWork {
...
         let static_metadata = context.ir.static_metadata.get();
         let axis_order: Vec<_> = static_metadata.axes.iter().map(|a| a.tag).collect();
+        let axis_count: u16 = axis_order
+            .len()
+            .try_into()
+            // in the unlikely event that we have more than 65535 axes...
+            .map_err(|_| Error::OutOfBounds {
+                what: "axis count".into(),
+                value: axis_order.len().to_string(),
+            })?;
```
cmyr, line 62: I would personally unwrap here, since I can't imagine this being possible for any conceivable font, but 🤷

cmyr (approved):
looks good!

fontbe/src/gvar.rs
anthrotype: just wait a few years :)

rsheeter: I strongly prefer Result unless it _can't_ happen. Rationale: if we unwrap all the things that shouldn't happen but plausibly can then it's all but assured we'll hit _some_ of them in production and rue the day we didn't use Result.

PR #840, opened 2024-05-30
[fea-rs] Tweak contextual lookup format selection

cmyr:
- keep track of the number of subtables generated by a given format
- account for the cost of additional offsets when formats generate multiple subtables

Noticed we were overlooking this while trying to debug some GSUB differences; it's not the culprit but we might as well try to be correct.

### DIFF
--- a/fea-rs/src/compile/lookups/contextual.rs
+++ b/fea-rs/src/compile/lookups/contextual.rs
@@ -688,30 +688,37 @@ impl Builder for SubChainContextBuilder {
 }
 
 // invariant: at least one item must be Some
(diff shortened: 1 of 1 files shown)


anthrotype (approved):

cmyr:
@anthrotype can we match this implementation in feaLib?

anthrotype:
> can we match this implementation in feaLib?

what are you referring to exactly? doesn't feaLib already "account for the cost of additional offsets when formats generate multiple subtables"?

cmyr:
sorry wrong PR! I was thinking about #842; the logic implemented there is not present in fonttools, but if it was I _think_ we would match 100% for GSUB.

PR #841, opened 2024-05-30
[fea-rs] Add script for building a test case via fontmake

cmyr:
This was useful for me in trying to debug our deviation from fontmake, and I suspect it might be useful again soon.


anthrotype:
You aren’t calling fontmake the command line tool, you are using fontTools.feaLib API (which fontmake among others also uses), I suggest you rename and replace mentions of fontmake with fonttools (feaLib)

anthrotype (approved):
LGTM except for s/fontmake/fonttools/

PR #842, opened 2024-05-30
[fea-rs] More reuse of lookups for inline multisub rules

cmyr:
With this change, if an existing lookup already contains the substitution that are trying to add, we will just reuse that lookup.

Our behaviour here was already slightly different from fonttools; this is a further optimization. I've opened https://github.com/fonttools/fonttools/issues/3551 there, and believe fonttools should match this behaviour.

### DIFF
--- a/fea-rs/src/compile/lookups/contextual.rs
+++ b/fea-rs/src/compile/lookups/contextual.rs
@@ -213,7 +213,7 @@ impl ContextualLookupBuilder<SubstitutionLookup> {
                 SubstitutionLookup::Multiple(subtables) => subtables
                     .subtables
                     .iter()
-                    .all(|subt| !subt.contains_target(target)),
+                    .all(|subt| subt.can_add(target, &replacements)),
                 _ => false,
             },
             |flags, mark_set| SubstitutionLookup::Multiple(LookupBuilder::new(flags, mark_set)),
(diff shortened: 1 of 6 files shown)


fea-rs/src/compile/lookups/contextual.rs
```diff
@@ -213,7 +213,7 @@ impl ContextualLookupBuilder<SubstitutionLookup> {
                 SubstitutionLookup::Multiple(subtables) => subtables
                     .subtables
                     .iter()
-                    .all(|subt| !subt.contains_target(target)),
+                    .all(|subt| subt.can_add(target, &replacements)),
```
anthrotype, line 216: why should this reuse of chained inline/anonymous lookups be limited to multiple substitutions? in theory the same treatment could be applied to other GSUB types like single or ligature substituions?

cmyr: we already have this implemented for single sub, matching fonttools. We do not have it implemented for ligature substitutions, although we probably should, but I'd want to know what fonttools is doing and do that in a separate PR with good testing etc.

anthrotype:
> believe fonttools should match this behaviour.

sounds good, I will take a look next week

anthrotype (approved):
LGTM

I have a WIP locally to patch fonttools feaLib which I'll PR soon. With this current PR applied (plus my own fonttools patch), I get _zero_ GSUB diffs b/w fontmake/fontc Oswald!

anthrotype:
> I have a WIP locally to patch fonttools feaLib which I'll PR soon

this one https://github.com/fonttools/fonttools/pull/3559

Note I have modified/renamed the test .fea/.ttx files slightly (included idotbelow, idotless, iogonek etc. to make the substitutions more realistic, closer to Oswald's original)

PR #846, opened 2024-06-03
[glyphs] Glyphs with no category can still be bases

cmyr:
This was a little thinko: we were returning early when determining GDEF categories if the _glyphs.app_ category was `None`, but in this case we should make our determination based on the presence of the relevant anchors.

(This was the issue I had earlier assumed was related to the glyphs not having assigned unicode values)


rsheeter (approved):
Nit: it looks to me like your test data is not minimal, you just need the most minimal possible contour plus an anchor. Might be nice to use a minimal input.

Issue #724, opened 2024-03-04
Test a lot of builds


rsheeter:
@simoncozens has been doing work on automating builds across the collection that we can exploit to build and ttx_diff a lot of fonts. "build me a VF" is an operation (implemented here: https://github.com/googlefonts/gftools/blob/builder2/Lib/gftools/builder/operations/buildVariable.py). Replace that line with a fontc equivalent, and off we go.

This should enable us to test "all the fonts for which we have a known upstream repository and where there's a gftools-bulider configuration file in that repo." At time of writing that means 482 fonts out of 1603.

TODO: add end to end instructions to acquire and build as many google fonts as possible.

simoncozens:
Here is how I would do this:

* https://github.com/simoncozens/gf-dashboard/blob/main/gfautobuilder.py is a driver script which clones each repo with a known upstream and runs gftools-builder on it.
  * It requires a cache file with information about which upstream repos seem like they can be built using gftools-builder.
  * You can download the [cache file directly](https://github.com/simoncozens/gf-dashboard/blob/main/cache.json) or you can generate it yourself with [this script](https://github.com/simoncozens/gf-dashboard/blob/main/list-known-upstreams.py).
    * If you choose to generate the cache file yourself, the `list-known-upstreams.py` script in turn requires the `googlefonts` Python module in the same repo, it requires you to set `GFDIR` on line 13 to your google/fonts checkout directory, and also requires you to set the environment variable `GITHUB_TOKEN` to a valid authentication token.
* Next you will need a version of gftools-builder which uses fontc instead of fontmake.
  * By the time you act on this, builder2 should be merged into the main branch, but in case it's not, start by checking out the builder2 branch of gftools.
  * Edit the files `Lib/gftools/builder/operations/buildVariable.py`, `Lib/gftools/builder/operations/buildTTF.py` and `Lib/gftools/builder/operations/buildOTF.py`, replacing the fontmake command line with the equivalent fontc command line. (The `$fontmake_type` variable means `-g` for Glyphs source, `-m` for designspace and `-u` for ufo; you don't need that font fontc.)
  * Install this modified gftools-builder into your Python environment; you now have a fontc-powered gftools-builder
* Run `gfautobuilder.py` and watch the pretty error messages.

rsheeter:
[...]

PR #852, opened 2024-06-13
Checkpoint: bulk compilation abilities

cmyr:
I initially took the approach that simon described in #724, but this ended up compiling everything multiple times (since gftools builder builds a bunch of different targets for each source, which we're not interested in yet) and getting it to work involved patching gftools in a way that seemed hard to distribute.

In the end, then, I've added a new `bulk_runner` crate. This calls into fontc directly (which I think will be useful in the future when we want to collect stats about what esoteric corners of OpenType are used by various fonts); with this binary running in release mode it takes ~5 minutes to compile the 480~ google fonts fonts we have sources for, versus an hour + from the previous script.

The reporting here isn't great! It's hard to get high-level information from our errors, and many of them don't print very well. My current approach is to dump the results to json, and then use ipython to explore them; but we can iterate from here.

To run:

```sh
cargo run --release --p=bulk_runner PATH_TO_CHECKOUT_FONT_REPOS compile --out=results.json
```


README.md
```diff
@@ -52,6 +52,15 @@ Once you have them you could try building them:
 cargo run --package fontc -- ../google_fonts_sources/sources/ofl/notosanskayahli/sources/NotoSansKayahLi.designspace
 ```
 
+### Building lots of fonts at once
```
rsheeter, line 55: Can we be slightly more specific? What lots of fonts? I assume all Google Fonts for which we have sources?

fontc/src/bin/fontc_bulk/README.md
```diff
@@ -0,0 +1,26 @@
...
+bulk.
+
+It takes as input a JSON dictionary (as a text file) containing key/value pairs
+where the keys are the names of fonts and the values are the /username/repo
+portion of a github URL.
+
+This data is derived from the cache files used by `gfautobuilder.py`: see [this
+comment](https://github.com/googlefonts/fontc/issues/724#issuecomment-1977431334)
+for details on how to get or rebuild this. (We restructure the data slightly to
+include only the 'real_upstream' value.)
```
rsheeter, line 13: Does this mean the command to run bulk in the root index won't actually work, there's some other gymnastics one has to do first?

If so lets change the root index to say how to build all know Google Fonts sources and point to a script that runs all the steps.

fontc/src/bin/fontc_bulk/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+We include a file containing information for all of the known google fonts font
+repositories as of June 2024, at resources/testdata/gf_fonts.json.
```
rsheeter, line 16: Let's add `resources/scripts/make-me-that-file` instead of committing the giant json file.

README.md
rsheeter: As per below I suggest we update this to be how to build all Google Fonts we have sources for and point to a script that walks through it, including the construction of the json file if a recent copy is not locally available.

rsheeter (approved):
LGTM if we don't commit gf_fonts.json and instead commit the python code to produce it. Let's make it easy for people to do that as it's the most common (possibly only) bulk compile people will do.

cmyr:
@rsheeter this has all been updated to use the `google-fonts-sources` crate.

I had one realization just now: maybe it _would_ make sense for this to be separate crate, since we wouldn't need to publish it and then we could mor easily have extra deps here that we might not want in `fontc` itself?

PR #854, opened 2024-06-20
[fontir] Start to refactor errors

cmyr:
This is some initial cleanup of the errors in fontir; the next step is going to be to unify Error and WorkError.

- Replace various Error variants with a 'BadSource' error that packages up a particular error cause with the path of the problematic source
- Remove the general 'io error' variant from Error in favour of finer-grained errors that include additional context.
- get rid of `dyn Error`


rsheeter (approved):
LGTM once ci is happy

PR #855, opened 2024-06-21
[fontir] Unify Error & WorkError

cmyr:
This also makes errors more granular:

- errors related to finding, loading, or parsing a source file are their own 'BadSource' error type, which always preserves the path of the file in question;
- errors that occur during conversion of a specific glyph are now their own 'BadGlyph' error type, which means we always preserve the name of the glyph in question when such an error occurs.

A few other variants were removed or combined.

### DIFF
--- a/fontir/src/error.rs
+++ b/fontir/src/error.rs
@@ -9,6 +9,74 @@ use smol_str::SmolStr;
 use thiserror::Error;
 use write_fonts::types::{InvalidTag, Tag};
 
+#[derive(Debug, Error)]
+pub enum Error {
+    /// A source file was not understood
+    #[error(transparent)]
+    BadSource(#[from] BadSource),
+    /// An error occured while attempting to track a file
+    #[error(transparent)]
+    TrackFile(#[from] TrackFileError),
+    /// An error occured while converting a glyph to IR
+    #[error(transparent)]
+    BadGlyph(#[from] BadGlyph),
+    #[error("Failed to delete file {path}: '{source}'")]
+    DeleteFailed {
+        path: PathBuf,
+        #[source]
+        source: io::Error,
+    },
+    #[error("UPEM value {0} outside valid range 16..=16384")]
+    InvalidUpem(f64),
+    #[error("Inconsistent UPEM values: {0:?}")]
+    InconsistentUpem(Vec<u16>),
+    #[error("Variation model error: '{0}'")]
+    VariationModelError(
+        #[from]
+        #[source]
+        VariationModelError,
+    ),
+    #[error("feature files are non-identical: {0}, {1}")]
+    NonIdenticalFea(PathBuf, PathBuf),
+    #[error("axis '{0}' missing at least one of default/min/max mapping")]
+    MissingAxisMapping(Tag),
+    #[error("no glyph for name '{0}'")]
+    NoGlyphForName(GlyphName),
+    #[error("Missing required axis values for {0}")]
+    NoAxisDefinitions(String),
+    #[error("Axis {0} has no entry in axes")]
+    NoEntryInAxes(String),
+    #[error("Axis definitions are inconsistent: '{0}'")]
+    InconsistentAxisDefinitions(String),
+    #[error("Missing layer '{0}'")]
+    NoSuchLayer(String),
+    #[error("No files associated with glyph {0}")]
+    NoStateForGlyph(GlyphName),
+    #[error("No design space location(s) associated with glyph {0}")]
+    NoLocationsForGlyph(GlyphName),
+    #[error("Asked to create work for something other than the last input we created")]
+    UnableToCreateGlyphIrWork,
+    #[error("Unexpected state encountered in a state set")]
+    UnexpectedState,
+    #[error("Duplicate location for {what}: {loc:?}")]
+    DuplicateUserLocation { what: String, loc: UserLocation },
+    #[error("Global metadata very bad, very very bad")]
+    InvalidGlobalMetadata,
+    #[error("No default master in {0}")]
+    NoDefaultMaster(PathBuf),
+    #[error("Missing mapping on {axis_name} for {field} at {value:?}. Mappings {mappings:?}")]
+    MissingMappingForDesignCoord {
+        axis_name: String,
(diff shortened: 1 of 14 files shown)


fontra2fontir/src/fontra.rs
```diff
@@ -225,16 +225,17 @@ pub(crate) struct FontraPoint {
 impl FontraPoint {
     /// <https://github.com/googlefonts/fontra/blob/a4edd06837118e583804fd963c22ed806a315b04/src/fontra/core/path.py#L396-L406>
     #[allow(dead_code)] // TEMPORARY
-    pub(crate) fn point_type(&self) -> Result<PointType, Error> {
+    pub(crate) fn point_type(&self) -> Result<PointType, String> {
```
rsheeter, line 228: You updated to import BadSource but didn't actually use it?

fontra2fontir/src/toir.rs
```diff
@@ -159,66 +155,59 @@ fn to_ir_glyph(
...
-                glyph_name: glyph_name.clone(),
-                message: format!("No point type for {first:?}: {e}"),
-            })?;
+            .map_err(|e| BadGlyph::parse(glyph_name.clone(), e))?;
         if first_type.is_off_curve() {
-            return Err(WorkError::InvalidSourceGlyph {
-                glyph_name: glyph_name.clone(),
-                message: String::from("Open path starts with off-curve points"),
-            });
+            return Err(BadGlyph::parse(
```
rsheeter, line 193: I find the use of parse very weird here. Why isn't this just constructing an error?

fontir/src/error.rs
```diff
@@ -249,12 +225,53 @@ impl TrackFileError {
...
+impl BadGlyph {
+    /// Convenience method for creating a `BadGlyph` error.
+    ///
+    /// To create a general-purpose parse method, use [`BadGlyph::parse`]
+    pub fn new(name: impl Into<GlyphName>, kind: impl Into<BadGlyphKind>) -> Self {
+        Self {
+            name: name.into(),
+            kind: kind.into(),
+        }
+    }
```
rsheeter, line 237: Not loving the "smart" error, just construct it at the callsite with a glyph name and a kind, why do we need to be fancy?

fontir/src/ir.rs
```diff
@@ -2151,49 +2138,49 @@ mod tests {
...
-            panic!("unexpected error: {:?}", result);
-        };
-        assert_eq!("test", glyph_name.as_str());
-        assert_eq!((4.0, 4.0), (point.x, point.y));
+        assert_eq!(
+            result,
+            Err(PathConversionError::MoveAfterFirstPoint {
+                point: (4.0, 4.0).into()
+            })
+        );
```
rsheeter, line 2154: Why did we drop the glyph name?

fontir/src/error.rs
cmyr: The rationale for this is that it cleans up the call site significantly, so when you're reading the code you can focus on the logic without getting too lost in the error handling.

Compare,

```rust
        path_builder
            .move_to((first.x, first.y))
            .map_err(|e| BadGlyph {
                name: glyph_name.clone(),
                kind: BadGlyphKind::PathConversion(e),
            })?;
        add_to_path(&mut path_builder, contour.points[1..].iter()).map_err(|e| BadGlyph {
            name: glyph_name.clone(),
            kind: BadGlyphKind::PathConversion(e),
        })?;
    } else {
        add_to_path(&mut path_builder, contour.points.iter()).map_err(|e| BadGlyph {
            name: glyph_name.clone(),
            kind: BadGlyphKind::PathConversion(e),
        })?;
    }

    let path = path_builder.build().map_err(|e| BadGlyph {
        name: glyph_name.clone(),
        kind: BadGlyphKind::PathConversion(e),
    })?;
```

and,

```rust
        path_builder
            .move_to((first.x, first.y))
            .map_err(|e| BadGlyph::new(glyph_name.clone(), e))?;
        add_to_path(&mut path_builder, contour.points[1..].iter())
            .map_err(|e| BadGlyph::new(glyph_name.clone(), e))?;
    } else {
        add_to_path(&mut path_builder, contour.points.iter())
            .map_err(|e| BadGlyph::new(glyph_name.clone(), e))?;
    }

    let path = path_builder
        .build()
        .map_err(|e| BadGlyph::new(glyph_name.clone(), e))?;
```

Is there anything in particular you don't like about this, or is it just a pattern you aren't familiar with?

fontra2fontir/src/toir.rs
cmyr: `parse` is a method for constructing `BadGlyph::Parse`.

fontir/src/ir.rs
cmyr: because `PathConversionError` now only exists inside `BadGlyph`, which stores the glyph name.

fontir/src/error.rs
rsheeter: If you dropped new and made BadGlyph a tuple-struct you would get equivalent readability, no? `BadGlyph(glyph_name.clone(), e.into())`?

rsheeter (approved):
LGTM although I might prefer if we dropped `parse` - very strongly associated with actual parsing imho - and possibly also `new`

fontir/src/error.rs
cmyr: In general I only use tuple structs when there's a single field, and that would also require these fields to be public, which they are not currently. Certainly this could be an exception, but it seems worse to me in identifiable ways. Maybe I'm overlooking some issue with `new`? Are you worried about the costs of monomorphism? Is there another concern I've overlooked?

fontir/src/error.rs
```diff
@@ -249,12 +225,53 @@ impl TrackFileError {
...
+
+    /// Create a custom parse error with any printable thing
+    pub fn parse(name: impl Into<GlyphName>, msg: impl Display) -> Self {
```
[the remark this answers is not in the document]
cmyr: Looking at the source again we shouldn't even have `Glyph::Parse`, these should all be `PathConversionError`s, so can delete some code (including this method)

PR #856, opened 2024-06-26
[fontir] Rename SourceError::parse -> custom

cmyr:
per conversation in #855, 'parse' is a confusing name for a constructor. This renames that method for `SourceError`.

JMM


rsheeter (approved):
Ty, this makes it much easier for a population of _at least_ me :D

PR #858, opened 2024-06-28
Add cursive feature writer

simoncozens:
- Add cursive entry kinds to marks
- Make the cursive builder public, documented and useable from feature writers
- Add cursive feature writer


fontir/src/ir.rs
```diff
@@ -1035,19 +1035,31 @@ pub enum AnchorKind {
...
+    Ligature {
+        group_name: GroupName,
+        index: usize,
+    },
     /// An anchor marking the presence of a ligature component with no anchors.
     ///
     /// These are names like '_3'.
     ComponentMarker(usize),
-    // we will add more anchor kinds in the future like entry/exit
+    CursiveEntry,
```
cmyr, line 1046: is it possible to have multiple entry/exist anchors? Fine if we don't handle that on a first pass, but is it something we need to consider?

cmyr (approved):
Looks good! one little question inline but happy to merge, the one CI failure is the result of the PR originating from outside of this repo, so we can ignore it.

PR #860, opened 2024-07-03
[ttx_diff] Initial refactor for machine-readable output

cmyr:
This splits the generating of the different objects that we compare (ttx tables and layout_normalizer output) from the printing of them, which is the first step towards not printing them.

There is one little addition: it now prints a % representing the diff ratio when two outputs differ.

----

Thinking about what we'll want for the bulk compiler, I realize maybe it's more significantly different than I had thought at first? Basically I'm wondering if, when we're called from the bulk compiler we shouldn't just return some kind of JSON summary that just lists the tables and their difference ratio, and we shouldn't really bother writing all the individual files to disk, since it's all going to be nuked on the next font anyway?

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -22,6 +22,7 @@ from pathlib import Path
 import shutil
 import subprocess
 import sys
+from difflib import SequenceMatcher
 from typing import MutableSequence
 
 
@@ -30,6 +31,8 @@ _COMPARE_GFTOOLS = "gftools"
 
 
 FLAGS = flags.FLAGS
+# used instead of a tag for the normalized mark/kern output
+MARK_KERN_NAME = "(mark/kern)"
 
 
 flags.DEFINE_enum(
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -314,6 +317,76 @@ def reduce_diff_noise(build_dir, fontc, fontmake):
...
+    for tag in sorted(e.tag for e in font_xml.getroot()):
+        table_str = etree.tostring(tables[tag])
+        path = build_dir / f"{compiler}.{tag}.ttx"
+        path.write_bytes(table_str)
+        comparables[tag] = table_str
+
+    return comparables
+
+
+def diff_ratio(text1: str, text2: str):
```
rsheeter, line 381: Suggest a return type annotation and a comment to indicate how to interpret the value

resources/scripts/ttx_diff.py
```diff
@@ -314,6 +317,76 @@ def reduce_diff_noise(build_dir, fontc, fontmake):
...
+    m = SequenceMatcher(None, text1, text2)
+    return m.quick_ratio()
```
rsheeter, line 383: Suggest comment explaining why it's ok that we aren't using the real ratio as quick ratio is commented as producing an upper bound on the real ratio

resources/scripts/ttx_diff.py
rsheeter: (or just switching to use `ratio`)

rsheeter (approved):

resources/scripts/ttx_diff.py
cmyr: ratio is just really slow, but I've switched this to use ratio but operate line-wise instead of char-wise.

PR #861, opened 2024-07-04
[ttx_diff] Add support for outputing JSON

cmyr:
This adds a `--json` flag to the ttx_diff script. When set, the script will output a JSON dictionary describing the results of a particular operation.

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -56,11 +56,13 @@ FLAGS = flags.FLAGS
 # used instead of a tag for the normalized mark/kern output
 MARK_KERN_NAME = "(mark/kern)"
 
+
 # we don't print to stdout of we're generating JSON
 def maybe_print(*objects):
     if FLAGS.json:
         return
     print(*objects)
+
 
 flags.DEFINE_enum(
     "compare",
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -52,22 +79,30 @@
...
+                text=True,
+                check=True,
+                cwd=working_dir,
+                stdout=log_file,
+                stderr=log_file,
+                **kwargs,
+            )
+    except subprocess.CalledProcessError as e:
+        # hack: in the JSON case I want to be able to return the stashed output
+        # in the error, so we go full python :)
```
rsheeter, line 102: "full python" seems needlessly disparaging of Python.

Suggest declaring a custom error type and raising it.

resources/scripts/ttx_diff.py
```diff
@@ -388,6 +445,17 @@ def path_for_output_item(tag_or_normalizer_name: str, compiler: str) -> str:
...
 
+
+# if we're outputing json, format error and output as json, else reraise
+def handle_compiler_error(error: Exception, compiler_name: str, json: bool):
+    if not json:
+        raise error
+    message = getattr(error, "_secret_log_stash", "") or str(error)
+    out = {"error": {"source": compiler_name, "message": message}}
+    print_json(out)
+    sys.exit(0)
```
rsheeter, line 456: Exit hidden here is a nasty surprise, I don't think we want that

rsheeter: Ideally the control flow for json vs not json would be less different

rsheeter: Hm. I also think it's informative to know whether both compilers fail regardless of json mode.

rsheeter: I also really dislike that `--json` changes the exit code for the same build. I propose we use the same codes either way, e.g.:

0 - successful execution, diffs match
1 - an unexpected error occurred (such as an unhandled exception)
2 - avoid because https://docs.python.org/3/library/sys.html#sys.exit suggests there are conventions to it's use
3 - fontc or fontmake failed. Could split into 3 codes if we want to distinguish which one or both.

A tool like fontc crater would be aware of these codes and report based on them. 

This is similar to `diff` which uses "Exit status is 0 if inputs are the same, 1 if different, 2 if trouble." except I avoided 1 (typical python result for unhandled exception, which is more similar to "trouble" as diff defines it) and 2 (due to sys.exit note about interpretation of 2).

Control flow for json vs non-json would become the same. json would print json, non-json can use traceback.print_exception to print the error.

resources/scripts/ttx_diff.py
```diff
@@ -407,17 +475,30 @@ def main(argv):
...
+            fontc_ttf = build_fontc(source.resolve(), build_dir, compare)
+        except Exception as e:
+            handle_compiler_error(e, "fontc", FLAGS.json)
+        try:
+            fontmake_ttf = build_fontmake(source.resolve(), build_dir, compare)
 
-        fontc_ttf = build_fontc(source.resolve(), build_dir, compare)
-        fontmake_ttf = build_fontmake(source.resolve(), build_dir, compare)
+        except Exception as e:
+            handle_compiler_error(e, "fontmake", FLAGS.json)
```
rsheeter, line 495: The blocks are identical other than which build function is called, could be a function.

resources/scripts/ttx_diff.py
```diff
@@ -13,12 +13,34 @@
...
     # useful if you are making changes to fontc meant to narrow the diff
     python resources/scripts/ttx_diff.py --rebuild fontc ../OswaldFont/sources/Oswald.glyphs
+
+JSON:
+    If the `--json` flag is passed, this tool will output JSON.
+
+    If the script exits with a `0` status code, then `stdout` is a JSON dictionary.
+
+    If both compilers ran successfully, this dictionary will have a single key,
+    "success", which will contain a dictionary, where keys are the tags of tables
```
rsheeter, line 23: Highly recommend something like "result" that is a bit more neutral wrt the nature of the result (e.g. it might be a catastrophic failure of a diff)

resources/scripts/ttx_diff.py
cmyr: > Exit hidden here is a nasty surprise, I don't think we want that

The goal was to keep the behaviour the same as it is right now without `--json` (that is, the exception is unhandled and we exit) and to output valid JSON communicating the error otherwise.

> Hm. I also think it's informative to know whether both compilers fail regardless of json mode.

Do you mean that if one compiler fails we should still try the other one, just to see? I think this makes sense, but want to make sure I understand the comment.

On error codes:

Something that is not well communicated in this patch is that I'm using the `0` exit code to communicate that the script completed in a controlled manner, _and the output is valid JSON_, which makes it easier for the consumer.

Ultimately we need to think about what this script is for, what it needs to communicate, and how best to do that. I'm not sure that adding a whole bunch of extra exit codes is going to be that helpful.


So currently, the codes are:

- 0 if exit was controlled (output is valid json if `--json` is set)
- 1 if unhandled exception

And some alternatives are (I'm using `2` here but we can skip it if you prefer)

- 0 if equal
- 1 if unhandled exception (no guarantees on output)
- 2 if different (output is valid json with --json flag)
- 3 if a compiler failed (output is valid json with --json flag)

or,

- 0 if equal
- 1 if unexpected exception
- 2 if controlled failure, which includes both if we get a diff or if a compiler fails (output is valid json)


and of these options I prefer the second.

resources/scripts/ttx_diff.py
cmyr: I think it's more recognizing python's unique strengths, but 🤷

PR #862, opened 2024-07-04
[fontc_bulk] Call out to ttx_diff script

cmyr:
This adds the ability to diff lots of fonts and collect the aggregate results.

- this is based on #852, but also requires #860 and #861.

- This currently just outputs JSON, instead of fancier reports
- `fontmake` fails on a ton of sources, presumably because we are passing the wrong args? I think we should be using `gftools builder --generate` to generate recipes from config files or something? For instance we always pass `-o variable` to fontmake, even when we aren't building a variable font.
- but we get reasonable output for some fonts, and in that case it looks like,

```json
{
  "success": {
    "../gf_fonts/Advent/sources/AdventPro.designspace": {
      "diffs": {
        "GPOS": 1.0,
        "HVAR": 0.99195546,
        "OS_2": 0.9946581,
        "gasp": "fontmake",
        "glyf": 0.9999832,
        "gvar": 0.9998928,
        "head": 0.9967949,
        "hhea": 0.9981982
      }
    },
    "../gf_fonts/Briem-Hand/sources/BriemHand.glyphs": {
      "diffs": {
        "(mark/kern)": 0.99426824,
        "GDEF": 0.9837696,
        "GPOS": 0.9656813,
        "GSUB": 0.9259056,
        "OS_2": 0.99946666,
        "avar": 0.95606697,
        "fvar": 0.99673414,
        "glyf": 1.0,
        "gvar": 0.99999917,
        "head": 0.9984051,
        "maxp": 0.9983165,
        "name": 0.96020263
      }
    }
}
```

where the float values are a handwavey 'difference ratio' between the outputs, and if only one compiler generated a given table the value is that compiler's name. Identical tables are omitted (maybe they shouldn't be?).

### DIFF
--- a/fontc_crater/Cargo.toml
+++ b/fontc_crater/Cargo.toml
@@ -11,6 +11,8 @@ readme = "README.md"
 fontc = { version = "0.0.1", path = "../fontc" }
 
 google-fonts-sources = "0.1.0"
+
+rayon.workspace = true
 write-fonts.workspace = true
 serde.workspace = true
 thiserror.workspace = true
--- a/fontc_crater/src/args.rs
+++ b/fontc_crater/src/args.rs
@@ -21,10 +21,13 @@ pub(super) struct Args {
     /// Optional path to write out results (as json)
     #[arg(short = 'o', long = "out")]
     pub(super) out_path: Option<PathBuf>,
+    /// for debugging, execute only a given number of fonts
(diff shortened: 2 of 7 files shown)


dfrg (approved):
I won’t claim to understand everything that’s going on here but this looks reasonable to me.

PR #869, opened 2024-07-17
[fontc] Use explicit threadpool

cmyr:
I ran into an issue where both fontc and fontc_crater (using fontc as a dependency) were using the implicit global threadpool, and this led to a deadlock where a new compilation operation was started on each available thread, which meant there were no threads available for the actual work to be done.

Using an explicit threadpool in fontc ensures this can't happen in the future.


dfrg (approved):
Approving to unblock progress but this feels like an architectural footgun if we can deadlock based on threadpool size.

PR #870, opened 2024-07-17
[glyphs] Ignore unknown glyph categories

cmyr:
A number of our crater fails are the result of glyphs that have categories or subcategories that do not exist in the current version of glyphs.

With this patch, instead of erroring when we encounter these categories we will just ignore them, and fall through as we would when no explicit category was set (in which case we infer it based on unicodes and/or glyph name.)

This also changes the sub_category field on Glyph to be non-optional, since `Subcategory` has its own `None` variant, and it feels like asking for trouble if we have two different ways of indicating none-ness.


dfrg (approved):
Looks good!

PR #874, opened 2024-07-18
[glyphs] Improve g2 -> g3 metric conversion

### LINKED ISSUE #873: [glyphs] v2 to v3 metrics conversion looks buggy
This is acknowledged by an inline TODO:

https://github.com/googlefonts/fontc/blob/ff40153a6e77a211d55d5a7e29a5d278c87d55ca/glyphs-reader/src/font.rs#L1234

and some manual testing is backing up that suspicion.

I currently see two likely issues with the current approach:

- glyphs v2 allowed for adding custom alignment zones, so the assertion that `len(zones) <= len(metrics)` is not assured, which causes crashes on some inputs (https://github.com/googlefonts/fontc/blob/ff40153a6e77a211d55d5a7e29a5d278c87d55ca/glyphs-reader/src/font.rs#L1235-L1240)
- and it looks like at least some fonts do not use the metric ordering that we assume (for instance Alexandria (https://github.com/Gue3bara/Alexandria) has ascender, cap, x-, baseline, descender) and so we should assign them by lining up the lhs and rhs as suggested in the todo.
- there cannot be an alignment zone for 'italic angle', but we try to assign it one anway.

- In addition to this, I wonder if we should be saving any custom alignment zones so they can be set as blue zones in the `post` table? I notice that `glyphsLib` does this: https://github.com/googlefonts/glyphsLib/blob/cedaacdc2bd17e879c3332b30e567ce5bafa0f37/Lib/glyphsLib/builder/blue_values.py#L18

cmyr:
In particular we are now more careful about associating alignment zones with the appropriate metrics.

This mostly resolves #873, with the exception that we do not correctly add additional metrics for any defined alignment zones which do not correspond to standard metrics.

This issue (in particular the removed assert) was causing us to panic on eight inputs.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -1226,30 +1226,36 @@ impl RawFont {
                 master.italic_angle,
             ]
             .into_iter()
-            .map(|pos| pos.and_then(|v| (v != 0.0).then_some(v)))
-            .map(|pos| RawMetricValue { pos, over: None })
+            .map(|pos| RawMetricValue {
+                pos: pos.filter(|x| *x != 0.),
+                over: None,
+            })
             .collect();
 
             // "alignmentZones is now a set of over (overshoot) properties attached to metrics"
-            // TODO: are these actually aligned by index or do you just lookup lhs to get over from rhs
-            assert!(
-                master.alignment_zones.len() <= metric_values.len(),
-                "{} should be <= {}",
-                master.alignment_zones.len(),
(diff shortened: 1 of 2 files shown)


dfrg (approved):
Seems like a weird way to specify these things but the fix lgtm.

rsheeter:
> This issue (in particular the removed assert) was causing us to panic on eight inputs.

I love that we can now make statements like this! We've come a long way :)

PR #875, opened 2024-07-18
[crater] report failures as both/only fontc/only fontmake

cmyr:
This is more useful when deciding what issues to triage.


dfrg (approved):
Keep on keepin' on.

PR #876, opened 2024-07-18
[glyphs] Convert non-metric alignment zones to metrics

### LINKED ISSUE #873: [glyphs] v2 to v3 metrics conversion looks buggy

cmyr:
This is followup on #874 and should close #873.

The implementation is based on looking at what actually happens to glyph sources if they contain additional alignment zones and are converted from v2 to v3, in glyphs.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -1213,8 +1213,10 @@ impl RawFont {
             })
             .collect();
 
+        let mut non_metric_alignment_zones = vec![vec![]; self.font_master.len()];
+
         // in each font master setup the parallel array
-        for master in self.font_master.iter_mut() {
+        for (i, master) in self.font_master.iter_mut().enumerate() {
             // Copy the v2 metrics from actual fields into the parallel array rig
             // the order matters :(
             let mut metric_values: Vec<_> = [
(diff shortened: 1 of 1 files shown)


glyphs-reader/src/font.rs
```diff
@@ -1256,10 +1258,48 @@ impl RawFont {
...
+        let mut new_metrics = HashMap::new();
+        for pos in non_metric_alignment_zones
+            .iter()
+            .flat_map(|master| master.iter().map(|(pos, _)| *pos))
+        {
+            if !new_metrics.contains_key(&pos) {
+                let next_zone = new_metrics.len() + 1;
+                let idx = self.metrics.len();
+                self.metrics.push(RawMetric {
+                    type_: Some(format!("zone {next_zone}")),
```
dfrg, line 1279: If we’re going to retain these, I’d suggest capturing them into `blue_values` and `other_blues` fields as glyphslib does. I don’t think storing them as “zone _n_” has any useful meaning.

cmyr: as best I understand things that would happen during during the glyphs2fontir stage? All I'm trying to do here is convert v2 to v3, so that when we read glyphs sources we have a single normalized representation at the end.

dfrg: Ah, okay. That makes sense then. I’ll admit to not being familiar with the differences between these formats.

dfrg (approved):
Ignoring my misunderstanding of the purpose, code lgtm!

PR #880, opened 2024-07-23
[glyphs] Parse possible 'disabled' flag in custom params

cmyr:
This was causing us to fail to parse 28 inputs in our test set.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -12,7 +12,6 @@ use std::str::FromStr;
 use std::{fs, path};
 
 use crate::glyphdata::{Category, GlyphData, Subcategory};
-use crate::plist::FromPlist;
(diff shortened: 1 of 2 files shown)


glyphs-reader/src/font.rs
```diff
@@ -494,13 +501,15 @@ impl FromPlist for CustomParameters {
...
+                        tokenizer.eat(b';')?;
+                        break;
                     }
                     _ => return Err(Error::SomethingWentWrong),
                 }
-                tokenizer.eat(b';')?;
             }
 
-            if let (Some(name), Some(value)) = (name, value) {
+            if let Some((name, value)) = name.zip(value).filter(|_| !disabled) {
```
dfrg, line 512: Haven’t seen this pattern before but will definitely store for later :)

dfrg (approved):

PR #881, opened 2024-07-24
[crater] run targets in parallel, not repos

cmyr:
A repo can have multiple targets, and we were running all the targets in a given repo sequentially; this isn't a huge problem except that it meant that if a repo containing multiple sources was run near the end (or took a long time) we wouldn't compile those sources in parallel, causing unnecessary delays.

This is particularly annoying for the noto latin/greek/cryillic fonts, because ttx_diff on them can currently take multiple hours, and we were doing them one at a time.


dfrg (approved):
Seems like an obvious improvement!

PR #882, opened 2024-07-24
[glyphs] Remove Error::SomethingWentWrong

cmyr:
And replace it with two new variants which try to capture more context. This will hopefully make it easier to debug glyphs parsing failures.


dfrg (approved):
👍

PR #883, opened 2024-07-24
[glyphs] Parse plist data blocks

cmyr:
The ascii plist format has syntax for inline binary data, which our plist lib wasn't parsing. This adds support for parsing that data, which has compiling another 5 sources from our test set.

### DIFF
--- a/glyphs-reader/src/plist.rs
+++ b/glyphs-reader/src/plist.rs
@@ -21,14 +21,19 @@ pub enum Plist {
     String(String),
     Integer(i64),
     Float(OrderedFloat<f64>),
+    Data(Vec<u8>),
 }
 
-#[derive(Clone, Debug, thiserror::Error)]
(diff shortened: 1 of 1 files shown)


glyphs-reader/src/plist.rs
```diff
@@ -351,8 +369,43 @@ impl Plist {
...
+            0..=9 => ('0' as u32 as u8 + val).into(),
+            10..=15 => (('a' as u32 as u8) + val - 10).into(),
+            _ => unreachable!("only called with values in range 0..=15"),
+        }
+    }
+
+    [to_hex_digit(byte >> 4), to_hex_digit(byte & 0x0f)]
+}
+
+fn byte_from_hex(hex: [u8; 2]) -> Result<u8, Error> {
```
dfrg, line 394: I feel like I've written something like this often enough that it should be in core.

dfrg (approved):
Nice and clean!

PR #886, opened 2024-07-25
[chore] Update fontations deps

cmyr:
This brings in the split between GlyphId and GlyphId16, as well as minor changes to the LookupFlag API.


dfrg (approved):
And now the hard work of actually supporting big glyph ids can begin…

PR #888, opened 2024-07-30
[propagate_anchors] do not sort anchors by name in test

anthrotype:
this is unnecessary, the order we generate is not random, we use IndexMap and we append in the order in which components appear in the composite glyph, therefore it's better to keep the resulting ordering of propagated anchors (and adjust the test file accordingly) instead of artificially sorting the anchors by name in the test code only, just to make the test pass. In Python glyphsLib we use (ordered) dicts that remember insertion order so we will generate identical propagated anchors as well. The order of anchors is not semantically meaningful anyway, so this is a minor cosmetic change only.


cmyr (approved):
Welcome back! 🤝

PR #890, opened 2024-07-31
propagate_anchors: fix mis-named test case, add specific one for *origin anchors

anthrotype:
`origin_anchor` test case fails at the moment, still investigating why

### DIFF
--- a/glyphs-reader/src/propagate_anchors.rs
+++ b/glyphs-reader/src/propagate_anchors.rs
@@ -859,7 +859,7 @@ mod tests {
             .add_glyph("aacute", |glyph| {
                 glyph
                     .add_component("a", (0, 0))
-                    .add_component("acutecomb", (96, -32));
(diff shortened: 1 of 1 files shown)


anthrotype:
The test passes now. The code is correct, the input values were wrong (not matching the expected result).

But I have discovered something interesting relating to the `*origin` anchor. When you modify its position in the glyph editing view, the composite glyphs that reference that particular glyph you just edited do not immediately respond to the change, but only after they are double-clicked... I think it may have to do with the automatic component alignment feature that keeps components in precomposed glyphs (aacute) anchored to one another through the same anchors that are used for OTL mark positioning, are are also influenced by this *origin anchor but not immediately but only upon them subsequently being edited or modified...
That might explain why the position of acutecomb was kind of out of sync in the original test case (shifted -20 units) compared with its correct position.

cmyr (approved):
good catch!

PR #892, opened 2024-08-02
Glyphs fea syntax

cmyr:
This adds support in fea-rs and fontc for compiling basic glyphsapp number values. based on #891.

- includes logic for evaluating arbitrary (simple) arithmetic expressions (support for `+-*/`, no parens)
- named number values are resolved via a new method on `VariationInfo` trait. This is maybe a minor mixing of concerns, the alternative would be to have yet another trait that is only used when the source includes glyphsapp sytanx?
- we compile these values directly into variable metrics, like we do with the current variable FEA syntax proposal
- this gets us compiling another 10 sources


fea-rs/src/compile/glyphsapp_syntax_ext.rs
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    #[test]
+    fn float_math() {
+        let text = "${padding/2.4}";
+        let expr = parse_expr(text);
+        let val: ResolvedValue = resolve_glyphs_app_expr(&expr, |_| simple_number_value());
+        assert!(ordered_eq(&val, [4i16, 10]));
+    }
+}
```
rsheeter, line 222: I don't seem to see any tests for invalid states, malformed input, etc producing errors?

fea-rs/src/compile/validate.rs
```diff
@@ -44,6 +44,8 @@ pub struct ValidationCtx<'a, V: VariationInfo> {
     condition_set_defs: HashMap<SmolStr, Token>,
     aalt_referenced_features: HashMap<Tag, typed::Tag>,
     all_features: HashSet<Tag>,
+    // we cache these because they're reused a lot and slightly expensive to query
+    glyphs_app_number_idents: HashSet<SmolStr>,
```
rsheeter, line 48: suggest also explaining (or linking to explanation) what the heck a glyphs app number ident is

fea-rs/src/compile/glyphsapp_syntax_ext.rs
```diff
@@ -0,0 +1,222 @@
...
+            return false;
+        };
+        let mut ordered: Vec<_> = lhs.iter().collect();
+        ordered.sort_by_key(|(k, _)| *k);
+        let ordered = ordered.iter().map(|(_, v)| **v).collect::<Vec<_>>();
+        ordered == rhs
+    }
+
+    #[test]
+    fn scalar_arithmetic() {
```
rsheeter, line 174: Do these expressions allow brackets to influence precedence? - it seems like you'd want that but I don't see tests for it

fea-rs/src/compile/glyphsapp_syntax_ext.rs
```diff
@@ -0,0 +1,222 @@
...
+        let text = "${2 + 4 * 5 / 7 - 3}";
+
+        let expr = parse_expr(text);
+        let val = resolve_glyphs_app_expr(&expr, |_| Default::default());
+        let expected = (2f64 + 4. * 5. / 7. - 3.).round() as i16;
+        assert_eq!(val, ResolvedValue::Scalar(expected));
+    }
+
```
rsheeter, line 182: Missing tests: just one value, either literal or named

1. `${padding}` is valid?
1. `${42}` is valid?

fea-rs/src/compile/validate.rs
```diff
@@ -1322,6 +1332,60 @@ impl<'a, V: VariationInfo> ValidationCtx<'a, V> {
...
+    fn validate_glyphsapp_number_value(&mut self, value: &typed::GlyphsAppNumberValue) {
+        match value {
+            typed::GlyphsAppNumberValue::Expr(expr) => self.validate_glyphsapp_number_expr(expr),
+            typed::GlyphsAppNumberValue::Ident(ident) => self.validate_glyphsapp_number_name(ident),
+        }
+    }
+
+    fn validate_glyphsapp_number_expr(&mut self, expr: &typed::GlyphsAppNumberExpr) {
+        // ensure pattern is (value) (op) (value) [(op) (value)]*
+        // and ensure names exist.
```
rsheeter, line 1345: The implementation seems a touch more complicated than the comment would lead me to expect?

1. I suspect pattern is `value [op value]*` unless just a value is meant to be illegal?
   * We should have a test for this case, noted @ https://github.com/googlefonts/fontc/pull/892/files#r1703336046
1. If we iterate `chunks(2)` `[0]` must be a value and `[1]` must be an op if present
1. A separate loop to validate name might be tidier and likely of immaterial impact on perf

fea-rs/src/parse/grammar/metrics.rs
```diff
@@ -131,6 +146,138 @@ fn eat_metric(parser: &mut Parser, recovery: TokenSet) -> bool {
     true
 }
 
+// $pad or ${pad * 2 + 5}
```
rsheeter, line 149: What about `${pad}`?

fea-rs/src/parse/grammar/metrics.rs
```diff
@@ -131,6 +146,138 @@ fn eat_metric(parser: &mut Parser, recovery: TokenSet) -> bool {
...
+            })
+        } else {
+            parser.err_recover("expected '$ident' or '${predicate}'", recovery);
+            false
+        }
+    })
+}
+
+// generally just eat values and operators; we will validate later
+// the one special case is '2-2' which we need to ensure gets parsed as (NUM)(-)(NUM)
```
rsheeter, line 174: Should we have a test case for something like `2 - -2` ?

rsheeter: Can we negate named values, e.g. `-padding`

fea-rs/test-data/parse-tests/bad/glyphsapp_number_bad.ERR
```diff
@@ -0,0 +1,23 @@
...
+in ./test-data/parse-tests/bad/glyphsapp_number_bad.fea at 3:13
+  | 
+3 |     pos A B $0;
+  |              ^
+
+error: expected ident or number
+in ./test-data/parse-tests/bad/glyphsapp_number_bad.fea at 4:22
+  | 
+4 |     pos C D ${pad * * 2};
+  |                       ^
```
rsheeter, line 11: Is `* * ` exponentiation? If not, shouldn't we get something like expected value for the second * per the `value [op value]*` rule?

fea-rs/test-data/parse-tests/good/glyphsapp_number_value.fea
```diff
@@ -0,0 +1,6 @@
+# https://glyphsapp.com/learn/tokens#g-number-values
+feature test {
+    pos A B $pad;
+    pos C D <$pad 0 ${pad * 2 - pad + 4 / 1 - 1} 0>;
+    pos E F <${pad-2} ${pad+2} ${pad/2.0-pad} ${pad*2}>;
+} test;
```
rsheeter, line 6: If valid please add `${pad}`, `${-pad}`, ${pad - -2}`, etc

rsheeter (approved):
Couple of small questions/comments but generally lgtm

fea-rs/src/compile/glyphsapp_syntax_ext.rs
cmyr: I did not see brackets in any of the official samples so I did not implement them. Including them will be a parsing error.

If we ever see them used in a font we can revisit. I'll add a test to ensure this doesn't parse.

fea-rs/src/compile/glyphsapp_syntax_ext.rs
cmyr: good catch, added those tests.

fea-rs/src/parse/grammar/metrics.rs
cmyr: we'll treat it as a valid expression.

fea-rs/src/parse/grammar/metrics.rs
cmyr: > Should we have a test case for something like `2 - -2` ?

There's a unit test for this in `glyphsapp_syntax_ext.rs`.

> Can we negate named values, e.g. `-padding`

no, we don't parse this.

fea-rs/test-data/parse-tests/bad/glyphsapp_number_bad.ERR
cmyr: no, we don't support exponentiation; this is a test that we error when there are two operators in a row.

fea-rs/src/compile/glyphsapp_syntax_ext.rs
cmyr: those things are part of the parse tests; this is just testing that we are actually evaluating valid expressions as expected.

PR #893, opened 2024-08-05
[ttx_diff] Run otl-normalizer in parallel

cmyr:
For very large fonts one of the biggest time sinks is actually normalization. We need to normalize both outputs; previously we were doing this sequentially, which is totally unnecessary; with this patch we ensure that we normalize both fonts at the same time (using python threading, which is adequate since all the actual work is happening in a child process.)

For a simple test input this reduced runtime from 12m28s to 8m56s.

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -42,6 +42,7 @@ import json
 import shutil
 import subprocess
 import sys
+import threading
 from cdifflib import CSequenceMatcher as SequenceMatcher
 from typing import MutableSequence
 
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -115,6 +116,22 @@ def ttx(font_file: Path, can_skip: bool):
...
     return ttx_file
 
+# run normalizer on both fonts at once; because this is an external process
+# we can use the threading module
+def run_layout_normalizer(cargo_manifest_path: Path, build_dir: Path, fontc_ttf: Path, fontmake_ttf: Path):
+
+    can_skip_fontc = can_skip(build_dir, "fontc")
+    can_skip_fontmake = can_skip(build_dir, "fontmake")
+    t1 = threading.Thread(target=simple_gpos_output, args = [cargo_manifest_path, fontc_ttf, build_dir / "fontc.markkern.txt", can_skip_fontc])
+    t2 = threading.Thread(target=simple_gpos_output, args = [cargo_manifest_path, fontmake_ttf, build_dir / "fontmake.markkern.txt", can_skip_fontmake])
```
rsheeter, line 126: This looks sus, why would we need to start a thread to start a process? Just start the processes?

rsheeter (approved):
LGTM if one of:

1. The thread is removed; we just launch processes
1. A comment is added explaining why we need a thread whose sole job is to run a process

resources/scripts/ttx_diff.py
rsheeter: https://docs.python.org/3/library/subprocess.html#subprocess.Popen should be able to do non-blocking launches

PR #894, opened 2024-08-06
[fea-rs] Allow unescaped 'NULL' keyword as glyph name

cmyr:
The spec is explicit that if a keyword is used as a glyphname than it needs to be escaped with '\', but feaLib does not enforce this and a bunch of noto fonts use 'NULL' (unescaped) as a glyph name.

Making this work is pretty trivial, and this gets us compiling a bunch of (12) additional sources, so I'm happy to compromise principles for expediency.


dfrg (approved):
> so I'm happy to compromise principles for expediency.

Seems to be a requirement if you want to work with fonts :)

PR #895, opened 2024-08-07
[ttx_diff] Actually return otl-normalizer results

cmyr:
Apparently Thread.join does not actually return the output of whatever function was called; you need to do some extra work to get that information.


rsheeter (approved):
LGTM, still think a good followon would be to drop the threads. Exploratory work to that end in https://github.com/googlefonts/fontc/pull/896.

cmyr:
closing this because it afaict the otl-normalizer runtime really wasn't a major slowdown. If we want to do something along these lines we can go with #896.

PR #896, opened 2024-08-07
[ttx_diff] Launch processes directly instead of using threads

rsheeter:
Built on #895


cmyr:
So after applying #897 and comparing timing with and without this patch I'm not seeing much difference, so I think this was a white herring. Thanks for the cleanup though!

PR #897, opened 2024-08-08
[ttx_diff] Be less quadratic

cmyr:
While resolving off-by-ones we were doing a relatively expensive linear search for a matching item; this puts all the items into a dictionary ahead of time and queries that instead, which offers a huge speedup on large fonts (from a small number of hours to a small number of minutes)

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -307,12 +307,13 @@ def allow_some_off_by_ones(
         return
 
     coord_tag = coord_holder.rpartition("/")[-1]
+    # put all the containers into a dict to make querying more efficient:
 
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -328,12 +328,13 @@ def allow_some_off_by_ones(
         return
 
     coord_tag = coord_holder.rpartition("/")[-1]
+    # put all the containers into a dict to make qeurying more efficient:
```
dfrg, line 331:
```suggestion
    # put all the containers into a dict to make querying more efficient:
```

dfrg (approved):
seems like a modest improvement :)

PR #899, opened 2024-08-14
Checkpoint 'crater in CI' code.

cmyr:
This includes the code to run ttx_diff in a CI context, tracking changes overtime and generating a basic HTML page to display them.

With this landed, there are a few small tasks remaining before we can actually make this useful
- finish up the shell scripts that will be invoked to start a CI run
- set up github pages so we can publish the results
- iron out various little API questions that will arise once we know how this will actually be run
- document the process involved in setting this up and getting it to work

The reports are currently very bare bones: a simple table that looks like this:

Screenshot 2024-08-14 at 12 16 06 PM

but it will be easy to modify the table generation code going forward, to make it more useful.


fontc_crater/src/ci.rs
```diff
@@ -0,0 +1,118 @@
...
+    let fontc_rev = super::get_git_rev(None).unwrap();
+    if prev_runs
+        .last()
+        .map(|prev| prev.fontc_rev == fontc_rev)
+        .unwrap_or(false)
+    {
+        // eventually we will want an argument to force rerunning, which we'll
+        // use when we update fontmake or the input data set
+        eprintln!("fontc rev is unchange from last run, skipping");
+        std::process::exit(0);
```
rsheeter, line 82: I really really dislike functions that kill my process, perhaps we an return a specific error? Also ... just because fontc didn't change doesn't mean fontmake or the sources didn't?

rsheeter (approved):

rsheeter:
Nit: based on the screenshot we might be coloring green (good) backwards a few places, e.g. more got diff and more diff% are bad iiuc

fontc_crater/src/ci.rs
cmyr: This isn't an error condition, this just means we have no more work to do, and we're exiting happily.

cmyr:
'diff %' is actually 'similarity %', so that's just a confusing name. And the 'got diff' number going up is good if it means we had fewer errors, but bad if it means we had fewer identical matches? maybe I should just not color that one.

PR #902, opened 2024-08-16
Generate LigCaretList table

### LINKED ISSUE #809: Generate ligature caret table
This is something that fontmake does for oswald but that we don't. Closely related to #805, in pythonland both of these things are handled by the [`GdefFeatureWriter`](https://github.com/googlefonts/ufo2ft/blob/main/Lib/ufo2ft/featureWriters/gdefFeatureWriter.py)

cmyr:
~*This is currently patching fontations to bring in https://github.com/googlefonts/fontations/pull/1090, I'll do a release when that gets merged*~

This patch generates a `LigCaretList` entry for any glyphs which include an anchor name beginning with `caret_` or `vcaret_`.

- currently this happens during mark generation, which is convenient since we have access to all the anchors there.
- this matches the `LigCaretList` produced by `fontmake` for oswald.
- closes #809

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,8 +22,8 @@ rayon = "1.6"
 icu_properties = "1.4"
 
 # fontations etc
-write-fonts = { version = "0.28.1", features = ["serde", "read"] }
-skrifa = "0.20.0"
+write-fonts = { version = "0.29.0", features = ["serde", "read"] }
+skrifa = "0.21.0"
 norad = "0.12"
 
 # dev dependencies
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -165,9 +165,9 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
         // NOTE: this is the easiest place for us to do this, but we
         // could potentially be more performant by running this in parallel,
         // immediately after parsing?
-        self.run_feature_writer_if_present();
+        let lig_carets = self.run_feature_writer_if_present();
 
-        self.finalize_gdef_table();
+        self.finalize_gdef_table(lig_carets);
         self.features
             .finalize_aalt(&mut self.lookups, &self.default_lang_systems);
         self.features.dedupe_lookups();
@@ -263,9 +263,10 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
         ))
     }
 
-    fn run_feature_writer_if_present(&mut self) {
+    // returns the ligcaret values; we add them after finalizing gdef
+    fn run_feature_writer_if_present(&mut self) -> BTreeMap<GlyphId16, Vec<CaretValue>> {
         let Some(writer) = self.feature_writer else {
-            return;
+            return Default::default();
         };
 
         let mut builder = FeatureBuilder::new(
@@ -282,6 +283,7 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
             .values_mut()
             .for_each(|feat| feat.base.iter_mut().for_each(|id| *id = id_map.get(*id)));
         self.features.merge_external_features(builder.features);
+        builder.lig_carets
     }
 
     /// Infer/update GDEF table as required.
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -53,3 +53,7 @@ members = [
     "otl-normalizer",
     "fontc_crater",
 ]
+
+[patch.crates-io]
```
rsheeter, line 57: I think we should view this as submit-blocking, lets just release fontations.

fea-rs/src/compile/compile_ctx.rs
```diff
@@ -331,6 +333,10 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
             gdef.mark_glyph_sets = sorted.into_iter().map(|(_, cls)| cls).collect();
         }
 
+        if gdef.ligature_pos.is_empty() {
+            gdef.ligature_pos = user_lig_carets;
+        }
```
rsheeter, line 338: else ... warn if the user had specified carets? Or comment to explain why just ignoring the user carets is fine.

Cargo.toml
cmyr: totally, just waiting for a review.

fea-rs/src/compile/compile_ctx.rs
cmyr: oops this was just a bad name, the `user_lig_carets` are the ones generated by the compiler.

fea-rs/src/compile/tables/gdef.rs
```diff
@@ -40,17 +43,26 @@ pub struct GdefBuilder {
...
+            self.build_lig_caret_list(&mut var_store),
             self.build_mark_attach_class_def(),
         );
 
         table.mark_glyph_sets_def = self.build_mark_glyph_sets().into();
-        if let Some((var_store, key_map)) = self.var_store.clone().map(VariationStoreBuilder::build)
-        {
+        // only use the var_store if we had one set at the start of this fn
+        let var_store = Some(var_store).filter(|_| self.var_store.is_some());
+        if let Some((var_store, key_map)) = var_store.map(VariationStoreBuilder::build) {
```
rsheeter, line 63: Creating a mutable varstore that isn't self.var_store and then having to carefully remember if it's bogus or not seems fairly ugly. Why is that necessary? We can't or shouldn't code it to run against self.var_store and keep the original code here?

fontbe/src/features/marks.rs
```diff
@@ -621,6 +627,62 @@ fn resolve_anchor_once(
     Ok(anchor)
 }
 
+fn get_ligature_carets(
```
rsheeter, line 630: Is there a python equivalent to link to?

I suggest we should have a comment somewhere, perhaps here, that captures, or links to, the discussion of whether we need the table, why / who uses it.

fea-rs/src/compile/tables/gdef.rs
rsheeter: At a glance it seems like it might be ok for build to take `&mut self`, as CompilationCtx::build - which appears to be the primary caller - does?

cmyr: agree that this is annoying, and is the consequence of a few things colliding:
- we need to mutate the var store to add deltas to it, in `build_lig_caret_list`.
- for Annoying Rust Reasons, `Option<&mut T>` is not `Copy`. This means you can't do something like,

```rust
let mut thing = Thing;
let thing_ref = Some(&mut Thing);
for _ in 0..5 {
    do_thing(thing_ref); // < error: `thing_ref` moved in previous loop pass
}
```
- and as a consequence of this, `build_lig_caret_list` (along with many other similar functions) take `&mut VariationStoreBuilder` instead of `Option<&mut VariationStoreBuilder>`
- and this means that even if there is no varstore at this point (and this is late in compilation, if this was a variable font we would have a varstore) we need to create one in order to call that method; but if we created one just for this we don't care to keep it on, and so we discard it afterwards.

having `&mut self` wouldn't really change anything; we would still need to create a dummy varstore in order to build the lig caret list, and we would still discard it afterwards.

rsheeter: Per IM checking if the varstore is empty instead of `Some(var_store).filter(|_| self.var_store.is_some())` should suffice.

rsheeter (approved):
LGTM subject to `Some(var_store).filter(|_| self.var_store.is_some())` switching to just a check for var_store not being empty. One hopes 0 axis varstores are always empty since no delta that makes sense seems possible.

cmyr:
changed to a simpler version of this original check; if for some crazy reason there were no variations for any values in gpos but there happen to be variations in the ligature carets the empty check would fail, and we'd compile garbage, so I prefer to just always chuck the var_store if we have no deltas at this point.

PR #907, opened 2024-08-21
[glyphs] Detect cycles during anchor propagation

cmyr:


anthrotype:
I think your current cycle detection logic is leading to false positives, have a look at https://github.com/googlefonts/fontc/pull/909

cmyr:
oh yes great catch, the impl was broken. I've fixed it by tracking not just the names of glyphs but also the length of the queue when a glyph is pushed to the back; if we see the same glyph twice without the queue getting smaller we must have a cycle.

anthrotype (approved):
lgtm, I copied your solution in https://github.com/googlefonts/glyphsLib/pull/1023

I think we should also skip these cyclical composite glyphs in the main propagate anchors loop. In the glyphsLib PR I exclude them from the list returned from depth_sorted_composite_glyphs. There's no point attempting to process them here, there are no anchors to inherit if we can't dereference their components.

cmyr:
good idea, I'll do the same.

PR #911, opened 2024-08-22
[ttx_diff] Make fontmake lookuplist order match fontc

cmyr:
The order of items in this list doesn't matter, so we can put it in ascending order, matching fontc.

This doesn't move the needle too much, but I was finding the difference confusing.

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -347,7 +347,23 @@ def allow_some_off_by_ones(
         )
 
 
+# the order of lookups in a feature's lookuplist do not matter;
+# fontc always has them in sorted order but fontmake doesn't, so sort them
+def sort_fontmake_feature_lookups(ttx):
+    gpos = ttx.find("GPOS")
+    if not gpos:
+        return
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -347,7 +347,23 @@ def allow_some_off_by_ones(
...
+# fontc always has them in sorted order but fontmake doesn't, so sort them
+def sort_fontmake_feature_lookups(ttx):
+    gpos = ttx.find("GPOS")
+    if not gpos:
+        return
+    features = gpos.xpath("//Feature")
+
+    for feature in features:
+        items = list(feature.iter())
+        # the first item is 'Feature', the second always a comment
```
anthrotype, line 360: relying on a particular ordering of the elements is a bit brittle; how about collecting the list of all the LookupListIndex elements in this Feature and only then read/write their "value" attrib?

cmyr: that does seem better, yes :)

anthrotype (approved):

PR #913, opened 2024-08-23
glyphs2fontir: add missing 'description' name record nameID=10

anthrotype:
JMM


glyphs2fontir/src/source.rs
```diff
@@ -278,6 +278,7 @@ fn try_name_id(name: &str) -> Option<NameId> {
         "trademarks" => Some(NameId::TRADEMARK),
         "manufacturers" => Some(NameId::MANUFACTURER),
         "designers" => Some(NameId::DESIGNER),
+        "descriptions" => Some(NameId::DESCRIPTION),
```
cmyr, line 281: Just want to double check that it's not 'description'?

cmyr (approved):
one QQ inline

glyphs2fontir/src/source.rs
anthrotype: yeah, like the other localizable strings, it's a plural.

PR #914, opened 2024-08-23
[fea-rs] Match fontmake for class-based pairpos

cmyr:
In particular this means that we will not try to be smart about grouping rules into subtables based on their actual value formats; we will instead minimize the number of subtables by choosing the union of the value formats for each rule.

### DIFF
--- a/fea-rs/src/compile/lookups/gpos_builders.rs
+++ b/fea-rs/src/compile/lookups/gpos_builders.rs
@@ -143,7 +143,7 @@ impl Default for ClassPairPosSubtable {
 
 #[derive(Clone, Debug, Default, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-struct ClassPairPosBuilder(BTreeMap<(ValueFormat, ValueFormat), Vec<ClassPairPosSubtable>>);
+struct ClassPairPosBuilder(Vec<ClassPairPosSubtable>);
 
 impl ClassPairPosBuilder {
     fn insert(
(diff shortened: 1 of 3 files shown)


fea-rs/src/compile/lookups/gpos_builders.rs
```diff
@@ -188,6 +185,14 @@ impl ClassPairPosSubtable {
             .or_default()
             .insert(class2, (record1, record2));
     }
+
+    // determine the union of each of the two value formats
```
rsheeter, line 189: Suggest explaining why, e.g. to match fontmake [link to relevant code]

behdad:
Was the previous logic even correct? Or did it split by value format even for the same first-glyph?

fea-rs/src/compile/lookups/gpos_builders.rs
```diff
@@ -153,19 +153,16 @@ impl ClassPairPosBuilder {
...
             Some(subtable) => !subtable.can_add(&class1, &class2),
         };
-        if add_sub {
-            entry.push(Default::default());
+        if needs_new_subtable {
+            self.0.push(Default::default());
+        }
+        if let Some(subtable) = self.0.last_mut() {
+            subtable.add(class1, class2, record1, record2);
         }
```
rsheeter, line 165: At a glance this seems to have more branching than strictly necessary?

```rust
// If we don't have any subtables or we can't add to the last one make a new one
if !self.0.last().map(|s| s.can_add(...).unwrap_or_default() {
  self.0.push(...)
}
// We definitely have a last now
self.0.last_mut().add(...)

cmyr:
> Was the previous logic even correct? Or did it split by value format even for the same first-glyph?

I'm not sure I understand, but I'm also definitely not sure it was correct. The splitting by value format would only happen for class-based rules.

fea-rs/src/compile/lookups/gpos_builders.rs
cmyr: I wrote it this way instead of using an unwrap, but we could definitely safely unwrap.

rsheeter (approved):
I'm mildly uncomfortable that we can't resolve this by improving normalization if we believe the behavior is equivalent. If it *isn't* then it would be nice to see a test demonstrating there was a bug and it's now fixed. EDIT: per IM, now convinced on basis that matching fontmake is the safest path unless there is a specific, strong, reason to do otherwise.

fea-rs/src/compile/lookups/gpos_builders.rs
rsheeter: Note that this replaces match/if/if not *only* the final unnecessary if

Issue #919, opened 2024-08-27
Glyphs instances' width_class should be mapped to 'wdth' % scale while inferring axis mappings


anthrotype:
When no Axis Mappings nor Axis Location custom parameters are used, we fall back to inferring the axis mappings (for wght and wdth only) from the instances' user-space locations as implicitly defined by their weight_class and width_class properties.

The instance width_class is an int from 1 to 9 (like the OS/2.usWidthClass):
https://docu.glyphsapp.com/#GSInstance.widthClass

we need to map this value to the `wdth` registered axis' percent scale, where 5 is 100% etc.
https://learn.microsoft.com/en-gb/typography/opentype/spec/os2#uswidthclass

for reference, glyphsLib does it here:
https://github.com/googlefonts/glyphsLib/blob/7041311e2658326c419f02ecdc3a040aa094d0b2/Lib/glyphsLib/builder/axes.py#L27

If we don't do that, we end up building incorrect fvar and avar.

e.g. compare fontc vs fontmake's fvar and avar tables of NotoSans.glyphspackage from https://github.com/notofonts/latin-greek-cyrillic

I presume it's this TODO here:

https://github.com/googlefonts/fontc/blob/a85d946e60fff2766e19f07e0e3a75e6162e95a4/glyphs-reader/src/font.rs#L1954

Issue #904, opened 2024-08-19
should update OS/2 {weight,width}Class and post.italicAngle to match wght/wdth/slnt defaults


anthrotype:
when fonttools varLib builds a VF, it ensures that the OS/2 weight and width classes and post.italicAngle match respectively the [wght](https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxistag_wght), [wdth](https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxistag_wdth) and [slnt](https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxistag_slnt) registered axes' default values from fvar:

https://github.com/fonttools/fonttools/blob/0c2652011e51bd78d153b41287bb845530dd0fdf/Lib/fontTools/varLib/__init__.py#L1016-L1038

https://github.com/fonttools/fonttools/blob/0c2652011e51bd78d153b41287bb845530dd0fdf/Lib/fontTools/varLib/__init__.py#L1237-L1239

Right now, fontc doesn't even set the OS/2 weight/width class (it does set the post.italicAngle).

anthrotype:
so when the default master is not the Regular weight and/or normal width, we get different OS/2 classes from the ones that fontmake generates

anthrotype:
Oswald's default happen to be the Regular so this issue doesn't show.

It does show with e.g. notofonts/hentaigana/sources/NotoSerifHentaigana.glyphspackage (where the default is the ExtraLight)

PR #921, opened 2024-08-28
Map Glyphs instance width_class to wdth percent scale when loading axis mappings

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/919

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -13,6 +13,7 @@ use std::{fs, path};
 
 use crate::glyphdata::{Category, GlyphData, Subcategory};
 use ascii_plist_derive::FromPlist;
+use fontdrasil::types::WidthClass;
 use kurbo::{Affine, Point, Vec2};
 use log::{debug, warn};
 use ordered_float::OrderedFloat;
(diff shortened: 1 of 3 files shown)


anthrotype:
failed as expected, reproducer for #919 

https://github.com/googlefonts/fontc/actions/runs/10598084282/job/29369795269?pr=921#step:6:171
```
---- font::tests::loads_global_axis_mappings_from_instances_wdth_glyphs3 stdout ----
thread 'font::tests::loads_global_axis_mappings_from_instances_wdth_glyphs3' panicked at glyphs-reader/src/font.rs:2750:9:

Diff < left / right > :
 RawUserToDesignMapping(
     {
         "Width": RawAxisUserToDesignMap(
             [
                 (
<                    50.0,
>                    1.0,
                     22.0,
                 ),
                 (
                     100.0,
                     41.0,
                 ),
                 (
<                    200.0,
>                    9.0,
                     62.0,
                 ),
             ],
         ),
     },
 )

```

glyphs-reader/src/font.rs
```diff
@@ -1971,7 +1973,22 @@ impl Instance {
...
+                        2 => 62.5,  // Extra-condensed
+                        3 => 75.0,  // Condensed
+                        4 => 87.5,  // Semi-condensed
+                        5 => 100.0, // Medium (normal)
+                        6 => 112.5, // Semi-expanded
+                        7 => 125.0, // Expanded
+                        8 => 150.0, // Extra-expanded
+                        9 => 200.0, // Ultra-expanded
+                        _ => {
+                            panic!("Unsupported instance width_class {}", v);
```
anthrotype, line 1988: glyphsLib also raises NotImplementedError in this situation, which is unlikely enough to warrant an error, not just a warning. The Glyphs Python scripting API docs say this field only has int values from 1 to 9 corresponding to OS/2 usWidthClass.

glyphs-reader/src/font.rs
```diff
@@ -1971,7 +1973,22 @@ impl Instance {
...
+                    match i32::from_str(v).unwrap() {
+                        1 => 50.0,  // Ultra-condensed
+                        2 => 62.5,  // Extra-condensed
+                        3 => 75.0,  // Condensed
+                        4 => 87.5,  // Semi-condensed
+                        5 => 100.0, // Medium (normal)
+                        6 => 112.5, // Semi-expanded
+                        7 => 125.0, // Expanded
+                        8 => 150.0, // Extra-expanded
+                        9 => 200.0, // Ultra-expanded
```
rsheeter, line 1986: Optional, it would be nicer to enumerate the OS/2 values, parse str to that, and then ask the enum for the width %

resources/testdata/glyphs3/WdthVar.glyphs
rsheeter, line 1: I wonder if this could be minimized? - I think it has a bunch of stuff you don't need for the test

rsheeter (approved):

glyphs-reader/src/font.rs
anthrotype: I did that here [190c059](https://github.com/googlefonts/fontc/pull/921/commits/190c059ff4cd7a8bfb238dae3b7fb77b774d8c45), but I'm not sure I like it better than, the previous code was more concise

anthrotype: actually, I'm going to move this enum WidthClass to fontdrasil so that I can use in the FE as well as the BE, I need to map back from a percent value to u16 in order to fix https://github.com/googlefonts/fontc/issues/904

PR #922, opened 2024-08-29
fontbe/os2: set us{Weight,Width}Class to wght/wdth axis defaults

### LINKED ISSUE #904: should update OS/2 {weight,width}Class and post.italicAngle to match wght/wdth/slnt defaults

anthrotype:
Fixes #904

### DIFF
--- a/fontbe/src/os2.rs
+++ b/fontbe/src/os2.rs
@@ -820,17 +820,13 @@ impl Work<Context, AnyWorkId, Error> for Os2Work {
         // in the same way fontmake does when building a VF:
         // https://github.com/fonttools/fonttools/blob/770917d8/Lib/fontTools/varLib/__init__.py#L1016-L1032
         let default_wght = static_metadata
-            .axes
-            .iter()
-            .find(|axis| axis.tag == Tag::new(b"wght"))
+            .axis(&Tag::new(b"wght"))
             .map(|axis| axis.default.into_inner().0)
             .unwrap_or(400.0);
         let us_weight_class: u16 = default_wght.clamp(1.0, 1000.0).ot_round();
 
         let default_wdth = static_metadata
-            .axes
-            .iter()
-            .find(|axis| axis.tag == Tag::new(b"wdth"))
+            .axis(&Tag::new(b"wdth"))
             .map(|axis| axis.default.into_inner().0)
             .unwrap_or(100.0);
         let us_width_class = WidthClass::nearest(default_wdth) as u16;
--- a/fontbe/src/os2.rs
+++ b/fontbe/src/os2.rs
@@ -2,7 +2,10 @@
 
 use std::{cmp::Ordering, collections::HashSet};
 
-use fontdrasil::orchestration::{Access, AccessBuilder, Work};
+use fontdrasil::{
+    orchestration::{Access, AccessBuilder, Work},
+    types::WidthClass,
+};
 use fontir::{ir::GlobalMetricsInstance, orchestration::WorkId as FeWorkId};
 use log::warn;
 use write_fonts::{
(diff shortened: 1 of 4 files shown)


fontbe/src/os2.rs
```diff
@@ -813,6 +816,25 @@ impl Work<Context, AnyWorkId, Error> for Os2Work {
...
+            .map(|axis| axis.default.into_inner().0)
+            .unwrap_or(400.0);
+        let us_weight_class: u16 = default_wght.clamp(1.0, 1000.0).ot_round();
+
+        let default_wdth = static_metadata
+            .axes
+            .iter()
+            .find(|axis| axis.tag == Tag::new(b"wdth"))
+            .map(|axis| axis.default.into_inner().0)
+            .unwrap_or(100.0);
```
rsheeter, line 835: These lines repeat almost identically here and above. I suggest extracting a function or even adding it to static metadata, e.g. `static_metadata.axis(Tag) -> Option<Axis>`

fontdrasil/src/types.rs
```diff
@@ -156,4 +157,18 @@ impl WidthClass {
             WidthClass::UltraExpanded => 200.0,
         }
     }
+
+    pub fn from_percent(percent: f32) -> Self {
```
rsheeter, line 161: Should document what happens if pct doesn't exactly match a class. Changing the name might suffice, e.g. `nearest_class(pct: f32)` I'm immediately clear what I get.

fontdrasil/src/types.rs
```diff
@@ -156,4 +157,18 @@ impl WidthClass {
...
+        let mappings = WidthClass::all_values()
+            .iter()
+            .map(|wc| (wc.to_percent().into(), (*wc as u16).into()))
+            .collect();
+        let v: u16 = PiecewiseLinearMap::new(mappings)
```
rsheeter, line 166: I don't think I'd have expected this to allocate. 

This is testable, I suggest adding tests including out of bounds values. At a glance I'm slightly skeptical this works if the value if out of bounds.

rsheeter: I tried it, `WidthClass::from_percent(42.0)` fails (bonus points: not the way I expected  ) with `attempt to subtract with overflow`

rsheeter: Over-large values also fail.

rsheeter: Exploratory fix (PR to this branch) in https://github.com/googlefonts/fontc/pull/923

rsheeter (approved):
LGTM with a couple of suggestions

fontdrasil/src/types.rs
anthrotype: but why doesn't it work? I thought PiecewiseLinearMap worked the same way as fonttools' equivalent. I used the same function that varLib uses for the same task and naively assumed they'd work the same. Should they not work the same?

anthrotype:
as for post.italicAngle, fontbe already sets that one, I don't think we need to do anything about it. The font developer should already ensure that it matches the slnt default.

> In a variable font that implements 'slnt' variations, the value in the italicAngle field of the 'post' table must match the default 'slnt' value specified in the ['fvar'](https://learn.microsoft.com/en-gb/typography/opentype/spec/fvar) table.

https://learn.microsoft.com/en-gb/typography/opentype/spec/dvaraxistag_slnt

fontdrasil/src/types.rs
rsheeter: Fair point, the non-handling of our of bounds values should be viewed as a bug then. I will file an issue. I still think we prefer not to allocate here.

rsheeter: https://github.com/googlefonts/fontc/issues/924

PR #927, opened 2024-09-03
[ttx_diff] Drop kern/mark lookups in GPOS

cmyr:
Since these are handled by layout normalizer.

This keeps the lookup element itself as well as the 'LookupKind', but removes all other children.

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -362,6 +362,22 @@ def sort_fontmake_feature_lookups(ttx):
         for (i, lookup_index) in enumerate(has_value):
             lookup_index.attrib['value'] = str(values[i])
 
+LOOKUPS_TO_SKIP = set([2, 4, 5, 6]) # pairpos, markbase, marklig, markmark
+
+def remove_mark_and_kern_lookups(ttx):
+    gpos = ttx.find("GPOS")
+    if gpos is None:
+        return
+    for lookup in gpos.xpath("//Lookup"):
+        lookup_type = int(lookup.find("LookupType").attrib["value"])
+        if lookup_type not in LOOKUPS_TO_SKIP:
+            continue
+        # remove all the elements but the type:
+        to_remove = [child for child in lookup if child.tag != "LookupType"]
+        for child in to_remove:
+            lookup.remove(child)
+
+
 def reduce_diff_noise(build_dir, fontc, fontmake):
     sort_fontmake_feature_lookups(fontmake)
     for ttx in (fontc, fontmake):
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -362,6 +362,22 @@ def sort_fontmake_feature_lookups(ttx):
...
             lookup_index.attrib['value'] = str(values[i])
 
+LOOKUPS_TO_SKIP = set([2, 4, 5, 6]) # pairpos, markbase, marklig, markmark
+
+def remove_mark_and_kern_lookups(ttx):
+    gpos = ttx.find("GPOS")
+    if gpos is None:
+        return
+    for lookup in gpos.xpath("//Lookup"):
+        lookup_type = int(lookup.find("LookupType").attrib["value"])
```
rsheeter, line 372: nit: `attrib.get("value", 0)` so a lookup type with no value doesn't blow you up.

resources/scripts/ttx_diff.py
```diff
@@ -362,6 +362,22 @@ def sort_fontmake_feature_lookups(ttx):
...
+        if lookup_type not in LOOKUPS_TO_SKIP:
+            continue
+        # remove all the elements but the type:
```
rsheeter, line 375: explain why it matters to leave the type behind? - my naive thought is it wouldn't

rsheeter: if you did remove the type and you then had an empty lookup could you delete the entire lookup?

rsheeter (approved):

resources/scripts/ttx_diff.py
cmyr: this field should always exist and have a value; if it doesn't I would prefer to know about that and figure out why...

resources/scripts/ttx_diff.py
cmyr: That was my first thought, but if we remove the lookups completely we're leaving invalid ttx, since there will be lookup ids that reference lookups that don't exist. We do expect to always create the same number and type of lookups, so leaving that information in place seemed like a reasonable compromise.

anthrotype:
The GDEF and its ItemVariationStore will still be there and may produce a ttx diff even after the GPOS kern/mark lookups have been expunged from GPOS.
GDEF VarStore holds deltas for the GDEF ligature carets and for the GPOS (and JSTF table, but he latter we never build currently in fontmake so we can expect it to not be there - unless put in place by some ad-hoc postprocess).
Does otl-normalizer support GDEF ligature carets already? If so, then if we have removed all the GPOS lookups with variations that may reference the GDEF VarStore, then we can also omit the entire GDEF VarStore and its ligature carets from the ttx diff comparison.

PR #928, opened 2024-09-04
[fontbe] Match fonttools behaviour for LigCaret rounding

### LINKED ISSUE #810: checklist: Get TTX diff matching for oswald
Once we get #806 merged, we'll be into the home stretch of matching fontmake for oswald. This issue is intended to track the remaining sources of difference.

**Diffs in the normalized `markkern` layout-normalizer output:**

- [x] bigraphs: there is a difference in anchor propagation between fontmake & fontc for glyphs with multiple base components that are not ligatures. The best issue for this is https://github.com/googlefonts/glyphsLib/issues/368#issuecomment-2103445759. I believe cosimo has a fix ready for fontmake.
- [x] anchor propagation for marks: Anchor propagation behaviour differs between fontmake & fontc for some mark glyphs, which causes fontc to treat certain spacing non-combining marks in oswald as bases for the mark feature.  (#811)
- [x] marks for glyphs with no glyph category: in glyphs2fontir we were failing to treat glyphs with no (glyphs.app) category as bases where appropriate. fixed in #846 

**Diffs in GPOS.ttx**
- [ ] there are various diffs here that were the original impetus for layout-normalizer: some lookups have different ids for instance, and delta indexes differ. I think long term we want to just use normalizer on this full table, and ignore it?

**Diffs in GSUB.ttx**
- [x] fontmake is generating an additional `ccmp` feature (two total), but this looks like a bug in fontmake, since the additional feature seems to be identical to the other one, except for a duplicate entry for one lookup? (fixed in https://github.com/fonttools/fonttools/pull/3520)
- [x] fontc and fontmake are choosing different formats for a contextual lookup. Looks like this is related to suboptimal inline multiple substitution lookup reuse; explanation at https://github.com/fonttools/fonttools/issues/3551 and improvement for fontc at #842.
- [x] various other differences similar to GPOS, where some ids are different. I think like with GPOS, the answer here is going to be to just improve layout-normalizer and ignore this table (#800)

**Diffs in GDEF.ttx**
- [x] missing the ligature caret table: (#809)
- [ ] some differences in the GDEF categories; I'm not actually sure how we're generating these, it might be that we're sort of getting it by accident via fea-rs? Some of these differences are related to things like the different handling of digraphs, others will need a bit of investigation. (#805)
- [ ] the ItemVariationStore differs significantly, but this is probably mostly the result of the GPOS diffs, where we have more deltas because we have more items/rules; we can revisit this after GPOS is matching.

cmyr:
When generating variable FEA the various Feature Writers use otround, since variable FEA syntax does not support floats.

This is not an issue with kerning, since glyphsapp does not allow for non-integer kerning values, but it is an issue for LigatureCarets and potentially for other variable mark positions (this needs more investigation)

- closes #810

### DIFF
--- a/fontbe/src/features/marks.rs
+++ b/fontbe/src/features/marks.rs
@@ -20,6 +20,7 @@ use smol_str::SmolStr;
 use write_fonts::{
     tables::{gdef::GlyphClassDef, layout::LookupFlag},
     types::{GlyphId16, Tag},
+    OtRound,
 };
 
 use crate::{
@@ -671,6 +672,10 @@ fn make_caret_value(
         .iter()
         .map(|(loc, pt)| {
             let pos = if is_vertical { pt.y } else { pt.x };
+            // we round here to match fonttools, which can't express non-integer
+            // values in the variable fea that it generates.
+            // https://github.com/googlefonts/ufo2ft/blob/12b68f0c69/Lib/ufo2ft/featureWriters/baseFeatureWriter.py#L435-L436
+            let pos: i16 = pos.ot_round();
 
             (loc.clone(), OrderedFloat::from(pos as f32))
         })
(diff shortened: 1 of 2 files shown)


fontbe/src/features/marks.rs
```diff
@@ -776,16 +781,41 @@ mod tests {
         }
     }
 
+    // when describing an anchor position we almost always use integers, but
+    // for some tests we want to accept floats so we can verify rounding behaviour.
+    // we define our own trait and impl it for only these two types so that the
+    // compiler doesn't need additional type info when we use int/float literals.
```
rsheeter, line 787: This makes me a little uncomfortable but I guess it's not public

cmyr: yes this is just for a test helper.

rsheeter (approved):

PR #931, opened 2024-09-04
[crater] Use a timeout when running ttx_diff

cmyr:
Sometimes ttx_diff hangs, in which case CI never finishes. This sets a timeout for each execution of ttx_diff.


fontc_crater/src/ttx_diff_runner.rs
```diff
@@ -43,11 +48,9 @@ pub(super) fn run_ttx_diff(source: &Path) -> RunResult<DiffOutput, DiffError> {
             }
             Ok(RawDiffOutput::Error(error)) => RunResult::Fail(DiffError::CompileFailed(error)),
         },
+        Some(124) => RunResult::Fail(DiffError::Other("ttx_diff timed out".to_string())),
```
rsheeter, line 51: Nit: seems worth a dedicated error code?

fontc_crater/src/ttx_diff_runner.rs
```diff
@@ -7,11 +7,15 @@ use std::{
 use crate::{Results, RunResult};
 
 static SCRIPT_PATH: &str = "./resources/scripts/ttx_diff.py";
+// in the format expected by timeout(1)
+static TIMEOUT: &str = "5h";
```
rsheeter, line 11: TIMEOUT_FOR_WHAT_EXACTLY?

rsheeter (approved):

fontc_crater/src/ttx_diff_runner.rs
cmyr: this should be uncommon and adding a new category requires modifying a bunch of other code (like the html generation would need to have a new section etc)

PR #932, opened 2024-09-04
[crater] Never show '100%' diff

cmyr:
Since it can only happen as a result of rounding; so do an explicit check and skip rounding if we're rounding to 100.

### DIFF
--- a/fontc_crater/src/ci/html.rs
+++ b/fontc_crater/src/ci/html.rs
@@ -215,11 +215,14 @@ fn make_diff_report(current: &DiffResults, prev: &DiffResults) -> Markup {
             Some(d) => html! ( span.worse { (format!("{d:+.3}")) } ),
         };
         let details = format_diff_report(diff_details);
+        // avoid .9995 printing as 100%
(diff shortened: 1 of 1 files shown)


fontc_crater/src/ci/html.rs
```diff
@@ -142,7 +142,12 @@ fn make_table_body(runs: &[RunSummary]) -> Markup {
             prev.map(|p| p.stats.diff_perc_including_failures),
             true,
         );
-        let diff_fmt = format!("{:.3}", run.stats.diff_perc_including_failures);
```
rsheeter, line 145: I don't love the explicit if 100, how about we just truncate after 3 decimal places, e.g. `format!("{:.3}%", (pct * 1000.0).floor() / 1000.0)  // avoid .9995 printing as 100%` 

The results are slightly more conservative but given the nature of the tool that's OK

rsheeter (approved):
LGTM with a suggested alternate approach

cmyr:
cool yea I prefer your way too.

PR #935, opened 2024-09-05
[fontbe/avar] Fix bug where PLM supplies a bad input to binary_search

### LINKED ISSUE #933: Piecewise linear map behaviour differs on 1.83.0-nightly?!
Using `fontc` @ 1c75077e on NotoSerifRegular.glyphspackage from notofonts/latin-greek-cyrillic produces an avar table like so:

```xml
  <avar>
    <version major="1" minor="0"/>
    <segment axis="wght">
      <mapping from="-0.6667" to="-0.7879"/>
      <mapping from="-0.3333" to="-0.4545"/>
      <mapping from="0.2" to="0.16"/>
      <mapping from="0.4" to="0.36"/>
      <mapping from="0.6" to="0.58"/>
      <mapping from="0.8" to="0.81"/>
    </segment>
    <segment axis="wdth">
      <mapping from="-0.6667" to="-0.7"/>
      <mapping from="-0.3333" to="-0.36664"/>
      <!-- line conspicuously absent here -->
    </segment>
  </avar>
```

As there is no 0->0 mapping for the wdth axis, the default width=100 position is normalised to -0.27496338, all the normal width instances are made too narrow, and the diffenator explodes.

anthrotype:
Fixes #933

the first commit is the test case which currently fails on nightly (our CI probably doesn't test this though), to be followed by the actual fix

### DIFF
--- a/fontbe/src/avar.rs
+++ b/fontbe/src/avar.rs
@@ -1,7 +1,7 @@
 //! Generates a [avar](https://learn.microsoft.com/en-us/typography/opentype/spec/avar) table.
 
 use fontdrasil::{
-    coords::{CoordConverter, DesignCoord, NormalizedCoord},
+    coords::NormalizedCoord,
     orchestration::{Access, Work},
     types::Axis,
 };
@@ -40,18 +40,7 @@ fn default_segment_map() -> SegmentMaps {
 }
 
 fn to_segment_map(axis: &Axis) -> SegmentMaps {
-    // default normalization
-    let mut default_mappings = Vec::new();
-    let mut default_idx = 0;
-    if axis.min < axis.default {
-        default_mappings.push((axis.min, DesignCoord::new(-1.0)));
-        default_idx = 1;
-    }
-    default_mappings.push((axis.default, DesignCoord::new(0.0)));
-    if axis.max > axis.default {
-        default_mappings.push((axis.max, DesignCoord::new(1.0)));
-    }
-    let default_converter = CoordConverter::new(default_mappings, default_idx);
+    let default_converter = axis.default_converter();
 
     // We have to walk twice but we don't expect there to be a lot of values so don't stress
 
--- a/fontdrasil/src/coords.rs
+++ b/fontdrasil/src/coords.rs
@@ -191,6 +191,28 @@ impl CoordConverter {
         }
     }
 
+    /// Initialize default converter from user to predefined -1.0/0.0/1.0 normalized coords.
+    ///
+    /// This is normally the first step of `avar` normalization.
+    /// <https://learn.microsoft.com/en-us/typography/opentype/spec/avar#overview>
+    pub fn default_normalization(
+        min: UserCoord,
+        default: UserCoord,
+        max: UserCoord,
+    ) -> CoordConverter {
+        let mut default_mappings = Vec::new();
+        let mut default_idx = 0;
+        if min < default {
+            default_mappings.push((min, DesignCoord::new(-1.0)));
+            default_idx = 1;
+        }
+        default_mappings.push((default, DesignCoord::new(0.0)));
+        if max > default {
+            default_mappings.push((max, DesignCoord::new(1.0)));
+        }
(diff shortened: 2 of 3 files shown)


anthrotype:
FWIW, it fails like this on Rust version 1.83.0-nightly (4ac7bcbaa 2024-09-04) on my local Linux machine

```rust
running 1 test
test avar::tests::zero_zero_map_should_always_be_present ... FAILED

failures:

---- avar::tests::zero_zero_map_should_always_be_present stdout ----
thread 'avar::tests::zero_zero_map_should_always_be_present' panicked at fontbe/src/avar.rs:216:9:
  left: [(-1.0, -1.0), (-0.6667, -0.7), (-0.3333, -0.3666), (0.0, 0.0), (1.0, 1.0)]
 right: [(-1.0, -1.0), (-0.6667, -0.7), (-0.3333, -0.3666), (1.0, 0.0), (1.0, 1.0)]


failures:
    avar::tests::zero_zero_map_should_always_be_present

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 55 filtered out; finished in 0.00s
```

anthrotype:
I now also wonder if PiecewiseLinearMap::new (and as a result CoordConverter::new) should be fallible, in case we ever construct them with duplicate `from` inputs which would lead to a similar ambiguous situation.

anthrotype:
in fonttools this situation never occurs because mappings are a dict where keys are unique by definition.
When reading a designspace axis' maps, the list of (user, design) tuples is turned into a dict (with a dict comprehension, where last key wins in case of duplicates):
https://github.com/fonttools/fonttools/blob/c2119229cfb02cdb7c5a63374ef29d3d514259e8/Lib/fontTools/designspaceLib/__init__.py#L1031-L1047

not sure if fontmake warns about this ambiguity (maybe glyphsLib does when constructing the DS axes)

rsheeter:
> I now also wonder if PiecewiseLinearMap::new (and as a result CoordConverter::new) should be fallible, in case we ever construct them with duplicate from inputs which would lead to a similar ambiguous situation.

I suggest filing an issue, we should add a test and potentially fix.

I also suggest tweaking the title to more clearly indicate this is a bug in our code, "Fix bug where PLM supplies a bad input to binary_search" or some such. Currently it reads a bit as if we're working around a bug in binary_search which isn't the case.

fontbe/src/avar.rs
```diff
@@ -199,4 +202,29 @@ mod tests {
             dump(to_segment_map(&axis(mappings, 0)))
         );
     }
+
```
rsheeter, line 205: I suggest two more tests: an axis with min = default should not produce duplicate mappings and an axis with default = max should not produce duplicate mappings. Might have to hoist out the code to build a map vec to a fn to ease testing. IMHO this more directly validates the root cause o fthe problem than the test below, you could elect to remove it although retaining is also fine.

rsheeter (approved):
LGTM subject to adding a more direct test of the root cause

fontbe/src/avar.rs
anthrotype: > default = max

I think my current test already hits this case (max wdth == default == 100)

>  should not produce duplicate mappings

I'm confused by the "duplicate mappings"... That's not what the issue manifested itself. Do you mean, should not omit the default 0:0 avar segment map?

rsheeter: IIUC the root cause of the problem is that the list of mapping tuples produced by `default_converter` can contain duplicate keys (e.g. min or max might be default). If you extract a function to produce the default mappings you can more directly expose the cause and prove it's fixed.

anthrotype: ah you mean, take that out of `to_segment_map`! ok

anthrotype: @rsheeter how about now https://github.com/googlefonts/fontc/pull/935/commits/07e3505ae843f135617be9e6aba225a57a59ade3 ?

rsheeter: lgtm

PR #938, opened 2024-09-05
[crater] Show per-table deltas in diff detail view

cmyr:
That is, in the detail view where it shows the per-table diff percentages for a given font, if there have been any changes from the previous run display the delta at the per-table level, so we can better understand what changed.

You can see how this looks at https://googlefonts.github.io/fontc_crater/

### DIFF
--- a/fontc_crater/src/ci/html.rs
+++ b/fontc_crater/src/ci/html.rs
@@ -9,7 +9,7 @@ use std::{
 
 use crate::{
     error::Error,
-    ttx_diff_runner::{CompilerFailure, DiffError, DiffOutput},
+    ttx_diff_runner::{CompilerFailure, DiffError, DiffOutput, DiffValue},
 };
 use maud::{html, Markup};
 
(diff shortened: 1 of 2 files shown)


fontc_crater/src/ci/html.rs
```diff
@@ -172,6 +153,27 @@ fn make_table_body(runs: &[RunSummary]) -> Markup {
     }
 }
 
+/// for values that change between runs, make the little decoration span that says like,
+/// "+0.333" or "-5"
+fn make_delta_decoration<T: PartialOrd + Copy + Sub<Output = T> + Display + Default>(
+    current: T,
+    prev: Option<T>,
+    more_is_better: bool,
```
rsheeter, line 161: Suggest enum instead of bool, the callsites are a bit opaque. A trait might be nice? - `ratio.delta_decoration(prev, Better::More)`

fontc_crater/src/ci/html.rs
```diff
@@ -172,6 +153,27 @@ fn make_table_body(runs: &[RunSummary]) -> Markup {
...
+    let delta = prev.map(|prev| current - prev).unwrap_or(current);
+    let diff = format!("{delta:+.3}");
+    match prev.and_then(|prev| current.partial_cmp(&prev)) {
+        Some(std::cmp::Ordering::Equal) => Default::default(),
+        // don't write +0
+        None if current == T::default() => Default::default(),
+        None | Some(std::cmp::Ordering::Greater) if more_is_better => {
+            html! { span.better { (diff) } }
+        }
+        Some(std::cmp::Ordering::Less) if !more_is_better => html! { span.better { (diff) } },
```
rsheeter, line 172: shouldn't this be |'d onto the case above?

rsheeter (approved):

fontc_crater/src/ci/html.rs
cmyr: added the enum, trait is definitely an option but would rather just get this merged than noodle on it much more. :)

fontc_crater/src/ci/html.rs
cmyr: there can only be a single `if` clause on a match arm, and it applies after the match (it is not associated with a specific match item, see https://doc.rust-lang.org/1.80.1/reference/expressions/match-expr.html#match-expressions)

rsheeter: TIL. Or re-learned I suppose since I'm sure I did read that at some point.

PR #939, opened 2024-09-05
Fix the NameBuilder family/subfamily fallback logic 

### LINKED ISSUE #906: name table for NotoSerifHentaigana doesn't match fontmake's
I suspect that fontir's NameBuilder isn't matching 100% the ufo2ft's default fallbacks for family/style

```diff
$ diff -u /Users/clupo/oss/fontc/build/default/fontc.name.ttx /Users/clupo/oss/fontc/build/default/fontmake.name.ttx
--- /Users/clupo/oss/fontc/build/default/fontc.name.ttx	2024-08-19 16:52:22
+++ /Users/clupo/oss/fontc/build/default/fontmake.name.ttx	2024-08-19 16:52:22
@@ -3,22 +3,22 @@
       Copyright 2023 The Noto Project Authors (https://github.com/notofonts/hentaigana)
     </namerecord>
     <namerecord nameID="1" platformID="3" platEncID="1" langID="0x409">
-      Noto Serif Hentaigana
+      Noto Serif Hentaigana EL
     </namerecord>
     <namerecord nameID="2" platformID="3" platEncID="1" langID="0x409">
       Regular
     </namerecord>
     <namerecord nameID="3" platformID="3" platEncID="1" langID="0x409">
-      1.000;GOOG;NotoSerifHentaigana-Regular
+      1.000;GOOG;NotoSerifHentaigana-EL
     </namerecord>
     <namerecord nameID="4" platformID="3" platEncID="1" langID="0x409">
-      Noto Serif Hentaigana Regular
+      Noto Serif Hentaigana EL
     </namerecord>
     <namerecord nameID="5" platformID="3" platEncID="1" langID="0x409">
       Version 1.000
     </namerecord>
     <namerecord nameID="6" platformID="3" platEncID="1" langID="0x409">
-      NotoSerifHentaigana-Regular
+      NotoSerifHentaigana-EL
     </namerecord>
     <namerecord nameID="7" platformID="3" platEncID="1" langID="0x409">
       Noto is a trademark of Google Inc.
\ No newline at end of file
@@ -38,6 +38,12 @@
     <namerecord nameID="14" platformID="3" platEncID="1" langID="0x409">
       https://scripts.sil.org/OFL
     </namerecord>
+    <namerecord nameID="16" platformID="3" platEncID="1" langID="0x409">
+      Noto Serif Hentaigana
+    </namerecord>
+    <namerecord nameID="17" platformID="3" platEncID="1" langID="0x409">
+      EL
+    </namerecord>
     <namerecord nameID="256" platformID="3" platEncID="1" langID="0x409">
       Weight
     </namerecord>
```

anthrotype:
Fixes #906 

We should now match fontmake (ufo2ft) logic for determining the two sets of family and subfamily names depending on which one is present or not.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -738,6 +738,7 @@ pub struct GlobalMetricsInstance {
 /// Helps accumulate 'name' values.
 ///
 /// See <https://github.com/googlefonts/ufo2ft/blob/fca66fe3ea1ea88ffb36f8264b21ce042d3afd05/Lib/ufo2ft/outlineCompiler.py#L367>.
+#[derive(Default)]
 pub struct NameBuilder {
     names: HashMap<NameKey, String>,
     /// Helps lookup entries in name when all we have is a NameId
@@ -746,18 +747,11 @@ pub struct NameBuilder {
     version_minor: u32,
 }
 
-impl Default for NameBuilder {
-    fn default() -> Self {
-        let mut builder = Self {
-            names: Default::default(),
-            name_to_key: Default::default(),
-            version_major: Default::default(),
-            version_minor: Default::default(),
-        };
-        // Based on diffing fontmake Oswald build
-        builder.add(NameId::SUBFAMILY_NAME, "Regular".to_string());
-        builder
-    }
+const STYLE_MAP_STYLE_NAMES: [&str; 4] = ["regular", "bold", "italic", "bold italic"];
+
+/// Returns true if the style name is one of the four standard style names.
+fn is_ribbi(style_name: &str) -> bool {
+    return STYLE_MAP_STYLE_NAMES.contains(&style_name.to_lowercase().as_str());
 }
 
 impl NameBuilder {
@@ -774,24 +768,17 @@ impl NameBuilder {
     }
 
     pub fn add_if_present(&mut self, name_id: NameId, value: &Option<String>) {
-        let value = if let Some(value) = value {
-            Some(value.clone())
-        } else {
-            default_value(name_id).map(String::from)
-        };
-        if let Some(value) = value {
+        if let Some(value) = value.as_ref().cloned() {
             self.add(name_id, value);
         }
     }
 
     pub fn apply_fallback(&mut self, name_id: NameId, fallbacks: &[NameId]) {
-        if let Some(fallback_id) = fallbacks.iter().find(|n| {
-            let Some(key) = self.name_to_key.get(*n) else {
-                return false;
-            };
-            self.names.contains_key(key)
-        }) {
-            self.add(name_id, self.names[&self.name_to_key[fallback_id]].clone());
+        if self.contains_key(name_id) {
+            return;
+        }
+        if let Some(fallback) = self.get_fallback_or_default(name_id, fallbacks) {
+            self.add(name_id, fallback.to_string());
         }
     }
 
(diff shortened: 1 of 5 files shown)


anthrotype:
ignore the clippy lint errors, nothing to do with this PR

fontir/src/ir.rs
```diff
@@ -746,18 +747,11 @@ pub struct NameBuilder {
...
-            names: Default::default(),
-            name_to_key: Default::default(),
-            version_major: Default::default(),
-            version_minor: Default::default(),
-        };
-        // Based on diffing fontmake Oswald build
-        builder.add(NameId::SUBFAMILY_NAME, "Regular".to_string());
-        builder
-    }
+const STYLE_MAP_STYLE_NAMES: [&str; 4] = ["regular", "bold", "italic", "bold italic"];
```
rsheeter, line 750: Would RBBI_STYLE_NAMES make sense? - I find the current name a touch opaque

rsheeter: (optional)

fontir/src/ir.rs
```diff
@@ -809,7 +796,60 @@ impl NameBuilder {
...
+            let (fallback_subfamily, family_suffix) = if is_ribbi(&fallback_subfamily) {
+                (fallback_subfamily, None)
+            } else {
+                ("Regular".to_string(), Some(fallback_subfamily))
+            };
+
+            self.add(NameId::SUBFAMILY_NAME, fallback_subfamily);
+
+            family_suffix
+        };
```
rsheeter, line 838: This feels ... more complicated than the work being done but I'm not immediately sure how to fix it.

Adding a helper for the get/unwrap/to-string helps a little maybe.

```rust
    /// Fetch a fallback that definitely has a value in a String to avoid an immutable borrow on self
    /// Only safe to use with names that have default values
    fn fallback_string(&mut self, name_id: NameId, fallback: NameId) -> String {
        self.get_fallback_or_default(name_id, &[fallback]).unwrap().to_string()
    }
```

Optional given I have no real suggestion for improvement.

fontir/src/ir.rs
```diff
@@ -809,7 +796,60 @@ impl NameBuilder {
...
+            let mut fallback_family = self
+                .get_fallback_or_default(NameId::FAMILY_NAME, &[NameId::TYPOGRAPHIC_FAMILY_NAME])
+                .unwrap() // safe because FAMILY_NAME has a default_value
+                .to_string();
+            if let Some(family_suffix) = family_suffix.as_ref() {
+                fallback_family.push(' ');
+                fallback_family.push_str(family_suffix);
+            }
+            self.add(NameId::FAMILY_NAME, fallback_family);
+        };
```
rsheeter, line 851: This feels ... more complicated than the work being done but I'm not immediately sure how to fix it.

Adding a helper for the get/unwrap/to-string helps a little maybe.

```rust
    /// Fetch a fallback that definitely has a value in a String to avoid an immutable borrow on self
    /// Only safe to use with names that have default values
    fn fallback_string(&mut self, name_id: NameId, fallback: NameId) -> String {
        self.get_fallback_or_default(name_id, &[fallback]).unwrap().to_string()
    }
```

Optional given I have no real suggestion for improvement.

rsheeter (approved):

fontir/src/ir.rs
cmyr, line 750: in general for arrays or strings I prefer `static` to `const` (since in theory a `const` is instantiated in each place it is referenced), and so this would probably be a static slice of `&str`, like:

```rust
static STYLE_MAP_STYLE_NAMES: &[&str] = &[..];
```

fontir/src/ir.rs
```diff
@@ -746,18 +747,11 @@ pub struct NameBuilder {
...
-        };
-        // Based on diffing fontmake Oswald build
-        builder.add(NameId::SUBFAMILY_NAME, "Regular".to_string());
-        builder
-    }
+const STYLE_MAP_STYLE_NAMES: [&str; 4] = ["regular", "bold", "italic", "bold italic"];
+
+/// Returns true if the style name is one of the four standard style names.
+fn is_ribbi(style_name: &str) -> bool {
+    return STYLE_MAP_STYLE_NAMES.contains(&style_name.to_lowercase().as_str());
```
cmyr, line 754: but if _this_ is the only place where you're referencing this static than I would actually just keep it inline in this function, either by having the `static` declared here or by just writing,

```rust
["regular", "etc"].contains(&thing)
```

and just for fun you could also express this using `match`, which _might_ generate slightly better code (very unlikely to matter in this case):

```rust
matches!(
        style_name.to_lowercase().as_str(),
        "regular" | "bold" | "italic" | "bold italic"
    )
```

cmyr (approved):
looks reasonable, some rust golf comments inline :)

fontir/src/ir.rs
rsheeter: Good catch. I'd have thought to use const as well, does it actually produce a worse result or does the compiler fix it for you?

fontir/src/ir.rs
rsheeter: oh good call, +1 to `matches!` if that works

anthrotype: I went for `matches!`

fontir/src/ir.rs
anthrotype: yeah, the original logic we're trying to match is indeed complicated. I've added the helper

PR #941, opened 2024-09-05
[clippy] Roll the rock back up the hill

cmyr:
JMM


rsheeter:
I wonder if we should use python rather than text but I don't want to block fixing CI on that so merging, we can send a followon if necessary

PR #942, opened 2024-09-05
[fontbe] Create GDEF for GlyphClassDef if needed

cmyr:
Previously we would only add a GlyphClassDef table to an existing GDEF table, but would not create a GDEF table if one had not been produced by feature compilation.

This patch matches the behaviour of fontmake.

This does raise the question of whether or not either compiler should bother generating a GlyphClassDef subtable in the absense of GPOS/GSUB, but that's something we can address separately.

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -553,23 +553,20 @@ impl Work<Context, AnyWorkId, Error> for FeatureCompilationWork {
         let marks = context.fea_rs_marks.get();
 
         let mut result = self.compile(&static_metadata, &ast, kerns.as_ref(), marks.as_ref())?;
-        if result.gdef_classes.is_none() {
-            // there were no explicit gdef categories, so let's use the ones
-            // that were in public.openTypeCatgories or which we computed from
-            // GlyphData.xml
+        if result.gdef_classes.is_none() && !static_metadata.gdef_categories.categories.is_empty() {
+            // the FEA did not contain an explicit GDEF block with glyph categories,
+            // so let's use the ones from the source, if present (i.e. from
+            // `public.openTypeCatgories` or computed from GlyphData.xml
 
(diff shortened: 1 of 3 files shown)


fontbe/src/features.rs
```diff
@@ -553,23 +553,20 @@ impl Work<Context, AnyWorkId, Error> for FeatureCompilationWork {
         let marks = context.fea_rs_marks.get();
 
         let mut result = self.compile(&static_metadata, &ast, kerns.as_ref(), marks.as_ref())?;
-        if result.gdef_classes.is_none() {
+        if result.gdef_classes.is_none() && !static_metadata.gdef_categories.categories.is_empty() {
             // there were no explicit gdef categories, so let's use the ones
             // that were in public.openTypeCatgories or which we computed from
             // GlyphData.xml
```
rsheeter, line 559: it's slightly difficult to line up the comment with the not-quite-matching fields. Can you update to specify. For example, is static_metadata.gdef_categories not the explicit gdef_categories? Is it the ones from public.openTypeCategories?

fontc/src/lib.rs
```diff
@@ -3049,4 +3049,16 @@ mod tests {
             // read-fonts would return an error when a glyph's variations are empty
             .is_err()));
     }
+
+    // if a font has custom gdef categories defined, create the gdef table and
+    // with the appropriate GlyphClassDef subtable, even if there's no GPOS/GSUB
+    // or no other stuff in GDEF
```
rsheeter, line 3055: ...to match fontmake

cmyr:
Opened https://github.com/googlefonts/fontmake/issues/1120 to discuss whether we should be creating GDEF at all in this case.

fontbe/src/features.rs
cmyr: updated this comment to better explain what's going on.

PR #946, opened 2024-09-06
[crater] Assert presence of timeout(1) before run

cmyr:
This isn't present by default on systems that don't have GNU Coreutils installed, and without this check we will just happily fail for every target.

(last night's crater run failed because I hadn't included `opt/homebrew/bin` in the PATH of the ci machine, and it hit this issue)

### DIFF
--- a/fontc_crater/src/ttx_diff_runner.rs
+++ b/fontc_crater/src/ttx_diff_runner.rs
@@ -253,8 +253,21 @@ pub(super) fn print_report(results: &Results<DiffOutput, DiffError>, verbose: bo
     }
 }
 
+fn assert_has_timeout_coreutil() {
+    match Command::new("which").arg("timeout").output() {
+        Ok(out) if out.status.success() => return,
+        Err(e) => eprintln!("'which' failed!? '{e}'"),
+        Ok(_) => eprintln!("could not find 'timeout'. You may need to install coreutils (on macos, `brew install coreutils`)"),
+    }
+    std::process::exit(1);
+}
+
 /// make sure we can find and execute ttx_diff script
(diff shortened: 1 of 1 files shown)


fontc_crater/src/ttx_diff_runner.rs
```diff
@@ -253,8 +253,21 @@ pub(super) fn print_report(results: &Results<DiffOutput, DiffError>, verbose: bo
     }
 }
 
+fn assert_has_timeout_coreutil() {
+    match Command::new("which").arg("timeout").output() {
+        Ok(out) if out.status.success() => return,
+        Err(e) => eprintln!("'which' failed!? '{e}'"),
+        Ok(_) => eprintln!("could not find 'timeout'. You may need to install coreutils"),
```
rsheeter, line 260: Suggest giving the exact command and indicating the advice is mac-specific

rsheeter (approved):

fontc_crater/src/ttx_diff_runner.rs
cmyr: sorry, exact command where? I thought the 'which' branch was academic since I thought `which` was in the POSIX spec, but apparently it's not, and just a shell builtin? but still hard to imagine hitting that branch.

I didn't provide mac specific advice since in theory this could also be hit on windows (or various other places that are not GNU/Linux)

rsheeter: I mean to install coreutils on a mac. I don't think anything but a mac is likely to be missing timeout.

I hadn't considered windows, I think deeming windows out of scope for crater is completely fine.

Issue #952, opened 2024-09-07
Anonymous lookups and lookupflag


khaledhosny:
The following code:
```fea
languagesystem arab dflt;

lookup LamAlef {
  sub lam-ar.init by lam-ar.init.1;
  sub lam-ar.medi by lam-ar.medi.1;
  sub alef-ar.fina by alef-ar.fina.1;
} LamAlef;

feature ccmp {
  sub hehHamzaabove-ar by heh-ar hamzaabove-ar;
  sub tehMarbuta-ar by heh-ar twodotshorizontalabove-ar;
} ccmp

feature rlig {
  lookupflag IgnoreMarks;
    sub [lam-ar.init lam-ar.medi]' lookup LamAlef alef-ar.fina' lookup LamAlef;

  @hehMarks = [twodotshorizontalabove-ar hamzaabove-ar];
  lookupflag UseMarkFilteringSet @hehMarks;
    ignore sub heh-ar' @hehMarks;
    sub heh-ar' by hehDoachashmee-ar;
} rlig;
```

Should result in two lookup in the rlig feature, one with `lookupflag IgnoreMarks` and the other with `lookupflag UseMarkFilteringSet @hehMarks`. However, fontc is creating one lookup with `lookupflag IgnoreMarks` and the `lookupflag UseMarkFilteringSet @hehMarks;` is completely ignored, resulting in broken wrong application.

If I put the last part in a explicit lookup block, I get th expected lookups in output:
```fea
feature rlig {
  lookupflag IgnoreMarks;
    sub [lam-ar.init lam-ar.medi]' lookup LamAlef alef-ar.fina' lookup LamAlef;

  lookup test {
    @hehMarks = [twodotshorizontalabove-ar hamzaabove-ar];
    lookupflag UseMarkFilteringSet @hehMarks;
      ignore sub heh-ar' @hehMarks;
      sub heh-ar' by hehDoachashmee-ar;
  } test;
} rlig;
```

(this is stripped-down version of code from [Mada](https://github.com/aliftype/mada/tree/main/sources))

PR #953, opened 2024-09-08
[fea-rs] Take lookup flags into account in ensure_current_lookup_type()

khaledhosny:
Checking lookup kind only will merge anonyms lookups with different lookup flags.

Fixes https://github.com/googlefonts/fontc/issues/952

### DIFF
--- a/fea-rs/src/compile/lookups.rs
+++ b/fea-rs/src/compile/lookups.rs
@@ -179,7 +179,7 @@ pub(crate) struct LookupIdMap {
 }
 
 /// Tracks the current lookupflags state
-#[derive(Clone, Copy, Debug, Default)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
 pub(crate) struct LookupFlagInfo {
     pub(crate) flags: LookupFlag,
     pub(crate) mark_filter_set: Option<FilterSetId>,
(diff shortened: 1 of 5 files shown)


fea-rs/src/compile/lookups.rs
```diff
@@ -430,8 +430,12 @@ impl AllLookups {
     }
 
     /// Returns `true` if there is an active lookup of this kind
-    pub(crate) fn has_current_kind(&self, kind: Kind) -> bool {
-        self.current.as_ref().map(SomeLookup::kind) == Some(kind)
+    pub(crate) fn has_current_kind(&self, kind: Kind, flags: LookupFlagInfo) -> bool {
```
cmyr, line 433: instead of changing this method I think I'd prefer to add a separate method with a name like `has_same_flags(&self, flags: LookupFlagInfo) -> bool`. This will let us continue to use this function in places where we don't care about the flags (like in the promotion logic below) and then in places where we can we can just write, `if !self.lookups.has_current_kind(kind) && !self.lookups.has_same_flags(flags) { ..` in the places where we care about both.

fea-rs/src/compile/lookups.rs
```diff
@@ -838,6 +842,58 @@ impl SomeLookup {
         }
     }
 
+    fn flags(&self) -> LookupFlag {
```
cmyr, line 845: This is gross and not your fault, I should really have a macro or something to handle these enums. In any case maybe we can improve it somewhat by returning `LookupFlagInfo`, so we don't need two separate branchy methods?

cmyr (commented):
Thanks for this, overall I think the logic makes sense although I have a couple little notes on the details.

I'm also glad you figured out how to add a new test. Was that painful at all? Did you also figure out how to use `FEA_WRITE_TEST_OUTPUT=1` to generate the ttx?

khaledhosny:
> Thanks for this, overall I think the logic makes sense although I have a couple little notes on the details.
> 
> I'm also glad you figured out how to add a new test. Was that painful at all? Did you also figure out how to use `FEA_WRITE_TEST_OUTPUT=1` to generate the ttx?

I followed the instructions in fea-rs/test-data/compile-tests/README.md and it was straightforward and everything worked right away.

cmyr (approved):
Looks good, thanks for catching this!

PR #956, opened 2024-09-10
Include links to source repo in crater reports

cmyr:
This is a bit involved since we didn't previously serialize this info; to avoid duplication we'll keep a single mapping that is shared between runs.


fontc_crater/src/ci.rs
```diff
@@ -114,7 +124,18 @@ fn run_crater_and_save_results(args: &CiArgs) -> Result<(), Error> {
...
+    super::try_write_json(&prev_runs, &summary_file)?;
+
+    // we write the map of target -> source repo to a separate file because
+    // otherwise we're basically duplicating it for each run.
+    let sources_file = args.out_dir.join(SOURCES_FILE);
+    let mut sources: BTreeMap<PathBuf, String> = load_json_if_exists_else_default(&sources_file)?;
+
+    // we extend here because it's possible there's a new repo, or a new source in an
+    // old repo; we'll just build up one big mapping.
+    sources.extend(results.source_repos);
```
rsheeter, line 137: Doesn't this mean stale entries last forever?

rsheeter (approved):

fontc_crater/src/ci.rs
cmyr: yes... I thought this would be helpful in ensuring that we would always know the repo for a source even if it disappears but I don't think that's very useful, should be fine if we just keep track of the repos/sources used for the most recent run.

Issue #958, opened 2024-09-11
Generated mark-to-mark attachment lookups do not work


khaledhosny:
The generated lookups have mark filtering sets that include only the mark glyphs but not the bases (which are also mark glyphs since this mark-to-mark), as a result the lookups are never applied since the base glyphs will be skipped.

khaledhosny:
This seemed drastic, but now I see it is broken only when the set of marks and bases are different, which might be a little uncommon thing to have.

PR #960, opened 2024-09-11
Add base marks to mark filtering sets of mark-to-mark attachment lookups

khaledhosny:
Otherwise the lookups will not be applied as the base marks will be skipped.

- Fixes https://github.com/googlefonts/fontc/issues/958

### DIFF
--- a/fontbe/src/features/marks.rs
+++ b/fontbe/src/features/marks.rs
@@ -80,6 +80,11 @@ impl MarkGroup<'_> {
             self.marks
                 .iter()
                 .map(|(name, _)| glyph_order.glyph_id(name).unwrap())
+                .chain(
+                    self.bases
+                        .iter()
+                        .map(|(name, _)| glyph_order.glyph_id(name).unwrap()),
+                )
                 .collect()
         })
     }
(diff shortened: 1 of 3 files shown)


fontbe/src/features/marks.rs
```diff
@@ -80,6 +80,11 @@ impl MarkGroup<'_> {
             self.marks
                 .iter()
                 .map(|(name, _)| glyph_order.glyph_id(name).unwrap())
+                .chain(
```
cmyr, line 83: okay cool I think I'd just misunderstood the logic here, and thought that only the marks which were 'acting as marks' should go in the mark filtering set in this case.

cmyr (approved):
Thanks!

fontbe/src/features/marks.rs
khaledhosny: The idea is that you want the mark-to-mark attachment to apply to these bases and marks only and ignore any possible intervening mark. Since any GDEF-mark glyph not in the mark attachment set will be ignored, the base marks will be ignored as well and the lookup will not be applied. Mark filtering sets do not care about the rule the mark is taking in this lookup (base or mark) only its GDEF class.

PR #962, opened 2024-09-11
[fontbe] Look at GPOS when computing OS/2 usMaxContext

### LINKED ISSUE #961: OS/2 'usMaxContext' value differs from fontmake for some fonts
~I believe we just aren't computing this value at all.~

Okay, this logic exists in fontbe/os2.rs.

It looks like this isn't handling GPOS, though, so that might be the issue?

cmyr:
This will (with #959) get us identical for at least one more font.

- closes #961


rsheeter (approved):
LGTM subject to unit test(s) being added

Issue #949, opened 2024-09-07
Some font-level custom paramaters are ignored in Glyphs sources


khaledhosny:
In [Mada](https://github.com/aliftype/mada/tree/main/sources), the following custom parameters are defined at the font instead of for each master, but fontc ignores them unless copied to each master:

typoAscender
typoDescender
typoLineGap
hheaAscender
hheaDescender
hheaLineGap
winAscent
winDescent
underlineThickness
underlinePosition

PR #964, opened 2024-09-11
[glyphs2fontir] Fallback to setting GlobalMetrics from font

khaledhosny:
If the corresponding custom parameter is missing from the master, try it from the parent font.

- Fixes https://github.com/googlefonts/fontc/issues/949

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -664,20 +664,6 @@ impl GlobalMetrics {
         }
     }
 
-    pub fn set_if_some_or_fallback(
-        self: &mut GlobalMetrics,
-        metric: GlobalMetric,
-        pos: NormalizedLocation,
-        maybe_value: Option<impl Into<OrderedFloat<f64>>>,
-        maybe_fallback: Option<impl Into<OrderedFloat<f64>>>,
-    ) {
-        if let Some(value) = maybe_value {
-            self.set(metric, pos, value.into().into_inner() as f32);
-        } else if let Some(fallback) = maybe_fallback {
-            self.set(metric, pos, fallback.into().into_inner() as f32);
-        }
-    }
-
     pub fn at(&self, pos: &NormalizedLocation) -> GlobalMetricsInstance {
         GlobalMetricsInstance {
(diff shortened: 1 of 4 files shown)


fontir/src/ir.rs
```diff
@@ -664,6 +664,20 @@ impl GlobalMetrics {
         }
     }
 
+    pub fn set_if_some_or_fallback(
+        self: &mut GlobalMetrics,
```
cmyr, line 668: I'm... not sure why this is like this above, but this signature is just a weird way of writing `&mut self` as the first param, and we should stick to that.

(unless I'm missing something)

glyphs2fontir/src/source.rs
```diff
@@ -545,105 +585,125 @@ impl Work<Context, WorkId, Error> for GlobalMetricWork {
 
             metrics.set_if_some(GlobalMetric::CapHeight, pos.clone(), master.cap_height());
             metrics.set_if_some(GlobalMetric::XHeight, pos.clone(), master.x_height());
-            metrics.set_if_some(
+            metrics.set_if_some_or_fallback(
                 GlobalMetric::Os2TypoAscender,
                 pos.clone(),
                 master.typo_ascender.map(|v| v as f64),
+                font.typo_ascender.map(|v| v as f64),
```
cmyr, line 592: okay so the other way to write all of these would be to stick with `set_if_some`, but then do,

```rust
metrics.set_if_some(
    MetricName,
    pos.clone(),
    master.typo_ascender.or(font.typo_ascender).map(|v| v as f64),
)
```

which might be... slightly better, on balance? at least it saves us adding the additional method. This whole big function is kind of calling out for a macro, but that's another story; let's just get it working first.

khaledhosny: Sounds good. I’ll try to do this in the morning.

fontir/src/ir.rs
khaledhosny: I just copied the above function and added a new parameter without giving it much thought.

cmyr (approved):
thanks!

Issue #950, opened 2024-09-07
Stylistic Set labels are not compiled from Glyphs sources


khaledhosny:
[Mada](https://github.com/aliftype/mada/tree/main/sources) has `ss01`, `ss03`, and `ss03` with label name entries, but fontc does not seem to support this (there is not corresponding name records in the compiled font).

Issue #406, opened 2023-08-17
Use other tables generated by fea-rs


cmyr:
feature files can end up creating or modifying a bunch of different tables; see https://github.com/cmyr/fea-rs/blob/main/fea-rs/src/compile/output.rs#L22 for the list. It is likely that many of these are rarely used, but at the very least we should be checking if they're present.

rsheeter:
I'm inclined to say we should start by erroring if fea gave us back any table we don't deal with

rsheeter:
We'll want to check which things fontmake accepts to be defined by fea and how it handles multiple definitions

anthrotype:
Fontmake accepts anything that fonttools feaLib returns with no questions, those are not multiple definitions but rather "overrides"

anthrotype:
That's how Adobe FEA spec describes these, and also how makeotf interprets them

rsheeter:
Is there any case where you have to _merge_ the results or can you always take teh whole table from fea? - for example, I think Colin said you can set modified time in fea and that made me think you'd need to take *just* that field and copy it into the table you built from non-fea sources. If we can just always take the whole table that would be nice.

rsheeter:
If the fea compiler is to apply overrides it would have to have access to the thing it is to apply them to? And that makes for all sorts of execution order limitations, e.g. if fea can override modified in `head` then we must run fea after we make `head` and we have to somehow make our `head available to `fea-rs`?

cmyr:
I think that having the fea compiler be responsible for merging is a bit of a layering violation; you should be able to compile FEA without knowing much about the final font. I would propose a design where `fontc` inspects the tables returned by `fea-rs` and merges them as required.

cmyr:
this is superseded by issues for the individual missing tables: #1391 & #1390

PR #969, opened 2024-09-18
[glyphs2fontir] Support stylistic set labels

khaledhosny:
Convert them to featureNames.

Fixes https://github.com/googlefonts/fontc/issues/950

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -1964,26 +1964,25 @@ impl RawFeature {
     // https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/Lib/glyphsLib/builder/features.py#L134
     fn feature_names(&self) -> String {
         if self.labels.is_empty() {
-            String::new()
-        } else {
-            let labels = self
-                .labels
-                .iter()
-                .map(|label| {
-                    let language_id = GLYPHS_TO_OPENTYPE_LANGUAGE_ID
-                        .iter()
-                        .find(|entry| entry.0 == label.language)
-                        .map(|entry| entry.1)
-                        .unwrap_or_else(|| {
-                            panic!("Unknown feature label language: {}", label.language);
-                        });
-                    let name = label.value.replace("\\", "\\005c").replace("\"", "\\0022");
-                    format!("  name 3 1 0x{:04X} \"{}\";", language_id, name)
-                })
-                .collect::<Vec<_>>()
-                .join("\n");
-            format!("featureNames {{\n{}\n}};\n", labels)
+            return String::new();
         }
+        let labels = self
+            .labels
+            .iter()
+            .map(|label| {
+                let language_id = GLYPHS_TO_OPENTYPE_LANGUAGE_ID
+                    .iter()
+                    .find(|entry| entry.0 == label.language)
+                    .map(|entry| entry.1)
+                    .unwrap_or_else(|| {
+                        panic!("Unknown feature label language: {}", label.language);
+                    });
+                let name = label.value.replace("\\", "\\005c").replace("\"", "\\0022");
+                format!("  name 3 1 0x{:04X} \"{}\";", language_id, name)
+            })
+            .collect::<Vec<_>>()
+            .join("\n");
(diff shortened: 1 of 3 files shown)


khaledhosny:
This change seems to work (I see `featureNames` blocks in the generated feature code), but the featureParams name IDs seems to get re-used by fvar table, so the final font has the wrong names.

khaledhosny:
> This change seems to work (I see `featureNames` blocks in the generated feature code), but the featureParams name IDs seems to get re-used by fvar table, so the final font has the wrong names.

That is probably the culprit as it does not use NameBuilder like other tables:
https://github.com/googlefonts/fontc/blob/09025bf4bf94cbc214c6fe6daf14f63c99409ca2/fontir/src/ir.rs#L346-L360

khaledhosny:
Or rather, the fea-rs compiler have no knowledge of any user name IDs allocated by other tables it doesn’t build itself.

rsheeter:
Apologies, this came in while I was travelling and got missed. We'll get this reviewed.

glyphs-reader/src/font.rs
```diff
@@ -1862,6 +1885,31 @@ impl RawFeature {
     fn disabled(&self) -> bool {
         self.disabled == Some(1)
     }
+
+    // https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/Lib/glyphsLib/builder/features.py#L134
+    fn feature_names(&self) -> String {
+        if self.labels.is_empty() {
+            String::new()
+        } else {
```
cmyr, line 1893: total style nit, in a fn like this I generally prefer to use an explicit return for the guard case, so the rest of the logic isn't indented unnecessarily, so like

```rust
if self.labels.is_empty() {
    return String::new()
}
// actual logic here
```

glyphs-reader/src/font.rs
```diff
@@ -1862,6 +1885,31 @@ impl RawFeature {
...
+            let labels = self
+                .labels
+                .iter()
+                .map(|label| {
+                    let language_id = GLYPHS_TO_OPENTYPE_LANGUAGE_ID
+                        .iter()
+                        .find(|entry| entry.0 == label.language)
+                        .map(|entry| entry.1)
+                        .unwrap_or_else(|| {
+                            panic!("Unknown feature label language: {}", label.language);
```
cmyr, line 1903: I don't think we should panic here; instead I would skip this label and emit a warning. You can use `filter_map` instead of `map` to skip missing items.

glyphs-reader/src/font.rs
```diff
@@ -1862,6 +1885,31 @@ impl RawFeature {
...
+        if self.labels.is_empty() {
+            String::new()
+        } else {
+            let labels = self
+                .labels
+                .iter()
+                .map(|label| {
+                    let language_id = GLYPHS_TO_OPENTYPE_LANGUAGE_ID
+                        .iter()
+                        .find(|entry| entry.0 == label.language)
```
cmyr, line 1900: not critical since this is not a hot code path, but in general I'd prefer doing a binary search to a linear scan here; the one gotcha is we would want to ensure that the array is sorted (which we could do with a test case that created an explicitly sorted copy and compared this with the original)

cmyr (approved):
Yea sorry for letting this hang... a couple little notes inline, the only one that I think is important is the one regarding the panic.

There are at least two reasons that this might not result in the names ending up in the final name table; the issue you've identified with `fvar`, but also the simple fact that I don't believe we currently even _use_ the name table that is produced by `fea-rs` (currently we only use GDEF/GPOS/GSUB, iirc). There's an old issue up for this, at #406; maybe time to figure this one out.

cmyr (approved):
great, thanks.

Per  your earlier comment this should still not result in the names actually ending up in the final font, because of one or more of the fvar issue & ignoring the fea-rs name table; but let's get this merged and we can worry about that as followup.

khaledhosny:
> great, thanks.
> 
> Per your earlier comment this should still not result in the names actually ending up in the final font, because of one or more of the fvar issue & ignoring the fea-rs name table; but let's get this merged and we can worry about that as followup.

Yes, it already already affects feature labels written in the feature file directly (e.g. in UFO sources).

PR #971, opened 2024-09-18
[ttx_diff] Smarter selection of fontmake -o flag

cmyr:
Unless the source is a ufo, we will load it with glyphsLib or designspaceLib and decide if it's variable or not based on the number of masters.

This fixes the issue with all the fonts where fontmake was building fvar/hvar etc for non-variable fonts.

JMM


resources/scripts/ttx_diff.py
```diff
@@ -552,6 +566,8 @@ def main(argv):
     if not source.exists():
         sys.exit(f"No such source: {source}")
 
+    is_variable = source_is_variable(source)
```
rsheeter, line 569: Since this is a function of the source and only needed in build_fontmake why not call it from there? - seems unnecessary to compute here and pass in

rsheeter (approved):
LGTM, had a suggestion so didn't merge

resources/scripts/ttx_diff.py
cmyr: good call!

anthrotype:
sorry i'm late. i'm afraid it's not necessarily that simple. There could be n>1 masters with designs which are incompatible with each another, and only instances that sit on top of an existing master, and no interpolated instances... and Glyphs.app would let one export them as static fonts, one individual master at a time, whereas compiling a variable font from them will fail. IIRC some Noto fonts were like this. Because fontmake can't always guess whether the intent is to attempt to build a variable font or a bunch of statics, it errs on the side of caution and asks the font developer to tell it whether they want to build a variable font or not.
Also see this related discussion
https://github.com/googlefonts/fontmake/issues/821

I think we should change `fontmake -o variable` to drop all the VF-related tables which are in fact no-op when the source turns out to only contain one master. You mentioned fvar. What else?

Could you file an issue?

rsheeter:
@anthrotype can we add something conceptually equivalent to `-o auto`, causing fontmake to pick static or variable much as fontc does?

rsheeter:
@anthrotype is it fair to think this is a step forward even if not perfect? Or is it so wrong as to merit rollback?

anthrotype:
but maybe we don't need a new -o auto if we change -o variable to do the right thing in the presence of only 1 master.

remember that fontmake -o defaults to `ttf otf` which means it builds _both_ a ttf and otf _static_  font for each master (see https://github.com/googlefonts/fontmake/issues/821).

what would `-o auto` do when there's only one master then? Build _two_ static fonts, one ttf and one otf? Or only one ttf?

rsheeter:
> maybe we don't need a new -o auto if we change -o variable to do the right thing in the presence of only 1 master

If the right thing is to build a static I think that is roughly equivalent to what I'm imagining `-o auto` would do. IIUC it has the advantage of meaning ttf so the ambituity of ttf vs otf is eliminated?

rsheeter:
Updating based on f2f: the right way for crater is to read the gftools config

PR #978, opened 2024-10-01
Use our logging framework for logging

rsheeter:
It's quite confusing that you can set `RUST_LOG` to adjust how much you see except for one specific thing that relies on a flag.

Delete the flag, use the same logging as everything else, and tell the user how to see the error if we aren't showing them.

JMM

### DIFF
--- a/fontc/src/main.rs
+++ b/fontc/src/main.rs
@@ -10,16 +10,19 @@ fn main() {
     let args = Args::parse();
 
     // catch and print errors manually, to avoid just seeing the Debug impls
+    // The default log level is error so the user will see it unless they specifically turned off logging
     if let Err(e) = run(args) {
         let mut error_displayed = false;
         let mut additional = "";
         if let Error::Backend(fontbe::error::Error::FeaCompileError(e)) = &e {
             if log::log_enabled!(log::Level::Warn) {
-                warn!("{}", e.display_verbose());
+                error!("{e}");
+                if let Some(diagnostic) = e.diagnostics() {
+                    warn!("{}", diagnostic.display());
+                }
                 error_displayed = true;
             } else {
-                additional =
-                    ", set log level to warn or higher (export RUST_LOG=warn) for additional detail"
+                additional = ", set log level to warn or higher (--log warn) for additional detail"
             }
         }
         if !error_displayed {
@@ -39,20 +42,23 @@ fn run(args: Args) -> Result<(), Error> {
     let time = create_timer(AnyWorkId::InternalTiming("Init logger"), 0)
         .queued()
         .run();
-    env_logger::builder()
-        .format(|buf, record| {
-            let ts = buf.timestamp_micros();
-            let style = buf.default_level_style(record.level());
-            writeln!(
-                buf,
-                "[{ts} {:?} {} {style}{}{style:#}] {}",
-                std::thread::current().id(),
-                record.target(),
-                record.level(),
-                record.args()
-            )
-        })
-        .init();
+    let mut log_cfg = env_logger::builder();
+    log_cfg.format(|buf, record| {
+        let ts = buf.timestamp_micros();
+        let style = buf.default_level_style(record.level());
+        writeln!(
+            buf,
+            "[{ts} {:?} {} {style}{}{style:#}] {}",
+            std::thread::current().id(),
+            record.target(),
+            record.level(),
+            record.args()
+        )
+    });
+    if let Some(log_filters) = &args.log {
+        log_cfg.parse_filters(log_filters);
+    }
+    log_cfg.init();
     timer.add(time.complete());
 
     fontc::run(args, timer)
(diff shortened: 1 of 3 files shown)


fontc/src/main.rs
```diff
@@ -4,21 +4,27 @@ use clap::Parser;
...
+            if log::log_enabled!(log::Level::Warn) {
+                warn!("{}", e.display_verbose());
+                error_displayed = true;
+            } else {
+                additional =
+                    ", set log level to warn or higher (export RUST_LOG=warn) for additional detail"
             }
         }
-        // we have a CI check that forbids eprintln, but we're very clever, so
-        let _ = writeln!(std::io::stderr(), "{e}");
```
cmyr, line 21: This is unrelated to us being verbose, and is just us printing an error even if we don't complete, even if logging is not enabled

cmyr (commented):
With this change we won't show any message at all on error unless logging is enabled?

anthrotype:
I don't like this, tbh. It's standard for CLI tools of this sort to have their own --verbose flag. Can't the existing flag be used to set up the logging verbosity instead? In similar (e.g. python) tools, it's common to have -v, -vv, etc. to increase the verbosity (reduce the logging level).

anthrotype:
I personally don't like env vars, in general. The RUST_LOG environment variable is too broad and affects any rust executable that uses the same logging machinery, not just fontc. Sure, you can avoid exporting it for the entire session, and only set for the single invocation of fontc.. But it still feels not ideal to me. It's no big deal, so feel free to ignore my comments.

fontc/src/main.rs
```diff
@@ -4,21 +4,27 @@ use clap::Parser;
...
     // catch and print errors manually, to avoid just seeing the Debug impls
     if let Err(e) = run(args) {
+        let mut error_displayed = false;
+        let mut additional = "";
         if let Error::Backend(fontbe::error::Error::FeaCompileError(e)) = &e {
-            if verbose {
-                let _ = writeln!(std::io::stderr(), "{}", e.display_verbose());
-                std::process::exit(1);
+            if log::log_enabled!(log::Level::Warn) {
+                warn!("{}", e.display_verbose());
```
cmyr, line 18: the call of `e.display_verbose()` is required in order to see the details of the error that you are wanting to see (the full diagnostic message for each error encountered in the FEA)

rsheeter:
> Can't the existing flag be used to set up the logging verbosity instead?

A flag to set log level (or more broadly configure logging) is very normal in tools using a logging framework. I'm happy to add that.

cmyr:
for some background this flag was added in #901

cmyr:
The big question for me is if we want logging to be the only UI we provide? This doesn't really sound right to me; I think logging is a diagnostic tool useful to the developer but I wouldn't expect it to be necessary for a 'user' (which is vague in this case, I acknowledge) to think about log levels in order to see information about an error.

rsheeter:
I propose to add a flag for log level whose default is either error or warn. I feel strongly we don't want to partially use logging, I've seen that pattern end up confusing before and have personally experienced it to be so here. If I crank logging to debug and don't get any output from an error that's ... wrong. And that's exactly what we currently do.

rsheeter:
> even if logging is not enabled

According to https://docs.rs/env_logger/latest/env_logger/#enabling-logging the default global setting is error so by default the user will see the error.

rsheeter:
Revised per above. Behavior (compiling noise elided, a real user probably just runs fontc not cargo run):

```shell
# If the user does nothing, no flag and no RUST_LOG value
$ unset RUST_LOG
$ cargo run -- --source  ~/oss/font_repos/googlesans-flex/sources/GoogleSansFlex.designspace 
[2024-10-02T20:16:20.459878Z ThreadId(1) fontc ERROR] FEA validation failed with 1 errors, set log level to warn or higher (--log warn) for additional detail

# how nice, it told me how to get more details
$ cargo run -- --source  ~/oss/font_repos/googlesans-flex/sources/GoogleSansFlex.designspace --log warn
[2024-10-02T20:31:13.174641Z ThreadId(1) fontc ERROR] FEA validation failed with 1 errors
[2024-10-02T20:31:13.174708Z ThreadId(1) fontc WARN] error: Referenced feature not found.
in /usr/local/google/home/rsheeter/oss/font_repos/googlesans-flex/sources/GoogleSansFlex-wg1-wd25-oz6-GD0-RD0-sl0.ufo/features.fea at 72:8
   | 
72 | feature case;
   |   
```

fontc/src/main.rs
rsheeter: Logging will write out error unless the user explicitly sets logging to off, in which case I think WAI

fontc/src/main.rs
rsheeter: We're still making the call, it's just emitted via the logging system.

cmyr (approved):
This looks more reasonable. Ultimately there's a philosophical point here around the difference between logging and user-facing diagnostics, but happy to skip that for the time being and just keep the ball moving.

PR #979, opened 2024-10-02
ttx_diff: pass --keep-overlaps for static fontmake builds for now

### LINKED ISSUE #975: static fonts from single-master sources have overlaps, skewing glyf/gvar ttx diff
Following #971, ttx_diff.py now invokes fontmake `-o ttf` instead of `-o variable` when the input source contains only one master, in order to avoid fontmake adding unnecessary empty/noop variable tables (https://github.com/googlefonts/fontmake/issues/1125).

But fontmake `-o ttf` also passes the source through the ufo2ft's [remove overlaps filter](https://github.com/googlefonts/ufo2ft/blob/main/Lib/ufo2ft/filters/removeOverlaps.py) (using booleanOperations by default, or skia-pathops if requested via `--overlaps-backend pathops`).

fontc being variable-first font compile does not attempt nor support removing the overlaps (as it would destroy master compatibility for a VF target).

Therefore comparing fontc-built single-master sources vs fontmake-built ones will usually produce several glyph diffs in glyph-related tables like glyf, gvar, maxp, etc. because the number of outline points when glyphs contain overlapping contours will be different between the two fonts.

For example, take this test font, containing only one glyph "T" with overlapping contours, and compare it with ttx_diff.py:

[SingleMaster.glyphs.zip](https://github.com/user-attachments/files/17203384/SingleMaster.glyphs.zip)

```diff
--- build/default/fontc.ttf
+++ build/default/fontmake.ttf
@@ -12,12 +12,12 @@
     <!-- Most of this table will be recalculated by the compiler -->
     <tableVersion value="1.0"/>
     <fontRevision value="1.0"/>
-    <checkSumAdjustment value="0x0"/>
+    <checkSumAdjustment value="0xd88e9083"/>
     <magicNumber value="0x5f0f3cf5"/>
     <flags value="00000000 00000011"/>
     <unitsPerEm value="1000"/>
     <created value="Mon Sep 30 16:38:24 2024"/>
-    <modified value="Tue Oct  1 10:17:24 2024"/>
+    <modified value="Tue Oct  1 10:17:25 2024"/>
     <xMin value="26"/>
     <yMin value="-200"/>
     <xMax value="574"/>
@@ -53,7 +53,7 @@
     <!-- Most of this table will be recalculated by the compiler -->
     <tableVersion value="0x10000"/>
     <numGlyphs value="3"/>
-    <maxPoints value="10"/>
+    <maxPoints value="8"/>
     <maxContours value="2"/>
     <maxCompositePoints value="0"/>
     <maxCompositeContours value="0"/>
@@ -169,12 +169,10 @@
         <pt x="26" y="723" on="1"/>
         <pt x="574" y="723" on="1"/>
         <pt x="574" y="578" on="1"/>
-        <pt x="363" y="578" on="1"/>
-        <pt x="408" y="623" on="1"/>
+        <pt x="408" y="578" on="1"/>
         <pt x="408" y="0" on="1"/>
         <pt x="209" y="0" on="1"/>
-        <pt x="209" y="623" on="1"/>
-        <pt x="254" y="578" on="1"/>
+        <pt x="209" y="578" on="1"/>
         <pt x="26" y="578" on="1"/>
       </contour>
       <instructions/>
```

anthrotype:
Works around #975 (until we also have gain the ability to remove overlaps)

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -200,8 +200,9 @@ def build_fontc(source: Path, fontc_cargo_path: Path, build_dir: Path, compare: 
 
 
 def build_fontmake(source: Path, build_dir: Path, compare: str):
+    variable = source_is_variable(source)
     buildtype = "variable"
-    if not source_is_variable(source):
+    if not variable:
         buildtype = "ttf"
     cmd = [
         "fontmake",
@@ -217,6 +218,12 @@ def build_fontmake(source: Path, build_dir: Path, compare: str):
         "--debug-feature-file",
         "debug.fea",
     ]
+    if not variable:
+        # fontmake static builds perform overlaps removal, but fontc can't do that yet.
+        # Disable the filter to make the diff less noisy.
+        # TODO(anthrotype): Remove if/when fontc gains the ability to remove overlaps.
+        # https://github.com/googlefonts/fontc/issues/975
+        cmd.append("--keep-overlaps")
     if compare == _COMPARE_GFTOOLS:
         cmd += [
             "--filter",


cmyr (approved):
makes sense!

rsheeter:
I don't think we should mark the issue fixed?

PR #981, opened 2024-10-02
[crater] Tidy ci report html

cmyr:
Previously we would just write a big contatenated string of html, which was unreadable and also impossible to diff.

With this patch we clean up the html (newlines & indentation) using tidy.

This is particularly important because it means git can efficiently compute deltas on the index, instead of each new commit involving adding a new ~1MB blob to the repo.

JMM

### DIFF
--- a/fontc_crater/src/ci/html.rs
+++ b/fontc_crater/src/ci/html.rs
@@ -32,7 +32,7 @@ pub(super) fn generate(target_dir: &Path) -> Result<(), Error> {
         })
         .collect::<Result<HashMap<_, _>, _>>()?;
 
-    let html_text = make_html(&summary, &sources, &details);
+    let html_text = make_html(&summary, &sources, &details)?;
     let outpath = target_dir.join(HTML_FILE);
     crate::try_write_str(&html_text, &outpath)
 }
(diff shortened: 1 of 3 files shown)


fontc_crater/src/ci/html.rs
```diff
@@ -91,7 +91,22 @@ fn make_html(
...
+    tidy_html(&raw_html)
+}
+
+fn tidy_html(raw_html: &str) -> Result<String, Error> {
+    let opts = tidier::FormatOptions {
+        // indent with tabs to reduce file size.
+        // the '4' option is suggested by the docs; does not mean 4 tabs each indent?
+        indent: tidier::Indent {
+            size: 4,
+            tabs: true,
```
rsheeter, line 104: I cannot endorse use of tabs or 4 space indents, let the great war resume :D

rsheeter (approved):

fontc_crater/src/ci/html.rs
cmyr: not my preference either, but compared to 4 spaces this reduces file size from 1.9MB to 900kb 😒

PR #982, opened 2024-10-02
[normalizer] Consider lookup flags when merging rules

cmyr:
This was causing normalizer to be non-deterministic in some cases; because we were merging rules without considering the lookup flags, we would preserve the flag of the first rule we encountered, which would depend on the iteration order of the hashmap.

This fixes the underlying issue, by not merging any rules that do not have identical lookup flags.

### DIFF
--- a/otl-normalizer/src/gpos.rs
+++ b/otl-normalizer/src/gpos.rs
@@ -284,7 +284,11 @@ impl LookupRules {
             .filter(|lookup| lookups.contains(&lookup.lookup_id))
             .flat_map(|lookup| lookup.iter())
         {
-            match pairmap.entry((rule.rule().first, rule.rule().second.clone())) {
+            match pairmap.entry((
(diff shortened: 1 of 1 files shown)


rsheeter:
I think this also contributes to (fixes?) https://github.com/googlefonts/fontc/issues/643?

otl-normalizer/src/gpos.rs
```diff
@@ -301,6 +305,7 @@ impl LookupRules {
         for rule in pairmap.into_values() {
             match seen.entry((
                 rule.rule().first,
+                rule.lookup_flags(),
                 rule.rule().record1.clone(),
                 rule.rule().record2.clone(),
             )) {
```
rsheeter, line 311: Kinda wish I saw a unit test :)

rsheeter (approved):

PR #986, opened 2024-10-03
[normalizer] Standardize location ordering

cmyr:
The order of variation regions within the region list is left up to the compiler, and in some instances fontc and fontmake were choosing different orderings. This ensures that regardless of the ordering in the region list, deltas are presented in a consistent ordering in the output.


dfrg (approved):
This might deserve an award for smallest functional change PR in fontations & co history :)

cmyr:
and it is also the largest single increase in the crater total diff % to date, I think 😆

PR #989, opened 2024-10-03
[fea-rs] Correctly handle singlesub singleton class

### LINKED ISSUE #988: FEA compilation fails for contextual single sub with singleton target class
Specifically we're failing to compile [Dekko](https://github.com/EbenSorkin/Dekko) because of this.

Per the spec, if you have the following FEA,

```fea
@class1 = [a b c];
@class2 = [ g ];
sub @class1' a by @class2;
```

it is identical to just writing,

```fea
sub @class1' a by a;
```

but we're treating it as an error because the classes have a different length.

I'm not sure if we are also doing this incorrectly for non-contextual single sub statements, but it seems likely?

cmyr:
- fixes #988 

this will get us compiling at least one more font, which is a nice way to make the diff % go up

### DIFF
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -641,6 +641,14 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
                 self.error(replace.unwrap().range(), "cannot sub glyph by glyph class");
                 None
             }
+            // treat singleton class as single glyph, per the spec:
+            // "If the replacement is a singleton glyph class, then the rule
+            // will be treated identically to a format B rule":
+            // http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#5a-gsub-lookuptype-1-single-substitution
+            (GlyphOrClass::Class(c1), GlyphOrClass::Class(c2)) if c2.len() == 1 => {
+                let g2 = *c2.into_iter().next().unwrap();
+                Some((GlyphOrClass::Class(c1), GlyphOrClass::Glyph(g2)))
+            }
             (GlyphOrClass::Class(c1), GlyphOrClass::Class(c2)) if c1.len() != c2.len() => {
                 self.error(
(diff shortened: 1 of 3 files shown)


anthrotype (approved):

anthrotype:
> this will get us compiling at least one more font

@cmyr which one is that, out of curiosity?

cmyr:
I believe it's [Dekko](https://github.com/EbenSorkin/Dekko).

fea-rs/src/compile/compile_ctx.rs
```diff
@@ -641,6 +641,14 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
                 self.error(replace.unwrap().range(), "cannot sub glyph by glyph class");
                 None
             }
+            // treat singleton class as single glyph, per the spec:
+            // "If the replacement is a singleton glyph class, then the rule
+            // will be treated identically to a format B rule":
+            // http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#5a-gsub-lookuptype-1-single-substitution
```
rsheeter, line 647: Optional, it would be nice to write an example as some readers [self] don't know what a format B rule is off the top of their head

fea-rs/test-data/compile-tests/mini-latin/good/single_sub_singleton_class.fea
```diff
@@ -0,0 +1,6 @@
+feature test {
+    # "if the replacement class is a singleton glyph class, the rule is treated
+    #  as identicalto format B" (as if it were a single glyph)
```
rsheeter, line 3: nit: missing space after identical

fea-rs/src/compile/compile_ctx.rs
cmyr: I worry if we follow this practice in general we will end up completely reproducing the spec in our docs; if someone is curious the full section from which this is extracted is available at the provided link.

PR #997, opened 2024-10-03
[fea-rs] Fix include path lexing

cmyr:
statements such as `include(something.fea)` require special treatment, because the stuff in the braces are irregular, and can contain arbitrary symbols. To handle this we attempt to keep track of whether or not we are currently lexing an include statement, but this logic was broken if there was whitespace between the 'include' keyword and the parens.

In addition, we were not correctly handling leading whitespace _within_ the parens; to handle this (along with trailing whitespace) we now lex everything in the parens as a path, and then split the token up if necessary during the parsing pass.


fea-rs/src/parse/lexer.rs
```diff
@@ -446,4 +449,18 @@ mod tests {
...
+        let tokens = tokenize(fea);
+        let token_strs = debug_tokens2(&tokens, fea);
+        assert_eq!(token_strs[0], "IncludeKw");
+        assert_eq!(token_strs[1], "WS( )");
+        assert_eq!(token_strs[2], "(");
+        assert_eq!(token_strs[3], "Path( path.fea )");
+        assert_eq!(token_strs[4], ")");
+        assert_eq!(token_strs[5], ";");
+        assert!(token_strs.get(6).is_none());
+    }
```
rsheeter, line 465: Would love to see the tests here (this and others) change to assert everything at once. A small helper for the tokenenize, debug_tokens2 might be worth considering.

`assert_eq!(vec!["IncludeKw", "WS( )", etc], debug_tokens("fea fragment"));` type of deal.

I see that we sometimes skip indices to avoid writing out the whitespace token, I suggest we either spell it out or add a method to give us the non-whitespace tokens.

rsheeter: (I'm find with deferring this to an issue since this PR follows the current convention)

rsheeter (approved):

fea-rs/src/parse/lexer.rs
cmyr: ya i think the most elegant solution would be some sort of custom macro...  but this module is touched very rarely, and the tests for actual parsing are generally where this sort of stuff lives, and those have much better ergonomics. 🤷

PR #998, opened 2024-10-03
Expose and fix rounding difference vs python.

rsheeter:
Manifests in [soft-type-jacquard/sources/Jacquard12Charted.glyphs](https://github.com/scfried/soft-type-jacquard) as `ySuperscriptYOffset` 451 (fontmake) vs 452 (fontc)


ufo2fontir/src/source.rs
```diff
@@ -1576,7 +1576,10 @@ mod tests {
         types::GlyphName,
     };
     use fontir::{
-        ir::{AnchorKind, GlobalMetricsInstance, GlyphOrder, NameKey, PostscriptNames},
+        ir::{
+            test_helpers::Round2, AnchorKind, GlobalMetricsInstance, GlyphOrder, NameKey,
```
cmyr, line 1580: you're probably going to get yelled at for unused imports, I'd import this down in the test module..

fontir/src/ir.rs
```diff
@@ -719,33 +718,86 @@ impl GlobalMetrics {
...
+    pub superscript_x_size: OrderedFloat<f64>,
+    pub superscript_y_offset: OrderedFloat<f64>,
+    pub superscript_y_size: OrderedFloat<f64>,
+    pub underline_thickness: OrderedFloat<f64>,
+    pub underline_position: OrderedFloat<f64>,
+}
+
+// It's not *that* fun to write long decimals in unit tests
+#[doc(hidden)]
+pub mod test_helpers {
```
cmyr, line 752: could this be `#[cfg(test)]`?

fontir/src/ir.rs
```diff
@@ -719,33 +718,86 @@ impl GlobalMetrics {
...
+    use ordered_float::OrderedFloat;
+
+    use super::GlobalMetricsInstance;
+
+    pub trait Round2 {
```
cmyr, line 757: staring at this it isn't really evident how it's helpful. Maybe a little bit of exposition?

cmyr (approved):
seems reasonable, a few notes inline but nothing blocking :)

rsheeter:
As far as I can tell it can't be cfg test because it is used from a
different crate. It didn't work when I tried it. Maybe there is some
incantation to say please depend on the test version in test?

fontir/src/ir.rs
rsheeter: Ty, will add

Issue #264, opened 2023-04-18
Match base style during .glyphs regular master selection


rsheeter:
In #252 we added a partial implementation of finding the regular master, see https://github.com/googlefonts/fontmake-rs/pull/252#pullrequestreview-1390381752. In particular, we don't implement the full "find a base style shared between all masters" logic. Create a source to prove this is broken and fix it, or ... perhaps compellingly argue we don't need to implement that logic because no known source actually uses it?

anthrotype:
I did find a source that is affected by this:

https://github.com/marcologous/hanken-grotesk/blob/master/sources/HankenGrotesk-Italic.glyphs

crater currently reports it as the lower scorer, with 18.226%

![Image](https://github.com/user-attachments/assets/d2147d78-d146-41c5-a2fb-06069d84c423)

This has three masters, Thin Italic, Italic and Black Italic. fontc currently picks the first one as the base or default master, whereas fontmake (glyphsLib) actually picks the second one named Italic (because that's the style name particle that is shared by all the master names).

picking a different default master obviously yields dramatic diffs when comparing two VF, both the default tables and variable ones will be very different.

If I modify the source file to add a Variable Font Origin custom parameter set to "Italic" master, fontc will produce a font that is almost identical to fontmake's, as they will both use the same master as the default one.

anthrotype:
I also see a bunch of other `*-Italic.glyphs` around the bottom of the https://googlefonts.github.io/fontc_crater/ list with low diff scores, they must be following similar pattern so presumably we are picking the wrong master there as well.

rsheeter:
Love having the concrete examples, crater all the things

anthrotype:
#999 fixes this for Glyphs v3 sources (like HankenGrotesk-Italic), where GSFontMasters have an explicit name attribute which we can use to find the common style name.

For Glyphs v2, masters don't have a name attribute. Their name is the concatenation of three other attributes, `width`, `weight` and `custom` (in this order), with "Regular" stripped away.

search for these attributes under fontMaster in https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv2.md

[...]

PR #999, opened 2024-10-04
implement glyphsLib's logic to find base style shared by all masters

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/264

After this [HankenGrotesk-Italic.glyphs](https://github.com/marcologous/hanken-grotesk/blob/master/sources/HankenGrotesk-Italic.glyphs) is happier

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -1499,9 +1499,7 @@ fn default_master_idx(raw_font: &RawFont) -> usize {
             raw_font
                 .font_master
                 .iter()
-                .enumerate()
-                .find(|(_, master)| master.id == origin)
-                .map(|(idx, _)| idx)
+                .position(|master| master.id == origin)
         })
     {
         return master_idx;
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2740,7 +2740,7 @@ mod tests {
             "Expanded Italic",
             "Expanded Bold Italic",
         ],
-        1 // "Expanded Italic" is common and exactly matches [1]
+        "Expanded Italic"  // is common and exactly matches [1]
     )]
     #[case::base_style_contains_regular(
         &[
@@ -2748,7 +2748,7 @@ mod tests {
             "Regular Foo Baz",
             "Regular Foo",
         ],
-        2 // "Regular Foo" is common and exactly matches [2]
+        "Regular Foo" // is common and exactly matches [2]
     )]
     #[case::base_style_with_regular_omitted(
         &[
@@ -2756,13 +2756,17 @@ mod tests {
             "Condensed Light",
             "Condensed Regular",
         ],
-        2 // "Condensed" is common and matches "Condensed Regular"" when "Regular" is ignored
+        // "Condensed" is common and matches "Condensed Regular" when "Regular" is ignored
+        "Condensed Regular"
     )]
     // "" is common and matches "Regular when "Regular" is ignored
-    #[case::default_to_regular(&["Thin", "Light", "Regular", "Medium", "Bold"], 2)]
+    #[case::default_to_regular(
+        &["Thin", "Light", "Regular", "Medium", "Bold"],
+        "Regular"
+    )]
     // "" is common, nothing matches, just take the first
-    #[case::default_to_first(&["Foo", "Bar", "Baz"], 0)]
-    fn find_default_master(#[case] master_names: &[&str], #[case] expected_idx: usize) {
+    #[case::default_to_first(&["Foo", "Bar", "Baz"], "Foo")]
+    fn find_default_master(#[case] master_names: &[&str], #[case] expected: &str) {
         let mut font = RawFont::default();
         for name in master_names {
             let master = RawFontMaster {
@@ -2772,7 +2776,9 @@ mod tests {
             font.font_master.push(master);
         }
 
-        assert_eq!(expected_idx, default_master_idx(&font));
+        let idx = default_master_idx(&font);
+
+        assert_eq!(expected, font.font_master[idx].name.as_deref().unwrap());
     }
 
     #[test]
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -1492,10 +1492,10 @@ fn parse_codepoint_str(s: &str, radix: u32) -> BTreeSet<u32> {
 fn default_master_idx(raw_font: &RawFont) -> usize {
     // Prefer an explicit origin
     // https://github.com/googlefonts/fontmake-rs/issues/44
-    if let Some(Some(master_idx)) = raw_font
+    if let Some(master_idx) = raw_font
         .custom_parameters
         .string("Variable Font Origin")
-        .map(|origin| {
+        .and_then(|origin| {
             raw_font
                 .font_master
                 .iter()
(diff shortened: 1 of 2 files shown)


glyphs-reader/src/font.rs
```diff
@@ -2683,6 +2725,46 @@ mod tests {
...
+        &[
+            "Condensed Thin",
+            "Condensed Light",
+            "Condensed Regular",
+        ],
+        2
+    )]
+    #[case::default_to_regular(&["Thin", "Light", "Regular", "Medium", "Bold"], 2)]
+    #[case::default_to_first(&["Foo", "Bar", "Baz"], 0)]
+    fn find_default_master(#[case] master_names: &[&str], #[case] expected_idx: usize) {
```
rsheeter, line 2755: Optional: the cases would be easier to read if the expected answer was also a string

glyphs-reader/src/font.rs
```diff
@@ -1502,17 +1502,54 @@ fn default_master_idx(raw_font: &RawFont) -> usize {
...
+                        .join(" ")
+                })
+                .as_ref()
+                == Some(&base_style)
         })
-        .unwrap_or_default()
+        .unwrap_or_default() // just pick the first
+}
+
+// Direct port of glyphsLib's find_base_style:
```
rsheeter, line 1537: While true this doesn't actually help me understand what the relationship between master names and styles is and that makes interpreting the surrounding code difficult.

rsheeter: I think we're saying the base style is the word(s) (defined as stuff delimited by whitespace) that are common in all master names is the base style, maybe we could comment as much and add an example.

glyphs-reader/src/font.rs
```diff
@@ -1502,17 +1502,54 @@ fn default_master_idx(raw_font: &RawFont) -> usize {
...
-        // Still nothing? - just look for one called Regular
-        .or_else(|| {
-            raw_font
-                .font_master
-                .iter()
-                .position(|m| matches!(m.name.as_deref(), Some("Regular")))
+    {
+        return master_idx;
+    }
+    // Look for a master whose style is shared by all masters, defaulting to 'Regular'
```
rsheeter, line 1509: As below, this is factually correct but doesn't really help understand the rationale.

My kneejerk reaction is the code feels more complicated than the underlying operation ... let me think about that a bit. Probably more helpful if it comes with an actual suggestion :)

glyphs-reader/src/font.rs
rsheeter: I would also suggest a common on each for why, e.g. 

```rust
    #[case::base_style_without_regular(
        &[
            "Expanded Thin Italic",
            "Expanded Italic",
            "Expanded Bold Italic",
        ],
        1  // common words "Expanded Italic", exact match at [1]
    )]
    #[case::base_style_contains_regular(
        &[
            "Regular Foo Bar",
            "Regular Foo Baz",
            "Regular Foo",
        ],
        2. // common words "Regular Foo", exact match at [2]
    )]
    #[case::base_style_with_regular_omitted(
        &[
            "Condensed Thin",
            "Condensed Light",
            "Condensed Regular",
        ],
        2 // common words "Condensed", matches [2] when "Regular" dropped
    )]
    // No common words, "Regular" wins
    #[case::default_to_regular(&["Thin", "Light", "Regular", "Medium", "Bold"], 2)]
   // No common words, no "Regular", first entry wins
    #[case::default_to_first(&["Foo", "Bar", "Baz"], 0)]
```

glyphs-reader/src/font.rs
```diff
@@ -1492,7 +1492,7 @@ fn parse_codepoint_str(s: &str, radix: u32) -> BTreeSet<u32> {
 fn default_master_idx(raw_font: &RawFont) -> usize {
     // Prefer an explicit origin
     // https://github.com/googlefonts/fontmake-rs/issues/44
-    raw_font
+    if let Some(Some(master_idx)) = raw_font
```
cmyr, line 1495: instead of `Some(Some(` change the second `map` to `and_then`.

glyphs-reader/src/font.rs
```diff
@@ -1502,17 +1502,54 @@ fn default_master_idx(raw_font: &RawFont) -> usize {
                 .enumerate()
```
cmyr, line 1502: not your code, but instead of enumerate + find, we can use the `position` method..

glyphs-reader/src/font.rs
cmyr: I did also find this pretty confusing, and the logic is also kind of funny but I guess depends highly on the use of standard naming conventions? Agree that a few examples would be useful...

glyphs-reader/src/font.rs
```diff
@@ -1502,17 +1502,54 @@ fn default_master_idx(raw_font: &RawFont) -> usize {
...
+    let master_names: Vec<_> = masters.iter().filter_map(|m| m.name.as_deref()).collect();
+    if master_names.is_empty() {
+        return None;
+    }
+
+    let mut base_style: Vec<_> = master_names[0].split_whitespace().collect();
+
+    for master_name in master_names {
+        let style: Vec<_> = master_name.split_whitespace().collect();
+        base_style.retain(|word| style.contains(word));
```
cmyr, line 1549: no need to allocate here, we can just scan the iterator directly:
```suggestion
        base_style.retain(|word| master_name.whitespace().any(|style| style == word));
```

glyphs-reader/src/font.rs
```diff
@@ -1502,17 +1502,54 @@ fn default_master_idx(raw_font: &RawFont) -> usize {
...
+    }
+
+    Some(base_style.join(" ")).filter(|s| !s.is_empty())
```
cmyr, line 1552: filter should be unnecessary, `split_whitespace` doesn't return empty strings.

(also worth noting[1] just to be academic that this function is lossy if the input contained whitespace that weren't a space)

1. okay it's extremely debatable that this is 'worth noting'

cmyr (commented):
various little rust nits inside 🎁

glyphs-reader/src/font.rs
rsheeter: https://github.com/googlefonts/fontc/pull/1001 attempts to:

1. Implement in terms of picking the best master
1. Expand comments based on parts I found hard to follow
1. Operate on `Vec<&str>` rather than forming String

rsheeter (approved):
LGTM with a couple of suggestions

glyphs-reader/src/font.rs
anthrotype: > the cases would be easier to read if the expected answer was also a string

done in 4dcde7ac

glyphs-reader/src/font.rs
anthrotype: done, thanks!  0834b8b9

glyphs-reader/src/font.rs
anthrotype: of course, sorry! :) 0b352556

glyphs-reader/src/font.rs
anthrotype: thanks for the comment. it should no longer apply though, given that #1001 isn't doing that any more

glyphs-reader/src/font.rs
anthrotype: good to know! since we are now scanning more than once after #1001 I think we do need to allocate

PR #1005, opened 2024-10-07
fix parsing widthClass from glyphs2 instances

anthrotype:
we were only testing parsing instance widthClass for glyphs3 sources, where this property already holds a number 1..9
for glyphs2 this is an enumerated string that maps to numbers from 1 to 9.

`lookup_class_value` is incorrectly mapping the widthClass to numbers from 50 to 200 (i.e. using the corresponding percentages instead of the enumerated integer values)

https://github.com/googlefonts/fontc/blob/4a4019c66902e9ce68c7b98403a734b94a88eaf2/glyphs-reader/src/font.rs#L1971-L1980

... but then the WidthClass::try_from fails because only numbers 1 to 9 are supported:

https://github.com/googlefonts/fontc/blob/4a4019c66902e9ce68c7b98403a734b94a88eaf2/fontdrasil/src/types.rs#L110-L134

I first push a failing test case (simply duplicating an existing test for glyphs3/WdthVar.glyphs, coverted to glyphs2 using the Glyphs.app), then will follow up with the actual fix.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -2842,7 +2842,24 @@ mod tests {
     }
 
     #[test]
-    fn os2_width_class_matches_default_wdth() {
+    fn os2_width_class_matches_default_wdth_glyphs2() {
+        let compile = TestCompile::compile_source("glyphs2/WdthVar.glyphs");
+        let font = compile.font();
+
+        // the default value for 'wdth' is 50 (UltraCondensed) in this test font
+        assert_eq!(
+            vec![(Tag::from_str("wdth").unwrap(), 50.0, 50.0, 200.0)],
+            axes(&font),
+        );
+        // ... which is reflected in the OS/2 table usWidthClass (UltraCondensed = 1)
+        assert_eq!(
+            WidthClass::UltraCondensed as u16,
+            font.os2().unwrap().us_width_class()
+        );
+    }
+
+    #[test]
+    fn os2_width_class_matches_default_wdth_glyphs3() {
         let compile = TestCompile::compile_source("glyphs3/WdthVar.glyphs");
         let font = compile.font();
 
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -550,7 +550,6 @@ mod tests {
         fs::{self, File},
         io::Read,
         path::{Path, PathBuf},
-        str::FromStr,
         sync::Arc,
         time::Instant,
     };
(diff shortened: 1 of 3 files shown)


fontc/src/lib.rs
```diff
@@ -2842,7 +2842,24 @@ mod tests {
...
 
     #[test]
-    fn os2_width_class_matches_default_wdth() {
+    fn os2_width_class_matches_default_wdth_glyphs2() {
+        let compile = TestCompile::compile_source("glyphs2/WdthVar.glyphs");
+        let font = compile.font();
+
+        // the default value for 'wdth' is 50 (UltraCondensed) in this test font
+        assert_eq!(
+            vec![(Tag::from_str("wdth").unwrap(), 50.0, 50.0, 200.0)],
```
cmyr, line 2851:
```suggestion
            vec![(Tag::new(b"wdth"), 50.0, 50.0, 200.0)],
```

glyphs-reader/src/font.rs
```diff
@@ -1422,7 +1422,7 @@ impl RawFont {
                 let Some(value) = lookup_class_value(tag, value) else {
                     return Err(Error::UnknownValueName(value.clone()));
                 };
-                let _ = opt.insert(format!("{}", value as i64));
+                let _ = opt.insert(format!("{}", value));
```
cmyr, line 1425:
```suggestion
                let _ = opt.insert(value.to_string());
```

glyphs-reader/src/font.rs
```diff
@@ -1953,31 +1953,31 @@ fn lookup_class_value(axis_tag: &str, user_class: &str) -> Option<f32> {
...
-        ("wdth", "semicondensed") => Some(87.5),
-        ("wdth", "") => Some(100.0),
-        ("wdth", "Medium (normal)") => Some(100.0),
-        ("wdth", "semiexpanded") => Some(112.5),
-        ("wdth", "expanded") => Some(125.0),
-        ("wdth", "extraexpanded") => Some(150.0),
-        ("wdth", "ultraexpanded") => Some(200.0),
+        ("wght", "thin") => Some(100),
+        ("wght", "extralight") => Some(200),
+        ("wght", "ultralight") => Some(200),
```
cmyr, line 1958:
```suggestion
        ("wght", "extralight" | "ultralight") => Some(200),
```

glyphs-reader/src/font.rs
```diff
@@ -1953,31 +1953,31 @@ fn lookup_class_value(axis_tag: &str, user_class: &str) -> Option<f32> {
...
+        ("wght", "light") => Some(300),
+        ("wght", "") => Some(400),
+        ("wght", "normal") => Some(400),
+        ("wght", "regular") => Some(400),
```
cmyr, line 1962:
```suggestion
        ("wght", "" | "normal" | "regular") => Some(400),
```

..etc

cmyr (approved):
Just some rust idiom notes, otherwise looks good!

PR #1007, opened 2024-10-08
Fix bug where fontc didn't have Glyphs-specific underline defaults

rsheeter:
`fontc` and `fontmake` disagree about `yStrikeoutSize` for [soft-type-jacquard/sources/Jacquard24Charted.glyphs](https://github.com/scfried/soft-type-jacquard).

In outlineCompiler.py in ufo2ft `yStrikeoutSize` is set to `openTypeOS2StrikeoutSize` if present, otherwise `postscriptUnderlineThickness`. Turns out 
 https://github.com/googlefonts/glyphsLib/blob/9d5828d874110c42dfc5f542db8eb84f88641eb5/Lib/glyphsLib/builder/custom_params.py#L1136-L1156 captures parameters that differ from the ufo2ft defaults, notably underline thickness and position.

After this fix `OS/2` and `post` match for Jacquard 24 Charted. Presumably impacts others too.

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -704,15 +704,23 @@ impl Work<Context, WorkId, Error> for GlobalMetricWork {
                     .or(font.hhea_line_gap)
                     .map(|v| v as f64),
             );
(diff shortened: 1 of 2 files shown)


anthrotype (approved):

cmyr (approved):
nice!

PR #1013, opened 2024-10-08
[fea-rs] Simplify singlesub format selection

cmyr:
This matches the behaviour of fonttools; we not attempt to optimize size by splitting glyphs into format1 subtables, and will only use format1 if all the glyphs in a builder share a delta.


fea-rs/src/compile/lookups/gsub_builders.rs
```diff
@@ -67,82 +54,24 @@ impl Builder for SingleSubBuilder {
...
-            format1: BTreeMap<i16, Vec<(GlyphId16, GlyphId16)>>,
-            format2: Vec<(GlyphId16, GlyphId16)>,
+        if self.items.is_empty() {
+            return Default::default();
         }
+        let delta = self
+            .items
+            .iter()
+            .map(|(k, v)| v.to_u16() as i32 - k.to_u16() as i32)
+            .reduce(|acc, val| if acc == val { acc } else { i32::MAX })
```
rsheeter, line 64: maybe I'm just slow today but this reduce is throwing me, please add a comment explaining what's happening here

rsheeter: ...I just realized the comment in question is below (re equidistant), maybe move it up slightly?

rsheeter (approved):

PR #1014, opened 2024-10-08
generate master names when upgrading glyphs2 to 3 

### LINKED ISSUE #264: Match base style during .glyphs regular master selection

anthrotype:
Fixes #264 for glyphs2 as well.

the test currently fails because we aren't generating glyphs2 master names properly (they are all set to 'Regular' which doesn't help)

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2586,6 +2586,11 @@ mod tests {
         assert_load_v2_matches_load_v3("WghtVar_NoExport", LoadCompare::Glyphs);
     }
 
+    #[test]
+    fn read_master_names_2_and_3() {
+        assert_load_v2_matches_load_v3("MasterNames", LoadCompare::Glyphs);
+    }
+
     fn only_shape_in_only_layer<'a>(font: &'a Font, glyph_name: &str) -> &'a Shape {
         let glyph = font.glyphs.get(glyph_name).unwrap();
         assert_eq!(1, glyph.layers.len());
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -885,6 +885,8 @@ struct RawFontMaster {
     name: Option<String>,
 
     weight: Option<String>,
+    width: Option<String>,
+    custom: Option<String>,
 
     weight_value: Option<OrderedFloat<f64>>,
     interpolation_weight: Option<OrderedFloat<f64>>,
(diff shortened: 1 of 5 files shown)


anthrotype:
I made two test glyphs2 sources using master names found in the wild from projects where we weren't generating them correctly, then I saved them as glyphs3 where the masters gain the explicit `name` attribute, and added two tests where we load both v2 and v3 copy of the same source and assert they are the same (I reused the existing `assert_load_v2_matches_load_v3`).

After https://github.com/googlefonts/fontc/pull/1014/commits/f90605e4f33303f9a93c27d0b2fa791d416fed21 both test now pass, which means we generate the same master names for v2 as found in the v3 source.

glyphs-reader/src/font.rs
```diff
@@ -1439,7 +1470,7 @@ impl RawFont {
 
     /// `<See https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md#differences-between-version-2>`
     fn v2_to_v3(&mut self) -> Result<(), Error> {
-        self.v2_to_v3_weight()?;
+        self.v2_to_v3_master_names()?;
```
rsheeter, line 1473: I'm surprised v2_to_v3 weight was removed as opposed to just a new call being added, just want to confirm that was on purpose?

rsheeter (approved):

glyphs-reader/src/font.rs
anthrotype: yeah I did on purpose.. the old v2_to_v3_weight wasn't actually setting any `weight` attribute (which is only found in v2, not in v3), it was actually setting the master.name, but only taking into account `weight`. Whereas my code is also taking into account the additional `width` and `custom` attributes of v2 sources (we weren't parsing those before)

glyphs-reader/src/font.rs
```diff
@@ -1346,19 +1348,48 @@ impl RawFont {
...
-                master
-                    .weight
-                    .take()
-                    .unwrap_or_else(|| String::from("Regular")),
-            );
+            let width_name = master.width.take();
+            let weight_name = master.weight.take();
+            let custom_name = master.custom.take();
+            // Remove Nones, empty strings and redundant occurrences of 'Regular'
+            let mut names: Vec<_> = vec![width_name, weight_name, custom_name]
```
cmyr, line 1372:
```suggestion
            let mut names: Vec<_> = [width_name, weight_name, custom_name]
```

use an array instead of a vec, and skip an allocation

glyphs-reader/src/font.rs
```diff
@@ -1346,19 +1348,48 @@ impl RawFont {
...
+                .into_iter()
+                .flatten()
+                .filter(|x| !x.is_empty() && x != "Regular")
+                .collect();
+            // append "Italic" if italic angle != 0
+            if let Some(italic_angle) = master.italic_angle {
+                if italic_angle != 0.0
+                    && (names.is_empty()
+                        || (!names.contains(&"Italic".into())
+                            && !names.contains(&"Oblique".into())))
```
cmyr, line 1382:
```suggestion
                    && (names.is_empty()
                        || !names.iter().any(|name| name == "Italic" || name == "Oblique")
```

(avoiding allocating the two strings)

cmyr (approved):
some codegolf inline, no pressure though..

glyphs-reader/src/font.rs
anthrotype: oh sorry I just merged it, will follow up with a new PR thanks!

PR #1017, opened 2024-10-08
Take abs of win descent when reading .glyphs files

rsheeter:
OS/2 for Ga Maamli, Alumni Sans Pinstripe, and likely more should match after. Alumni Sans Pinstripe now reports identical in default mode locally.


cmyr (approved):
📈

PR #1025, opened 2024-10-12
Found another hidden cache of defaults for .glyphs

### LINKED ISSUE #1024: fontc doesn't use glyphsLib _defaultsForName for OS/2
https://github.com/googlefonts/glyphsLib/blob/1cb4fc5ae2cf385df95d2b7768e7ab4eb60a5ac3/Lib/glyphsLib/classes.py#L1590-L1601 causes different defaults:

```python
    _defaultsForName = {
        "weight": "Regular",
        "width": "Regular",
        "x-height": 500,
        "cap height": 700,
        "ascender": 800,
        "descender": -200,
        "italic angle": 0,
    }
```

This seems likely to explain OS/2 diffs for at least [oft-type-yarndings/sources/Yarndings12Charted.glyphs](https://github.com/scfried/soft-type-yarndings)

rsheeter:
FIxes #1024. [soft-type-yarndings/sources/Yarndings12Charted.glyphs](https://github.com/scfried/soft-type-yarndings) is now deemed identical locally.

JMM if happy.

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -564,8 +564,15 @@ impl Work<Context, WorkId, Error> for GlobalMetricWork {
         for master in font.masters.iter() {
             let pos = font_info.locations.get(&master.axes_values).unwrap();
 
-            metrics.set_if_some(GlobalMetric::CapHeight, pos.clone(), master.cap_height());
-            metrics.set_if_some(GlobalMetric::XHeight, pos.clone(), master.x_height());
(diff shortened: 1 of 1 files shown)


anthrotype (commented):
This one is unexpected at first look, I would have expected that if a metric value isn't defined then it is not there, period; instead of some arbitrary hardcoded defaults (which may not match the upem).
I'm typing from mobile, but I'll take a closer look next week when I am on laptop

rsheeter:
> I would have expected that if a metric value isn't defined then it is not there, period; instead of some arbitrary hardcoded defaults (which may not match the upem).

I agree wrt expectations but https://github.com/googlefonts/glyphsLib/blob/1cb4fc5ae2cf385df95d2b7768e7ab4eb60a5ac3/Lib/glyphsLib/classes.py#L1590-L1601 - observed to impact real sources that don't match before and do after - does not appear to match our expectations.

PR #1037, opened 2024-10-15
Support panose in .glyphs customParameters

### LINKED ISSUE #1031: support panose font custom parameter for Glyphs sources
this is a 10 value array of integers, search "panose" in

https://handbook.glyphsapp.com/custom-parameter-descriptions/

similar to #1026

rsheeter:
* Fixes #1031
* Makes [are-you-serious/sources/AreYouSerious.glyphs](https://github.com/googlefonts/are-you-serious) ttx_diff identical

JMM if happy.


anthrotype:
+1 for the best font name!

PR #1039, opened 2024-10-15
Tweaks to make reproducing builds from crater easier

rsheeter:
* Set defaults on crater arguments and simplify example commands
* Use the same default for local font repos in ttx_diff
* Default ttx_diff to compare default because gftools doesn't actually do what it says at time of writing
* Add the ability to run ttx_diff against a specific source in a repository
   * `python resources/scripts/ttx_diff.py https://github.com/cyrealtype/Alice#sources/Alice.glyphs` will check for a local copy of the repo, clone it if there isn't one, and then build/compare the glyphs file specified in the fragment
* Document how to run ci locally to get an html report; very useful when trying to alter the html!
* Add a copy repro button to the results

image

Copy repro gets you something like `python resources/scripts/ttx_diff.py https://github.com/googlefonts/are-you-serious#sources/AreYouSerious.glyphs` which ttx_diff now supports.

### DIFF
--- a/fontc_crater/README.md
+++ b/fontc_crater/README.md
@@ -7,13 +7,16 @@ fonts.
 
 
 ```sh
-$ cargo run --release -p=fontc_crater -- compile FONT_CACHE --fonts-repo GOOGLE/FONTS  -o results.json
-```
+# By default font repositories and results will be written to ~/.fontc_crater_cache
 
-or, to run ttx_diff (comparing with fontmake)
+# Just see if we can compile with fontc
+$ cargo run --release -p=fontc_crater -- compile
 
-```sh
-$ cargo run --release -p=fontc_crater -- diff FONT_CACHE --fonts-repo GOOGLE/FONTS  -o results.json
+# To take it for a test-spin do a limited # of fonts
+$ cargo run --release -p=fontc_crater -- compile --limit 8
+
+# Build with fontmake and fontc and compare the results with ttx_diff
+$ cargo run --release -p=fontc_crater -- diff
 ```
 
 This is a binary for executing font compilation (and possibly other tasks) in
@@ -39,7 +42,7 @@ You can generate a report from the saved json by passing it back to
 `fontc_crater`:
 
 ```sh
-$ cargo run -p fontc_crater -- report results.json
+$ cargo run -p fontc_crater -- report
 ```
 
 ## CI
@@ -47,6 +50,17 @@ $ cargo run -p fontc_crater -- report results.json
 This binary is also run in CI. In that case, the execution is managed by a
 script in the [`fontc_crater` repo][crater-repo] and results are posted to
 [github pages][crater-results].
+
+To run in CI mode locally to play with the html output:
+
+```shell
+# clone git@github.com:googlefonts/fontc_crater.git somewhere, we'll assume at ../fontc_crater
+# CI currently has it's own format for the repo list, use the file from ^
+
+$ cargo run --release -p=fontc_crater -- ci ../fontc_crater/gf-repos-2024-08-12.json -o ../fontc_crater/results/ --html_only
+
+# Review ../fontc_crater/results/index.html (NOT ../fontc_crater/index.html)
+```
 
 [google-fonts-sources]: https://github.com/googlefonts/google-fonts-sources
 [google/fonts]: https://github.com/google/fonts
--- a/fontc_crater/src/args.rs
+++ b/fontc_crater/src/args.rs
@@ -21,18 +21,19 @@ pub(super) enum Commands {
 
 #[derive(Debug, PartialEq, clap::Args)]
 pub(super) struct RunArgs {
-    /// Directory to store font sources.
+    /// Directory to store font sources and the google/fonts repo.
     ///
     /// Reusing this directory saves us having to clone all the repos on each run.
     ///
     /// This directory is also used to write cached results during repo discovery.
-    pub(super) font_cache: PathBuf,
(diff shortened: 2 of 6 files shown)


fontc_crater/README.md
```diff
@@ -7,13 +7,16 @@ fonts.
 
 
 ```sh
-$ cargo run --release -p=fontc_crater -- compile FONT_CACHE --fonts-repo GOOGLE/FONTS  -o results.json
-```
+# By default font repositories and results will be written to ./font_repos
```
cmyr, line 10: The repo cache is expected to be transient, I wouldn't want to write anything important there..

fontc_crater/src/args.rs
```diff
@@ -26,21 +26,22 @@ pub(super) struct RunArgs {
     /// Reusing this directory saves us having to clone all the repos on each run.
     ///
     /// This directory is also used to write cached results during repo discovery.
+    #[arg(default_value = "./font_repos")]
     pub(super) font_cache: PathBuf,
     /// Path to local checkout of google/fonts repository
-    #[arg(short, long)]
+    #[arg(short, long, default_value = "../fonts")]
```
cmyr, line 32: is this documented anywhere? feels like a potential footgun to decide we own some arbitrary sibling directory named 'fonts'

fontc_crater/src/args.rs
```diff
@@ -26,21 +26,22 @@ pub(super) struct RunArgs {
...
     pub(super) fonts_repo: Option<PathBuf>,
     /// Optional path to write out results (as json)
-    #[arg(short = 'o', long = "out")]
+    #[arg(short = 'o', long = "out", default_value = "./font_repos/results.json")]
```
cmyr, line 35: - If I've set a different cache dir above, this will still default to here? 
- The current default is to just print a summary to stdout, which is useful sometimes
- if we're deciding that this is always set than there are various places where we're checking if `out_path.is_some()` or similar, and changing our behaviour based on that...

overall my instinct would be to just leave this part, in my actual use of this tool (outside of CI) I am generally comparing multiple runs and so I almost always want the results to have descriptive names, and if a new run overwrote my last result I would be annoyed.

fontc_crater/src/args.rs
```diff
@@ -26,21 +26,22 @@ pub(super) struct RunArgs {
...
     /// for debugging, execute only a given number of fonts
     #[arg(long)]
-    pub(super) n_fonts: Option<usize>,
+    pub(super) limit: Option<usize>,
 }
 
 #[derive(Debug, PartialEq, clap::Args)]
 pub(super) struct CiArgs {
-    /// Path to a json list of repos + revs to run.
+    /// Path to a json list of repos + revs to run.
```
cmyr, line 44: ?

fontc_crater/src/args.rs
```diff
@@ -54,6 +55,7 @@ pub(super) struct CiArgs {
 
 #[derive(Debug, PartialEq, clap::Args)]
 pub(super) struct ReportArgs {
+    #[arg(default_value = "./font_repos/results.json")]
```
cmyr, line 58: I think drop this bit and make it be explicit. The report stuff isn't used for CI anyway and we can probably get rid at some point.

cmyr (commented):
overall looks like some nice improvements but I have concerns about the defaults, specifically `../fonts` and to a lesser extent the `results.json`, more detail inline.

fontc_crater/src/args.rs
rsheeter: Well that right there is a mistake!

fontc_crater/README.md
rsheeter: I think if I address the don't default results there this is addressed?

fontc_crater/src/args.rs
rsheeter: My naive expectation was `--help` would reflect the default. It's a pretty natural place for it to land.

cmyr: i mean it's a reasonable guess, but it still feels sketchy?

bear in mind that,

- if you don't pass anything we'll just use a transient tempdir
- this arg only impacts the non-ci behaviour

fontc_crater/src/args.rs
rsheeter: I think I can live without this one. Having font_repos and the location of google fonts default is very helpful because it makes a copy/pasteable command just about plausible instead of having to copy, fiddle with, and run.

fontc_crater/src/args.rs
rsheeter: google/fonts is relatively heavy, I don't think re-pullking it all the time is a good default. I changed the default to under font_repos which is now gitignore'd.

fontc_crater/src/args.rs
rsheeter: (removed)

fontc_crater/src/args.rs
rsheeter: Done.

fontc_crater/src/args.rs
cmyr: just want to point out that for CI you only fetch this repo when you're making the input list, which you do rarely.

cmyr (approved):
Okay so to summarize a discussion from chat:

- we can just get rid of the 'fonts_repo' arg completely, it's a distraction. I've done that in https://github.com/googlefonts/fontc/pull/1041
- for hard-coding a default cache location I would prefer something like `.fontc_crater_cache`, under the user's home directory. This isn't perfect, but a relative path won't work in the CI case (since we clone the `fontc` repo into a new working directory and then rm -rf it after each run)

PR #1040, opened 2024-10-15
[ttx_diff] Retry normalizer on failure

cmyr:
We have spurious normalizer failures sometimes, and this will now check for those cases and retry (up to five times) if they occur.

JMM

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -46,6 +46,7 @@ from cdifflib import CSequenceMatcher as SequenceMatcher
 from typing import MutableSequence
 from glyphsLib import GSFont
 from fontTools.designspaceLib import DesignSpaceDocument
+import time
 
 
 _COMPARE_DEFAULTS = "default"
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -121,11 +121,23 @@ def ttx(font_file: Path, can_skip: bool):
...
 
-# generate a simple text repr for gpos for this font
+# generate a simple text repr for gpos for this font, with retry
 def simple_gpos_output(cargo_manifest_path: Path, font_file: Path, out_path: Path, can_skip: bool):
+    NUM_RETRIES = 5
+    for i in range(NUM_RETRIES+1):
+        try:
+           return simple_gpos_output_impl(cargo_manifest_path, font_file, out_path, can_skip)
+        except subprocess.CalledProcessError as e:
+            if i == NUM_RETRIES:
```
rsheeter, line 131: I know it shouldn't happen but ... >= perhaps

Also I think you should wait if you failed, maybe a random period from 100 to 2500 ms or something of that nature

rsheeter (approved):

resources/scripts/ttx_diff.py
```diff
@@ -121,11 +122,24 @@ def ttx(font_file: Path, can_skip: bool):
...
+            time.sleep(0.1)
+            if i >= NUM_RETRIES:
+                raise e
+            print(f"normalizer failed with code '{e.returncode}'', retrying", file=sys.stderr)
+
+def simple_gpos_output_impl(cargo_manifest_path: Path, font_file: Path, out_path: Path, can_skip: bool):
     if not (can_skip and out_path.is_file()):
         temppath = font_file.parent / "markkern.txt"
         cmd = [
+            "timeout",
```
anthrotype, line 141: Maybe this was already metnioned somewhere but, I couldn't find this `timeout` command on my macOS 14.7 environment. I had to do `brew install coreutils` and then I had to symlink gtimeout to timeout. I suppose the 'g' stands for GNU and brew doesn't want to step on macos feet.

cmyr: oh yes you will need `coreutils`, but I don't recall having to do an additional symlink? I would only expect the `g` prefix on commands that are also included by the system (and I just double checked this and `brew install coreutils` was enough for me?)

rsheeter: I think it would be OK to either make timeout optional, off by default, or to only do it if `which timeout` indicates it exists. I think I prefer the flag option so CI can't accidentally lose timeout.

anthrotype: oh actually subprocess.run has a timeout parameter so we might not even need to use an external command for this, since we are actually relying on subprocess already?

https://docs.python.org/3/library/subprocess.html#subprocess.run

PR #1050, opened 2024-10-17
Expose and fix broken processing of bbox for components whose transform is more than scale and transform

rsheeter:
Simplified reproduction of problem impacting Luxurious Roman (`python resources/scripts/ttx_diff.py https://github.com/googlefonts/luxurious-roman#sources/Luxurious-Roman.glyphs`)

```shell
$ python resources/scripts/ttx_diff.py resources/testdata/glyphs2/MatrixComponent.glyphs
...glyf, head, hhea, hmtx diff...

$ diff ~/oss/fontc/build/default/fontc.glyf.ttx ~/oss/fontc/build/default/fontmake.glyf.ttx
38c38
<     <TTGlyph name="infinity" xMin="750" yMin="272" xMax="877" yMax="443">
---
>     <TTGlyph name="infinity" xMin="767" yMin="284" xMax="863" yMax="431">
```

This appears to be due to fontc assuming it can simply apply the component transform to the control box when in fact it needs to transform the points and compute the control box of the new points. https://codepen.io/rs42/pen/wvVqVPL?editors=1000 sketches the impact of transformation for the new test font.

Reproduced in simplified test that now passes. Luxurious Roman is now identical when ttx_diff'd locally.

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -3,7 +3,10 @@
 //! Each glyph is built in isolation and then the fragments are collected
 //! and glued together to form a final table.
 
-use std::collections::{BTreeSet, HashMap, HashSet};
+use std::{
+    collections::{BTreeSet, HashMap, HashSet},
+    sync::Arc,
+};
 
 use fontdrasil::{
     coords::NormalizedLocation,
@@ -11,7 +14,7 @@ use fontdrasil::{
     types::GlyphName,
 };
 use fontir::{
-    ir,
+    ir::{self, GlyphOrder},
     orchestration::{Flags, WorkId as FeWorkId},
     variations::{VariationModel, VariationRegion},
 };
@@ -30,6 +33,7 @@ use write_fonts::{
         },
         gvar::{iup::iup_delta_optimize, GlyphDelta},
     },
+    types::GlyphId16,
     OtRound,
 };
 
@@ -77,19 +81,10 @@ fn can_reuse_metrics(
     coeffs == Affine::IDENTITY.as_coeffs()
 }
 
-fn create_component(
-    context: &Context,
-    ref_glyph_name: &GlyphName,
+fn create_component_ref_gid(
+    gid: GlyphId16,
     transform: &Affine,
 ) -> Result<(Component, Bbox), GlyphProblem> {
-    // Obtain glyph id from static metadata
-    let gid = context
-        .ir
-        .glyph_order
-        .get()
-        .glyph_id(ref_glyph_name)
-        .ok_or(GlyphProblem::NotInGlyphOrder)?;
-
     // No known source does point anchoring so we just turn transform into a 2x2 + offset
     let [a, b, c, d, e, f] = transform.as_coeffs();
     let flags = ComponentFlags {
@@ -99,8 +94,8 @@ fn create_component(
     let component = Component::new(
         gid,
         Anchor::Offset {
-            x: e as i16,
-            y: f as i16,
+            x: e.ot_round(),
+            y: f.ot_round(),
         },
         Transform {
             xx: F2Dot14::from_f32(a as f32),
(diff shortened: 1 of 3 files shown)


fontbe/src/glyphs.rs
```diff
@@ -731,22 +742,62 @@ fn affine_for(component: &Component) -> Affine {
...
+    affine: Affine,
+) -> Result<Option<Rect>, Error> {
+    // For simple scale+translate transforms, which seem to be common, we could just transform a bbox
+    // Let's wait to see if that pops out in a profile and do the simple solution for now
+    // Because transforms can skew/rotate the control box computed for the simple glyph isn't always reusable
+
+    let mut bbox: Option<Rect> = None;
+    for component in composite.components() {
+        // The transform we get here has changed because it got turned into F2Dot14 and i16 parts
+        // We could go get the "real" transform from IR but ... this seems to match fontmake so far
```
anthrotype, line 766: the 'real' transform from source or IR doesn't actually matter any more, it's the one that's stored in the glyf table that matters for bbox computation, that's what fonttools TTGlyph.recalcBounds() does (and ufo2ft uses the same method on the compiled TTGlyphs when building the glyf table).

fontbe/src/glyphs.rs
```diff
@@ -763,71 +814,13 @@ fn compute_composite_bboxes(context: &Context) -> Result<(), Error> {
...
-        if pending == composites.len() {
-            return Err(Error::CompositesStalled(
-                composites.iter().map(|g| g.name.clone()).collect(),
-            ));
-        }
+    for (glyph_name, glyph) in glyphs.iter().filter_map(|(gn, g)| match &g.data {
+        RawGlyph::Composite(composite) => Some((*gn, composite)),
+        RawGlyph::Simple(..) | RawGlyph::Empty => None,
+    }) {
+        let bbox = bbox_of_composite(&glyph_order, &glyphs, glyph, Affine::IDENTITY)?;
```
anthrotype, line 822: you went for a recursive traversal instead of iterative one. though unlikely, should you prevent from a component cycle triggering an infinite recursion? Python has a max recursion depth after which it raises an exception. Does Rust protect you as well?

fontbe/src/glyphs.rs
```diff
@@ -731,22 +742,62 @@ fn affine_for(component: &Component) -> Affine {
...
 
+/// See <https://github.com/fonttools/fonttools/blob/42c1a52c5facd0edbc9c685787b084af44f6f607/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L1244>
+fn bbox_of_composite(
+    glyph_order: &GlyphOrder,
+    glyphs: &HashMap<&GlyphName, Arc<Glyph>>,
+    composite: &CompositeGlyph,
+    affine: Affine,
+) -> Result<Option<Rect>, Error> {
+    // For simple scale+translate transforms, which seem to be common, we could just transform a bbox
+    // Let's wait to see if that pops out in a profile and do the simple solution for now
```
anthrotype, line 760: > Let's wait to see if that pops out in a profile

did you actually try to profile before/after this? I suggest filing an issue

fontbe/src/glyphs.rs
anthrotype: I think for glyphs sources at least we already check against component cycles when we propagate anchors, see glyphs-reader's propagate_anchors.rs (depth_sorted_composite_glyphs)

anthrotype (approved):
LGTM with minor comments

fontbe/src/glyphs.rs
rsheeter: Let me take confirmation of no cycles as a follow-on. I have the vague notion we might already do it but I'll double-check.

fontbe/src/glyphs.rs
rsheeter: Good idea, https://github.com/googlefonts/fontc/issues/1061

fontbe/src/glyphs.rs
rsheeter: https://github.com/googlefonts/fontc/issues/1062

PR #1051, opened 2024-10-17
[ttx_diff] Remove manifest path arguments

cmyr:
I'm not really sure what these were doing there? My best guess is that I'd had a little brain fart and forgot that I could just pass `-p` to specify the package that cargo should build, since package selection is the only thing we're using this for.


cmyr:
Oops: for the record we need to use an explicit manifest path because we are running in a tempdir in crater.

PR #1052, opened 2024-10-17
[ttx_diff] ensure only the GPOS lookups get xpath'ed for pruning

anthrotype:
right now the `gpos.xpath("//Lookup")` ends up selecting and removing GSUB lookups of type 2, 4, 5 and 6 as well, even though we only intended to prune the GPOS ones.

That's the way `//` works, it will select all matching elements following the current node anywhere they appear, not only their descendants. GSUB follows GPOS since ttx sorts table alphabetically.
Putting a dot in front `.//` ensures only the descendants of the current node are selected.

see https://www.w3schools.com/xml/xpath_syntax.asp

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -412,7 +412,10 @@ def remove_mark_and_kern_lookups(ttx):
     gpos = ttx.find("GPOS")
     if gpos is None:
         return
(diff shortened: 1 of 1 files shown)


cmyr:
well that is certainly nasty...

rsheeter (approved):

PR #1055, opened 2024-10-17
[ttx_diff] Better printing logic

cmyr:
Instead of only printing if we aren't outputting json, we can always print, but to stderr.


rsheeter (approved):
When it grows up it'll be just like Rust

PR #1056, opened 2024-10-18
[ttx_diff] Simplify reuse logic

cmyr:
At the beginning of the run, we now check the 'rebuild' flag, and explicitly delete anything we are required to rebuild. This lets us simplify later logic by knowing we're allowed to reuse anything that exists.

This is part of a larger refactor to support the CI use case; because CI builds each individual font in a transient temporary directory, these artifacts do not persist between runs. To reuse artifacts CI is going to need to start storing these artifacts somewhere other than in the build directory.

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -42,6 +42,7 @@ import json
 import shutil
 import subprocess
 import sys
+import os
 from urllib.parse import urlparse
 from cdifflib import CSequenceMatcher as SequenceMatcher
 from typing import MutableSequence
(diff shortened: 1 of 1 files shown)


rsheeter (approved):

resources/scripts/ttx_diff.py
```diff
@@ -246,11 +249,7 @@ def build_fontmake(source: Path, build_dir: Path, compare: str):
         ]
     cmd.append(str(source))
 
-    return build(
-        cmd,
-        build_dir,
-        "fontmake",
-    )
+    return build(cmd, build_dir)
```
anthrotype, line 252: the `build` function doesn't return anything as far as I can see

anthrotype (approved):
LGTM with minor comment

(I'm starting to wonder if ttx_diff has grown so much that it may use some tests)

PR #1058, opened 2024-10-18
[ttx_diff] More cleanup

cmyr:
- Improve some type annotations
- rename 'run' to 'log_and_run'
- use kwargs to forward more args to subprocss.run
- use subprocess.run timeout arg instead of gnu timeout

JMM

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -45,7 +45,7 @@ import sys
 import os
 from urllib.parse import urlparse
 from cdifflib import CSequenceMatcher as SequenceMatcher
-from typing import MutableSequence
+from typing import Sequence
 from glyphsLib import GSFont
 from fontTools.designspaceLib import DesignSpaceDocument
 import time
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -90,18 +90,15 @@ def eprint(*objects):
...
-# The 'check' argument is passed to subprocess.run; if it is true than an
-# exception will be raised if the command does not exit successfully; otherwise
-# the caller can check the status via the returned `CompletedProcess` object.
-def run(cmd: MutableSequence, working_dir: Path, check=False, **kwargs):
+# execute a command after logging it to stderr.
+# All additional kwargs are passed to subprocess.run
+def log_and_run(cmd: Sequence, cwd=None, **kwargs):
     cmd_string = " ".join(str(c) for c in cmd)
-    eprint(f"  (cd {working_dir} && {cmd_string})")
+    eprint(f"  (cd {cwd} && {cmd_string})")
```
anthrotype, line 97: `cd None` would look weird, maybe don't echo cd something if cwd is None?

anthrotype (approved):

PR #1065, opened 2024-10-23
[crater] Initial support for richer targets

cmyr:
_this is blocked until I do a release of google-fonts-sources_

This will facilitate gftools mode.

Ultimately I decided to just rip out a bunch of the old functionality, because keeping it working was becoming a chore and it really wasn't being used much.

### DIFF
--- a/fontc_crater/Cargo.toml
+++ b/fontc_crater/Cargo.toml
@@ -10,7 +10,7 @@ readme = "README.md"
 [dependencies]
 fontc = { version = "0.0.1", path = "../fontc" }
 
-google-fonts-sources = "0.3.1"
+google-fonts-sources = "0.5.0"
 maud = "0.26.0"
 tidier = "0.5.3"
 
--- a/fontc_crater/src/args.rs
+++ b/fontc_crater/src/args.rs
@@ -1,8 +1,13 @@
 //! CLI args
 
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use clap::{Parser, Subcommand};
+
+// this env var can be set by the runner in order to reuse git checkouts
+// between runs.
+static GIT_CACHE_DIR_VAR: &str = "CRATER_GIT_CACHE";
+const DEFAULT_CACHE_DIR: &str = "~/.fontc_crater_cache";
 
 #[derive(Debug, PartialEq, Parser)]
 #[command(about = "compile multiple fonts and report the results")]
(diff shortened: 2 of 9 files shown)


fontc_crater/Cargo.toml
```diff
@@ -10,7 +10,7 @@ readme = "README.md"
 [dependencies]
 fontc = { version = "0.0.1", path = "../fontc" }
 
-google-fonts-sources = "0.3.1"
+google-fonts-sources = { path = "../../google-fonts-sources" }
```
rsheeter, line 13: Reminder to move back to a released version pre-submit

fontc_crater/src/args.rs
```diff
@@ -13,33 +18,21 @@ pub(super) struct Args {
...
-    pub(super) cache_dir: PathBuf,
-    /// Optional path to write out results (as json)
-    #[arg(short = 'o', long = "out")]
-    pub(super) out_path: Option<PathBuf>,
-    /// for debugging, execute only a given number of fonts
-    #[arg(long)]
-    pub(super) limit: Option<usize>,
-}
+    #[arg(short, long = "cache")]
+    cache_dir: Option<PathBuf>,
```
rsheeter, line 34: Why did we remove the default value from here? - now it won't show up in `--help`

fontc_crater/src/args.rs
```diff
@@ -1,9 +1,14 @@
 //! CLI args
 
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use clap::{Parser, Subcommand};
 
+// this env var can be set by the runner in order to reuse git checkouts
+// between runs.
+static GIT_CACHE_DIR_VAR: &str = "CRATER_GIT_CACHE";
```
rsheeter, line 9: It's not obvious why we need to probe an environment variable *and* support a cli option. Tbh, I'd prefer only cli arg but if we think we really need this lets document why.

fontc_crater/src/args.rs
rsheeter: If we have to keep the env var lets document it and the interaction with this flag here

fontc_crater/src/ci.rs
```diff
@@ -140,11 +132,54 @@ fn run_crater_and_save_results(args: &CiArgs) -> Result<(), Error> {
...
+                Ok(x) => x,
+                Err(e) => {
+                    log::warn!("failed to load config '{}': '{e}'", config_path.display());
+                    continue;
+                }
+            };
+            for source in &config.sources {
+                let src_path = config_path
+                    .parent()
+                    .expect("config path always in sources dir")
```
rsheeter, line 166: nit: here and below we don't actually verify what dir, should it just be `config path is a file in a directory`?

rsheeter (approved):

fontc_crater/src/args.rs
cmyr: The reason this uses an env var is because in the CI environment we don't call this executable directly; we call the `run.sh` script in the `fontc_crater` repo, which in turn calls the `ci.sh` script in this repo, which executes this binary. The environment variable is set in the config for the scheduler that launches the CI task on the host machine.

fontc_crater/src/ci.rs
cmyr: I use `expect` as basically an alternative to an assert, or an unwrap + comment.

These are both trying to communicate the invariant that the paths returns from `iter_configs` always start with the `cache_dir` argument, and necessarily include the `source` (or `Source`) directory component.

fontc_crater/src/args.rs
cmyr: added a note about the default to the help docs.

As I think you inferred this was removed so that we could check for `None` and then use the env var if it's set.

PR #1069, opened 2024-10-27
gftools mode for crater

cmyr:
This wires everything up to get gftools working in crater.

- currently blocked on ~https://github.com/googlefonts/google-fonts-sources/pull/24 and~ https://github.com/googlefonts/gftools/pull/1047
- but maybe we don't actually need to merge the gftools patch? If we can pip install from a git rev or branch, we could just maintain our own branch of gftools and not need to dirty up the main tree.
- we skip noto fonts and other fonts that have a custom recipe provider
- there are a number of failures because it turns out that in lots of cases one source does produce multiple fonts; a bunch of these I could address by turning off config options (so to not build small caps or italics) but a handful of fonts are building a regular and a bold, and I didn't see a config option for this.

### DIFF
--- a/fontc_crater/src/ci.rs
+++ b/fontc_crater/src/ci.rs
@@ -18,7 +18,7 @@ use crate::{
     args::CiArgs,
     error::Error,
     ttx_diff_runner::{DiffError, DiffOutput},
-    Results, Target,
+    BuildType, Results, Target,
 };
 
 mod html;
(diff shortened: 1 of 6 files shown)


fontc_crater/src/ci.rs
```diff
@@ -176,12 +176,46 @@ fn make_targets(cache_dir: &Path, repos: &[RepoInfo]) -> (Vec<Target>, BTreeMap<
...
+    let gftools = should_build_in_gftools_mode(src_path, config).then(|| Target {
+        config: config_path.to_owned(),
+        source: src_path.to_owned(),
+        build: BuildType::GfTools,
+    });
+    [default, gftools].into_iter().flatten()
+}
+
+fn should_build_in_gftools_mode(src_path: &Path, config: &Config) -> bool {
+    // skip noto, which have an implicitly different recipe provider
```
rsheeter, line 206: can we link to where python does this implicit recipe provider selection?

fontc_crater/src/ci.rs
```diff
@@ -176,12 +176,46 @@ fn make_targets(cache_dir: &Path, repos: &[RepoInfo]) -> (Vec<Target>, BTreeMap<
...
+    if src_path
+        .file_stem()
+        .and_then(|stem| stem.to_str().map(|s| s.to_lowercase().contains("noto")))
```
rsheeter, line 209: starts with "noto " might be slightly safer

fontc_crater/src/ci.rs
```diff
@@ -176,12 +176,46 @@ fn make_targets(cache_dir: &Path, repos: &[RepoInfo]) -> (Vec<Target>, BTreeMap<
...
+        .unwrap_or(false)
+    {
+        return false;
+    }
+
+    // if there is a recipe provider other than googlefonts, we skip
```
rsheeter, line 215: should tersely say why, e.g. because we have no idea what it will do, it's arbitrary code

resources/scripts/ttx_diff.py
```diff
@@ -259,6 +285,9 @@ def copy(old, new):
     return new
 
 
+# def find_and_copy_one_file(from_dir: Path, to_file: Path):
```
rsheeter, line 288: leftover?

rsheeter (approved):

fontc_crater/src/ci.rs
cmyr: it doesn't, gftools cannot build these fonts, even with python.

fontc_crater/src/ci.rs
cmyr: oh definitely, for some reason I thought 'noto' wasn't always at the start...

fontc_crater/src/ci.rs
rsheeter: Maybe we should refer the reader to https://github.com/googlefonts/oxidize/blob/main/text/2024-06-26-fixes-and-nonstandard-builds.md#noto ?

Issue #1043, opened 2024-10-16
master values should be ot-rounded *before* computing deltas in order to match fontmake


anthrotype:
take https://github.com/SorkinType/Gelasio/tree/main/sources/Gelasio.glyphspackage and run through ttx_diff.py

You'll notice MVAR has one delta related to "stro" tag (strikeout offset or position) which differs by 1 unit when compared with fontmake's

```diff
--- /Users/clupo/oss/fontc/build/default/fontc.MVAR.ttx	2024-10-16 12:21:15
+++ /Users/clupo/oss/fontc/build/default/fontmake.MVAR.ttx	2024-10-16 12:21:15
@@ -22,7 +22,7 @@
         <NumShorts value="0"/>
         <!-- VarRegionCount=1 -->
         <VarRegionIndex index="0" value="0"/>
-        <Item index="0" value="[5]"/>
+        <Item index="0" value="[4]"/>
         <Item index="1" value="[8]"/>
       </VarData>
     </VarStore>
\ No newline at end of file
```

The reason for this is that fontmake builds TTFs for each master and then uses varLib to build MVAR among other tables, and the global metrics values as stored in the master TTFs are already rounded to integers (using otRound) so the input to varLib are ot-rounded integers, not floats.

Whereas fontc keeps the global metrics value as floats (the sources allow to store these as floats even though basically all these metrics are going to end up being rounded to int16 or uint16 when serialised to OpenType) and then it generates the deltas from these float metrics, and only at the end it ot-rounds the deltas themselves.

In this particular case, Gelasio doesn't actually contain floats for strikeout offset; in fact it doesn't specify any strikeout offset, so it is the GlobalMetrics::populate_defaults() method that is introducing the floats, here:

https://github.com/googlefonts/fontc/blob/37143a0c39535aaf8eaa42be53fa028bd6746cc3/fontir/src/ir.rs#L698-L701

Now, in order to match fontmake we could either

1) change `ir::GlobalMetrics` so that it stores integer values instead of f64, and do an ot_round() for all incoming values (from any frontends, not just glyphs or ufo) as soon as they get set, eg

```diff
- HashMap<GlobalMetric, HashMap<NormalizedLocation, OrderedFloat<f64>>>
+ HashMap<GlobalMetric, HashMap<NormalizedLocation, i32>>
```

[...]

PR #1070, opened 2024-10-28
Fix rounding of VariationModel::deltas

anthrotype:
the first commit is a repro for https://github.com/googlefonts/fontc/issues/1043#issuecomment-2426907952

the lack of rounding within the delta computation can lead to deltas that are off-by-one by the time they are eventually ot-rounded to i16, as it is shown in Gelasio's MVAR table (from which the reproducer is drawn).

The commits that follow will apply the fix for #1043 as well as for #235, and add more tests borrowed from fonttools models_test.py

Basically we need to apply the rounding within the very same loop in which deltas get subtracted away from the absolute master values. This is to make sure that the rounding errors are accounted for and don't accummulate beyond 0.5 when multiple regions overlap.

The use of the banker's "round half to even" method (`round_ties_even` in Rust) matches the fonttools varLib's choice of Python's built-in `round()` function. The latter is preferred over OtRound for _relative_ values such as deltas primarily because of its symmetry: e.g. round(1.5) == 2 _and_ round(-1.5) == -2, not -1. OtRound (`(v + 0.5).floor()`) is biased towards higher, positive values and makes more sense for absolute coordinates.

/cc @behdad

### DIFF
--- a/fontir/src/variations.rs
+++ b/fontir/src/variations.rs
@@ -21,11 +21,11 @@ use write_fonts::{
 
 use crate::error::VariationModelError;
 
-/// Generic trait for rounding half-way values to the nearest even number.
+/// Trait for rounding half-way values to the nearest even number.
 ///
 /// For example, 2.5 rounds to 2.0, 3.5 rounds to 4.0, and -2.5 rounds to -2.0.
 /// This is the same rounding mode as [`f64::round_ties_even`], which was
-/// stabilized in Rust 1.77.0. The trait provides a generic way to apply this
+/// stabilized in Rust 1.77.0. The trait provides a common way to apply this
 /// to different types besides `f64` e.g. `kurbo::Vec2`.
 /// We use it below in [`VariationModel`] for rounding variation deltas.
 ///
@@ -39,25 +39,25 @@ use crate::error::VariationModelError;
 /// For more info, see:
 /// - discussion in a related FontTools PR <https://github.com/fonttools/fonttools/pull/2214>
 /// - 'Rounding half to even' section in <https://en.wikipedia.org/wiki/Rounding>.
-pub trait RoundTiesEven<U, T = Self> {
-    fn round_ties_even(self) -> U;
+pub trait RoundTiesEven {
+    fn round_ties_even(self) -> Self;
 }
 
-impl RoundTiesEven<f64> for f64 {
+impl RoundTiesEven for f64 {
     #[inline]
     fn round_ties_even(self) -> f64 {
         self.round_ties_even()
     }
 }
 
-impl RoundTiesEven<f32> for f32 {
+impl RoundTiesEven for f32 {
     #[inline]
     fn round_ties_even(self) -> f32 {
         self.round_ties_even()
     }
 }
 
-impl RoundTiesEven<kurbo::Vec2> for kurbo::Vec2 {
+impl RoundTiesEven for kurbo::Vec2 {
     #[inline]
     fn round_ties_even(self) -> kurbo::Vec2 {
         kurbo::Vec2::new(self.x.round_ties_even(), self.y.round_ties_even())
@@ -214,7 +214,7 @@ impl VariationModel {
     ) -> Result<Vec<(VariationRegion, Vec<V>)>, DeltaError>
     where
         P: Copy + Default + Sub<P, Output = V>,
-        V: Copy + Mul<f64, Output = V> + Sub<V, Output = V> + RoundTiesEven<V>,
+        V: Copy + Mul<f64, Output = V> + Sub<V, Output = V> + RoundTiesEven,
     {
         if point_seqs.is_empty() {
             return Ok(Vec::new());
@@ -1419,7 +1419,7 @@ mod tests {
     #[derive(Debug, Default, Copy, Clone, PartialEq)]
     struct NoRoundF64(f64);
 
-    impl RoundTiesEven<NoRoundF64> for NoRoundF64 {
+    impl RoundTiesEven for NoRoundF64 {
         #[inline]
         fn round_ties_even(self) -> NoRoundF64 {
             self
--- a/fontir/src/variations.rs
+++ b/fontir/src/variations.rs
@@ -297,6 +297,8 @@ impl VariationModel {
     /// Convert relative deltas to absolute values at the given location.
     ///
     /// Rust version of <https://github.com/fonttools/fonttools/blob/4ad6b0db/Lib/fontTools/varLib/models.py#L514-L545>
+    ///
+    /// TODO: document invariants and what we are returning
     pub fn interpolate_from_deltas<V>(
         &self,
         location: &NormalizedLocation,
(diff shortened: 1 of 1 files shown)


fontir/src/variations.rs
```diff
@@ -21,6 +21,44 @@ use write_fonts::{
...
+/// 3.5 rounds to 4.0, and -2.5 rounds to -2.0.
+///
+/// This is the same rounding mode as [`f64::round_ties_even`], which was
+/// stabilized in Rust 1.77.0. It also matches Python's buit-in `round` function.
+/// It is usually preferred over alternative approaches because it reduces bias
+/// and errors in calculations.
+///
+/// This trait provides a generic way to apply banker's rounding to different types,
+/// such as `f64` and `kurbo::Vec2`.
+pub trait BankerRound<U, T = Self> {
```
rsheeter, line 37: Suggest naming to match the rust method, e.g. RoundTiesEven not "BankerRound." I notice the documentation for round ties even doesn't even mention bankers, it's not key to the concept.

fontir/src/variations.rs
```diff
@@ -21,6 +21,44 @@ use write_fonts::{
...
+/// Trait for performing banker's rounding on a value.
+///
+/// Banker's rounding rounds half to even. This means that half-way values
+/// are rounded to the nearest even number. For example, 2.5 rounds to 2.0,
+/// 3.5 rounds to 4.0, and -2.5 rounds to -2.0.
+///
+/// This is the same rounding mode as [`f64::round_ties_even`], which was
+/// stabilized in Rust 1.77.0. It also matches Python's buit-in `round` function.
+/// It is usually preferred over alternative approaches because it reduces bias
+/// and errors in calculations.
```
rsheeter, line 33: Suggest elaborating on when this makes sense and ideally adding a citation. Perhaps we can reference the python PR or other artifacts where it was adopted?

fontir/src/variations.rs
rsheeter: Also someone familiar with the Rust library are far more likely to recognize round_ties_even

fontir/src/variations.rs
```diff
@@ -1314,13 +1393,174 @@ mod tests {
...
+
+        fn mul(self, rhs: f64) -> Self::Output {
+            NoRoundF64(self.0 * rhs)
+        }
+    }
+
+    #[test]
+    fn modeling_error_within_tolerance() {
+        // Compare interpolating from un-rounded float deltas vs rounded deltas,
+        // and check that rounding errors don't accummulate but stay within <= 0.5.
```
rsheeter, line 1441: Suggest putting discussion of accumulating rounding error in the RoundTiesEven trait documentation

rsheeter (approved):

fontir/src/variations.rs
anthrotype: I thought I could not shadow the built-in f64::round_ties_even method with my own trait, but turns out it works.. I'll rename

fontir/src/variations.rs
```diff
@@ -243,6 +286,41 @@ impl VariationModel {
...
+    where
+        V: Copy + Mul<f64, Output = V> + Add<V, Output = V>,
+    {
+        let mut result = None;
+        for (region, deltas) in deltasets.iter() {
+            let scalar = region.scalar_at(location).into_inner() as f64;
+            if scalar == 0.0 {
+                continue;
+            }
+            let contribution = deltas.iter().map(|d| *d * scalar).collect::<Vec<_>>();
```
cmyr, line 307: taking this as written, there's no need to collect here; you only need to actually allocate a new vec if the result is currently none.

fontir/src/variations.rs
```diff
@@ -243,6 +286,41 @@ impl VariationModel {
...
+            if result.is_none() {
+                result = Some(contribution);
+            } else {
+                result = Some(
```
cmyr, line 311: taken as written, you can write this as a match statement, like:

```rust
result = match result {
    None => Some(thing),
    Some(thing) => Some(thing.modified()),
};
```

fontir/src/variations.rs
```diff
@@ -243,6 +286,41 @@ impl VariationModel {
...
+                    result
+                        .unwrap()
+                        .iter()
+                        .zip(contribution.into_iter())
+                        .map(|(r, c)| *r + c)
+                        .collect(),
```
cmyr, line 317: this is allocating a new vec to store the result on each iteration, but it would be more efficient to just modify the existing values in place.

fontir/src/variations.rs
anthrotype: I'll add link to the wikipedia entry on Rounding and a link to the fonttools PR

fontir/src/variations.rs
```diff
@@ -243,6 +286,41 @@ impl VariationModel {
...
+    /// Rust version of <https://github.com/fonttools/fonttools/blob/4ad6b0db/Lib/fontTools/varLib/models.py#L514-L545>
+    pub fn interpolate_from_deltas<V>(
+        &self,
+        location: &NormalizedLocation,
+        deltasets: &[(VariationRegion, Vec<V>)],
+    ) -> Vec<V>
+    where
+        V: Copy + Mul<f64, Output = V> + Add<V, Output = V>,
+    {
+        let mut result = None;
```
cmyr, line 301: This looks funny.

- declaring something as a mutable `Option` that always gets unwrapped is confusing, and it is not clear to me that the unwrap is okay; wouldn't this panic if scalar was always `0.0`?
- it looks a lot like what you're trying to do is a `fold` or `reduce` operation, and it at least _looks_ like it could be expressible via iteration? I would think of something like,

```rust
deltasets
    .iter()
    .map(|(region, deltas)| (region.scalar_at(location).0 as f64, deltas))
    .filter(|(scalar, _deltas)| *scalar != 0.0)
    .fold(None, |result, (scalar, deltas)| {
        let contribution = deltas.iter().map(|d| *d * scalar);
        match result {
            None => Some(contribution.collect::<Vec<_>>()),
            Some(mut existing) => {
                existing
                    .iter_mut()
                    .zip(contribution)
                    .for_each(|(acc, val)| *acc = *acc + val);
                Some(existing)
            }
        }
    })
// .unwrap_or(what?)
```
- and then I think it would be useful to document all of the invariants of this method, specifically: should `deltasets` all be the same length? can `deltasets` be empty? 
- and finally also documenting what we are returning, as in 'a vec of the same length as the input `deltasets`, containing the accumulated deltas.

fontir/src/variations.rs
```diff
@@ -243,6 +286,41 @@ impl VariationModel {
...
+                        .unwrap()
+                        .iter()
+                        .zip(contribution.into_iter())
+                        .map(|(r, c)| *r + c)
+                        .collect(),
+                );
+            }
+        }
+
+        result.unwrap()
```
cmyr, line 322: as noted above, I'm not convinced this can't crash?

cmyr (approved):
A bunch of Rust style nits inline. Thanks for taking this one on, it looks fiddly...

fontir/src/variations.rs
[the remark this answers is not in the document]
anthrotype: I dunno I copied this pattern from OtRound in write_fonts lol

I'll simplify, thanks :)

fontir/src/variations.rs
anthrotype: > should deltasets all be the same length?

indeed. I wish Rust had something like zip(strict=True) that would fail if the two iterators have different lenghts. Looks like there's zip_eq in itertools crate but don't want to add a new dep for that. Given this is supposed to be used for deltas that are in turn generated by VariationModel itself where this invariant already holds, it should be fine just documenting thus.

fontir/src/variations.rs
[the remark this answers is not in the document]
cmyr: OtRound is different because the output is not always the same as the implementing type. :)

fontir/src/variations.rs
anthrotype: I made it private for now, until we'll have the need to make it pub and polish it more

PR #1071, opened 2024-10-28
crater html display tweaks

cmyr:
- add a list of tables that differ inline
- move the reproduction link into the details view (we were running out of space)

Screenshot 2024-10-28 at 5 37 38 PM

### DIFF
--- a/fontc_crater/resources/style.css
+++ b/fontc_crater/resources/style.css
@@ -64,13 +64,21 @@ td.rev {
   font-family: monospace;
 }
 
-.font_path .diff_result {
+.diff_result {
+  width: 10em;
   display: inline-block;
 }
 
 .font_path {
   width: 50%;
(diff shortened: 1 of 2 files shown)


rsheeter:
> move the reproduction link into the details view (we were running out of space)

I suggest instead of moving this we might show only the filename of the source (drop the path/segments/before/filename)

rsheeter (approved):

cmyr:
There are several sources in our list that have identical names, and differ only based on the repo.

rsheeter:
> several sources in our list that have identical names, and differ only based on the repo

I should have written when the name is unique :)

cmyr:
going to merge this and see how it feels, and then we can always followup. The link to the repo might not even be necessary if we have one-click reproduction.

PR #1074, opened 2024-10-29
[crater] Prebuild fontc & normalizer

cmyr:
Currently each invocation of ttx_diff attempts to `cargo build` these targets. This is a problem if you want to run crater locally and also continue working on fontc, since ttx_diff will attempt to recompile any saved changes.

It is also a possible source of various intermittent 'other' errors including some very clear instances where the binaries are not in the target dir as expected.

this also removes use of `timeout` in ci; we now timeout otl-normalizer in the ttx_diff script, and I don't think we need two timeouts (but if it turns out we do, we can put it back)

### DIFF
--- a/fontc_crater/src/ci.rs
+++ b/fontc_crater/src/ci.rs
@@ -9,6 +9,7 @@ use std::{
     collections::BTreeMap,
     fmt::Write,
     path::{Path, PathBuf},
+    process::Command,
 };
 
 use chrono::{DateTime, TimeZone, Utc};
@@ -103,10 +104,26 @@ fn run_crater_and_save_results(args: &CiArgs) -> Result<(), Error> {
     let cache_dir = args.cache_dir();
     log::info!("using cache dir {}", cache_dir.display());
 
+    // we want to build fontc & normalizer once, and then move them out of the
+    // build directory so that they aren't accidentally rebuilt or deleted
+    // while we're running
+    let temp_bin_dir = tempfile::tempdir().expect("couldn't create tempdir");
+    let (fontc_path, normalizer_path) = precompile_rust_binaries(temp_bin_dir.path());
+
+    log::info!("compiled fontc to {}", fontc_path.display());
+    log::info!("compiled otl-normalizeer to {}", normalizer_path.display());
+
     let (targets, source_repos) = make_targets(&cache_dir, &inputs);
     let n_targets = targets.len();
+
+    let context = super::ttx_diff_runner::TtxContext {
+        fontc_path,
+        normalizer_path,
+        cache_dir,
+    };
+
     let began = Utc::now();
-    let results = super::run_all(targets, &cache_dir, super::ttx_diff_runner::run_ttx_diff)?
+    let results = super::run_all(targets, &context, super::ttx_diff_runner::run_ttx_diff)?
         .into_iter()
         .map(|(target, result)| (target.id(), result))
         .collect();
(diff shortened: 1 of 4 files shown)


fontc_crater/src/ci.rs
```diff
@@ -246,3 +263,44 @@ fn format_elapsed_time<Tmz: TimeZone>(start: &DateTime<Tmz>, end: &DateTime<Tmz>
     write!(&mut out, "{secs}s").unwrap();
     out
 }
+
+fn precompile_rust_binaries(temp_dir: &Path) -> (PathBuf, PathBuf) {
+    fn copy_file_into_dir(file_path: &Path, dir_path: &Path) -> PathBuf {
+        let new_file_path = dir_path.join(file_path.file_name().unwrap());
+        std::fs::copy(file_path, &new_file_path).expect("failed to copy binary to tempdir");
+        new_file_path
+    }
```
rsheeter, line 272: I find this style *far* harder to read than top-level functions, can we push this up a level to stand alone?

fontc_crater/src/ci.rs
```diff
@@ -246,3 +263,44 @@ fn format_elapsed_time<Tmz: TimeZone>(start: &DateTime<Tmz>, end: &DateTime<Tmz>
...
+
+    let fontc = compile_crate_or_die_trying("fontc");
+    let normalizer = compile_crate_or_die_trying("otl-normalizer");
```
rsheeter, line 275: we're php now (or die). Nit: _trying isn't really adding anything, x_or_die would be equally clear

fontc_crater/src/ci.rs
```diff
@@ -246,3 +263,44 @@ fn format_elapsed_time<Tmz: TimeZone>(start: &DateTime<Tmz>, end: &DateTime<Tmz>
...
+        .status()
+        .expect("failed to run cargo build");
+    if !status.success() {
+        panic!("cargo build '{name}' failed");
+    }
+
+    find_binary_target(name)
+}
+
+fn find_binary_target(name: &str) -> PathBuf {
```
rsheeter, line 296: "find" implies a level of searching but we're just predicting, perhaps drop find_?

For consistency with compile or die perhaps it shoudl be locate_binary_or_die or some such?

fontc_crater/src/main.rs
```diff
@@ -86,10 +86,10 @@ enum SkipReason {
...
 
 #[allow(clippy::type_complexity)] // come on, it's not _that_ bad
-fn run_all<T: Send, E: Send>(
+fn run_all<T: Send, E: Send, Cx: Sync>(
     targets: Vec<Target>,
-    cache_dir: &Path,
-    runner: impl Fn(&Path, &Target) -> RunResult<T, E> + Send + Sync,
+    context: &Cx,
+    runner: impl Fn(&Cx, &Target) -> RunResult<T, E> + Send + Sync,
 ) -> Result<Vec<(Target, RunResult<T, E>)>, Error> {
```
rsheeter, line 93: I'm with clippy, this seems over-complicated. Do we really need it to be so generic?

fontc_crater/src/ci.rs
rsheeter: Also ... should this just be directly in compile crate?

resources/scripts/ttx_diff.py
```diff
@@ -641,6 +651,29 @@ def build_crate(manifest_path: Path):
...
+    if norm_path is None:
+        otl_norm_manifest_path = root_dir / "otl-normalizer" / "Cargo.toml"
+        norm_path = root_dir / "target" / "release" / "otl-normalizer"
+        build_crate(otl_norm_manifest_path)
+        assert norm_path.is_file(), "failed to build otl-normalizer?"
+    else:
+        norm_path = Path(norm_path)
+        assert norm_path.is_file(), f"normalizer path '{norm_path}' does not exist"
+
+    return (fontc_path, norm_path)
```
rsheeter, line 674: This is basically the same thing repeated twice, make a fn?

resources/scripts/ttx_diff.py
```diff
@@ -650,14 +683,8 @@ def main(argv):
...
-    fontc_manifest_path = root / "fontc" / "Cargo.toml"
-    fontc_bin_path = root / "target" / "release" / "fontc"
-    otl_norm_manifest_path = root / "otl-normalizer" / "Cargo.toml"
-    otl_bin_path = root / "target" / "release" / "otl-normalizer"
-    build_crate(otl_norm_manifest_path)
-    build_crate(fontc_manifest_path)
-    assert otl_bin_path.is_file(), "failed to build otl-normalizer?"
-    assert fontc_bin_path.is_file(), "failed to build fontc?"
+
+    (fontc_bin_path, otl_bin_path) = get_fontc_and_normalizer_binary_paths(root)
```
rsheeter, line 687: Seems clearer to have a function that takes in the crate name and the root path and call it twice

```python
fontc_file = get_binary(root, FLAGS.fontc)
# etc
```

rsheeter (approved):
LGTM subject to comments being addressed

fontc_crater/src/main.rs
cmyr: it's a legacy of the earlier design, where this one function was responsible for executing various different runners that would produce different types of output. It could be simplified but doesn't feel like a priority...

fontc_crater/src/ci.rs
cmyr: I mean it could be, but that doesn't really seem better in any meaningful way?

PR #1075, opened 2024-10-29
[crater] Compute and store 'hash' of python packages

cmyr:
This is a coarse way of determining whether or not any python packages have been modified, which we can use for invalidating a potential cache.

Also adds a env var to force CI to rerun, which is something I often want during development.

### DIFF
--- a/fontc_crater/src/main.rs
+++ b/fontc_crater/src/main.rs
@@ -4,6 +4,7 @@ use std::{
     collections::BTreeMap,
     fmt::Display,
     path::{Path, PathBuf},
+    process::{Command, Stdio},
     str::FromStr,
     sync::atomic::{AtomicUsize, Ordering},
 };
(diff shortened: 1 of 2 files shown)


rsheeter:
> Also adds a env var to force CI to rerun, which is something I often want during development.

Google generally favors flags to environment variables. I recommend we do the same.

fontc_crater/src/main.rs
```diff
@@ -135,6 +136,22 @@ fn get_git_rev(repo_path: Option<&Path>) -> Option<String> {
...
+        .unwrap();
+    let sha1sum = Command::new("shasum")
+        .stdin(Stdio::from(pipfreeze.stdout.unwrap()))
+        .output()
+        .expect("shasum should be preinstalled everywhere");
+    assert!(sha1sum.status.success());
+    std::str::from_utf8(sha1sum.stdout.trim_ascii())
+        .expect("shasum output always ascii")
+        .to_owned()
+}
```
rsheeter, line 153: You don't *really* need a crytographic hash or to spawn although I suppose it's harmless enough

rsheeter (approved):

fontc_crater/src/main.rs
cmyr: this was just the quickest way I could think of to do this in a consistent and portable way. 🤷

PR #1079, opened 2024-10-30
[crater] Cache output of fontmake between runs

cmyr:
This clears the cache if the pip freeze output is different.

It's possible that this isn't worth the complexity, but it should offer meaningful runtime speed improvements, so let's give it a try?

### DIFF
--- a/fontc_crater/src/ci/results_cache.rs
+++ b/fontc_crater/src/ci/results_cache.rs
@@ -1,0 +1,84 @@
+//! Caching fontmake's output between runs
+
+use std::path::{Path, PathBuf};
+
+use crate::Target;
+
+static CACHE_DIR_NAME: &str = "crater_cached_results";
+
+// the files that we cache for each target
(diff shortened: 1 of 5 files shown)


fontc_crater/src/ci/results_cache.rs
```diff
@@ -0,0 +1,118 @@
...
+        std::fs::copy(font, to_dir.join(FONT_FILE)).map(|_| ())?;
+        std::fs::copy(ttx, to_dir.join(TTX_FILE)).map(|_| ())?;
+        std::fs::copy(markkern, to_dir.join(MARKKERN_FILE)).map(|_| ())?;
+        Ok(true)
+    } else {
+        Ok(false)
+    }
+}
+
+fn all_exist(paths: [&Path; 3]) -> bool {
```
rsheeter, line 111: I don't see any reason to limit to 3 and I'm suprrised this isn't done via Iterator, e.g. paths.iter.all

rsheeter (approved):

PR #1082, opened 2024-10-31
make sure masters for global/glyph/layout metrics are rounded before computing deltas

### LINKED ISSUE #1043: master values should be ot-rounded *before* computing deltas in order to match fontmake

anthrotype:
Fixes #1043

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ rayon = "1.6"
 icu_properties = "1.4"
 
 # fontations etc
-write-fonts = { version = "0.29.0", features = ["serde", "read"] }
+write-fonts = { version = "0.30.0", features = ["serde", "read"] }
 skrifa = "0.24.0"
 norad = "0.14.2"
 
@@ -53,10 +53,3 @@ members = [
     "otl-normalizer",
     "fontc_crater",
 ]
-
-# TEMPORARY
-[patch.crates-io]
-font-types = { git="https://github.com/googlefonts/fontations.git" }
-read-fonts = { git="https://github.com/googlefonts/fontations.git" }
-write-fonts = { git="https://github.com/googlefonts/fontations.git" }
-skrifa = { git="https://github.com/googlefonts/fontations.git" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ icu_properties = "1.4"
 
 # fontations etc
 write-fonts = { version = "0.29.0", features = ["serde", "read"] }
-skrifa = "0.21.0"
+skrifa = "0.24.0"
 norad = "0.14.2"
 
 # dev dependencies
@@ -53,3 +53,10 @@ members = [
     "otl-normalizer",
     "fontc_crater",
 ]
+
+# TEMPORARY
+[patch.crates-io]
+font-types = { git="https://github.com/googlefonts/fontations.git" }
+read-fonts = { git="https://github.com/googlefonts/fontations.git" }
+write-fonts = { git="https://github.com/googlefonts/fontations.git" }
+skrifa = { git="https://github.com/googlefonts/fontations.git" }
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -158,7 +158,15 @@ pub(crate) fn resolve_variable_metric<'a>(
 ) -> Result<(i16, Vec<(VariationRegion, i16)>), DeltaError> {
     let point_seqs: HashMap<_, _> = values
         .into_iter()
-        .map(|(pos, value)| (pos.to_owned(), vec![value.0 as f64]))
+        .map(|(pos, value)| {
+            // The master values for anchor positions or kerning adjustments are
+            // expected to be rounded before computing the deltas, because instancing
+            // a VF at the masters' location is expected to be equivalent to building
+            // individual masters as static fonts. fontmake does the same, see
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -53,3 +53,10 @@ members = [
     "otl-normalizer",
     "fontc_crater",
 ]
+
+# TEMPORARY
```
rsheeter, line 57: Please release fontations and remove this

fontbe/src/features.rs
```diff
@@ -158,7 +158,15 @@ pub(crate) fn resolve_variable_metric<'a>(
...
     let point_seqs: HashMap<_, _> = values
         .into_iter()
-        .map(|(pos, value)| (pos.to_owned(), vec![value.0 as f64]))
+        .map(|(pos, value)| {
+            // The master values for anchor positions or kerning adjustments are
+            // expected to be rounded before computing the deltas, because instancing
+            // a VF at the masters' location is expected to be equivalent to building
+            // individual masters as static fonts. fontmake does the same, see
+            // https://github.com/googlefonts/fontc/issues/1043
+            let value: f32 = value.into_inner().ot_round();
```
rsheeter, line 167: should we ot_round to f64 and avoid f32 then f64? - I recall we need to add that to otround, maybe we should do that first?

rsheeter (approved):
LGTM if tested. Seems like there should be inputs with observably different results?

fontbe/src/features.rs
anthrotype: I already added OtRound<f64> for f64 and OtRound<f32> for f32 here: 
https://github.com/googlefonts/fontations/pull/1209 (unreleased yet, which is why this PR is fetching fontations from github)

Are you saying we should also add OtRound<f64> for f32? Then why not any other combinations of primitive number types? It feels to me a bit out of scope for a rounding trait that we are casting f32 as f64 because of some other API which works best with those.

anthrotype:
> Seems like there should be inputs with observably different results?

yes, I mentioned Teachers-Italic.glyphs in https://github.com/googlefonts/fontc/issues/1043#issuecomment-2444388789, which contains a composite with a scaled component that gets its anchors propagated as float coordinates and produces different deltas when these get rounded beforehand. I can add a subset of that one the fontc/lib.rs integration tests.
It's harder to find real world examples of masters where advance widths or kerning values or global font metrics contain floats, because designers normally use whole integer numbers.

anthrotype:
I made a test in a separate branch to show it failing on `main`: https://github.com/googlefonts/fontc/pull/1087
but I'll cherry pick it here anyway so I'll close that as soon CI fails

Cargo.toml
anthrotype: there are several other changes to read-fonts and write-fonts in fontations and I am unsure about bumping minor vs patch or whether there are other pending stuff that waits to be merged before release. Maybe @cmyr or @dfrg can take care of that?

anthrotype: done

PR #1085, opened 2024-10-31
[crater] Use timestamp to key crater results

cmyr:
We were using the fontc rev as the key to a map of runs -> detailed results, which doesn't work because it's possible for multiple runs to share the same git rev (e.g. if the inputs or the installed python packages have changed)

JMM


rsheeter (approved):
I would suggest just giving each successive run an id, perhaps simply assigning count(results) +1. Time is notoriously prone to snap when you least expect it.

cmyr:
I agree that time isn't the best choice in a vacuum but our timestamps are normalized to utc and runs occur every hour or so at the most so... it should be okay?

- It's also not a huge deal if it breaks, since this is only used for loading the results of the previous run, for figuring out deltas
- and the summary file is already on disc, so if we want to change its format (by adding a field) we'd also need code to handle the version transition, which seems not worth?


tldr if this ends up breaking as a result of this choice I accept full responsibility.

PR #1090, opened 2024-11-04
[ttx_diff] Use pip-compile and requirements.in

cmyr:
This is a more consistent way of ensuring we have pinned our python dependencies.


rsheeter (approved):

fontc_crater/README.md
```diff
@@ -22,14 +22,17 @@ script in the [`fontc_crater` repo][crater-repo] and results are posted to
...
+(part of [`pip-tools`]) to generate a pinned `requirements.txt` from a
+`requirements.in` file in the root `resources/scripts` directory. To update the
+set of python packages in use, you need to update this file:
 
 - `rm -rf` your existing venv
 - create a new venv and activate it
-- `pip install -r resources/scripts/requirements.txt`
-- `pip freeze > resources/scripts/constraints.txt`
+- install pip-tools with `python -m pip install pip-tools`
+- delete `resources/scripts/requirements.txt`
```
anthrotype, line 33: I don't think you need to delete it as it will be overwritten, but it's ok

fontc_crater/README.md
```diff
@@ -22,14 +22,17 @@ script in the [`fontc_crater` repo][crater-repo] and results are posted to
...
-python pacakges we are comparing against; to ensure this we use `constraints.txt`
-in the root `resources/scripts` directory. To update the set of python packages
-in use, you need to update this file:
+python pacakges we are comparing against; to ensure this we use `pip-compile`
+(part of [`pip-tools`]) to generate a pinned `requirements.txt` from a
+`requirements.in` file in the root `resources/scripts` directory. To update the
+set of python packages in use, you need to update this file:
 
 - `rm -rf` your existing venv
 - create a new venv and activate it
```
anthrotype, line 31: one doesn't need to delete and create a new venv when using pip-compile, it may well run from a global site-packages (e.g. I install it globally with `pipx install pip-tools`) as pip-compile does not modify its environment.

anthrotype: installing something in a venv is generally good practice especially when developing onto something, in this case we are simply using a development tool and it's not a requirement that it runs within a venv, so I don't think you should mention this here

anthrotype (approved):

fontc_crater/README.md
cmyr: My intention is to provide instructions that will work. If I omit installing `pip-compile`, these instructions won't work, and it feels more complicated to try and explain that there are multiple ways I can install pip-compile? If the reader knows how this all works then I think they'll still be okay...

anthrotype: no I didn't mean you should omit installing pip-compile, of course. I mean you can omit the suggestion to (re-)create a venv before running the usual pip install something.

cmyr: ooooh yes okay I understand, good point.

fontc_crater/README.md
anthrotype: have you actually tried _not_ to delete the existing requirements.txt? I think pip-compile is smart enough to only update what's strictly needed. Unless you really want to achieve the equivalent of `pip-compile --upgrade requirements.in` (which tries to upgrade all deps to latest version, not only the ones that are required) with your manual removal? You may well use the pip-compile --upgrade flag then.

cmyr: if there's an existing requirements.txt it doesn't modify it, it just prints to stdout..

anthrotype: Did you provide an output path -o requirements.txt? I think it prints to stout if you don't

cmyr: it creates `requirements.txt` if it doesn't exist, if there are no arguments.

PR #1091, opened 2024-11-05
[glyphs] Generate tuples instead of bincode

cmyr:
This removes the need to deserialize bincode in order to load the data. Ideally we could just generate the types directly but my various (extensive) attempts to do that kept running into extremely slow compile times; this is a compromise.


rsheeter:
Per f2f we landed #1066

PR #1093, opened 2024-11-05
Compact generated data

rsheeter:
Per conversation with @cmyr on #1066, try to compact data generated from glyphsLib xml files. Don't rust format, don't space nicely, just ram it all together in a compact form.


cmyr (approved):
lgtm

PR #1096, opened 2024-11-06
Add a missing +1

rsheeter:
Key fix is a missing +1:

```rust
    fn sequence_len(&self) -> usize {
        self.input_sequence.len() + 1
    }
```

Makes OS/2 match for 
```shell
$ python resources/scripts/ttx_diff.py https://github.com/SorkinType/Alata#sources/Alata.glyphs
```

### DIFF
--- a/fontbe/src/os2/max_context.rs
+++ b/fontbe/src/os2/max_context.rs
@@ -48,21 +48,37 @@ pub(super) fn compute_max_context_value(gpos: Option<&Gpos>, gsub: Option<&Gsub>
         .unwrap_or_default()
 }
 
+/// The "chain" in Python
+#[derive(Debug, Copy, Clone)]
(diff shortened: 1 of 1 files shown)


anthrotype:
so basically, since we are computing the ChainedClassSequenceRule's InputGlyphCount from the length of the inputSequence and because the latter, by definition, skips the first glyph, we need to +1

https://learn.microsoft.com/en-us/typography/opentype/spec/chapter2#chained-sequence-context-format-2-class-based-glyph-contexts

image

```diff
 impl Rule for ChainedClassSequenceRule {
     fn sequence_len(&self) -> usize {
-        self.input_sequence.len()
+        self.input_sequence.len() + 1
     }
```

anthrotype (approved):

PR #1099, opened 2024-11-06
Include empty glyphs when calculating max width, to match fonttools

Hoolean:
Most fields in `hhea` are set based [only on glyphs with bounds](https://github.com/fonttools/fonttools/blob/e04dfaab/Lib/fontTools/ttLib/tables/_h_h_e_a.py#L101-L119), but `advanceWidthMax` is an exception to this, as it considers [every glyph that has an entry in `hmtx`](https://github.com/fonttools/fonttools/blob/e04dfaab/Lib/fontTools/ttLib/tables/_h_h_e_a.py#L73).

This commit includes empty glyphs in the calculation of the field, which matches fontmake's behaviour. After this change, Fleur De Leah compiles identically:

```bash
$ python resources/scripts/ttx_diff.py https://github.com/googlefonts/fleurdeleah#sources/FleurDeLeah.glyphs
  ...
output is identical
```

### DIFF
--- a/fontbe/src/metrics_and_limits.rs
+++ b/fontbe/src/metrics_and_limits.rs
@@ -81,6 +81,9 @@ impl GlyphLimits {
 
 impl FontLimits {
     fn update(&mut self, id: GlyphId16, advance: u16, glyph: &Glyph) {
+        // max advance width should consider every glyph that has an hmtx entry
+        self.advance_width_max = max(self.advance_width_max, advance);
+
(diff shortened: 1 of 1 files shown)


Hoolean:
(forgot to add: [the spec confirms that fonttools is doing the right thing here too](https://learn.microsoft.com/en-us/typography/opentype/spec/hhea) :balance_scale:)

cmyr (approved):
good catch, this makes sense!

PR #1101, opened 2024-11-07
use gftools==0.9.74, update glyphsLib==6.9.4

anthrotype:
we no longer need to install gftools from github, as our patches were released upstream.

also, glyphsLib [6.9.4 reverted a change in 6.9.3](https://github.com/googlefonts/glyphsLib/pull/1044) that caused lots of unnecessary diffs in GlyphOrder

JMM


anthrotype:
I just merged this to kick a new fontc_crater run

PR #1102, opened 2024-11-07
Support long parameter name for PANOSE, to match glyphsLib

Hoolean:
glyphsLib supports longer alternative names for some of its parameters, [including PANOSE](https://github.com/googlefonts/glyphsLib/blob/050ef62c52906bc8ca820e7cfc6963138253f41f/Lib/glyphsLib/builder/custom_params.py#L322-L323) (albeit with [the short name taking precedence](https://github.com/googlefonts/glyphsLib/blob/050ef62c52906bc8ca820e7cfc6963138253f41f/Lib/glyphsLib/builder/custom_params.py#L258-L269)). This PR adds support for the longer name, and tests that it is parsed and has its precedence weighed correctly.

With this change, most of the OS/2 diff for Noto Sundanese is resolved; an unrelated diff in `xAvgCharWidth` remains.

```bash
$ python resources/scripts/ttx_diff.py https://github.com/notofonts/sundanese#sources/NotoSansSundanese.glyphs
```

Before:
```
...
DIFF 'OS_2', [...]/build/default/fontc.OS_2.ttx [...]/build/default/fontmake.OS_2.ttx (79.6%)
...
```

After:
```
...
DIFF 'hmtx', [...]/build/default/fontc.hmtx.ttx [...]/build/default/fontmake.hmtx.ttx (98.9%)
...
```

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -437,10 +437,10 @@ impl CustomParameters {
     }
 
     fn panose(&self) -> Option<&Vec<i64>> {
-        let Some(CustomParameterValue::Panose(values)) = self.get("panose") else {
-            return None;
-        };
-        Some(values)
+        match self.get("panose").or_else(|| self.get("openTypeOS2Panose")) {
+            Some(CustomParameterValue::Panose(values)) => Some(values),
+            _ => None,
+        }
     }
 }
 
@@ -569,7 +569,9 @@ impl FromPlist for CustomParameters {
                                 value =
                                     Some(CustomParameterValue::CodepageRange(tokenizer.parse()?));
                             }
-                            _ if name == Some(String::from("panose")) => {
+                            _ if name == Some(String::from("panose"))
+                                || name == Some(String::from("openTypeOS2Panose")) =>
+                            {
                                 let Token::OpenParen = peek else {
                                     return Err(Error::UnexpectedChar('('));
                                 };
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -1966,8 +1966,18 @@ mod tests {
 
     #[test]
     fn captures_panose() {
+        let expected: Option<Panose> = Some([2, 0, 5, 3, 6, 0, 0, 2, 0, 3].into());
+
+        // short parameter name
         let (_, context) = build_static_metadata(glyphs3_dir().join("WghtVarPanose.glyphs"));
-        let expected: Panose = [2, 0, 5, 3, 6, 0, 0, 2, 0, 3].into();
-        assert_eq!(Some(expected), context.static_metadata.get().misc.panose);
+        assert_eq!(expected, context.static_metadata.get().misc.panose);
+
+        // long parameter name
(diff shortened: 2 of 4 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -1966,8 +1966,18 @@ mod tests {
...
     fn captures_panose() {
+        let expected: Option<Panose> = Some([2, 0, 5, 3, 6, 0, 0, 2, 0, 3].into());
+
+        // short parameter name
         let (_, context) = build_static_metadata(glyphs3_dir().join("WghtVarPanose.glyphs"));
-        let expected: Panose = [2, 0, 5, 3, 6, 0, 0, 2, 0, 3].into();
-        assert_eq!(Some(expected), context.static_metadata.get().misc.panose);
+        assert_eq!(expected, context.static_metadata.get().misc.panose);
+
+        // long parameter name
```
rsheeter, line 1975: Could you please make these separate tests instead of having one test that does three seperable checks?

glyphs-reader/src/font.rs
```diff
@@ -437,10 +437,10 @@ impl CustomParameters {
     }
 
     fn panose(&self) -> Option<&Vec<i64>> {
-        let Some(CustomParameterValue::Panose(values)) = self.get("panose") else {
-            return None;
-        };
-        Some(values)
+        match self.get("panose").or_else(|| self.get("openTypeOS2Panose")) {
```
rsheeter, line 440: Add a comment citing https://github.com/googlefonts/glyphsLib/blob/050ef62c52906bc8ca820e7cfc6963138253f41f/Lib/glyphsLib/builder/custom_params.py#L322-L323

For context we've been trying hard to keep the breacrumbs back to python (or C++ or w/e)

rsheeter (changes requested):
LGTM other than a couple of nits

glyphs2fontir/src/source.rs
Hoolean: Yes, good shout - I've split them now

glyphs-reader/src/font.rs
Hoolean: I've now linked the relevant sections for the names and their precedence

Hoolean: I was unable to find this prior to making the PR but have since identified the section of the Glyphs handbook where these longer parameter names are described, in case a later refactor applies them more broadly and needs somewhere to refer back to as well:

https://handbook.glyphsapp.com/custom-parameter-descriptions/
> Custom parameters in camelcase ("exampleName") are defined in the UFO specification and change the font information. In contrast, capitalized ones ("Example Name") are specific to Glyphs and usually change something in the font (for instance, run a filter on the outlines). UFO parameters follow the naming convention for Font Info properties outlined in the UFO 3 specification published in March 2012. Glyphs also makes use of a simplified naming convention. Wherever possible, leave out the prefix of the keyword. For example, instead of "openTypeNameDescription", use "description", or "blueScale" instead of "postscriptBlueScale". Both long and short versions work side by side.

rsheeter: Ty!

rsheeter (approved):

rsheeter:
Thank you!

PR #1108, opened 2024-11-13
Normalize the order of glyf contours

### LINKED ISSUE #1107: Contour order may not match fontmake
We have some `glyf` diffs that are purely contour order. For example,

```shell
$ python resources/scripts/ttx_diff.py 'https://github.com/googlefonts/arizonia#sources/Arizonia.glyphs'
```

has the same contours in a different order. It isn't clear what "correct" order is as source order may mix contours and components but we don't preserve that in fontmake. Per consultation with experts it doesn't matter so eliminate this in ttx_diff.

rsheeter:
Fixes #1107

Arizonia, and hopefully more, now ttx_diff equivalent.

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -371,6 +371,22 @@ def stat_like_fontmake(ttx):
     ver.attrib["value"] = "0x00010001"
 
 
+# https://github.com/googlefonts/fontc/issues/1107
+def normalize_glyf_contours(ttx):
+    for glyph in ttx.xpath("//glyf/TTGlyph"):
+        contours = glyph.xpath("./contour")
+        if len(contours) < 2:
+            continue
+        normalized = sorted(contours, key=lambda c: etree.tostring(c))
+        if normalized == contours:
+            continue
+        for contour in contours:
+            glyph.remove(contour)
+        for contour in normalized:
+            glyph.append(contour)
+    
+    
+
 def allow_some_off_by_ones(fontc, fontmake, container, name_attr, coord_holder):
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -755,7 +773,7 @@ def main(argv):
     root = Path(".").resolve()
     if root.name != "fontc":
         sys.exit("Expected to be at the root of fontc")
-
+
```
cmyr, line 776: noise?

resources/scripts/ttx_diff.py
```diff
@@ -371,6 +371,22 @@ def stat_like_fontmake(ttx):
     ver.attrib["value"] = "0x00010001"
 
 
+# https://github.com/googlefonts/fontc/issues/1107
+def normalize_glyf_contours(ttx):
+    for glyph in ttx.xpath("//glyf/TTGlyph"):
+        contours = glyph.xpath("./contour")
+        if len(contours) < 2:
+            continue
+        normalized = sorted(contours, key=lambda c: etree.tostring(c))
```
cmyr, line 380: do we even need to `tostring`? not sure how expensive it is, but all we need is some total order, can we get that from `c` itself?

cmyr (approved):
cool this looks painless, and i'm very curious to see how many fonts are affected.

resources/scripts/ttx_diff.py
rsheeter: It wasn't entirely clear to me what that would do whereas tostring is simple to reason about

PR #1114, opened 2024-11-14
Update requirements and instructions to update

rsheeter:
JMM


fontc_crater/README.md
```diff
@@ -24,13 +24,19 @@ To run in CI mode locally to play with the html output:
...
-- run `pip-compile resources/scripts/requirements.in`
-- commit the new `requirements.txt` file.
+To update the set of python packages in use, you need to update this file:
+
+```shell
+# Make sure you have pip-tools
+$ python -m pip install pip-tools
+
+# Rebuild the file
+$ pip-compile resources/scripts/requirements.in
```
cmyr, line 36: this shouldn't work without either redirection, `--output-file`, or deleting the existing `requirements.txt` ahead of time?

rsheeter: It does update the file but per IM it may not be the update we want so added back the rm plus a comment

fontc_crater/README.md
```diff
@@ -24,13 +24,20 @@ To run in CI mode locally to play with the html output:
...
-- commit the new `requirements.txt` file.
+To update the set of python packages in use, you need to update this file:
+
+```shell
+# Make sure you have pip-tools
+$ python -m pip install pip-tools
+
+# Rebuild the file
+# Remove it to force a complete update rather than minimal
+$ rm -rf resources/scripts/requirements.txt && pip-compile resources/scripts/requirements.in
```
cmyr, line 37: why -rf? It's a single file and we should own it, so `rm` should be fine, and avoids the risk of fat-fingering..

rsheeter:
This is to pull in https://github.com/googlefonts/glyphsLib/pull/1048

fontc_crater/README.md
rsheeter: Removed r. I think we do want f so the command works even if someone already removed the file.

rsheeter: For context, I've seen a ton of these example commands and scripts fail over the years because they were fragile and someone ran them from a slightly odd starting state.

Issue #1117, opened 2024-11-14
Rounding differences in fvar


cmyr:
In several fonts we have a very slight difference in values in the fvar table, that looks to be a rounding issue?

For instance for [GrenzeGotisch.glyphs](https://github.com/Omnibus-Type/Grenze-Gotisch) in default mode we end up with a value of `200.00002` for `wght` for the `ExtraLight` named instance, where fontmake has `200`.

The exact same problem shows up in at least one other font, [BeVietnamPro.glyphs](https://github.com/bettergui/BeVietnamPro).

PR #1120, opened 2024-11-15
piecewise_linear_map: perform internal computations as f64 to increase precision

anthrotype:
but keep the outward interface as f32 to limit the scope of changes (otherwise callers would have to deal with f64s and there are a lot)

This seems to fix https://github.com/googlefonts/fontc/issues/1117 in the sense that ttx_diff.py now reports [Grenze-Gotisch](https://github.com/Omnibus-Type/Grenze-Gotisch) as "output is identical".

Basically, the loss of precision happened while converting glyphs sources to ir: the instance locations are in design-space in the source; we store them as normalized location

https://github.com/googlefonts/fontc/blob/53151a1af8b2977d3aaca66dd6d7d13c43a80ec9/glyphs2fontir/src/toir.rs#L288

... but then in the ir::NamedInstance they are in user-space (like in fvar) so they must be converted again `to_user`:

https://github.com/googlefonts/fontc/blob/53151a1af8b2977d3aaca66dd6d7d13c43a80ec9/glyphs2fontir/src/source.rs#L397-L404

In this particular font, design- and user-space location are the same (there are no axis mappings). The instance at wght=200 (design) is converted to 0.666667 (normalized) and then again to 200 (user). If the lerp is done with f32 we lose precision and instead of 200 we get back 200.00002

I'm not entierly sure this is the best fix but it does seem to improve the situation at least.

### DIFF
--- a/fontdrasil/src/piecewise_linear_map.rs
+++ b/fontdrasil/src/piecewise_linear_map.rs
@@ -10,15 +10,28 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct PiecewiseLinearMap {
     // these two mappings have identical lengths, by construction
-    from: Vec<OrderedFloat<f32>>, // sorted, ||'s to
-    to: Vec<OrderedFloat<f32>>,   // sorted, ||'s from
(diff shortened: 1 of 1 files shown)


cmyr (approved):
This seems like a reasonable way to handle this, thanks for taking the ticket. 🏆

PR #1131, opened 2024-11-18
Remove `set_unconditionally` & `BeValue`

cmyr:
How this happened:

- I was looking into an issue with name records generated from FEA
- I got confused about why we were calling a method called `set_unconditionally`, instead of just `set`
- The docs for that method had a TODO saying it was (mostly?) only necessary because `write-fonts` types don't impl `PartialEq`
- ... except that they have, for over a year
- so I went and tried to impl that trait, except we don't impl it for them directly, we impl it for a wrapper type, `BeValue`
- ...and the docs for _that_ type said it was only necessary to get around the orphan rules
- ... which wouldn't be a problem if we just impl'd the `Persistable` trait for the write-fonts types in the same crate where the trait is defined (`fontir`)
- So I did _that_, and then ran into an issue where for `avar` we want to be able to store an explicit `Option<T>` (the only case like this) and we were repurposing the `BeValue` to also do that job
- ~so I added a `PersistableOption` type to explicitly handle this case (which I still do not fully understand)~
- so I just made this an explicit special case, with a `MaybeEmptyAvar` type.
- and now everything compiles again, and overall feels simpler.
- ... profit?

Bonus:

Why do we have this special case for avar? why do we need an explicit option? Why can't we just serialize an empty table in the case where we want to have an empty table in the final font? cc @rsheeter

### DIFF
--- a/fontbe/src/avar.rs
+++ b/fontbe/src/avar.rs
@@ -5,9 +5,10 @@ use fontdrasil::{
     orchestration::{Access, Work},
     types::Axis,
 };
-use fontir::orchestration::WorkId as FeWorkId;
+use fontir::orchestration::{Persistable, WorkId as FeWorkId};
 use log::debug;
 use write_fonts::{
+    read::FontRead,
     tables::avar::{Avar, AxisValueMap, SegmentMaps},
     types::F2Dot14,
 };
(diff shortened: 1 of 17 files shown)


rsheeter:
Oh nice, if we now have partialeq everywhere we should be able to tear down a bunch of that indeed.

Memory is failing me wrt avar weirdness.

fontbe/src/avar.rs
```diff
@@ -17,6 +18,42 @@ use crate::{
     orchestration::{AnyWorkId, BeWork, Context, WorkId},
 };
 
+/// Avar is a special case where sometimes we want to explicitly have an empty one
```
rsheeter, line 21: Note in comment why this exists, e.g. why not just `Option<Avar>`

rsheeter (approved):
Definitely looks like an improvement and makes sense if write-fonts now has partialeq we shouldn't need to work around it not having it.

As you suggest, I don't _think_ we need to special case avar so. Maybe as a follow-on we can delete that enum and let avar be like everything else?

cmyr:
I tried just deleting the option but it broke some tests that did not seem immediately related, so I settled on this.

PR #1133, opened 2024-11-19
Merge names from FEA into final table

### LINKED ISSUE #1130: GSUB stylistic set names not being added to name table
noticed in Questrial (repo with `python resources/scripts/ttx_diff.py 'https://github.com/googlefonts/questrial#sources/Questrial.glyphspackage'`).

I think this might've been brought up before, but can't find the issue.

I haven't dug into this but I suspect we're just generally not even looking at the name table that is produced by `fea-rs`. A larger problem is that fea-rs is assigning its own name values to things without knowing what is going on elsewhere, which means it's very easy for it to use nameIds that collide with those used elsewhere in the compilation process.

There are various possible solutions; one might be to have a 'PendingNameId' type that is used during compilation, and then these all get assigned their final values at the end; alternatively we could possible remap name values generated by fea-rs once we know what ids are used elsewhere.

cmyr:
Previously we were just dropping them on the floor.

This is one of a number of cases where it is possible for FEA to generate a table that is also generated (or could be generated) through the main compilation process.

This introduces an overall strategy for handling this: when FEA generates these extra tables we now stash them in the main context, where we can reference them when building the final font.

This patch also uses this mechanism to merge name records generated from FEA into the main name table; this looks to get us at least a +7 on crater.

As a bonus this includes a commit that fixes a little bug in ttx_diff where we were accidentally sorting name ids as if they were lookup ids, and which very briefly had me doubting my sanity.

- fixes #1130

### DIFF
--- a/fea-rs/src/compile/tables/name.rs
+++ b/fea-rs/src/compile/tables/name.rs
@@ -10,7 +10,8 @@ use crate::compile::tags::{MAC_PLATFORM_ID, WIN_PLATFORM_ID};
 #[derive(Clone, Debug)]
 pub(crate) struct NameBuilder {
     records: Vec<(NameId, NameSpec)>,
-    last_anon_id: u16,
+    // the last used non-reserved nameid value.
+    last_nonreserved_id: NameId,
 }
 
 #[derive(Clone, Debug, Default)]
@@ -25,14 +26,15 @@ impl Default for NameBuilder {
     fn default() -> Self {
         NameBuilder {
             records: Vec::new(),
-            last_anon_id: 255,
+            // incremented first time we call `next_name_id`
+            last_nonreserved_id: NameId::LAST_RESERVED_NAME_ID,
         }
     }
 }
 
 impl NameBuilder {
     pub(crate) fn add(&mut self, name_id: NameId, name_spec: NameSpec) {
-        self.last_anon_id = self.last_anon_id.max(name_id.to_u16());
+        self.last_nonreserved_id = self.last_nonreserved_id.max(name_id);
         self.records.push((name_id, name_spec));
     }
 
@@ -49,7 +51,9 @@ impl NameBuilder {
     }
 
     pub(crate) fn next_name_id(&self) -> NameId {
-        NameId::new(self.last_anon_id + 1)
+        self.last_nonreserved_id
+            .checked_add(1)
+            .unwrap_or(self.last_nonreserved_id)
     }
 
     pub(crate) fn build(&self) -> Option<write_fonts::tables::name::Name> {
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -37,13 +37,15 @@ use fontdrasil::{
 };
 use write_fonts::{
     tables::{layout::ClassDef, variations::VariationRegion},
-    types::Tag,
+    types::{NameId, Tag},
     OtRound,
 };
 
 use crate::{
     error::Error,
-    orchestration::{AnyWorkId, BeWork, Context, FeaAst, FeaRsKerns, FeaRsMarks, WorkId},
+    orchestration::{
+        AnyWorkId, BeWork, Context, ExtraFeaTables, FeaAst, FeaRsKerns, FeaRsMarks, WorkId,
+    },
 };
 
 mod kern;
@@ -550,6 +552,7 @@ impl Work<Context, AnyWorkId, Error> for FeatureCompilationWork {
             WorkId::Gpos.into(),
             WorkId::Gsub.into(),
             WorkId::Gdef.into(),
+            WorkId::ExtraFeaTables.into(),
         ]
     }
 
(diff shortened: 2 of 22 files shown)


fea-rs/src/compile/tables/name.rs
```diff
@@ -7,10 +7,13 @@ use write_fonts::{read::tables::name::Encoding, types::NameId};
 
 use crate::compile::tags::{MAC_PLATFORM_ID, WIN_PLATFORM_ID};
 
+pub(crate) const FIRST_NON_RESERVED_NAME_ID: u16 = 256;
```
rsheeter, line 10: Suggest comment referencing https://learn.microsoft.com/en-us/typography/opentype/spec/name#name-ids

fea-rs/src/compile/tables/name.rs
```diff
@@ -7,10 +7,13 @@ use write_fonts::{read::tables::name::Encoding, types::NameId};
...
+
 #[derive(Clone, Debug)]
 pub(crate) struct NameBuilder {
     records: Vec<(NameId, NameSpec)>,
-    last_anon_id: u16,
+    // the last used non-reserved nameid value.
+    last_nonreserved_id: u16,
```
rsheeter, line 16: Optional, making this be _next_ rather than last even when there was no last would make more sense to me?

fontbe/src/features.rs
```diff
@@ -56,6 +58,8 @@ pub use marks::create_mark_work;
 
 const DFLT_SCRIPT: Tag = Tag::new(b"DFLT");
 const DFLT_LANG: Tag = Tag::new(b"dflt");
+// https://learn.microsoft.com/en-us/typography/opentype/spec/name#name-ids
+const FIRST_NON_RESERVED_NAME_ID: u16 = 256;
```
rsheeter, line 62: I think I've seen this twice, I suggest putting it in fontdrasil which both fea-rs and fontbe are entitled to reference

fontbe/src/name.rs
```diff
@@ -25,14 +27,18 @@ impl Work<Context, AnyWorkId, Error> for NameWork {
     }
 
     fn read_access(&self) -> Access<AnyWorkId> {
-        Access::Variant(AnyWorkId::Fe(FeWorkId::StaticMetadata))
+        AccessBuilder::new()
+            .variant(AnyWorkId::Fe(FeWorkId::StaticMetadata))
+            .variant(AnyWorkId::Be(WorkId::ExtraFeaTables))
```
rsheeter, line 32: This is an unfortunate dependency; fea is often long pole so blocking things on fea is undesirable. I suppose to avoid you'd have to fix name in final font assembly which seems even worse.

fontbe/src/name.rs
```diff
@@ -42,11 +48,25 @@ impl Work<Context, AnyWorkId, Error> for NameWork {
...
 
-        context
-            .name
-            .set(Name::new(name_records.into_iter().collect()));
+        context.name.set(Name::new(name_records));
         Ok(())
     }
 }
+
+fn merge_name_records(records: &mut BTreeSet<NameRecord>, fea_names: &Name) {
```
rsheeter, line 66: I think we're just overwriting not meaningfully merging so perhaps upsert_name_records would be more descriptive?

fontbe/src/name.rs
```diff
@@ -42,11 +48,25 @@ impl Work<Context, AnyWorkId, Error> for NameWork {
...
+    // what should we be doing, here?
+    // for now I am going to use the same logic as we have for handling manual
+    // GlyphClassDef statements in FEA, and assume that the FEA is correct?
+    // My confidence in this approach is medium.
```
rsheeter, line 70: is that what fontmake does? - if so we should be more confident and ideally cite the fontmake code

fontbe/src/name.rs
```diff
@@ -42,11 +48,25 @@ impl Work<Context, AnyWorkId, Error> for NameWork {
...
             })
-            .collect::<Vec<_>>();
+            .collect();
+
+        if let Some(name) = extra_tables
+            .as_ref()
+            .and_then(|tables| tables.name.as_ref())
+        {
+            log::info!("merging name records from FEA");
+            merge_name_records(&mut name_records, name);
```
rsheeter, line 58: We don't really seem to need a function, just call extend here?

resources/testdata/WghtVar-Bold.ufo/features.fea
rsheeter, line 1: How about a new minimal test font? - a nop .glyphs file with a name seems like it would suffice and would be quite terse. If we keep adding things to our existing test fonts then they test every more things at once which is harder to reason about.

rsheeter (approved):

resources/scripts/ttx_diff.py
rsheeter, line 1: Nit: this would have been nicer as it's own PR

resources/testdata/WghtVar-Bold.ufo/features.fea
```diff
@@ -0,0 +1,10 @@
+languagesystem DFLT dflt;
+
+feature ss01 {
+        featureNames {
+        name "my fun feature";
+    };
+
+    sub bar by plus;
+} ss01;
+
```
khaledhosny, line 10: Does fea_rs support [`table name {} name`](https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#9.e) blocks? If so, I think it deserves a test, as they would override existing names in the font.

fea-rs/src/compile/tables/name.rs
cmyr: This is just a rename; changing the semantics causes breakages elsewhere (I had tried this)

fontbe/src/name.rs
cmyr: this is as separate function because I am unclear of what the logic should actually be, and the name was chosen to reflect the semantics, not the specifics of the current implementation.

resources/testdata/WghtVar-Bold.ufo/features.fea
khaledhosny: Names defined in the feature file override those already in the name table. This [what feaLib does](https://github.com/fonttools/fonttools/blob/b90ac3c29f6030ec7309b044c41a72d9f586163f/Lib/fontTools/feaLib/builder.py#L452), and I believe this is what makeotf does as well (though it does not allow overriding name IDs 2 and 6, or at least that [what the spec says](https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#:~:text=are%20reserved%20by%20the%20implementation%20and%20cannot%20be%20overridden), I haven’t tested it recently).

resources/scripts/ttx_diff.py
cmyr: As a compromise it's in its own commit 😬

Issue #1113, opened 2024-11-14
Support meta table


anthrotype:
https://learn.microsoft.com/en-us/typography/opentype/spec/meta

Glyphs.app has a custom parameter named "meta Table" (https://handbook.glyphsapp.com/custom-parameter-descriptions/)

Ufo has `public.openTypeMeta` in the lib.plist (https://unifiedfontobject.org/versions/ufo3/lib.plist/#publicopentypemeta)

This would require supports in write-fonts first: https://github.com/googlefonts/fontations/pull/1237

searching https://googlefonts.github.io/fontc_crater/ reveals quite a few fonts define this table

PR #1136, opened 2024-11-20
Add initial implementation of `meta` compilation

Hoolean:
Hello!

This PR begins implementation of `meta` table compilation for the Glyphs and UFO formats. This is a larger end-to-end implementation than my previous PRs, and so I am making an early draft to work on it publicly to make sure that it can be shaped from the beginning, and that it stays on track; all feedback and contributions welcome :)

## Implemented

* Full happy path
* Produces an identical `meta` table for the top few fontc_crater reports

## TODO

* [ ] Confirm semantics of Glyphs custom parameter (Glyphs.app vs glyphsLib)
* [ ] Confirm approach for `StateSet` construction in `glyphs2fontir`
* [ ] Make error handling idiomatic to this project (no `expect(...)`, no type conversion clutter, etc)
* [x] Rebase on `main`
* [x] Wait for fontations release that includes https://github.com/googlefonts/fontations/pull/1244
* [x] Fix lints
* [x] Fix failing tests
* [ ] Add tests (!)

(most of these are also flagged as inline TODOs, to be resolved before merge)

---

Will implement #1113


Hoolean:
(rebased on `main` to adapt for #1131 and to use full fontations release instead of bleeding edge git reference)

fontbe/src/meta.rs
```diff
@@ -0,0 +1,56 @@
...
+            } else {
+                // Instantiate DataMapRecords from tuples.
+                Some(Meta::new(
+                    records
+                        .iter()
+                        .map(|(tag, metadata)| DataMapRecord::new(*tag, metadata.clone()))
+                        .collect(),
+                ))
+            }
+        };
```
rsheeter, line 48: This looks like it should be possible to unindent, e.g.

```rust
let static_metadata = context.ir.static_metadata.get();
let records = &static_metadata.misc.meta;

// Guard clause: nop if there are no meta records
if records.is_empty() {
  return Ok(());
}

context.meta.set(Meta::new(...));
```

glyphs2fontir/src/source.rs
```diff
@@ -130,6 +134,8 @@ impl GlyphsIrSource {
             unicode_range_bits: font.unicode_range_bits.clone(),
             codepage_range_bits: font.codepage_range_bits.clone(),
             panose: font.panose.clone(),
+            // TODO: Is this correct?
```
rsheeter, line 137: looks right at a glance

glyphs2fontir/src/source.rs
```diff
@@ -491,6 +499,36 @@ impl Work<Context, WorkId, Error> for StaticMetadataWork {
             })
             .or(static_metadata.misc.created);
 
+        // TODO: Use idiomatic error handling.
+        // TODO: This diverges from glyphsLib, which must convert a list of
+        //       records into UFO's intermediate dict representation; is this
+        //       actually an improvement, or a breaking change? What does Glyphs
+        //       do?
```
rsheeter, line 506: this seems like an @anthrotype question

rsheeter (commented):
Looks promising!

cmyr:
I'm sorry, I somehow missed this PR and went and impl'd this for glyphs sources in #1160

PR #1137, opened 2024-11-21
[glyphs] Remove 'Raw' from non-raw type names

cmyr:
_I'm trying to clean this module up a bit; this is a preliminary PR to just doublecheck my understanding_

As I understand it 'Raw' means something parsed right out of the plist data, and these types do not seem to conform to this, and they are only used in types that are themselves not raw.


anthrotype (approved):
I think you're right, they look cooked to me

PR #1141, opened 2024-11-21
[glyphs] Clean up custom params

cmyr:
We're going to need to support a bunch of of these, and it was becoming hard to keep track of what was being handled and what wasn't. This moves all the custom params (at the font level) into a new CustomFontParameters struct.

It also reduces boilerplate when these are used in glyphs2fontir.

The ultimate goal here is to do smarter parsing of these custom paramaters in such a way as to be able to warn if there are any we aren't handling.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -41,12 +41,9 @@ pub struct AxisUserToDesignMap(Vec<(OrderedFloat<f32>, OrderedFloat<f32>)>);
 /// A tidied up font from a plist.
 ///
 /// Normalized representation of Glyphs 2/3 content
-#[derive(Debug, PartialEq, Hash)]
+#[derive(Debug, Default, PartialEq, Hash)]
 pub struct Font {
     pub units_per_em: u16,
-    pub fs_type: Option<u16>,
-    pub use_typo_metrics: Option<bool>,
-    pub has_wws_names: Option<bool>,
     pub axes: Vec<Axis>,
     pub masters: Vec<FontMaster>,
     pub default_master_idx: usize,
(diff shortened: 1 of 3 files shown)


glyphs-reader/src/font.rs
```diff
@@ -65,6 +62,15 @@ pub struct Font {
     // master id => { (name or class, name or class) => adjustment }
     pub kerning_ltr: Kerning,
 
+    pub custom_parameters: CustomFontParameters,
```
anthrotype, line 65: how about naming them "FontCustomParameters" instead? To my non-native Engish speaking mind, it reads slightly better as it makes it easier to compare/contrast to the other masters' and instances' custom parameters.
By the way, don't we need to also group the latter two?

glyphs2fontir/src/source.rs
```diff
@@ -598,166 +550,64 @@ impl Work<Context, WorkId, Error> for GlobalMetricWork {
...
-            metrics.set_if_some(
-                GlobalMetric::HheaLineGap,
-                pos.clone(),
-                master
-                    .hhea_line_gap
-                    .or(font.hhea_line_gap)
-                    .map(|v| v as f64),
-            );
+
+            macro_rules! set_metric {
```
anthrotype, line 564: nice!

anthrotype (approved):
+1 much better when they are logically grouped under font.custom_parameters

glyphs-reader/src/font.rs
cmyr: name sounds good.

And yes, the goal is also to group the other ones, I can take a look at that too.

PR #1143, opened 2024-11-26
Glyphs custom params parsing rework

cmyr:
This is a rollup of three commits that rework how we parse custom parameter values from glyphs sources.

- This now parses all custom param values to the `Plist` type, which represents any plist data
- This means we do not need to know the expected type for a given key in order to parse it; we will parse anything that is in the source.
- *exhaustive handling*: Instead of querying into the custom parameters with the keys that we know, we now iterate through all the parameters in sequence, which forces us to handle them all (or report where we don't).
- This reuses the same parsing logic and `CustomParameters` type for the top-level `Font` as well as each individual master, which lets us delete a bunch of code

This PR doesn't change any observable behaviour, but there is a followup that will add handling for some additional params.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -62,12 +62,12 @@ pub struct Font {
     // master id => { (name or class, name or class) => adjustment }
     pub kerning_ltr: Kerning,
 
-    pub custom_parameters: FontCustomParameters,
+    pub custom_parameters: CustomParameters,
 }
 
 /// Custom parameter options that can be set on a glyphs font
 #[derive(Clone, Debug, PartialEq, Hash, Default)]
-pub struct FontCustomParameters {
+pub struct CustomParameters {
     pub use_typo_metrics: Option<bool>,
     pub fs_type: Option<u16>,
     pub has_wws_names: Option<bool>,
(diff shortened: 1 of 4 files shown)


glyphs-reader/src/font.rs
```diff
@@ -347,359 +352,275 @@ struct NumberName {
...
+            .map(|b| codepage_range_bit(*b as _))
+            .collect::<Result<_, _>>()
+            .ok()
     }
+}
 
-    fn panose(&self) -> Option<&Vec<i64>> {
+impl RawCustomParameters {
+    ////convert into the parsed params for a top-level font
+    fn cook(&self) -> Result<CustomParameters, Error> {
```
rsheeter, line 505: If I see `x.cook()` in code I wouldn't know what it did, can we get a more descriptive name? - for example, if I saw `.create_custom_parameters()?` that would be far clearer to me.

glyphs-reader/src/font.rs
```diff
@@ -347,359 +352,275 @@ struct NumberName {
...
+        let mut panose = None;
+        let mut panose_old = None;
+
+        for RawCustomParameterValue {
+            name,
+            value,
+            disabled,
+        } in &self.0
+        {
+            macro_rules! add_and_report_issues {
```
rsheeter, line 521: Can we do this without a macro? If not, can we add a comment explaining why not?

rsheeter (approved):
`+282 −404` - less lines ftw!

glyphs-reader/src/font.rs
cmyr: We need to use a macro because one of the inputs is the name of the field we're setting; can't pass that to a function. (Added a comment to this effect)

PR #1144, opened 2024-11-26
[glyphs] Handle additional custom params

### LINKED ISSUE #1128: handle `openTypeLowestRecPPEM` value in glyphs sources
Motivated in particular by `openTypeLowestRecPPEM`: I'm not sure exactly how this works, but it looks like something that exists in glyphsLib and that we aren't looking for in glyphs2fontir, and it is causing a diff in `NotoFangsongKSSRotated.glyphspackage`.

edit: it looks like there are a number of these values that we aren't handling, such as `openTypeHheaCaretSlopeRise` and `openTypeHheaCaretSlopeRun`. we should Probably just be making sure we're handling all the values listed there?

cmyr:
_This is based on #1143_

This handles additional custom parameters, specifically `openTypeHeadLowestRecPPEM`, and the various hhea & vhea caret properties (vhea isn't handled yet, but we can at least parse it now in preparation).

 These params are kind of funny because you don't seem to be able to add them via the glyphs.app UI? But they exist in some fonts, and were causing crater failures.

- closes #1128

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -95,6 +95,13 @@ pub struct CustomParameters {
     pub codepage_range_bits: Option<BTreeSet<u32>>,
     pub panose: Option<Vec<i64>>,
 
+    pub lowest_rec_ppem: Option<i64>,
+    pub hhea_caret_slope_run: Option<i64>,
+    pub hhea_caret_slope_rise: Option<i64>,
+    pub hhea_caret_offset: Option<i64>,
+    pub vhea_caret_slope_run: Option<i64>,
+    pub vhea_caret_slope_rise: Option<i64>,
+    pub vhea_caret_offset: Option<i64>,
     // these fields are parsed via the config, but are stored
     // in the top-level `Font` struct
     pub virtual_masters: Option<Vec<BTreeMap<String, OrderedFloat<f64>>>>,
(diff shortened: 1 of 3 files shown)


khaledhosny:
> These params are kind of funny because you don't seem to be able to add them via the glyphs.app UI?

You can add anything as Glyphs custom parameter. If Glyphs recognizes the name, it will give it the appropriate UI (e.g. a check box for boolean parameters), but otherwise you get a default text entry box that can have any value. Some of the parameters that Glyphs recognizes are hidden; if you search for them they will not show up, but if you go ahead and add it anyway it will get the appropriate UI.

They `openType` parameters, are, I think, present to parallel the corresponding UFO fontinfo keys, and I think Glyphs will use them when importing UFO files, rather than the “native” equivalent custom parameters.

resources/testdata/glyphs3/UnusualCustomParams.glyphs
rsheeter, line 1: I think this can be further minimized?

glyphs-reader/src/font.rs
```diff
@@ -520,7 +527,15 @@ impl RawCustomParameters {
         {
             macro_rules! add_and_report_issues {
```
rsheeter, line 528: Not a huge fan of the macro, lets at least add a comment explaining what precludes doing this without a macro

glyphs-reader/src/font.rs
```diff
@@ -570,14 +585,48 @@ impl RawCustomParameters {
                 "codePageRanges" => {
                     add_and_report_issues!(codepage_range_bits, Plist::as_codepage_bits)
                 }
+                // these values are not listed in the glyphs.app UI, but exist
+                // in some fonts. Maybe only fonts converted to glyphs format
+                // from something else?
```
rsheeter, line 590: update based on Khaled's comment

rsheeter (approved):

PR #1145, opened 2024-11-26
[glyphs] Parse kerning as floats

cmyr:
Glyphs kern values can be floats, but we were parsing them as integers.

This also uncovered something else fishy: our integer parsing logic in glyps-reader/ascii-plist-derive will happy parse something like "1.321" as an integer, by parsing to a float and casting.

I think this is incorrect? I think it is reasonable to go the other way (to accept an integer value as a float) and it _might_ be reasonable to accept an integer value with no fractional component as a float, but I definitely don't think we should parse a float with a fractional value as an integer; it hides bugs like this.

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -804,7 +804,7 @@ impl Work<Context, WorkId, Error> for KerningInstanceWork {
                 Some(((side1, side2), pos_adjust))
             })
             .for_each(|(participants, (_, value))| {
-                *kerning.kerns.entry(participants).or_default() = (value as f32).into();
+                *kerning.kerns.entry(participants).or_default() = (value.0 as f32).into();
             });
 
         context.kerning_at.set(kerning);
--- a/resources/testdata/glyphs3/KernFloats.glyphs
+++ b/resources/testdata/glyphs3/KernFloats.glyphs
@@ -1,0 +1,28 @@
+{
+.formatVersion = 3;
+familyName = KernFloats;
+fontMaster = (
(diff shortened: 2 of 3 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -804,7 +804,7 @@ impl Work<Context, WorkId, Error> for KerningInstanceWork {
                 Some(((side1, side2), pos_adjust))
             })
             .for_each(|(participants, (_, value))| {
-                *kerning.kerns.entry(participants).or_default() = (value as f32).into();
+                *kerning.kerns.entry(participants).or_default() = (value.0 as f32).into();
```
rsheeter, line 807: In the spirit of f64 all the things I wonder if we shoudl file an issue to get rid of the `as f32`

resources/testdata/glyphs3/KernFloats.glyphs
```diff
@@ -0,0 +1,29 @@
+{
+.formatVersion = 3;
+familyName = KernFloats;
+date = "2023-05-05 15:11:55 +0000";
```
rsheeter, line 4: optional, I think you can delete date

rsheeter (approved):
Good catch, lgtm

glyphs2fontir/src/source.rs
cmyr: I started out assuming that was the PR I was writing until I got in here and noticed there was a clearer root cause. I do  think getting rid of f32 makes sense, though.

PR #1148, opened 2024-11-28
[clippy] Roll the rock back up the hill

cmyr:


cmyr (commented):
self +1'ing this because it's trivial and everyone is eating turkey or football players or whatever

Issue #1062, opened 2024-10-22
Detect component cycles


rsheeter:
Double-check we prevent component cycles and add a test if we don't have one already.

Context https://github.com/googlefonts/fontc/pull/1050#pullrequestreview-2384815572.

PR #1150, opened 2024-11-28
Deterministic resolving of mixed component/contour glyphs

### LINKED ISSUE #1146: non-deterministic interpolation error
there's sime jitter on crater runs lately involving one particular font, which fails to compile some portion of the time.

repro with,

`python resources/scripts/ttx_diff.py 'https://github.com/ThomasJockin/readexpro#sources/ReadexPro.glyphs'`.

(it took me four tries to get a failure, but maybe you're luckier than I am?)

cmyr:
If a glyph has mixed components & contours, we need to convert it to use all of one or the other.

If such a glyph has a component that _also_ has mixed components & contours, we need to make sure we handle the child component first.

We handles this by first collecting the set of such glyphs, and then determining an appropriate ordering.

I tried to keep this as simple as possible; in particular I'm defensive against cycles but am not reporting those errors here, since I assume we catch them elsewhere.

- fixes #1146

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -3,7 +3,10 @@
 //! Notably includes splitting glyphs with contours and components into one new glyph with
 //! the contours and one updated glyph with no contours that references the new gyph as a component.
 
-use std::collections::{HashMap, HashSet, VecDeque};
+use std::{
+    collections::{HashMap, HashSet, VecDeque},
+    sync::Arc,
+};
 
 use fontdrasil::{
     coords::NormalizedLocation,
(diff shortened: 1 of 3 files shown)


rsheeter:
> I'm defensive against cycles but am not reporting those errors here

Making sure we handle cycles is tracked as #1062

rsheeter:
IIUC the root problem is:

> When converting components to contours we need to make sure to perform this operation in order, i.e. if glyph A has component B has component C we need to make sure we convert B before A.

This PR says:

> We handles this by first collecting the set of such glyphs, and then determining an appropriate ordering.

Maybe I'm missing something, isn't this resolved by traversing components depth-first? - would make watching for cycles easier too.

cmyr:
maybe just unclear PR text? I'm using the same approach to determining the ordering as in the anchor propogation (it is depth first).

The reason this isn't useful as general-purpose cycle detection is that we're only operating on the subset of glyphs which have mixed contours/components.

fontir/src/glyph.rs
```diff
@@ -168,6 +171,101 @@ fn components(
...
+    order: &mut GlyphOrder,
+) -> Result<(), BadGlyph> {
+    let items = todo
+        .into_iter()
+        .map(|(op, glyph)| (glyph.name.clone(), (op, glyph)))
+        .collect::<HashMap<_, _>>();
+
+    for name in component_ordering_within_glyphs(items.values().map(|(_, g)| g.clone())) {
+        let (op, glyph) = items.get(&name).unwrap();
+        match op {
```
rsheeter, line 199: Nit: I feel like `op.execute(context, glyph)` might be nicer than matching on op to call a fn?

rsheeter (approved):

PR #1153, opened 2024-11-29
We want more (f64)

### LINKED ISSUE #1152: We should use f64 everywhere
we're seeing various little things in diffs that sure look like they could be floating point error accumulation, and instead of dealing with these piecemeal we should just make sure we're f64 everywhere so we can rule that out for any remaining issues.

cmyr:
This replaces a bunch of use of f32 (in parsing and in coordinate types) with f64.

There are two tests that are failing which I don't want to touch, because I don't really understand what's best.

This doesn't have a dramatic impact on crater, but it does show improvements for fvar & HVAR for a number of fonts.

closes #1152 

(The main place where we still use f32 is when we are close to `F2Dot14` types, because f32 fully captures the precision these can represent)

### DIFF
--- a/fontir/src/variations.rs
+++ b/fontir/src/variations.rs
@@ -1201,17 +1201,17 @@ mod tests {
                 ],
                 vec![
                     (0_usize, OrderedFloat(1.0)),
-                    (3_usize, OrderedFloat(0.7555555)),
-                    (4_usize, OrderedFloat(0.24444449)),
+                    (3_usize, OrderedFloat(0.7555555555555555)),
+                    (4_usize, OrderedFloat(0.24444444444444444)),
                     (5_usize, OrderedFloat(1.0)),
                     (6_usize, OrderedFloat(0.66))
                 ],
                 vec![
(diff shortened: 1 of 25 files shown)


rsheeter (approved):

fontir/src/variations.rs
```diff
@@ -50,12 +50,12 @@ impl RoundTiesEven for f64 {
     }
 }
 
-impl RoundTiesEven for f32 {
-    #[inline]
-    fn round_ties_even(self) -> f32 {
-        self.round_ties_even()
-    }
-}
+//impl RoundTiesEven for f64 {
```
dfrg, line 53: Was this always unnecessary? Can it just be removed?

dfrg (approved):
This change makes sense and the code lgtm. Just one question about a commented impl.

fontir/src/variations.rs
cmyr: oops good catch! this used to say `f32` before I /s/f32/f64, and then became redundant.

PR #1154, opened 2024-11-30
Add links to compilers to crater output

rsheeter:
When showing crater to someone new to fonts they wondered what those things were, make it easier to find out

JMM


cmyr (approved):
seems reasonable 💁‍♂️

PR #1155, opened 2024-11-30
Show only latest crater results plus a link to show more

rsheeter:
When I load crater for my own use or to show others I typically want to see the latest results and the failures. As the list of results has grown this means load then scroll down almost every time. To mitigate this hide most of the results by default and provide an option to see all.

JMM if happy


cmyr (approved):
natural that as time goes on we'll run into these sorts of growing pains, this looks good.

PR #1156, opened 2024-12-01
Add a count of targets impacted by tag

rsheeter:
I find myself counting targets impacted using find, make crater do it for me.

JMM if happy

### DIFF
--- a/fontc_crater/src/ci/html.rs
+++ b/fontc_crater/src/ci/html.rs
@@ -106,13 +106,25 @@ fn make_html(
             body {
                 h1 { "fontc_crater" }
                 div #explain {
-                    "compiling a large number of target fonts with both "
+                    "Compiling all known Google Fonts that have sources with both "
                     {a href = "https://github.com/googlefonts/fontc" { "fontc" } }
                     " and "
                     {a href = "https://github.com/googlefonts/fontmake" { "fontmake" } }
                     ", comparing the results."
                 }
                 (table)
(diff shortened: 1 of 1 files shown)


fontc_crater/src/ci/html.rs
```diff
@@ -106,12 +106,24 @@ fn make_html(
...
                 h1 { "fontc_crater" }
                 div #explain {
-                    "compiling a large number of target fonts with both "
+                    "Compiling all known Google Fonts that have sources with both "
                     {a href = "https://github.com/googlefonts/fontc" { "fontc" } }
                     " and "
                     {a href = "https://github.com/googlefonts/fontmake" { "fontmake" } }
                     ", comparing the results."
                 }
+                div #explain {
```
cmyr, line 115: I would probably put this underneath the main table? It is closely related to the specific current run. Medium term, I think it would make sense to have an additional 'summary pane that just shows additional stats related to the latest run (and could include the # of matching families, for instance)

cmyr (approved):
As a meta comment: I think it's helpful for PRs that are changing the html output to just include a screenshot, since how it looks is an important part of the change.

One suggestion, and longer term I think we have some UI work to do here but happy to merge this now as an improvement over the current situation.

fontc_crater/src/ci/html.rs
rsheeter: Done

PR #1159, opened 2024-12-02
Exploratory fix for non-deterministic processing of mixed contours and paths

rsheeter:
I'm trying to wrap my head around #1146

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -181,12 +181,13 @@ enum GlyphOp {
 }
 
 /// Fix glyphs with mixed components/contours.
-/// 
+///
 /// We presume component cycles are checked elsewhere and do not check for them here
 fn resolve_inconsistencies(
     context: &Context,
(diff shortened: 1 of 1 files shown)


fontir/src/glyph.rs
```diff
@@ -181,89 +181,46 @@ enum GlyphOp {
...
         }
-    }
-    Ok(())
-}
-
-// given a sequence of glyphs, return an ordering such that any glyph that
-// is a component of another glyph in the input set occurs before it.
-//
-// this has a funny signature so I can unit test it.
-fn component_ordering_within_glyphs(
```
cmyr, line 213: fwiw the only real reason that this existed as a separate fn was so that I could write a unit test against it. 🤷

cmyr (approved):
works for me, although this is normally the kind of thing it's worth having a regression test for.

fontir/src/glyph.rs
rsheeter: Added back one of the tests

PR #1160, opened 2024-12-03
Compile 'meta' table from glyphs sources

cmyr:
This handles the 'meta Table' custom param, and generates a meta table in this case. It doesn't handle this for UFO sources.

- progress on #1113


resources/testdata/glyphs3/MetaTable.glyphs
```diff
@@ -0,0 +1,46 @@
...
+width = 600;
+}
+);
+unicode = 32;
+}
+);
+unitsPerEm = 1000;
+versionMajor = 1;
+versionMinor = 1;
+}
```
rsheeter, line 46: this can be trimmed further, for example I don't believe you need appVersion, DisplayStrings, date, lastChange of glyph, possibly width of glyph, and version major/minor

glyphs2fontir/src/source.rs
```diff
@@ -446,6 +446,13 @@ impl Work<Context, WorkId, Error> for StaticMetadataWork {
             })
             .or(static_metadata.misc.created);
 
+        if let Some(meta_table) = &font.custom_parameters.meta_table {
+            static_metadata.meta_table = Some(MetaTableValues {
+                dlng: meta_table.dlng.clone(),
+                slng: meta_table.slng.clone(),
+            });
+        }
```
rsheeter, line 454: I wonder if meta should go in misc? - fairly obscure and rarely used afaik

fontc/src/lib.rs
```diff
@@ -3400,4 +3410,26 @@ mod tests {
             ]
         );
     }
+
+    #[test]
```
rsheeter, line 3414: I suggest adding a test that just says some other source doesn't produce a meta table. And maybe another for empty meta where the glyphs file has "meta Table"  but nothing in it.

rsheeter (approved):

PR #1165, opened 2024-12-03
Prefer default master panose

rsheeter:
Makes `python resources/scripts/ttx_diff.py 'https://github.com/notofonts/bamum#sources/NotoSansBamum.glyphs'` identical which hopefully tips us from 199 to 200+ identical on https://googlefonts.github.io/fontc_crater/

JMM


cmyr (approved):
🚢

PR #1179, opened 2024-12-05
[glyphs] Ignore *empty* legacy fea name records

cmyr:


rsheeter:
It doesn't say JMM but it feels JMM

PR #1182, opened 2024-12-06
Fix the pitch!

rsheeter:
```shell
$ python resources/scripts/ttx_diff.py 'https://github.com/JetBrains/JetBrainsMono#sources/JetBrainsMono-Italic.glyphs' --compare gftools --config ~/.fontc_crater_cache/JetBrains/JetBrainsMono/sources/config.yaml

# main
-    <isFixedPitch value="0"/>
+    <isFixedPitch value="1"/>

# branch
output is identical
```


glyphs2fontir/src/source.rs
```diff
@@ -1981,4 +1983,20 @@ mod tests {
             static_metadata.misc.panose
         );
     }
+
+    fn assert_fixed_pitch(glyphs_file: PathBuf, expected: Option<bool>) {
+        let (_, context) = build_static_metadata(glyphs_file);
+        let static_metadata = context.static_metadata.get();
+        assert_eq!(expected, static_metadata.misc.is_fixed_pitch);
+    }
```
cmyr, line 1991: totally optional, but: it would be exactly as concise to make this be something like `fn get_fixed_pitch(glyphs_file: PathBuf) -> Option<bool)`, and then have the `assert_eq` in the specific test cases; the advantage to this is that it means the backtrace will reference a line in the actual test case, instead of in this helper fn. (The test output will still name the failing test so it doesn't make a huge difference, but I think is still some marginal bit more useful)

cmyr (commented):
looks good, one optional thing inline.

glyphs2fontir/src/source.rs
rsheeter: I like it slightly better, making it so!

rsheeter:
> looks good, one optional thing inline.

Hm, I took this to mean approved and merged ... and have now noticed it's not actually marked approved. Hopefully that was just an oversight, if not happy to fix any further comments in a follow-on.

PR #1187, opened 2024-12-11
[fontir] Add ir::path_builder module

cmyr:
And move all the GlyphPathBuilder stuff there.

No functional change. I'd like to hang the 'erase open corners' code off of this, and the ir module is already huge.


dfrg (approved):
Lovely bit of gardening

PR #1190, opened 2025-01-03
Update README.md hyperfine example

rsheeter:
Bake the entire run and compare branches process into the command

JMM if happy.


cmyr (approved):
👍

PR #1193, opened 2025-01-04
Obey source OS/2 family class

rsheeter:
Helps with `python resources/scripts/ttx_diff.py 'https://github.com/wix/wixmadefor#sources/WixMadeforText-SemiBoldItalic.ufo'`

JMM if happy


cmyr (approved):
looks good, although merging is blocked.

PR #1194, opened 2025-01-05
Drop quotes from names parsed from fea earlier so we detect empty ones correctly

rsheeter:
Parsing NameSpec from fea retains the double quotes, dropping them only when generating final `name` records. That causes fontc to produce unwanted empty name records because it filters using is_empty. Naive solution: drop the quotes earlier so is_empty works as expected.

Makes `name` identical for `python resources/scripts/ttx_diff.py 'https://github.com/marcologous/hanken-grotesk#sources/HankenGrotesk.glyphs'`.

### DIFF
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -1819,7 +1819,28 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
     }
 
     fn resolve_name_spec(&mut self, node: &typed::NameSpec) -> super::tables::NameSpec {
-        resolve_name_spec(node)
+        const WIN_DEFAULT_IDS: (u16, u16) = (1, 0x0409);
+        const MAC_DEFAULT_IDS: (u16, u16) = (0, 0);
+
+        let platform_id = node
+            .platform_id()
+            .map(|n| n.parse().unwrap())
+            .unwrap_or(tags::WIN_PLATFORM_ID);
+
+        let (encoding_id, language_id) = match node.platform_and_language_ids() {
+            Some((platform, language)) => (platform.parse().unwrap(), language.parse().unwrap()),
+            None => match platform_id {
+                tags::MAC_PLATFORM_ID => MAC_DEFAULT_IDS,
+                tags::WIN_PLATFORM_ID => WIN_DEFAULT_IDS,
+                _ => panic!("missed validation"),
+            },
+        };
+        super::tables::NameSpec {
+            platform_id,
+            encoding_id,
+            language_id,
+            string: node.string().into(),
+        }
     }
 
     fn resolve_lookup_ref(&mut self, lookup: typed::LookupRef) {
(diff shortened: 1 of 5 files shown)


fea-rs/src/compile/compile_ctx.rs
```diff
@@ -2140,6 +2119,43 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
...
+        None => match platform_id {
+            tags::MAC_PLATFORM_ID => MAC_DEFAULT_IDS,
+            tags::WIN_PLATFORM_ID => WIN_DEFAULT_IDS,
+            _ => panic!("missed validation"),
+        },
+    };
+
+    // Drop wrapping quotes around names if present, it confuses subsequent processing of string
+    let string = match &node.string().text {
+        quoted if quoted.starts_with('"') && quoted.ends_with('"') && quoted.len() > 1 => quoted
```
cmyr, line 2143: parsing guarantees that the string has lead/trailing double quotes, so this check isn't necessary.

fea-rs/src/compile/compile_ctx.rs
```diff
@@ -2140,6 +2119,43 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
...
+            .strip_prefix('"')
+            .unwrap()
+            .strip_suffix('"')
+            .unwrap()
+            .into(),
```
cmyr, line 2148: `strip_prefix` is a very useful method in some cases, (specifically when you want to branch only if a prefix was present) but in this case you can just use `trim_matches('"')`.

fea-rs/src/compile/compile_ctx.rs
```diff
@@ -2222,4 +2238,21 @@ mod tests {
...
+    }
+
+    #[test]
+    fn parse_name_spec_drop_quotes() {
+        assert_eq!(
+            vec!["", "duck"],
+            vec![
+                parse_name_spec("3 1 0x409 \"\"").string,
+                parse_name_spec("3 1 0x409 \"duck\"").string
+            ]
```
cmyr, line 2255: trivia, but fwiw you don't need the `vec` here you can just have these be arrays..

fea-rs/src/token_tree/typed.rs
```diff
@@ -89,7 +89,7 @@ macro_rules! ast_node {
         #[derive(Clone, Debug)]
         #[allow(missing_docs)]
         pub struct $typ {
-            inner: Node,
+            pub(crate) inner: Node,
```
cmyr, line 92: we generate a `node()` method you can call, so this shouldn't be necessary.

cmyr (commented):
so I'm still slightly confused about where the `is_empty` check is, but if I'm understanding the intent of this patch correctly, I think that the cleanest fix is to replace the `string()` method on the `NameSpec` type in `token_tree::typed` with something like,


```rust
    pub(crate) fn string(&self) -> &str {
        self.find_token(Kind::String)
            .unwrap()
            .text
            .trim_matches('"')
    }
```

(although honestly just doing it inline in `resolve_name_spec` seems fine too.)

...and then since the actual bug is with something ending up in the binary, the best way to test this would be to add a new case in `fea-rs/test-data/compile-tests`; there's a readme there that explains this process but it's easy.

**BUT** if part of the complication here is that there is some filtering happening outside of fea-rs then that won't work, and in that case we should figure out if fea-rs should be doing this filtering or not.

I've left a few little comments on the PR as written, for informational purposes.


(i'm not marking this approved yet because I want to understand where the filtering you mention is happening first)

fea-rs/src/token_tree/typed.rs
rsheeter: This was to enable construction of a test value, not reading, but perhaps that will be unnecessary once revised

fea-rs/src/compile/compile_ctx.rs
rsheeter: I want to remove only the starting/ending " which is not quite the same as trim_matches

cmyr (approved):
nice okay ya I do think this is much cleaner. Thanks!

fea-rs/src/compile/compile_ctx.rs
rsheeter: Good catch, ty

PR #1195, opened 2025-01-08
Prefer variable default for usXClass when variable

rsheeter:
#1191 had a bug, add a test to expose and fix.

Makes `python resources/scripts/ttx_diff.py 'https://github.com/TypeNetwork/Josefinslab#sources/JosefinSlab.designspace'` identical.

JMM if happy.


cmyr (approved):
🫸 🪨 ⛰️

PR #1196, opened 2025-01-08
Compile gasp

rsheeter:
Makes `python resources/scripts/ttx_diff.py 'https://github.com/wix/wixmadefor#sources/WixMadeforText-Italic.designspace'`, `python resources/scripts/ttx_diff.py 'https://github.com/wix/wixmadefor#sources/WixMadeforText-MediumItalic.ufo'`, and hopefully the rest of the Wix families that differ only in `gasp` identical.

JMM if happy.


fontir/src/ir.rs
```diff
@@ -149,6 +150,19 @@ pub struct MiscMetadata {
...
+
+    // <https://learn.microsoft.com/en-us/typography/opentype/spec/gasp>
+    pub gasp: Vec<GaspRange>,
+}
+
+/// A `gasp` range
+///
+/// <https://learn.microsoft.com/en-us/typography/opentype/spec/gasp>
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct GaspRange {
```
cmyr, line 162: this looks like it's identical to the same type in `write-fonts`, so maybe we should just reuse that?

cmyr (approved):
one little note, but lgtm

fontir/src/ir.rs
rsheeter: oh right, we added serde. Fixing.

PR #1198, opened 2025-01-09
[glyphs] Impl 'erase open corners'

cmyr:
This implements the functionality to erase 'open corners', which are features of some outlines where the intended corner of two curves needs to be inferred.

This implementation is a bit more verbose than we'd like, because kurbo does not currently have an implementation of arbitrary curve-curve intersection, so we've ported the implementation from fonttools' bezierTools.

... this was way more of a headache than I'd expected, with many exciting wrong turns and dead-ends. In any case, it seems like finally this is working although it only gives us a +1 on crater 😭

### DIFF
--- a/glyphs2fontir/src/erase_open_corners.rs
+++ b/glyphs2fontir/src/erase_open_corners.rs
@@ -81,16 +81,26 @@ pub fn erase_open_corners(path: &BezPath) -> Option<BezPath> {
         ops[ix % segments.len()] = op;
     }
 
-    let mut els = Vec::with_capacity(path.elements().len());
-    let mut iter = segments
+    let segs = segments
         .iter()
         .zip(ops)
         .filter_map(|(seg, op)| op.apply(*seg))
-        .peekable();
+        .collect::<Vec<_>>();
 
-    // we will use `PathSeg::as_path_el` below, but that doesn't handle moveto
-    els.push(PathEl::MoveTo(iter.peek().unwrap().start()));
-    els.extend(iter.map(|seg| seg.as_path_el()));
+    let mut els = Vec::with_capacity(path.elements().len());
+
+    // now handle the move_to; for a closed path this is the last point,
+    // for an open path it was the start point of the first segment (which is
+    // ignored when converting segment->element)
+    let first_pt = if is_closed {
+        // unwrap is fine, we must have at least two segments if we had an open corner
+        segs.last().unwrap().end()
+    } else {
+        segs.first().unwrap().start()
+    };
+
+    els.push(PathEl::MoveTo(first_pt));
+    els.extend(segs.iter().map(PathSeg::as_path_el));
     if is_closed {
         els.push(PathEl::ClosePath);
     }
@@ -126,24 +136,32 @@ struct ErasureCandidate {
 }
 
 impl ErasureCandidate {
-    // https://github.com/googlefonts/glyphsLib/blob/74c63244fdb/Lib/glyphsLib/filters/eraseOpenCorners.py#L66-L71
     // Are the incoming point from the previous segment and the outgoing point
     // from the next segment both on the right side of the line?
     // (see discussion at https://github.com/googlefonts/glyphsLib/pull/663)
+    // https://github.com/googlefonts/glyphsLib/blob/74c63244fdb/Lib/glyphsLib/filters/eraseOpenCorners.py#L66-L71
     fn points_are_right_of_line(&self) -> bool {
-        // the last point in one segment is the first point in the next;
-        // so finding the prev or next point means ignoring the shared point.
-        let prev_point = match self.prev {
-            PathSeg::Line(line) => line.p0,
-            PathSeg::Quad(quad_bez) => quad_bez.p1,
-            PathSeg::Cubic(cubic_bez) => cubic_bez.p2,
-        };
-        let next_point = match self.next {
-            PathSeg::Line(line) => line.p1,
-            PathSeg::Quad(quad_bez) => quad_bez.p1,
-            PathSeg::Cubic(cubic_bez) => cubic_bez.p1,
-        };
+        // return the last point in this segment not shared with the next segment
+        // (i.e., ignoring the final point)
+        fn last_non_shared_point(seg: PathSeg) -> Point {
+            match seg {
+                PathSeg::Line(line) => line.p0,
+                PathSeg::Quad(quad_bez) => quad_bez.p1,
+                PathSeg::Cubic(cubic_bez) => cubic_bez.p2,
+            }
+        }
 
+        // return the first point in this segment not shared with the previous
+        // segment (i.e, ignoring the first point)
+        fn first_non_shared_point(seg: PathSeg) -> Point {
+            match seg {
+                PathSeg::Line(line) => line.p1,
+                PathSeg::Quad(quad_bez) => quad_bez.p1,
+                PathSeg::Cubic(cubic_bez) => cubic_bez.p1,
+            }
+        }
+        let prev_point = last_non_shared_point(self.prev);
+        let next_point = first_non_shared_point(self.next);
         !(point_is_left_of_line(self.seg, prev_point)
             || point_is_left_of_line(self.seg, next_point))
     }
@@ -435,15 +453,7 @@ mod tests {
 
     macro_rules! assert_approx {
         ($left:expr, $right:expr) => {
-            assert_approx!($left, $right, 1e-4)
-        };
-        ($left:expr, $right:expr, $accuracy:literal) => {
-            assert!(
-                ($left - $right).abs() < $accuracy,
-                "{} !~= {}",
-                $left,
-                $right
-            )
+            assert!(($left - $right).abs() < 1e-4, "{} !~= {}", $left, $right)
         };
     }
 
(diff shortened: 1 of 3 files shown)


glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -0,0 +1,524 @@
...
+        };
+        ($left:expr, $right:expr, $accuracy:literal) => {
+            assert!(
+                ($left - $right).abs() < $accuracy,
+                "{} !~= {}",
+                $left,
+                $right
+            )
+        };
+    }
```
rsheeter, line 424: I'm failing to grasp why this needs to be a macro? If you made it a free fn that took things that impl a trait you could implement for points which would be nicer than doing x and y as individual asserts.

glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -0,0 +1,524 @@
...
+
+    #[test]
+    fn test_curve_curve_glyph() {
+        let glyph = python_test_glyphs::curvyCornerGlyph();
+        let after = erase_open_corners(&glyph);
+        let new_pt = after.unwrap().segments().next().unwrap().start();
+        // teaching to the test: because our segment splitting impl is different
+        // from fonttools', we get slightly different values here; we might get
+        // a few off-by-ones but that's okay?
```
rsheeter, line 452: off by 0.2 seems like kind of a lot?

glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -0,0 +1,524 @@
...
+        else {
+            panic!("wrong path elements: {:?}", after.elements())
+        };
+        assert_eq!(one, Point::new(100., 0.,));
+        assert_eq!(two, Point::new(400., 0.,));
+        assert_eq!(tre, Point::new(400., 400.,));
+        assert_eq!(four, Point::new(100., 400.,));
+        // sanity check
+        assert_eq!(one, end);
+        assert!(erase_open_corners(&glyph.reverse_subpaths()).is_none());
```
rsheeter, line 472: I'd much rather see one assert_eq for the entire sequence, e.g. `assert_eq!(expected, after.elements())`

rsheeter: (here and below)

glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -0,0 +1,524 @@
...
+            SegmentOp::Replace(seg) => Some(seg),
+        }
+    }
+}
+
+/// A potentially open corner.
+///
+/// This is any line segment, along with the preceding and following path segments.
+struct ErasureCandidate {
+    // index of the *first* participating segment, which contains prev_pt
```
rsheeter, line 121: I'm not so sure I follow this, I was expecting it to be specified to be the index of one of seg, prev, and next in the segments of the original bezpath?

glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -0,0 +1,524 @@
...
+        let prev_point = match self.prev {
+            PathSeg::Line(line) => line.start(),
+            PathSeg::Quad(quad_bez) => quad_bez.p1,
+            PathSeg::Cubic(cubic_bez) => cubic_bez.p2,
+        };
+        let next_point = match self.next {
+            PathSeg::Line(line) => line.end(),
+            PathSeg::Quad(quad_bez) => quad_bez.p1,
+            PathSeg::Cubic(cubic_bez) => cubic_bez.p1,
+        };
```
rsheeter, line 140: At a glance it's hard to spot the difference between these matches. I suggest a trait implemented by PathSeg might be clearer, such that this would read

```rust
let prev_pt = self.prev.end_pt();
let next_pt = self.next.p1();
```

glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -0,0 +1,524 @@
...
+        let seg = iter.next()?;
+        let next = iter.peek()?;
+        let result = (prev, seg, *next);
+        prev = seg;
+        Some(result)
+    })
+}
+
+#[derive(Clone, Debug)]
+struct Intersection {
```
rsheeter, line 217: intersection point is obvious. `t` may be magical to some readers so I suggest a comment to clarify what t0/t1 are.

glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -0,0 +1,524 @@
...
+    t0: f64,
+    t1: f64,
+    point: Point,
+}
+
+//given two segments, return an intersection point.
+//
+// If there are multiple intersections, pick the one closest to the end of seg1.
```
rsheeter, line 225: Potential rephrase: Find the intesection nearest the end of seg1, if any

glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -0,0 +1,524 @@
...
+            .map(|hit| Intersection {
+                t0: hit.segment_t,
+                t1: hit.line_t,
+                point: line.eval(hit.line_t),
+            }),
+        (bez0, bez1) => curve_curve_intersection_py(bez0, bez1),
+    }
+}
+
+const PY_ACCURACY: f64 = 1e-3;
```
rsheeter, line 250: link to py?

glyphs2fontir/src/erase_open_corners.rs
cmyr: I agree, but I didn't want to try and port the fonttools impl just for the sake of matching... ultimately any difference is going to leave open the possibility of rounding errors, and i think this is a case where if we very occasionally use up one of our off-by-one budget here it's not the end of the world?

glyphs2fontir/src/erase_open_corners.rs
cmyr: ah the comment might be out of date. it's the index of `prev` (we used to store just the previous point before I realized we'd need the full segment)

glyphs2fontir/src/erase_open_corners.rs
cmyr: the thing is it isn't the end point, because we are also including off-curves. This threw me too, at first. The other thing to keep in mind is that segments are independent, so they contain duplicate points (the end of one segment is the start of the subsequent one). I can add a comment to this effect, but since this logic is only used in one place it feels premature to try and do something fancier, here...

glyphs2fontir/src/erase_open_corners.rs
cmyr: the macro is variadic (has an optional argument) and also generates the panic at the assert site, including the printed message; to achieve this with a fn we would be replacing,

```rust
assert_approx!(new_pt.x, 114.5417);
```
with,
```rust
assert!(approx_eq(new_pt.x, 114.5417, 1e-4), "{new_pt.x} !~= 114.5417");
```

(you could have the assert in the body of the free fn, but then your backtrace originates there, instead of at the call site, which is a small thing but bugs me.)


this could be rewritten to operate on points instead of floats, but i was pretty mindlessly porting the python and these are separate there, so 🤷

glyphs2fontir/src/erase_open_corners.rs
rsheeter: Having it pop out in review as well suggests its worth a little polish imho. Bias to address review feedback basically.

rsheeter: Oh I see it's the point before the end point ...what a weird process. self.prev.second_last_pt() or some such would be materially easier to read imho

glyphs2fontir/src/erase_open_corners.rs
rsheeter: Hm, I don't think `assert!(approx_eq(new_pt.x, 114.5417, 1e-4), "{new_pt.x} !~= 114.5417");` is on the table, I'd have expected something like `assert_approx_eq(pt, (42, 43));`. I hadn't considered callstack root, that seems worth a comment?

glyphs2fontir/src/erase_open_corners.rs
rsheeter: I still find the magnitude alarming. If you told me we are off by 0.001 ... sure, but 0.2 seems significant unless we modified the implementation in a way that we expect to produce a large increase in the error rate? Or conversely we believe fonttools accumulates more error than we do somehow?

glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -0,0 +1,524 @@
...
+// 'left' from the perspective of an observer standing on line.p0 and lookign at line.p1?
+fn point_is_left_of_line(line: Line, point: Point) -> bool {
+    let Line { p0: a, p1: b } = line;
+    (b.x - a.x) * (point.y - a.y) - (b.y - a.y) * (point.x - a.x) >= 0.0
+}
+
+/// given a slice of path elements, determine which *segments* might be open corners.
+fn iter_erasure_candidates(
+    els: &[PathEl],
+    is_closed: bool,
```
rsheeter, line 164: IIUC is_closed is a functions of els, I suggest computing it rather than passing it in

glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -0,0 +1,524 @@
...
+    // index of the *first* participating segment, which contains prev_pt
+    ix: usize,
+    seg: Line,
+    prev: PathSeg,
+    next: PathSeg,
+}
+
+impl ErasureCandidate {
+    //https://github.com/googlefonts/glyphsLib/blob/74c63244fdb/Lib/glyphsLib/filters/eraseOpenCorners.py#L66-L71
+    fn points_are_right_of_line(&self) -> bool {
```
rsheeter, line 130: I don't find this methods behavior very intuitive. I suggest copying over this comment:

```python
            # Are the incoming point from the previous segment and the outgoing point
            # from the next segment both on the right side of the line?
            # (see discussion at https://github.com/googlefonts/glyphsLib/pull/663)
```

rsheeter (approved):
In general I feel like it looks a little translated from Python and might benefit from a bit more polishing into Rust. More traits, calling x.intersect(y) instead of intersect(x, y), etc. Not so much as to be blocking so LGTM.

glyphs2fontir/src/erase_open_corners.rs
cmyr: okay your instincts here were good, i'd missed something that the glyphsLib impl was doing (explicitly rebuilding the path after finishing all the operations) and so we were inadvertently ending up with a closed path where the start point was different from the end point of the last segment; fixing that means we can remove that accuracy parameter completely.

PR #1199, opened 2025-01-15
[glyphs] Rework corner erasure to delete in-place

cmyr:
My initial implementation of corner erasure had a significant problem: in an attempt to be clever I was not erasing corners as they were found, but instead collecting 'deletion operations' that could be applied all at once at the end.

The problem with this, in hindsight, is obvious: In a closed path, it is possible for a single segment to have an open corner at both ends, and correctly handling this requires applying the deletion of the first corner before trying to calculate the position of the second.

This new approach solves this by modifying the path in place, as corners are found. To do this efficiently it uses a copy-on-write type so that we only ever allocate if we actually are going to erase a corner, which is not the common case.

Doing it this was makes the logic tricky, because of how `BezPath` and `PathEl` work. In particular, the presence of `ClosePath` elements (which are significant in postscript but are essentially metadata for our purposes) makes indexing tricky; similarly in a closed path the first `MoveTo` element is redundant, and we need to account for that as well.

In hindsight, if I were tasked with writing this again tomorrow, I would probably not do it this way; instead I would try work with `PathSeg`s, which are easier to reason about, and I would just figure out some way of writing it so that we still only allocated as needed.

### DIFF
--- a/glyphs2fontir/src/erase_open_corners.rs
+++ b/glyphs2fontir/src/erase_open_corners.rs
@@ -86,10 +86,7 @@ impl<'a> CornerErasureCtx<'a> {
         // we don't use `for _ in 0..self.n_possible_candidates()` because
         // if we erase a corner the number of candidates changes
         while ix <= self.n_possible_candidates() {
-            if let Some((candidate, hit)) = self.open_corner_at(ix) {
-                self.erase_corner(candidate, hit);
-                made_changes = true;
-            }
+            made_changes |= self.erase_open_corner_if_present(ix);
             ix += 1;
         }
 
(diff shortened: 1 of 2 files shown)


rsheeter (approved):
> In hindsight, if I were tasked with writing this again tomorrow, I would probably not do it this way; instead I would try work with PathSegs, which are easier to reason about, and I would just figure out some way of writing it so that we still only allocated as needed.

I'm inclined to agree, my intuition is if you could express the algorithm as something like

```rust
for each el in path
   if it's not a line continue
   get the prev PathSeg
   get the next PathSeg

   if prev/next form an open corner write down the modifications
```

It might be significnatly simpler to read. However, I don't think it's worth blocking on this.

I also (still) think more traits would be helpful. Think `seg1.intersection(seg2)` instead of `curve_curve_intersection_py`.

cmyr:
that more or less _is_ the algorithm, it's just spread across a few different methods, and some of the names might be confusing. For an alternative structure (with identical logic) you can take a look at https://github.com/googlefonts/fontc/compare/open-corners-again...open-corners-again-option2; maybe that's better?

On the question of traits: it's definitely a bit squishy in terms of when a new trait feels worth it, but as a rule of thumb I wouldn't make a trait if it's only going to be impl'd for one type, and is only going to be called once.

In the particular case of the `curve_curve_intersection_py`, this is attempting to be a pretty literal transcription of the equivalent python function... in the future (at least on a long enough timeline) I would expect kurbo to add its own `intersection` method to `PathSeg`, and we may well want to move to that, so I think it makes sense to keep this fn sort looking like a bit of an ugly duckling?

PR #1200, opened 2025-01-16
[crater] Support relative source paths

cmyr:


fontc_crater/src/ci/html.rs
```diff
@@ -231,7 +231,7 @@ fn make_table_body(runs: &[RunSummary]) -> Markup {
     let iter = runs.iter().enumerate().map(|(i, run)| {
         let prev = (i > 0).then(|| &runs[i - 1]);
         let is_last = i == runs.len() - 1;
-        let default_visibility = i >= runs.len() - 16;
```
rsheeter, line 234: idk what this has to do with relative paths?

rsheeter (approved):

fontc_crater/src/ci/html.rs
cmyr: yes it's an unrelated crash but was triggered here because I was testing in a new directory (and so this overflows)

PR #1202, opened 2025-01-17
[crater] Support for private fonts

cmyr:
That is, fonts that live in private repos under the googlefonts org.

### DIFF
--- a/fontc_crater/src/ci.rs
+++ b/fontc_crater/src/ci.rs
@@ -251,13 +251,19 @@ fn targets_for_source(
 }
 
 fn should_build_in_gftools_mode(src_path: &Path, config: &Config) -> bool {
+    let file_stem = src_path
+        .file_stem()
+        .map(|s| s.to_string_lossy())
+        .unwrap_or_default();
     // skip noto, which have an implicitly different recipe provider
     //https://github.com/googlefonts/oxidize/blob/main/text/2024-06-26-fixes-and-nonstandard-builds.md#noto
-    if src_path
-        .file_stem()
-        .and_then(|stem| stem.to_str().map(|s| s.to_lowercase().starts_with("noto")))
-        .unwrap_or(false)
-    {
+    if file_stem.to_lowercase().starts_with("noto") {
+        return false;
+    }
+
+    // skip Google Sans, which isn't built with gftools (but other fonts with
+    // names that begin with 'GoogleSans' might be!)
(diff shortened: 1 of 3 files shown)


fontc_crater/src/ci.rs
```diff
@@ -251,13 +251,19 @@ fn targets_for_source(
...
-        .and_then(|stem| stem.to_str().map(|s| s.to_lowercase().starts_with("noto")))
-        .unwrap_or(false)
-    {
+    if file_stem.to_lowercase().starts_with("noto") {
+        return false;
+    }
+
+    // skip Google Sans, which isn't built with gftools (but other fonts with
+    // names that begin with 'GoogleSans' might be!)
+    if ["GoogleSans", "GoogleSans-Italic"].contains(&file_stem.as_ref()) {
```
rsheeter, line 266: Let's file an issue to do this via a config marker rather than hard-coded list.

Why are we looking at the file stem, doesn't the config give us the family name?

fontc_crater/src/main.rs
```diff
@@ -159,9 +159,10 @@ impl<T, E> Default for Results<T, E> {
 impl From<LoadRepoError> for SkipReason {
     fn from(value: LoadRepoError) -> Self {
         match value {
-            LoadRepoError::Io(_) | LoadRepoError::GitFail(_) | LoadRepoError::NoCommit { .. } => {
-                SkipReason::GitFail
-            }
+            LoadRepoError::Io(_)
+            | LoadRepoError::GitFail(_)
+            | LoadRepoError::NoCommit { .. }
+            | LoadRepoError::MissingAuth => SkipReason::GitFail,
```
rsheeter, line 165: If skip reasons surface in the UI or logs I suggest having one specifically for missing auth

rsheeter (approved):

fontc_crater/src/ci.rs
cmyr: the family name isn't mandatory.


The config marker thing is a bit funny, in that the reason we're skipping gftools for these fonts is because they don't have gftools configs; the noto fonts have an extremely similar looking config file but don't build with gftools for legacy reasons, and GoogleSans has a stub I added manually.

basically: does it make sense to have a field in a gftools config file that says "don't build me with gftools"?

fontc_crater/src/main.rs
cmyr: actually I think SkipReason is now completely unused, can delete in a followup.

PR #1208, opened 2025-01-22
[glyphs] Don't rotate nodes if all are off-curve

### LINKED ISSUE #1207: Handling of contours with no on-curve points
Another funny little thing showing up in diffs:

- `ProtestRevolution.glyphs` includes a contour that has no on-curve points, but has two off-curves. This is meaningless and likely an artifact of some bug, but nonetheless it is in the source.
- In glyphsLib, the logic in the point pen (during conversion to ufo) reorders the points in closed paths, which this happens to be, which means fontmake puts the second point first.
- fontc doesn't reorder the points.

Ultimately, though, is there _any_ merit in including such a contour? I feel like we've had similar discussions in the past, e.g. #1173, but this is another case where I really don't see the value in putting this contour in the final font, and I think we should just skip it (as I think we should skip contours with a single point, and probably contours with two points?)

So... I could pretty easily patch fontc to match the python (and patching the python here would involve funny special-case logic) but it feels to me like a better solution is just to drop the contour.

Edit: we could also handle this during normalization, although the logic is a bit tricky: specifically the sort order of contours depends on the order of points within them (but also other things in the case of decomposition) and so we'd need to figure that out..

cmyr:
Although it isn't supported in the editor, the glyphs source format supports curves with only on-curves, using the truetype 'implied oncurve points' model.

When dealing with a contour of this type, we should skip rotating the nodes when generating the path, because the first point of these contours is already correct (being the off-curve point of the first segment).

- fixes #1207

### DIFF
--- a/fontir/src/ir/path_builder.rs
+++ b/fontir/src/ir/path_builder.rs
@@ -188,7 +188,7 @@ impl GlyphPathBuilder {
     /// It's called automatically by `build()` thus can be
     /// omitted when building one BezPath per contour, but can be called manually in
     /// order to build multiple contours into a single BezPath.
-    pub fn end_path(&mut self) -> Result<(), PathConversionError> {
+    fn end_path(&mut self) -> Result<(), PathConversionError> {
         // a contour that does *not* start with a move is assumed to be closed
         // https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#point-types
         if !self.first_oncurve.is_some_and(|on| on.is_move()) {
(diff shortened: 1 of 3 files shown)


rsheeter (approved):
Novel that we have sources that do things that cannot be produced in the editor

florianpircher:
While it cannot be produced in the editor, opening a TTF file and saving that to a Glyphs file can yield such paths. Some people like this style of paths, so I suspect the editor will gain support to produce them from scratch as well.

PR #1209, opened 2025-01-23
Implement gasp for .glyphs

rsheeter:
Makes `python resources/scripts/ttx_diff.py 'https://github.com/tokotype/PlusJakartaSans#sources/PlusJakartaSans-Italic.glyphs' --rebuild fontc` report the `gasp` table identical

JMM if happy

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -108,6 +108,7 @@ pub struct CustomParameters {
     // in the top-level `Font` struct
     pub virtual_masters: Option<Vec<BTreeMap<String, OrderedFloat<f64>>>>,
     pub glyph_order: Option<Vec<SmolStr>>,
+    pub gasp_table: Option<BTreeMap<i64, i64>>,
 }
 
 /// Values for the 'meta Table' custom parameter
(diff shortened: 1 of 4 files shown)


glyphs-reader/src/font.rs
```diff
@@ -540,6 +542,15 @@ impl PlistParamsExt for Plist {
             .collect::<Result<_, _>>()
             .ok()
     }
+
+    fn as_gasp_table(&self) -> Option<BTreeMap<i64, i64>> {
+        Some(
+            self.as_dict()?
+                .iter()
+                .filter_map(|(k, v)| Some((k.parse::<i64>().ok()?, v.as_i64()?)))
```
cmyr, line 550: An opportunity to use the lovely `Option::zip`, if you wish:
```suggestion
                .filter_map(|(k, v)| k.parse::<i64>().ok().zip(v.as_i64)) 
```

cmyr (approved):
one extremely optional idea, otherwise lgtm

glyphs-reader/src/font.rs
rsheeter: TIL

PR #1211, opened 2025-01-23
Update to fonttools 4.55.5 in requirements.txt

### LINKED ISSUE #1206: Difference in glyph bbox calculation due to rounding inconsistency
So this has taken me the better part of a day to track down, but I've maybe made some progress?

The issue shows up in 'Faustina Italic', repro'd with the following:

```sh
python resources/scripts/ttx_diff.py 'https://github.com/Omnibus-Type/Faustina#sources/Faustina-Italic.glyphs'
```

And the issue is that for one glyph, 'dblgravecomb.case', we are computing a slightly different bounding box. This glyph as a component with a non-trivial transform; and as far as I can tell the issue is that fontmake is computing the bounding box after doing the cu2qu conversion, and so is working on slightly different points, and in this case we are ending up with values on either side of `.5`, which are being rounded differently.

My instinct here is that we should probably allow a few of these off-by-ones, similarly to how we handle glyph point off-by-ones?

I have a very minimized version of this font I've been using for testing, attached below.

[Faustina-Italic-minimal.glyphs.zip](https://github.com/user-attachments/files/18496213/Faustina-Italic-minimal.glyphs.zip)

anthrotype:
Will fix #1206

### DIFF
--- a/resources/scripts/requirements.txt
+++ b/resources/scripts/requirements.txt
@@ -68,7 +68,7 @@ defcon[lxml,pens]==0.10.3
     #   ufoprocessor
 deprecated==1.2.15
     # via pygithub
-filelock==3.16.1
+filelock==3.17.0
     # via youseedee
 font-v==2.1.0
     # via gftools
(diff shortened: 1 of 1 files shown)


cmyr (approved):
👍

PR #1213, opened 2025-01-23
[ttx_diff] put normalized `<contour>` elements first, before `<instructions>`

### LINKED ISSUE #1212: [ttx_diff] appends normalized contours at the end; it should put them at beginning before instructions etc.
I noticed that, when ttx_diff.py runs normalize_glyf_contours to normalize the order of contours in a TTGlyph, if that glyph contains an `<instructions>` element, after the contours are normalized, they get appended at the end of the TTGlyph's list of children. This may lead to spurious diffs like this one from 

python resources/scripts/ttx_diff.py 'https://github.com/googlefonts/kreon#sources/Kreon.glyphs'

```diff
--- /Users/clupo/oss/fontc/build/default/fontc.glyf.ttx	2025-01-23 17:47:05
+++ /Users/clupo/oss/fontc/build/default/fontmake.glyf.ttx	2025-01-23 17:47:05
@@ -1320,8 +1320,7 @@
     </TTGlyph>

     <TTGlyph name="Tbar" xMin="18" yMin="0" xMax="529" yMax="691">
-      <instructions/>
-    <contour>
+      <contour>
         <pt x="108" y="360" on="1"/>
         <pt x="442" y="360" on="1"/>
         <pt x="442" y="295" on="1"/>
\ No newline at end of file
@@ -1345,7 +1344,8 @@
         <pt x="423" y="52" on="1"/>
         <pt x="423" y="0" on="1"/>
       </contour>
-      </TTGlyph>
+      <instructions/>
+    </TTGlyph>
```

There's no difference really, just that the fontc's contours were normalized while fontmake's were not, but the `<instructions/>` element ends up in different position.

Instead of just removing all contour elements and appending their normalized version at end, we should maintain ttx's ordering whereby all contours are dumped first, then the rest of the TTGlyph's children. As far as I am aware, that's just `<instructions>` basically, which is optional.

anthrotype:
Fixes #1212

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -387,10 +387,14 @@ def normalize_glyf_contours(ttx):
         normalized = sorted(contours, key=to_xml_string)
         if normalized == contours:
             continue
+        # normalized contours should be inserted before any other TTGlyph's
+        # subelements (e.g. instructions)
         for contour in contours:
             glyph.remove(contour)
-        for contour in normalized:
-            glyph.append(contour)
+        non_contours = list(glyph)
+        for el in non_contours:
+            glyph.remove(el)
+        glyph.extend(normalized + non_contours)
 
 
 # https://github.com/googlefonts/fontc/issues/1173


cmyr (approved):
seems reasonable!

resources/scripts/ttx_diff.py
```diff
@@ -387,10 +387,14 @@ def normalize_glyf_contours(ttx):
...
             continue
+        # normalized contours should be inserted before any other TTGlyph's
+        # subelements (e.g. instructions)
         for contour in contours:
             glyph.remove(contour)
-        for contour in normalized:
-            glyph.append(contour)
+        non_contours = list(glyph)
+        for el in non_contours:
+            glyph.remove(el)
```
rsheeter, line 396: Since you are now removing everything I would imagine you can remove both remove loops and just call something to drop all child elements before you add everything back?

anthrotype: we need a list of elements that are not `<contours>` to be placed at the end after the normalized ones, we could alternatively drop all child elements and collect the ones that aren't contours in one for loop.. but it wouldn't change much. feel free to edit

PR #1219, opened 2025-01-28
[glyphs-reader] normalize 90°/180°/270° rotated component transform

### LINKED ISSUE #1218: rounding diffs in glyf/gvar for components rotated along cardinal directions
Run

```
python resources/scripts/ttx_diff.py 'https://github.com/googlefonts/karla#sources/Karla-Italic.glyphs' --compare gftools --config ~/.fontc_crater_cache/googlefonts/karla/sources/config.yaml
```

and observe a bunch of off-by-one diffs in both glyf and gvar table, which exceed the tolerated budget of off-by-ones.

These affect a composite glyph "questiondown" that contain a component rotated 180° (which effectively flips the sign of both x and y coordinates). In gftools mode, we decompose all transformed components, so the rotated composite glyph is turned into a simple contour glyph with the rotation applied to the points. These contain cubic curves that are converted to quadratics. Finally the coordinates are rounded to integer before building glyf and computing gvar deltas.

The problem is that [kurbo::Affine::rotate](https://github.com/linebender/kurbo/blob/d7bb94dddc6364faa3302e4468e1fe66c7f08fc1/src/affine.rs#L74-L77) method simply calls the sin and cos functions to build the rotation matrix, but when the angle is a perfect 0°, 90°, 180° and 270° angle (or the equivalent in radians, 0, π/2, π, 3/2*π), the standard sine and cosine functions in Rust (as well as Python) return values that are very close but not exactly 0.0, +1.0 or -1.0. The sin/cos are implemented by the platform's standard math library (e.g. libc) and as such have ["unspecified precision"](https://doc.rust-lang.org/std/primitive.f64.html#unspecified-precision-11). E.g.

```rust
>> use std::f64::consts::PI;
>> PI.sin_cos()
(1.2246467991473532e-16, -1.0)
>> (PI/2.0).sin_cos()
(1.0, 6.123233995736766e-17)
>> (3.0*PI/2.0).sin_cos()
(-1.0, -1.8369701987210297e-16)
```

That's expected and fine. However in particular circumstances, as in the case of "questiondown", the rotated and cu2qu-ed points will end up like `(455.4999999999999, 533.5000000000002)` instead of say `(455.5, 533.5)`, thus by the time we do ot_round() on them, we will get diffs like the ones observed in Karla-Italic.glyphs, i.e.

```diff
-        <pt x="455" y="534" on="0"/>
+        <pt x="456" y="534" on="0"/>
```

In fontTools, the `Transform` class that corresponds to kurbo's Affine struct, contains code to ensure that after computing the sine and cosine of the rotation angle, these will snap to exactly 0, +1 and -1 if they are within some very small epsilon (`1e-15`), see:

https://github.com/fonttools/fonttools/blob/b7509b24a9ed415f1dd166ac51551df78a5ee43a/Lib/fontTools/misc/transform.py#L65-L77
https://github.com/fonttools/fonttools/blob/b7509b24a9ed415f1dd166ac51551df78a5ee43a/Lib/fontTools/misc/transform.py#L256-L257

That means that the 0s and 1s of perfect cardinal rotation angles are _exactly_ 0.0 and +/-1.0 in fontTools, and glyphLib/fontmake which use the latter: 

```python
In [1]: from fontTools.misc.transform import Transform
In [2]: from math import radians, pi
In [3]: Transform().rotate(pi)
Out[3]: <Transform [-1 0 0 -1 0 0]>
In [4]: Transform().rotate(pi/2)
Out[4]: <Transform [0 1 -1 0 0 0]>
In [5]: Transform().rotate(3*pi/2)
Out[5]: <Transform [0 -1 1 0 0 0]>
```

I think we can and should fix these diffs, instead of simply increase the tolerated budget, because it's trivial to do. But I think we should do it on our side instead of changing kurbo::Affine::rotate itself, since other kurbo clients may be ok with the current behavior.

The only place where we use Affine::rotate right now is in glyphs-reader/src/font.rs where we are composing the Glyphs 3 component's `pos`, `angle` and `scale` attributes into an Affine, so I think for now we can add the `norm_sin_cos` code in there.

I have a patch locally that makes Karla-Italic.glyphs ttx_diff go "output is identical", which I'll send now.

anthrotype:
Fixes #1218

This matches `fontTools.misc.transform::Transform.rotate()` method, and makes Karla-Italic.glyphs ttx_diff.py output identical.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2013,11 +2013,11 @@ impl TryFrom<RawShape> for Shape {
 
 // Direct port from
 // <https://github.com/fonttools/fonttools/blob/b7509b2/Lib/fontTools/misc/transform.py#L65-L77>
-const EPSILON: f64 = 1e-15;
-const ONE_EPSILON: f64 = 1.0 - EPSILON;
-const MINUS_ONE_EPSILON: f64 = -1.0 + EPSILON;
-
 fn norm_sin_cos(v: f64) -> f64 {
+    const EPSILON: f64 = 1e-15;
+    const ONE_EPSILON: f64 = 1.0 - EPSILON;
+    const MINUS_ONE_EPSILON: f64 = -1.0 + EPSILON;
+
     if v.abs() < EPSILON {
         0.0
     } else if v > ONE_EPSILON {
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2011,24 +2011,6 @@ impl TryFrom<RawShape> for Shape {
     }
 }
 
-// Direct port from
-// <https://github.com/fonttools/fonttools/blob/b7509b2/Lib/fontTools/misc/transform.py#L65-L77>
-fn norm_sin_cos(v: f64) -> f64 {
-    const EPSILON: f64 = 1e-15;
-    const ONE_EPSILON: f64 = 1.0 - EPSILON;
-    const MINUS_ONE_EPSILON: f64 = -1.0 + EPSILON;
-
-    if v.abs() < EPSILON {
-        0.0
-    } else if v > ONE_EPSILON {
-        1.0
-    } else if v < MINUS_ONE_EPSILON {
-        -1.0
-    } else {
-        v
-    }
-}
-
 /// Return [kurbo::Affine] rotation around angle (in degrees) with normalized sin/cos.
 ///
 /// This ensures that for the four "cardinal" rotations (0, 90, 180, 270), the values
@@ -2039,6 +2021,24 @@ fn norm_sin_cos(v: f64) -> f64 {
 /// It matches the output of the fontTools' Transform.rotate() used by glyphsLib.
 /// <https://github.com/fonttools/fonttools/blob/b7509b2/Lib/fontTools/misc/transform.py#L246-L258>
 fn normalized_rotation(angle_deg: f64) -> Affine {
+    // Direct port from
+    // <https://github.com/fonttools/fonttools/blob/b7509b2/Lib/fontTools/misc/transform.py#L65-L77>
+    fn norm_sin_cos(v: f64) -> f64 {
+        const EPSILON: f64 = 1e-15;
+        const ONE_EPSILON: f64 = 1.0 - EPSILON;
+        const MINUS_ONE_EPSILON: f64 = -1.0 + EPSILON;
+
+        if v.abs() < EPSILON {
+            0.0
+        } else if v > ONE_EPSILON {
+            1.0
+        } else if v < MINUS_ONE_EPSILON {
+            -1.0
+        } else {
+            v
+        }
+    }
+
     Affine::new(
         Affine::rotate(angle_deg.to_radians())
             .as_coeffs()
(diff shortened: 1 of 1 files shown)


glyphs-reader/src/font.rs
```diff
@@ -2011,6 +2011,41 @@ impl TryFrom<RawShape> for Shape {
     }
 }
 
+// Direct port from
+// <https://github.com/fonttools/fonttools/blob/b7509b2/Lib/fontTools/misc/transform.py#L65-L77>
+const EPSILON: f64 = 1e-15;
+const ONE_EPSILON: f64 = 1.0 - EPSILON;
+const MINUS_ONE_EPSILON: f64 = -1.0 + EPSILON;
```
cmyr, line 2018: since these are only used in side `norm_sin_cos` I would define them in there? It just makes it clear in the future that they're just for that fn, and makes it less likely they get separated in some future patch or refactor.

(I'd personally also define `norm_sin_cos` inside of `normalized_rotation` for the same reason, but that's a more contentious choice)

cmyr (approved):
makes sense! one little thought inline but looks good :)

glyphs-reader/src/font.rs
```diff
@@ -3656,4 +3691,34 @@ mod tests {
...
+        );
+        assert_eq!(
+            Affine::new([-1.0, 0.0, 0.0, -1.0, 0.0, 0.0]),
+            normalized_rotation(180.0)
+        );
+        assert_eq!(
+            Affine::new([0.0, -1.0, 1.0, 0.0, 0.0, 0.0]),
+            normalized_rotation(270.0)
+        );
+        // any other angles' sin and cos != (0, ±1) are passed through unchanged
```
rsheeter, line 3718: This is many tests in one, could we split them or use a parameterized test or make it one assert so it tells us all the problems if there are multiple failures?

Also, we should test values very close to multiples of 90. If we believe epsilon does anything - e.g. we can normalize a value that isn't an integer multiple of 90.0 - we should prove it.

glyphs-reader/src/font.rs
```diff
@@ -2011,6 +2011,41 @@ impl TryFrom<RawShape> for Shape {
...
+/// Return [kurbo::Affine] rotation around angle (in degrees) with normalized sin/cos.
+///
+/// This ensures that for the four "cardinal" rotations (0, 90, 180, 270), the values
+/// of sin(angle) and cos(angle) are rounded to *exactly* 0.0, +1.0 or -1.0, thus avoiding
+/// very small values (e.g. 1.2246467991473532e-16) whose effect may be amplified as some
+/// transformed point coordinates end up close to the 0.5 threshold before being rounded
+/// to integer later in the build.
+/// It matches the output of the fontTools' Transform.rotate() used by glyphsLib.
+/// <https://github.com/fonttools/fonttools/blob/b7509b2/Lib/fontTools/misc/transform.py#L246-L258>
+fn normalized_rotation(angle_deg: f64) -> Affine {
```
rsheeter, line 2041: I wonder if any font source actually benefits from the use of epsilon instead of just checking if the value is an exact multiple of 90? - it's so small that it seems unlikely a source has a value that isn't a multiple of 90 and matches. That woudl lead me to implement this as a lookup of a const Affine, perhaps similar to this [sketch](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e46e9d7288ccf00630d06138306781e3)

rsheeter (approved):
LGTM with a couple of small suggestions

glyphs-reader/src/font.rs
anthrotype: > I wonder if any font source actually benefits from the use of epsilon

I don't think so, good point! Since we already have angles in degrees, we can just test the 4 cardinal angles directly.

See https://github.com/googlefonts/fontc/pull/1219/commits/f178e899497f453efdc06ab05d3da221d326fee3

PR #1220, opened 2025-01-28
update glyphsLib 6.10.0 in requirements.txt

anthrotype:
includes fix for https://github.com/googlefonts/glyphsLib/issues/1059 which should improve fontc_crater's overall diff score


cmyr (approved):
excited to see what this does for crater :)

PR #1221, opened 2025-01-29
Work around icu4x issue with unicode 16 scriptext

cmyr:
This also bumps us to use the new icu4x 2.0 beta.

icu4x is not currently returning the correct script values for codepoints that were added to ScriptExtensions.txt in Unicode 16, so we have a temporary workaround that manually checks for those specific values before calling out to icu4x.

Note: this will cause a a big crater drop until we merge and bring in https://github.com/fonttools/fonttools/pull/3756.


rsheeter (approved):
LGTM, please make sure we have an issue open to update ICU and delete this code

PR #1224, opened 2025-01-30
update gftools and ufo2ft in requirements

anthrotype:
notably includes fixes for https://github.com/googlefonts/gftools/pull/1076 and https://github.com/googlefonts/ufo2ft/issues/896#event-16122153076


cmyr (approved):
👍

PR #1226, opened 2025-01-30
Don't encode all-zero LigCaret deltas

cmyr:
We handle this correctly everywhere else, but were not doing it here.

JMM


anthrotype (approved):
LGTM

(bonus point if you also added a little unit test)

PR #1227, opened 2025-01-30
Normalize Ligature Caret List

cmyr:
This table is a lot like mark/kern rules, in that it contains scalar values that can have variations applied. Like with those structures, this is an excellent case for normalization, since it means we can ignore differences related to the order in which deltas are stored.

This patch adds the ability to normalize this table, and updates ttx_diff to do that and to remove it as well as the VarStore from GDEF.

This gives us another +20 or so on crater.

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -61,6 +61,7 @@ GFTOOLS_FONTC_PATH = "GFTOOLS_FONTC_PATH"
 FLAGS = flags.FLAGS
 # used instead of a tag for the normalized mark/kern output
 MARK_KERN_NAME = "(mark/kern)"
+LIG_CARET_NAME = "ligcaret"
 # maximum chars of stderr to include when reporting errors; prevents
 # too much bloat when run in CI
 MAX_ERR_LEN = 1000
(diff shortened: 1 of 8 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -154,14 +155,20 @@ def run_ttx(font_file: Path):
 
 
 # generate a simple text repr for gpos for this font, with retry
-def run_normalizer_gpos(normalizer_bin: Path, font_file: Path):
-    out_path = font_file.with_suffix(".markkern.txt")
+def run_normalizer(normalizer_bin: Path, font_file: Path, table: str):
+    if table == "gpos":
+        out_path = font_file.with_suffix(".markkern.txt")
+    elif table == "gdef":
+        out_path = font_file.with_suffix(".norm_gdef.txt")
```
rsheeter, line 162: nit: I like the idea of the normalizer outputs having a consistent naming pattern. Why do we call gpos markkern and gdef norm_gdef instead of something consistent?

rsheeter (approved):
+20 ftw!

resources/scripts/ttx_diff.py
cmyr: I'd avoided this because I thought it would screw up the next crater run (since the old run would produce the first name, and the second would produce the new name, and they wouldn't know about each other) but this filename isn't actually the name that we use in crater so it should be fine, can change `markkern` to `norm_gpos`.

PR #1231, opened 2025-01-31
[fea-rs] Ensure feature lookups are always sorted

cmyr:
The order of lookups within a feature don't matter. In general we end up with a sorted order via the implementation, but in the case where there are multiple declarations of a given feature this was not always the case.

### DIFF
--- a/fea-rs/src/compile/features.rs
+++ b/fea-rs/src/compile/features.rs
@@ -398,6 +398,7 @@ impl ActiveFeature {
                 None => &mut feature.base,
             };
             to_add.extend(lookups);
+            to_add.sort();
         }
     }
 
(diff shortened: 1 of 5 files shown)


rsheeter:
Please update the comment to explain why, e.g. are we making the result deterministic where it previously wasn't, matching fonttools, ...?

rsheeter (approved):
LGTM subject to reason why, I'm a little unclear what the purpose is.

cmyr:
right, to clarify: there is no defined or correct order for lookup ids inside a feature; in fea-rs we made the decision to always have them in ascending order, since that's simple to do and at least gives us consistent behaviour. fonttools' ordering depends on implementation details, so we sort the fonttools lookups in ttx_diff. (updated a comment)

anthrotype:
>  fonttools' ordering depends on implementation details

fonttools stores them in a dict that remembers the insertion order in which the lookups are registered under a given feature

PR #1232, opened 2025-01-31
[glyphs] Don't error on invalid vendorID value

cmyr:
Since this is allowed to be missing; we can just log the error and continue.

This gets us compiling at least one more font, `madimi.glyphs`.

As a followup, the only diff between us and fontmake for this font is that we have the OS/2 `achVendID` value of `NONE` while fontmake has `    ` (0x20 * 4)`. I think fontmake is wrong here, since all spaces is not a valid tag, and the [spec clearly says that this should be a tag](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#achvendid).

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -301,11 +301,13 @@ impl Work<Context, WorkId, Error> for StaticMetadataWork {
         .map_err(Error::VariationModelError)?;
         static_metadata.misc.selection_flags = selection_flags;
         if let Some(vendor_id) = font.vendor_id() {
-            static_metadata.misc.vendor_id =
-                Tag::from_str(vendor_id).map_err(|cause| Error::InvalidTag {
-                    raw_tag: vendor_id.to_owned(),
-                    cause,
-                })?;
+            match Tag::from_str(vendor_id) {
+                Ok(id) => static_metadata.misc.vendor_id = id,
+                Err(e) => {
+                    // we can log and continue here, this value is not required
+                    log::warn!("Invalid vendorID: '{e}'");
+                }
+            }
         }
 
         // Default per <https://github.com/googlefonts/glyphsLib/blob/cb8a4a914b0a33431f0a77f474bf57eec2f19bcc/Lib/glyphsLib/builder/custom_params.py#L1117-L1119>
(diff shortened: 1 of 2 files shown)


rsheeter (approved):
LGTM but please add a test.

anthrotype:
> all spaces is not a valid tag, and the [spec clearly says that this should be a tag](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#achvendid)

in the page you linked, they say (emphasis mine)

> This field can also be left blank (set to null, or _a tag comprised of four space characters_).

anthrotype:
in this case the tag is not missing from the source, but it is set to an explicit four-space tag `    `, fontmake is simply writing that out.
Are you suggesting that we use the empty four-space tags instead of `NONE` when the vendorID is not specified in the source?
I don't actually know where the `NONE` came from to begin with, tbh.

cmyr:
Since four spaces is not a valid tag (and only indicates the absence of a tag) I don't think there is obvious value in passing it through to the binary, **if** we have decided that we use `NONE` to indicate the absence of a tag in the general case.

Basically we should think of the tag as an `Option<Tag>`, and `    ` means `None`. arguably in glyphs2fontir we should just treat `    ` the same as if the field wasn't present at all.

cmyr:
Also it looks like 'NONE' comes from here: https://github.com/googlefonts/ufo2ft/blob/d82816873ed40a8015159843167e4a19e2ca9f2f/Lib/ufo2ft/fontInfoData.py#L378

I think it would be more correct for this value to be `    `.

anthrotype:
> arguably in glyphs2fontir we should just treat      the same as if the field wasn't present at all.

yeah, that'd be better than issuing a warning that says this is invalid, given that the spec says four spaces are allowed and mean no vendor tag

As to whether we should change ufo2ft's default.. I'm not sure.

cmyr:
> > arguably in glyphs2fontir we should just treat      the same as if the field wasn't present at all.
> 
> yeah, that'd be better than issuing a warning that says this is invalid, given that the spec says four spaces are allowed and mean no vendor tag

Okay, opened [TK](https://github.com/googlefonts/fontc/pull/1238) to make that little change.

> As to whether we should change ufo2ft's default.. I'm not sure.

agreed; I went and played around with this and then noticed that this tag is reused in the name table as part of the 'UniqueID' fallback path, so we would also need to decide on our behaviour there (as in, do we start serializing "    " as part of this name, or do we replace it with the empty string? Something else?)

Issue #1215, opened 2025-01-24
Pervasive of OsString causes problems for wasm


simoncozens:
fontc uses `fontbe` which uses `fea-rs` which has a structure `Source` which stores feature file paths as `OsString`s. `fea-rs` uses the `serde` feature to serialize these paths for `fontbe` to do persistable job orchestration.

However,  `serde` does not implement `Serialize` and `Deserialise` for `OsString`s on the WASM platform, only Unix and Windows, so the whole thing falls down. We could replace it all with `String`s to work everywhere.

Additionally, `fontbe/src/feature.rs` has the concept of an `InMemoryResolver` which sounds perfect for things which don't have filesystems! Except that it also stores `content_path: OsString`, which makes it dependent on filesystems again.

davelab6:
This may be a blocker for using fontc within fontraweb; @rsheeter how important does this sound to you

simoncozens:
In general: attempting to build `fontc` and its dependent libraries with `--target wasm32-unknown-unknown` from time to time is a great way to shake out unintended assumptions.

cmyr:
Yea, we could definitely get rid of OsStr; the basic idea is that in general we're dealing with paths, but we wanted something that would also work for arbitrary identifiers for the in-memory case, and so I went with OsString. This is easy enough to fix though, if it's a problem!

rsheeter:
> This may be a blocker for using fontc within fontraweb; [@rsheeter](https://github.com/rsheeter) how important does this sound to you

"may be" or definitely is? - I'd have thought definitely is.

I think it's worth doing but lower priority than making the brand fonts identical on https://googlefonts.github.io/fontc_crater/. If that sounds wrong please lmk.

> attempting to build `fontc` and its dependent libraries with `--target wasm32-unknown-unknown` from time to time is a great way to shake out unintended assumptions

+1 to this, lets systematically work our way through making sure that works and adding it to CI. I imagine we'd want to start with the fontations libraries. I will rename this issue to reflect the more general problem.

simoncozens:
Adding CI to fontations is a good idea, but I don't anticipate any problems there as we are using read-fonts/write-fonts/skrifa heavily already in [Fontspector web](https://simoncozens.github.io/fontspector) (yes it still has Fontbakery branding but it's all Rust underneath).

rsheeter:
[...]

PR #1233, opened 2025-01-31
Add a wasm check for the parts that currently work

rsheeter:
Try not to backslide on wasm support (#1215 )

JMM


cmyr (approved):
good call!

PR #1235, opened 2025-01-31
Make fea-rs build for wasm with serde enabled

rsheeter:
By removing use of `OsStr(ing)?`

Also add the rust.yml checks to prove it

Another step toward building fontc for wasm (#1215)

### DIFF
--- a/fea-rs/src/parse/context.rs
+++ b/fea-rs/src/parse/context.rs
@@ -2,8 +2,8 @@
 
 use std::{
     collections::{HashMap, HashSet},
-    ffi::OsString,
     ops::Range,
+    path::{Path, PathBuf},
     sync::Arc,
 };
 
@@ -117,7 +117,7 @@ impl ParseContext {
     ///
     /// [`generate_parse_tree`]: ParseContext::generate_parse_tree
     pub(crate) fn parse(
-        path: OsString,
+        path: PathBuf,
         glyph_map: Option<&GlyphMap>,
         resolver: Box<dyn SourceResolver>,
     ) -> Result<Self, SourceLoadError> {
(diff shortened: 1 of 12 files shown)


fea-rs/src/parse/source.rs
```diff
@@ -334,15 +331,15 @@ impl SourceLoader {
 }
 
 impl SourceList {
-    pub(crate) fn id_for_path(&self, path: impl AsRef<OsStr>) -> Option<FileId> {
-        self.ids.get(path.as_ref()).copied()
+    pub(crate) fn id_for_path(&self, path: &Path) -> Option<FileId> {
+        self.ids.get(&path.to_path_buf()).copied()
```
cmyr, line 335: `to_path_buf` shouldn't be necessary?

fea-rs/src/parse/context.rs
```diff
@@ -431,9 +428,9 @@ mod tests {
         )
         .unwrap();
 
-        let a_id = parse.sources.id_for_path("file_a").unwrap();
-        let b_id = parse.sources.id_for_path("b").unwrap();
-        let c_id = parse.sources.id_for_path("c").unwrap();
+        let a_id = parse.sources.id_for_path(Path::new("file_a")).unwrap();
```
cmyr, line 431: I'd leave the signature of `id_for_path` to take `impl AsRef<Path>`, and then we wouldn't need any of these changes (since `str` impls `AsRef<Path>`).

cmyr (approved):
okay this ended up being more straightforward than I'd imagined, because I'd assumed that `PathBuf` was similarly unavailable in wasm, so that's nice to see.

I think you could simplify the patch by keeping the various `impl AsRef` and `impl Into` bits, and just changing `OsString` to `PathBuf`? It lets us keep our tests cleaner by just using `&str` there.

fea-rs/src/parse/context.rs
rsheeter: Made it so

fea-rs/src/parse/source.rs
rsheeter: Oops, good catch.

Issue #283, opened 2023-05-17
Update iup to remove macOS workaround


rsheeter:
Fonttools always writes gvar tuples for composites glyphs even when deltas are all zeros to support old macOS,
https://github.com/fonttools/fonttools/blob/65bc6105f7aec3478427525d23ddf2e3c8c4b21e/Lib/fontTools/varLib/__init__.py#L296-L314.

https://github.com/fonttools/fonttools/pull/1788 proposes to remove the workaround.

We should implement the workaround, and remove it if/when fonttools does.

PR #1237, opened 2025-02-03
[fontbe] remove workaround for old macOS bug with composites with all-zero deltas

anthrotype:
Now that fonttools has also removed this old workaround, we shall remove it as well in fontc.

See https://github.com/fonttools/fonttools/pull/1788

Fixes https://github.com/googlefonts/fontc/issues/283

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -303,7 +303,7 @@ fn process_composite_deltas(deltas: Vec<Vec2>) -> Vec<GlyphDelta> {
             (0, 0) => GlyphDelta::optional(0, 0),
             (x, y) => GlyphDelta::required(x, y),
         })
-        .collect::<Vec<_>>()
+        .collect()
 }
 
 impl Work<Context, AnyWorkId, Error> for GlyphWork {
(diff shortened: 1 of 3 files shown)


anthrotype:
maybe let's hold this until fonttools actually releases the workaround removal (only merged on main yet).

also, i'd like to conduct an experiment the all VFs on GF library to see how much bytes this will shave off from gvar

fontbe/src/glyphs.rs
```diff
@@ -295,26 +295,15 @@ fn compute_deltas(
...
-    // delta required, to ensure we write an entry for the glyf.
-    // <https://github.com/fonttools/fonttools/blob/0a3360e52727cdefce2e9b28286b074faf99033c/Lib/fontTools/varLib/__init__.py#L351>
-    // <https://github.com/fonttools/fonttools/issues/1381>
-    if deltas.iter().all(|delta| !delta.required) {
-        if let Some(first) = deltas.first_mut() {
-            first.required = true;
-        }
-    }
-    deltas
+        .collect::<Vec<_>>()
```
cmyr, line 306: optional but we no longer need to specify the type here, inference will figure out `collect()`.

cmyr (approved):
makes sense to hold off on fonttools release, but always happy to erase some funny edge case :)

PR #1241, opened 2025-02-04
[ttx_diff] Use same precision locally as with crater

cmyr:
This makes it much clearer to tell whether or not local output is the same as what is displayed on the crater site.

JMM


rsheeter (approved):
LGTM, didn't merge because CI is still running

PR #1244, opened 2025-02-05
Tweak GlyphOrder iteration

cmyr:
::iter() now iterates over both gids and names
::iter_names() iterates over only names

JMM, another little tweak from ongoing work.

### DIFF
--- a/fontbe/src/cmap.rs
+++ b/fontbe/src/cmap.rs
@@ -35,7 +35,7 @@ impl Work<Context, AnyWorkId, Error> for CmapWork {
         let glyph_order = context.ir.glyph_order.get();
 
         let mappings = glyph_order
-            .iter()
+            .names()
             .map(|glyph_name| context.ir.get_glyph(glyph_name.clone()))
             .enumerate()
             .flat_map(|(gid, glyph)| {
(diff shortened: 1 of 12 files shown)


rsheeter (approved):
IMHO what iter does now is more natural, the iteration of what is actually stored, so instead of renaming that iter_names it would make more sense to leave that alone and add a new method to iterate with gids. Don't feel strongly about it so lgtm either way.

cmyr:
My thinking:

- conceptually this is a mapping type, just the keys are implicit (they're the order)
- on mapping types, `iter()` returns (K, V) pairs (`keys()` returns just keys, `values()` returns just values)
- so `iter()` is  the natural name for a fn that returns both (gid, name) here.

The other issue is that if we use `iter()` for just names, there's no obvious name for a method that does both.

rsheeter:
> on mapping types ...

Interesting point. I don't hate the idea of embracing the maplike nature and providing `iter()`, `keys()` and `values()`, that feels more "normal" to me than iter_names. Or iter/gids/names.

> The other issue is that if we use iter() for just names, there's no obvious name for a method that does both.

Thinking aloud, another option might be to return a type with a method so you do something like `.iter().with_gids()` or the like?

cmyr:
or maybe `iter()` for (gid, name) and `names()` for just names? `keys()` and `values()` feels also a bit unclear, because it does imply you can manually set a value for a key, and realistically I don't expect us to want a method that does just gids.

`iter().names()` is interesting but harder to discover and not any easier to type, so I'd probably not go that route.

cmyr:
going to merge this with `iter()` and `names()` because I have other work on top of it I'd like to PR.

PR #1253, opened 2025-02-11
Use new `gvar` coordinates type with upstream intermediates optimisation

Hoolean:
Hello all!

This is the companion pull request for the upstream googlefonts/fontations#1345, after we rehomed the initial attempt to conditionally write `gvar` intermediates (#1225); thank you for your patience.

We will close the older PR in favour of this.

Until we land upstream this is still a draft, so happy to use it to guide that PR in terms of naming, documentation, and any other finishing touches.

### DIFF
--- a/fontbe/src/gvar.rs
+++ b/fontbe/src/gvar.rs
@@ -86,10 +86,7 @@ impl Work<Context, AnyWorkId, Error> for GvarWork {
 mod tests {
     use fontir::ir::GlyphOrder;
     use write_fonts::{
-        tables::{
-            gvar::{GlyphDelta, GlyphDeltas},
-            variations::Tuple,
-        },
+        tables::gvar::{AxisCoordinates, GlyphDelta, GlyphDeltas},
         types::F2Dot14,
     };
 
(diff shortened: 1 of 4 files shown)


cmyr:
okay I've released write-fonts 0.35.1 which includes the fix in fontations, so you should be able to remove the patch section and we can merge this :)

Hoolean:
Nice! Cheers Colin, I've now rebased us to bump instead of patch, and undrafted :)

fontbe/src/gvar.rs
```diff
@@ -110,9 +107,8 @@ mod tests {
                 v if v == glyph_without_var => Vec::new(),
                 // At the maximum extent (normalized pos 1.0) of our axis, add +1, +1
                 v if v == glyph_with_var => vec![GlyphDeltas::new(
-                    Tuple::new(vec![F2Dot14::from_f32(1.0)]),
+                    vec![gvar::Tent::new(F2Dot14::from_f32(1.0), None)],
```
anthrotype, line 110: i'm curious why not just import `Tent` like we already do for GlyphDelta and GlyphDeltas?

anthrotype: oh I see now, because fontir also defines a struct named Tent. (Oh look, that one also has fileds ordered as {min, peak, max}, not {peak, min, max}, but nevermind).

anthrotype (approved):
LGTM! 
GSF is now identical and the gvar table size is also the same! Good job 👍 

the CI failure is expected because this is a PR from a fork which hasn't access to the private token so we can merge regardless.

anthrotype:
maybe try rebase on main (or merge main) to see if the Build GS step is now skipped and CI goes green again

seems like bumping only patch (instead of minor) for write-fonts has broken CI on main as well

Hoolean:
Thanks Cosimo! I've rebased, let's see

PR #1254, opened 2025-02-11
[ttx_diff] Reorder class-based chain sub order

cmyr:
That is, the order in which classes are assigned, which is then also the order of class sets within the lookup.

fontmake uses glyph names but fontc uses glyph ids; this reorders the fontc lookups to match fontmake.

This pattern might occur in a few other lookup types, but I'll handle them one at a time.

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -34,6 +34,7 @@ JSON:
     is the command that was used to run that compiler.
 """
 
+from collections import defaultdict
 from absl import app
 from absl import flags
 from lxml import etree
@@ -45,7 +46,7 @@ import sys
 import os
 from urllib.parse import urlparse
 from cdifflib import CSequenceMatcher as SequenceMatcher
-from typing import Any, Optional, Sequence, Tuple
+from typing import Any, Dict, Optional, Sequence, Tuple
 from glyphsLib import GSFont
 from fontTools.designspaceLib import DesignSpaceDocument
 import time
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -519,9 +520,60 @@ def remove_gdef_lig_caret_and_var_store(ttx: etree.ElementTree):
         if subtable is not None:
             gdef.remove(subtable)
 
+# remap the elements, and return a map of new_ix -> prev_ix
+def remap_class_def_ids(class_def: etree.ElementTree) -> Dict[int, int]:
```
rsheeter, line 524: Nit: neither the name nor the commend is informative wrt what this is meant to do. The match sorting in fonttools is informative, perhaps the name could be closer to that? - `sort_class_def_like_fonttools` as it were

resources/scripts/ttx_diff.py
```diff
@@ -519,9 +520,60 @@ def remove_gdef_lig_caret_and_var_store(ttx: etree.ElementTree):
...
+def reorder_gsub_class_based_rules(ttx: etree.ElementTree):
+    gsub = ttx.find("GSUB")
+    if gsub is None:
+        return
+    for lookup in gsub.xpath(".//Lookup"):
+        chain_ctx = lookup.find("ChainContextSubst")
+        if chain_ctx is None or int(chain_ctx.attrib["Format"]) != 2:
+            continue
+        new_class_order = remap_class_def_ids(chain_ctx.find("InputClassDef"))
+        reorder_rules(chain_ctx, new_class_order, "ChainSubClassSet")
```
rsheeter, line 570: This feels like it wants to be one function; I'm failing to spot why it can't be?

resources/scripts/ttx_diff.py
```diff
@@ -519,9 +520,60 @@ def remove_gdef_lig_caret_and_var_store(ttx: etree.ElementTree):
...
+        lookup.remove(el)
+
+    for ix in range(len(orig_order)):
+        prev_ix = new_order.get(ix, ix)
+        el = orig_order[prev_ix]
+        el.set("index", str(ix))
+        prev_el.addnext(el)
+        prev_el = el
+
+    etree.indent(lookup, level=4)
```
rsheeter, line 556: This feels overcomplicated and it's not completely obvious to me if it's necessary. If we're just sorting all the children of some parent I would expect it to be simpler. If its' more complicated than that perhaps we could add comments to explain?

cmyr: This little patch took my waaay to long, and this logic is much trickier than expected; basically the elements are part of a list of children that includes other elements that we don't want to modify, so we can't just clear all children and then append them in the correct order; we need to insert them in the correct order into the middle of the list. I'll add a comment to try and clarify this..

resources/scripts/ttx_diff.py
cmyr: I expect there will be other lookups where rules are ordered based on class indices, and this logic will be reused.

rsheeter (approved):
Still feels like it might be more complicated than it has to be but I don't want to block on this and I don't have time to try to write a concrete suggestion for how to improve so approved.

PR #1259, opened 2025-02-13
Handle magic insertion marks when merging external features

### LINKED ISSUE #1257: handle "insert markers" when merging in feature code.
That is, if some fea like,

```fea
feature kern {
# Automatic Code
pos a b 10;
} kern;
```

then ufo2ft would insert any code generated for the kern feature _before_ the `kern` block. The logic here is a bit funny, see: https://github.com/googlefonts/ufo2ft/blob/5a606b7884bb6da594e3cc56a169e5c3d5fa267c/Lib/ufo2ft/featureWriters/baseFeatureWriter.py#L193

The other complication here is that ufo2ft is actually generating FEA directly, but _we_ are just generating lookups; however hopefully in practice this is fine, and we will just make sure that we track the last lookup ID that we see before the magic  comment, and then insert our external lookups at that point.

cmyr:
This was about as straight-forward as I could've hoped for, I think, and gives us a modest crater boost.

- closes #1257

### DIFF
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -92,6 +92,12 @@ pub struct CompilationCtx<'a, F: FeatureProvider, V: VariationInfo> {
     conditionset_defs: ConditionSetMap,
     mark_attach_class_id: HashMap<GlyphSet, u16>,
     mark_filter_sets: HashMap<GlyphSet, FilterSetId>,
+    // feature blocks can include `# Automatic Code` comments that instruct us
+    // on where we should merge in code generated from an external provider.
+    // when we encounter these we record what lookup id would be logically next,
+    // and we will use that for the generated lookups.
+    // We also store the start pos of the comment, to break ties.
+    insert_markers: HashMap<Tag, (LookupId, usize)>,
 }
 
 impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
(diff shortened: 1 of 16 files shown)


fea-rs/src/compile/compile_ctx.rs
```diff
@@ -1888,6 +1897,37 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
             }
         } else if item.kind() == Kind::Semi {
             // continue
+        } else if item.kind() == Kind::Comment {
+            assert!(item
+                .token_text()
+                .unwrap_or_default()
+                .contains("# Automatic Code"));
```
rsheeter, line 1904: I don't love the assert, shouldn't we just check if it matches and go about our business if it doesn't? Is it imposible to write some other comment in this position for some reason?

fea-rs/src/compile/compile_ctx.rs
```diff
@@ -1888,6 +1897,37 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
...
+                );
+                return;
+            };
+
+            // make sure we finish any active lookup before assigning the lookupid
+            if let Some((id, _name)) = self.lookups.finish_current() {
+                assert!(
+                    _name.is_none(),
+                    "named lookup should be finished before insert marker"
+                );
```
rsheeter, line 1918: If this is a bad state lets error. If it's not lets warn. Either way lets not assert.

rsheeter (approved):

fea-rs/src/compile/compile_ctx.rs
cmyr: I was thinking this is an invariant but actually this could occur in the wild if someone wrote,

```fea
feature test {
  lookup oops {
    # Automatic Code
  } oops;
} test;
```

that is, they put the comment inside a lookup block inside a feature.

Updated to just error in this case.

PR #1263, opened 2025-02-13
fix Glyphs 2 sources with implicit axes always getting indirect HVAR

### LINKED ISSUE #1256: Indirect HVAR is sometimes built where one without AdvWidthMap would be more efficient
Take e.g. 

python resources/scripts/ttx_diff.py 'https://github.com/Gue3bara/El-Messiri#sources/ElMessiri.glyphs'

fontmake produces a 'direct' HVAR table with implicit variation indices and no AdvanceWidthMap, which is 86 bytes smaller than the indirect one fontc produces for this particular font. The ttx diff for the HVAR is only 3.359%.

The problem in that, first of all this is a Glyphs 2 source and it has no Axes explicitly defined, so it gets the old three default Weight, Width and Custom axes. But the masters only define coordinates for the first, weight axis, so the other two axes get default values and as such they are not variable (we called them elsewhere 'point axes' when all min == default == max). When constructing a VariationModel we filter those out.
In the HVAR builder, we compare each glyph's advance width locations with the global model's locations (the latter only include coords for the variable axes, no point axes) to see if we can reuse the cached global model or we need to create a new (sparse) submodel.
Because the locations including the point axes appear to be different from the global model's location, we end up creating an additional submodel (which is actually the same as the global one) and therefore the HVAR builder (see `is_single_model` method) concludes that we must do an indirect HVAR with multiple sparse VarDatas and an AdvanceWidthMap to map glyph indices to variation indices.
But in fact, if we compared the locations excluding non-variable axes, we would end up with a single model that is suitable to encode with a simpler direct HVAR, which in this case is also more compact and would match 100% the ttx_diff.

I have a fix coming, just working on some tests before I make a PR.

anthrotype:
Fixes #1256 

The first commit reproduces the bug, followed by actual fix

### DIFF
--- a/resources/testdata/glyphs2/WghtVar_ImplicitAxes.glyphs
+++ b/resources/testdata/glyphs2/WghtVar_ImplicitAxes.glyphs
@@ -1,9 +1,5 @@
 {
-.appVersion = "3148";
-DisplayStrings = (
-"-",
-"!"
-);
+.appVersion = "3340";
 customParameters = (
 );
 date = "2022-12-01 04:52:20 +0000";
@@ -12,6 +8,8 @@ fontMaster = (
 {
 alignmentZones = (
 "{800, 16}",
+"{0, -16}",
+"{0, -16}",
 "{0, -16}",
 "{-200, -16}"
 );
@@ -50,7 +48,7 @@ unicode = 0020;
 },
 {
 glyphname = exclam;
-lastChange = "2022-12-01 05:10:49 +0000";
+lastChange = "2025-02-13 12:27:02 +0000";
 layers = (
 {
 layerId = m01;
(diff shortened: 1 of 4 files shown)


anthrotype:
ci failed as expected, the HVAR got an advance width mapping
https://github.com/googlefonts/fontc/pull/1263/commits/a0d4725969755485ce35dd70300544775fa9b013

```
assertion failed: hvar.advance_width_mapping().is_none()
```

resources/testdata/glyphs2/WghtVar_ImplicitAxes.glyphs
```diff
@@ -1,9 +1,5 @@
 {
-.appVersion = "3148";
-DisplayStrings = (
-"-",
-"!"
-);
+.appVersion = "3340";
```
cmyr, line 2: doesn't this mean glyphs3?

cmyr (approved):
looks good, one Q inline :)

resources/testdata/glyphs2/WghtVar_ImplicitAxes.glyphs
anthrotype: no, that's simply saying this particular glyphs2 source file was serialized by Glyphs 3 (3340). you can 'save as'. I don't have a Glyphs 2.app any more.
The .formatVersion key stores the source format version, it's omitted in Glyphs 2 (defaults to 2 in fact).

PR #1270, opened 2025-02-13
[fea-rs] Add ability to filter ttx tests with env var

cmyr:
This functionality existed but only worked for the bundled fonttools tests (for historical reasons) this expands on that so it can be used for all the ttx-based tests, and lets it be controlled via the FEA_FILTER_TESTS environment variable.

Also removes some redundant code, but otherwise no functional change.

JMM

### DIFF
--- a/fea-rs/src/bin/ttx_test.rs
+++ b/fea-rs/src/bin/ttx_test.rs
@@ -12,7 +12,7 @@ fn main() {
     env_logger::init();
     let args = Args::parse();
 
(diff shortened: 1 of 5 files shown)


rsheeter:
Unless this *has* to be an environment variable please use flags. That's basically google-wide guidance for good reason.

rsheeter (approved):
LGTM but lets move to a flag if possible

cmyr:
there's no simple way to pass a flag to `cargo test`...

PR #1271, opened 2025-02-13
[fea-rs] Handle edge case in chain context compilation

cmyr:
The way format 1 subtables work means they don't support input classes, _unless_ those classes contain a single glyph, in which case we can pretend they aren't classes.

### DIFF
--- a/fea-rs/src/common.rs
+++ b/fea-rs/src/common.rs
@@ -88,6 +88,15 @@ impl GlyphOrClass {
         }
     }
 
+    /// If this is a glyph or a class with exactly one, return it.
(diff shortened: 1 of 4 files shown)


fea-rs/src/common.rs
```diff
@@ -88,6 +88,15 @@ impl GlyphOrClass {
         }
     }
 
+    /// If this is a glyph or a class with only a single glyph, return it.
```
rsheeter, line 91: Optional, I suggest s/with only a single/exactly 1/

rsheeter (approved):

PR #1272, opened 2025-02-13
Support panose from instances

rsheeter:
Makes panose match for GSC. JMM if happy.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -1065,6 +1065,7 @@ pub struct Instance {
     pub type_: InstanceType,
     pub axis_mappings: BTreeMap<String, AxisUserToDesignMap>,
     pub axes_values: Vec<OrderedFloat<f64>>,
+    pub custom_parameters: CustomParameters,
 }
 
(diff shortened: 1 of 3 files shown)


glyphs-reader/src/font.rs
```diff
@@ -2590,6 +2593,12 @@ impl Font {
         &self.masters[self.default_master_idx]
     }
 
+    pub fn instance(&self, master: &FontMaster) -> Option<&Instance> {
+        self.instances
+            .iter()
+            .find(|i| i.active && i.type_ == InstanceType::Variable && i.name == master.name)
+    }
+
```
cmyr, line 2601: if this is only being used to find the default instance, maybe just a `default_instance` fn that didn't need to have the master passed in would be easier?

cmyr (approved):
thought, otherwise seems reasonable!

glyphs-reader/src/font.rs
rsheeter: I _think_ there are going to be mroe cases

anthrotype:
technically that's not an "instance" but a "variable font export setting" (they reused the same instance object and tacked on it a 'variable' attribute)
https://handbook.glyphsapp.com/exports/

PR #1278, opened 2025-02-14
Fix duplicate glyphs3 GSMetrics; take the first and ignore 'filter'

### LINKED ISSUE #1269: [glyphs-reader] duplicate GSMetric with 'filter' attribute leading to diffs in OS/2 and MVAR
in Glyhps3, global font metrics can define an optional 'filter' to make them apply only to particular classes of glyphs (e.g. the x-height of the small caps glyphs). For setting OS/2 or MVAR metrics, which is font-wide by definition, these should be ignored (i'm not even sure what they are used for within Glyphs.app itself, maybe hinting?).
However they way we collect these metrics into a BTreeMap keyed by metric name, makes is such that the _last_ metric with a given name wins and overwrite the ones that preceed it.

https://github.com/googlefonts/fontc/blob/51f1bd8bfefc008c292f7dd9485ad0662549be0d/glyphs-reader/src/font.rs#L2526-L2534

E.g. take this Etcetera-Type-Co/Epilogue

```
python resources/scripts/ttx_diff.py 'https://github.com/Etcetera-Type-Co/Epilogue#sources/Epilogue.glyphs' --compare gftools --config ~/.fontc_crater_cache/Etcetera-Type-Co/Epilogue/sources/config.yaml
```

you'll notice the OS/2 and MVAR tables differ from fontmake's. Some masters end up with a different strikeout offset and xheight, because there are GSMetrics with a filter that selects only the x-height of small caps glyphs, which are defined _after_ the default x-height and thus prevails.

In glyphsLib, only the first metric with a given name is actually used, and these `filter` attributes are completely unused.

So it's tempting to simply do this:

```diff
diff --git a/glyphs-reader/src/font.rs b/glyphs-reader/src/font.rs
index f39386de..789248e0 100644
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2525,7 +2525,12 @@ impl TryFrom<RawFont> for Font {
                             metric_names.get(&idx).map(|name| (name.clone(), value))
                         })
                         .filter(|(_, metric)| !metric.is_empty())
-                        .collect(),
+                        .fold(BTreeMap::new(), |mut acc, (name, value)| {
+                            if !acc.contains_key(&name) {
+                                acc.insert(name, value);
+                            }
+                            acc
+                        }),
                     number_values: from
                         .numbers
                         .iter()
```

and then this particular Epilogue.glyphs becomes 'output is identical' in ttx_diff

anthrotype:
GSMetric's `filter` attribute (that restrict a given metric to a group of glyphs, e.g. small caps) is not relevant to build OS/2 and MVAR tables. Some source contains those, normally following the same-named metric which is supposed to apply globally (and thus be stored in font wide metric tables).
glyphsLib does parse these filter attributes but doesn't even use them, when looking up a metric by name it simply returns the first one it finds. Similarly here when we collect metrics by name, we only take the first one we find and ignore any duplicate that may follow.

Fixes #1269  

After the fix, the following gives "output is identical"

```
$ python resources/scripts/ttx_diff.py 'https://github.com/Etcetera-Type-Co/Epilogue#sources/Epilogue.glyphs' --compare gftools --config ~/.fontc_crater_cache/Etcetera-Type-Co/Epilogue/sources/config.yaml
```

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -3586,6 +3586,19 @@ mod tests {
     }
 
     #[test]
+    fn v3_duplicate_metrics_first_wins() {
+        // In this test font, the default master contains two 'x-height' metric values,
+        // the first (501) that applies globally, and a second one (450) that applies
+        // only to small-caps, using GSMetric's `filter` attribute which we ignore as
+        // it is not relevant to build OS/2 and MVAR tables.
+        // We match glyphsLib and only consider the first metric with a given name.
+        let font = Font::load(&glyphs3_dir().join("WghtVar_OS2.glyphs")).unwrap();
+        let master = font.default_master();
+
+        assert_eq!(master.get_metric("x-height"), Some((Some(501.), None)));
+    }
+
+    #[test]
     fn v2_preserve_custom_alignment_zones() {
         let font = Font::load(&glyphs2_dir().join("alignment_zones_v2.glyphs")).unwrap();
         let master = font.default_master();
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2534,7 +2534,13 @@ impl TryFrom<RawFont> for Font {
                             metric_names.get(&idx).map(|name| (name.clone(), value))
                         })
                         .filter(|(_, metric)| !metric.is_empty())
-                        .collect(),
+                        .fold(BTreeMap::new(), |mut acc, (name, value)| {
+                            // only insert a metric if one with the same name hasn't been added
+                            // yet; matches glyphsLib's behavior where the first duplicate wins
+                            // https://github.com/googlefonts/fontc/issues/1269
+                            acc.entry(name).or_insert(value);
+                            acc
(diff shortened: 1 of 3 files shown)


anthrotype:
the test reproducer (https://github.com/googlefonts/fontc/pull/1278/commits/92b89b32c4d8b188f522853bae8012318cff1466) fails as expected (the last duplicate, the one with the 'filter', prevails)

```diff
---- font::tests::v3_duplicate_metrics_first_wins stdout ----
thread 'font::tests::v3_duplicate_metrics_first_wins' panicked at glyphs-reader/src/font.rs:3598:9:
assertion failed: `(left == right)`

Diff < left / right > :
 Some(
     (
         Some(
<            450.0,
>            501.0,
         ),
         None,
     ),
 )
```

https://github.com/googlefonts/fontc/pull/1278/commits/92b89b32c4d8b188f522853bae8012318cff1466

cmyr (approved):
looks good!

PR #1279, opened 2025-02-14
Tidy glyphs files a touch

rsheeter:
In poking at some of our .glyphs files I notice we have files with content no test needs, lets try to make them slightly more minimal. Manual editing can go much further but just deleting some stuff programmatically is a start.


cmyr:
A bunch of those fields are going to be added back automatically after editing, do we want a check in CI to prevent that?

rsheeter:
> A bunch of those fields are going to be added back automatically after editing, do we want a check in CI to prevent that?

Makes sense but I suggest we do that in a followon

cmyr (approved):
👍

PR #1284, opened 2025-02-14
Add a simple font that fontmake compiles to COLRv1

rsheeter:
Realized from crater that fontmake is actively producing COLRv1 fonts from .glyphs sources and created a simple example of linear and radial gradients:

image

### DIFF
--- a/resources/testdata/glyphs3/COLRv1-simple.glyphs
+++ b/resources/testdata/glyphs3/COLRv1-simple.glyphs
@@ -1,0 +1,963 @@
+{
+.appVersion = "3343";
+.formatVersion = 3;
+familyName = "New Font";
+fontMaster = (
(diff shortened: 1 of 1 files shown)


anthrotype:
> Realized from crater that fontmake is actively producing COLRv1 fonts from .glyphs sources

yeah, glyphsLib translates Glyphs.app's own representation into the format that ufo2ft and fontTools.colorLib expect (basically a json-like serialization of COLRv1 otData structures stored in the UFO lib).

https://github.com/googlefonts/glyphsLib/blob/main/Lib/glyphsLib/builder/color_layers.py

anthrotype:
and not just COLRv1, but also COLRv0, which Glyphs.app calls "color palette layers"

PR #1285, opened 2025-02-15
Parse basic linear/radial gradient specifications from .glyphs

rsheeter:
Read the example provided in #1284

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -316,13 +316,15 @@ impl Layer {
 #[derive(Clone, Default, Debug, PartialEq, Hash)]
 pub struct LayerAttributes {
     pub coordinates: Vec<OrderedFloat<f64>>,
-    // TODO: add axisRules, color, etc.
+    pub color: bool,
+    // TODO: add axisRules, etc.
 }
 
 // hand-parse because they can take multiple shapes
 impl FromPlist for LayerAttributes {
     fn parse(tokenizer: &mut Tokenizer<'_>) -> Result<Self, crate::plist::Error> {
         let mut coordinates = Vec::new();
+        let mut color = false;
 
         tokenizer.eat(b'{')?;
 
@@ -334,9 +336,8 @@ impl FromPlist for LayerAttributes {
             let key: String = tokenizer.parse()?;
             tokenizer.eat(b'=')?;
             match key.as_str() {
-                "coordinates" => {
-                    coordinates = tokenizer.parse()?;
-                }
+                "coordinates" => coordinates = tokenizer.parse()?,
+                "color" => color = tokenizer.parse()?,
                 // skip unsupported attributes for now
                 // TODO: match the others
                 _ => tokenizer.skip_rec()?,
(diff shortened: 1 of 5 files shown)


glyphs-reader/src/font.rs
```diff
@@ -333,24 +335,84 @@ impl FromPlist for LayerAttributes {
...
+    style: String,
+}
+
+#[derive(Clone, Default, Debug, PartialEq, Eq, Hash)]
+pub struct Color {
+    pub r: i64,
+    pub g: i64,
+    pub b: i64,
+    pub a: i64,
+    pub n: i64, // 0/1 for start/end
```
anthrotype, line 371: I think this would be better named `stop_offset`?

glyphs-reader/src/font.rs
```diff
@@ -333,24 +335,84 @@ impl FromPlist for LayerAttributes {
...
+        })
     }
 }
 
 #[derive(Debug, Clone, PartialEq, Hash)]
 pub enum Shape {
-    Path(Path),
-    Component(Component),
+    Path(Path, ShapeAttributes),
+    Component(Component, ShapeAttributes),
```
anthrotype, line 405: why did you add ShapeAttributes to the enum values instead of as actual fileds of Component and Path structs?

in the glyphs format both Component and Path have attributes directly, and glyphsLib also models them like that
https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md

cmyr: Agreed that this probably makes a bit more sense; it means that you can extract the inner type without having to discard the attributes, which might be useful at some point?

glyphs-reader/src/font.rs
cmyr: And are the only valid values `0` and `1`? Would it make more sense as an enum?

glyphs-reader/src/font.rs
```diff
@@ -333,24 +335,84 @@ impl FromPlist for LayerAttributes {
             let key: String = tokenizer.parse()?;
             tokenizer.eat(b'=')?;
             match key.as_str() {
-                "coordinates" => {
-                    coordinates = tokenizer.parse()?;
-                }
+                "coordinates" => coordinates = tokenizer.parse()?,
+                "color" => color = tokenizer.parse::<i64>()? == 1,
```
cmyr, line 339: you can just do `tokenizer.parse()?`, `FromPlist` is impl'd for `bool`.

glyphs-reader/src/font.rs
anthrotype: no no, it's not an enum, and neither it is an integer, it's a float that ranges from 0.0 (start) to 1.0 (end) and corresponds to COLRv1 StopOffset (along the 'color line'); if Rod only encountered whole 0 and 1 it's because he did not add intermediate stops

rsheeter: That's right, this PR was only really meant to parse the one example I have so far and then incrementally grow more capable over time. That said, I have updated to stop_offset: f64.

glyphs-reader/src/font.rs
```diff
@@ -304,7 +304,7 @@ impl Layer {
 
     pub(crate) fn components(&self) -> impl Iterator<Item = &Component> + '_ {
         self.shapes.iter().filter_map(|shape| match shape {
-            Shape::Path(_) => None,
+            Shape::Path(..) => None,
```
cmyr, line 307: this change is left over from an earlier version?

cmyr (approved):
looks good with two little nits inline.

glyphs-reader/src/font.rs
```diff
@@ -333,17 +335,67 @@ impl FromPlist for LayerAttributes {
...
+    style: String,
+}
+
+#[derive(Clone, Default, Debug, PartialEq, Eq, Hash)]
+pub struct Color {
+    pub r: i64,
+    pub g: i64,
+    pub b: i64,
+    pub a: i64,
+    pub stop_offset: OrderedFloat<f64>, // 0/1 for start/end
```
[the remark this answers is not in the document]
rsheeter: My bad, added a reference that I think will help. Unfortunatley it rapidly needs a paragraph of explanation so I think the link is the best path.

glyphs-reader/src/font.rs
rsheeter: Ty, fixed

PR #1286, opened 2025-02-19
[fea-rs] Better merging of contextual ligature rules

cmyr:
- Improves the logic for determining when we can reuse a lookup for an inline ligature sub rule, which will let us be much more aggressive in reusing these lookups,
- And adds logic to skip duplicate identical rules in ligature substitution tables.


dfrg (approved):
LGTM!

PR #1289, opened 2025-02-20
[features] Share more code, do less work

cmyr:
In order to compile both marks & kern we need to have already compiled the GSUB table. To avoid doing this repeatedly, we now compile it immediately after parsing the AST, and then make it available to the mark & kern feature writers.

We also similarly store the explicit GDEF classes, if they were in the FEA, and try to share more code between mark & kern generation in general.

This also includes a commit that makes us convert glyph names to glyph ids earlier in the marks compilation process, which similarly helps us share more code with kerning.

(all of this is motivated by work on abvm and blwm)

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -16,8 +16,8 @@ use ordered_float::OrderedFloat;
 
 use fea_rs::{
     compile::{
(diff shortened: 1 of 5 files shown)


dfrg (approved):
I'm not very familiar with the details of fea compilation but the changes here make sense to me and the code looks good.

Issue #945, opened 2024-09-06
head.checkSumAdjustment not set, always 0x0


anthrotype:
fontc is not computing the [head](https://learn.microsoft.com/en-us/typography/opentype/spec/head) table's `checkSumAdjustment` field, so it defaults to 0x0.
And neither is write_fonts. Not sure whose responsibility it should be, whether the client (fontc) or write_fonts itself.
Probably the latter?

Anyway, ttx_diff.py currently erases the field altogether so we may get "fonts are identical" but actually the checkSumAdjustment is plain wrong.

Ok, you may argue it's garbage anyway and nobody should trust that, but if we want to be nice we should just compute that correctly, e.g. the same way fontmake/fonttools would, with all the "BiboAfba" and things like that:

https://github.com/fonttools/fonttools/blob/c2119229cfb02cdb7c5a63374ef29d3d514259e8/Lib/fontTools/ttLib/sfnt.py#L375-L406

rsheeter:
https://github.com/googlefonts/fontations/blob/6d01f7425b52f06d23d8e24f433f09d2e3779ac5/write-fonts/src/font_builder.rs#L102 is computing table directory checksums and strikes me as the appropriate place to do it, update head if there is a head.

anthrotype:
sgtm. we could transfer the issue over to the fontations repo maybe

rsheeter:
Maybe open a new one there and reference it? - it's still true that fontc has a bug so I think we shoudl keep this one even if the fix ends up just being an update to our fontations version.

Issue #1268, opened 2025-02-13
Sort tables in the recommended OT order


rsheeter:
See https://github.com/googlefonts/fontations/issues/1357. Adding here so it shows up as required for the 1.0 milestone.

PR #1293, opened 2025-02-24
bump write-fonts 0.36.1: computes checksum_adjustment and sorts tables as recommended

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/945 and https://github.com/googlefonts/fontc/issues/1268

maybe since this was a patch (compatible) release, cargo should pick it up in any case, but I suppose it doesn't hurt if we explicitly do bump

### DIFF
--- a/fea-rs/src/util/ttx.rs
+++ b/fea-rs/src/util/ttx.rs
@@ -439,6 +439,8 @@ fn rewrite_ttx(input: &str) -> String {
     for line in input.lines() {
         if line.starts_with("<ttFont") {
             out.push_str("<ttFont>\n");
+        } else if line.starts_with("    <checkSumAdjustment value=\"0x") {
+            out.push_str("    <checkSumAdjustment value=\"0x0\"/>\n");
         } else {
             out.push_str(line);
             out.push('\n')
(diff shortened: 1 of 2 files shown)


fea-rs/src/util/ttx.rs
```diff
@@ -439,6 +439,8 @@ fn rewrite_ttx(input: &str) -> String {
     for line in input.lines() {
         if line.starts_with("<ttFont") {
             out.push_str("<ttFont>\n");
+        } else if line.starts_with("    <checkSumAdjustment value=\"0x") {
+            out.push_str("    <checkSumAdjustment value=\"0x0\"/>\n");
```
cmyr, line 443: that seems like a reasonable approach.

cmyr (approved):
👍

anthrotype:
actually, a couple of fea-rs fonttools_tests were failing because we know do compute the checksum_adjustment whereas we didn't before:

https://github.com/googlefonts/fontc/actions/runs/13501572022/job/37721051261

the expected. ttx files are copied from fonttools feaLib as is, without modifications, I can't simply edit them to match the computed checksum, so it's preferable to ignore that field when comparing ttx files
https://github.com/googlefonts/fontc/pull/1293/commits/92e1bc6119737a22c5b0574f9efd72f40e568be5

cmyr:
yes I agree, I think your approach is correct!

PR #1295, opened 2025-02-24
[glyphs] Improve parsing of escape sequences

cmyr:
This will now handle unicode escapes, including surrogate pairs. Will error on unpaired surrogate pairs.

In the process of this work I noticed a few other omissions related to string parsing; the major one is that the reference impl supports the use of single quotes to delimit strings. This doesn't seem to be in any of our sources, but it might be worth fixing now?

### DIFF
--- a/glyphs-reader/src/plist.rs
+++ b/glyphs-reader/src/plist.rs
@@ -36,6 +36,8 @@ pub enum Error {
     BadData,
     #[error("Unknown escape code")]
     UnknownEscape,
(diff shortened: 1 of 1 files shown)


rsheeter (approved):

anthrotype (approved):

anthrotype:
> This doesn't seem to be in any of our sources, but it might be worth fixing now?

Maybe file an issue. If no source used that, it's probably low priority

PR #1297, opened 2025-02-26
Generate abvm & blwm features

### LINKED ISSUE #1086: handle abvm/blwm for indic scripts when generating mark features
For some "indic" scripts, ufo2ft markFeatureWriters generates [`abvm`](https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae#tag-abvm) (above mark) and [`blwm`](https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae#tag-blwm) (below mark) features instead of `mark`.

The list of scripts is here:
https://github.com/googlefonts/ufo2ft/blob/d15ef8f6f235804733c6e2ea5f01f39aafb92f45/Lib/ufo2ft/featureWriters/markFeatureWriter.py#L337

To decide whether an anchor should go to abvm or blmw there's a hard-coded list of anchor names that are assigned "below" status, the rest are assumed to be "above", see:
https://github.com/googlefonts/ufo2ft/blob/d15ef8f6f235804733c6e2ea5f01f39aafb92f45/Lib/ufo2ft/featureWriters/markFeatureWriter.py#L948-L961

This is similar to #753, where the kernFeatureWriters choose the 'dist' tag instead of 'kern' for some Indic scripts.

One affected project that I stumbled upon is https://github.com/notofonts/takri
But I expect all projects containing one of the indic scripts to be also affected.

cmyr:
This was one of the big things that was missing from layout, and gives us a healthy bump on crater.

- closes #1086

### DIFF
--- a/fontbe/src/features/marks.rs
+++ b/fontbe/src/features/marks.rs
@@ -605,7 +605,7 @@ impl<'a> MarkLookupBuilder<'a> {
             // no abvm scripts: everything is a mark
             return Ok((
                 Default::default(),
-                self.char_map.values().copied().collect(),
+                self.glyph_order.iter().map(|(gid, _)| gid).collect(),
             ));
         }
 
@@ -730,7 +730,7 @@ fn resolve_anchor(
             resolve_anchor_once(anchor, static_metadata).map(BaseOrLigAnchors::Base)
         }
         BaseOrLigAnchors::Ligature(anchors) => anchors
-            .into_iter()
+            .iter()
             .map(|a| {
                 a.map(|a| resolve_anchor_once(a, static_metadata))
                     .transpose()
@@ -963,19 +963,14 @@ mod tests {
             ("ka-kannada", '\u{0C95}'),
         ];
 
-        static UNMAPPED: &[&str] = &["ka-kannada.base"];
+        static UNMAPPED: &[&str] = &["ka-kannada.base", "a.alt"];
 
         let c = agl::char_for_agl_name(name.as_str()).or_else(|| {
             MANUAL
                 .iter()
                 .find_map(|(n, uv)| (name.as_str() == *n).then_some(*uv))
         });
-        if c.is_none()
-            && UNMAPPED
-                .iter()
-                .find(|except| **except == name.as_str())
-                .is_none()
-        {
+        if c.is_none() && !UNMAPPED.iter().any(|except| *except == name.as_str()) {
             panic!("please add a manual charmap entry for '{name}'");
         }
         c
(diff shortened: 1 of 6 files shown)


fontbe/src/features/marks.rs
```diff
@@ -462,6 +567,80 @@ impl<'a> MarkLookupBuilder<'a> {
...
+        )?;
+
+        let mut non_abvm_glyphs = super::properties::glyphs_matching_predicate(
+            &self.char_map,
+            unicode_is_non_abvm,
+            gsub.as_ref(),
+        )?;
+        // https://github.com/googlefonts/ufo2ft/blob/5a606b7884bb6da/Lib/ufo2ft/featureWriters/markFeatureWriter.py#L1156
+        // I do not understand why we have all this fancy logic when it feels like
+        // 'not abvm' is implicit?
```
anthrotype, line 626: because https://github.com/googlefonts/ufo2ft/blob/16ed156bd6a8b9bc035d0aa8045a1271ef79a52e/Lib/ufo2ft/featureWriters/markFeatureWriter.py#L1148-L1150

fontbe/src/features/marks.rs
```diff
@@ -1069,6 +1334,192 @@ mod tests {
...
+                anchors
+                    .add("top", [(0, 810)])
+                    .add("top.a", [(0, 810)])
+                    .add("_top.a", [(0, 578)]);
+            })
+            .get_normalized_output();
+
+        assert_eq_ignoring_ws!(
+            out,
+            // this gets an abvm feature because we don't specify any languagesystems
```
anthrotype, line 1379: I don't understand this, why would this get "an abvm feature because we don't specify any languagesystems"?

fontbe/src/features/marks.rs
```diff
@@ -1069,6 +1334,192 @@ mod tests {
...
+
+                "#
+        );
+    }
+
+    #[test]
+    fn mark_on_gsub_glyph() {
+        let mut out = MarksInput::default();
+        let out = out
+            // a.alt is only reachable via substitution
```
anthrotype, line 1492: I don't think this is testing the gsub closure, the same mark feature would be generated even without user FEA snippet sub a by a.alt below. Maybe you wanted to test `abvm_on_gsub_glyph` instead?

fontbe/src/features/marks.rs
anthrotype: oh I see.. because this font doesn't have explicit languagesystems declared in the FEA, which would filter the scripts actually used, and because acutecomb's script extensions include 'Tale' (Tai Le) script and macroncomb's script extensions includ 'Tfng' (Tifinagh) script, they get slotted in the 'abvm' feature... Very edge case, tbh. More confusing than it is helpful. But I suppose this matches ufo2ft so, ok.

anthrotype (approved):

fontbe/src/features/marks.rs
cmyr: I was also confused by this, but it matches fonttools. Basically because of this: https://github.com/googlefonts/ufo2ft/blob/16ed156bd6a8b9bc035d0aa8045a1271ef79a52e/Lib/ufo2ft/featureWriters/markFeatureWriter.py#L1122

if there are no feaScripts declared, we don't modify the `scriptsUsingAbvm` value, and then several of these combining accents are used in abvm scripts.

fontbe/src/features/marks.rs
cmyr: I need to explain this a bit better; this is checking is that if we _don't_ have any abvm scripts, we put all glyphs into the 'non abvm' bucket, even if they are only reachable via substitution. (This exists because I had a bug where I wasn't handing this correctly)

(will add a better explanation)

PR #1300, opened 2025-02-26
[ttx_diff] Ignore lookup indices of normalized lookups

cmyr:
More specifically, for lookups that are generated via feature writers in python (mark/kern). There are differences between how fontc and ufo2ft handle generating these features, and it would add unnecessary complexity to try and make either match the other, so I think it's simpler to just ignore them in the diff.

JMM

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -490,7 +490,9 @@ def remove_mark_and_kern_lookups(ttx):
     # './/Lookup' xpath selects all the 'Lookup' elements that are descendants of
     # the current 'GPOS' node - no matter where they are under it.
     # Most importantly, this _excludes_ GSUB lookups, which shouldn't be pruned.
+    removed_indices = set()
     for lookup in gpos.xpath(".//Lookup"):
+        lookup_index = lookup.attrib["index"]
         lookup_type_el = lookup.find("LookupType")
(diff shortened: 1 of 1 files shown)


anthrotype:
the order of the Lookup list does matter though.. A feature's LookupListIndex doesn't, as they will be sorted by index in the global Lookup list before application. Are we sure this won't hide potential differences in the Lookup order?

resources/scripts/ttx_diff.py
```diff
@@ -508,6 +510,15 @@ def remove_mark_and_kern_lookups(ttx):
             lookup.remove(child)
         if is_extension:
             lookup_type_el.attrib["value"] = str(lookup_type)
+        removed_indices.add(lookup_index)
+
+    # finally go and replace any of the removed indices where they are referenced
+    # in features; our lookup order won't always match fonttools for mark/kern
+    # because of the particulars of the feature writers, so we just ignore this
```
rsheeter, line 517: This should address why this isn't hiding problems. Do we compare the normalized mark/kern data so this doesn't matter? - then this should explicitly say so

cmyr: yes, these lookups are all normalized and normalization accounts for lookup order. Will clarify the comment.

anthrotype (approved):

cmyr:
I'm going to hold this for now, it isn't really solving all my problems anyway..

PR #1301, opened 2025-02-26
blacken glyphs-reader/data/update.py

anthrotype:
my vscode likes to run black (ruff) autoformatter automatically on save, which skews diffs since I want to touch this file

JMM


anthrotype:
not sure who summoned @gemini-code-assist, but it wasn't me...

PR #1305, opened 2025-02-27
Better ordering of generated features

cmyr:
So... it turns out that this is extremely fiddly and complicated, and this patch isn't getting it completely correct; but I think it's worth checkpointing this, because I think the next step is a larger refactor, and I'll want to include some of this work in that.

So the problem we're trying to solve, here, is how best to match fonttools lookup ordering for lookups that are generated by 'feature writers'.

This is made more complicated by the fact that fonttools is generating FEA which it passes off to the feature compiler, but fontc is generating feature & lookup types directly, and then trying to merge them into the compiled FEA in such a way as to replicate the lookup ordering used in fonttools.

Some things that make this problem tricky:
- some lookups are declared inside a feature block, and so their IDs depend on where that feature block gets inserted (`abvm`, `blwm`).
- some features are prepended before the first feature generated by the feature writer, and so depend on the position of other features (`mark` always precedes `mkmk`, `kern` precedes `dist`)
- within these feature groups, the ordering of lookups depends on which feature writer generated them: mark lookups always precede mkmk lookups; kern and dist lookups are ordered based on the unicode shortname of the script they are assigned to.

All of this adds up to... a lot of complexity. I do now tentatively believe I know how to resolve this, but it's going to require some followup. (In particular I want to move this logic out of a single method on `AllLookups` and into its own file, maybe using a context or similar, so that I can split things up and try to make it a bit more intelligible).

In any case, this patch _is_ an improvement, and gives us a modest crater boost.

### DIFF
--- a/fea-rs/src/compile/lookups.rs
+++ b/fea-rs/src/compile/lookups.rs
@@ -746,11 +746,12 @@ impl AllLookups {
         let mut insert_markers = insert_markers.to_owned();
 
         // Also: in fonttools, dist/kern are added before marks, so let's do
-        // that first
-        if let Some((kern_id, kern_pos)) = insert_markers.get(&KERN).copied() {
+        // that first; and kern has to go before dist, if both is present
+        // https://github.com/googlefonts/ufo2ft/blob/16ed156bd6a8b9bc035d0/Lib/ufo2ft/featureWriters/kernFeatureWriter.py#L311
+        if let Some((kern_id, kern_pos)) = insert_markers.get(&DIST).copied() {
             // if kern has a marker but dist doesn't, insert dist first
             insert_markers
-                .entry(DIST)
+                .entry(KERN)
                 .or_insert((kern_id, kern_pos - 1));
         }
 
@@ -779,7 +780,7 @@ impl AllLookups {
         // for adding extras, mimic the order ufo2ft uses, which is alphabetical
         // but grouped by feature writer. We add markers for all features,
         // even if they dont exist in the font; we'll ignore those later
-        for feature in [CURS, DIST, KERN, ABVM, BLWM, MARK, MKMK] {
+        for feature in [CURS, KERN, DIST, ABVM, BLWM, MARK, MKMK] {
             insert_markers.entry(feature).or_insert_with(|| {
                 next_pos += 1;
                 (last_id, next_pos)
@@ -1543,7 +1544,7 @@ mod tests {
 
         assert_eq!(
             feature_order_for_test(&map, &features),
-            [CURS, DIST, KERN, ABVM, BLWM, MARK, MKMK]
+            [CURS, KERN, DIST, ABVM, BLWM, MARK, MKMK]
         );
     }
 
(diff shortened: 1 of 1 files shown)


rsheeter:
To ask the obvious, why can't we solve this with normalization?

fea-rs/src/compile/lookups.rs
```diff
@@ -724,41 +733,89 @@ impl AllLookups {
...
+
+        // Also: in fonttools, dist/kern are added before marks, so let's do
+        // that first; and kern has to go before dist, if both is present
+        // https://github.com/googlefonts/ufo2ft/blob/16ed156bd6a8b9bc035d0/Lib/ufo2ft/featureWriters/kernFeatureWriter.py#L311
+        if let Some((kern_id, kern_pos)) = insert_markers.get(&DIST).copied() {
+            // if kern has a marker but dist doesn't, insert dist first
+            insert_markers
+                .entry(KERN)
+                .or_insert((kern_id, kern_pos - 1));
+        }
```
anthrotype, line 756: I think the variable names and comments are confusing here.
I read this code as meaning "if dist has a marker but kern doesn't, insert dist first" (not the other way around). Also `kern_id` and `kern_pos` should be more logically named `dist_id` and `dist_pos`, no?

anthrotype: now I remembered that the ufo2ft kernFeatureWriter defines one set of lookups and then register them _all_ to either kern or dist or both, depeding on which script is "dist-enabled" or not. Those lookups are defined as standalone, outside feature blocks. So what matters is the order in which these get _defined_ explicitly (as standalone lookups), and not the order in which the feature blocks (that may contain inlined, implicit lookup definitions as well as named lookups, but in this particular case they do not). There's no ordering dependency between kern and dist features because, even if both are present, they will share exactly the same lookups (which will be applied in the order defined by the Lookup list); the order of the feature does not matter at all in this sense, from a functional POV. This is even before considering that pos adjustments are additive (commutative).

In the case of mark/mkmk, the ufo2ft feature writer doesn't define standalone lookups outside of the feature blocks and then references those from within the features; it instead declares them inside the features so the order of the feature definitions does matter, because when they occur within a feature block, either implicit lookups or even named ones will be as if they are defined and referenced at the same time.
Because we want mark lookups to be applied before mkmk (for there is a logical dependency there, with anchor positioning the last wins), and the feature writer defines those lookups within the feature block itself, then the order in which _features_ are defined in the FEA matters for abvm/blwm/mark/mkmk.

cmyr: yea so I noticed the thing with kern/dist generating lookups that are outside the feature blocks, and that's sort of what's being handled here; we don't really need to distinguish between kern/dist lookups, we group them together and put them before whichever of kern/dist comes first. (the order of the lookups is determined by the code that generates the lookups, before we get to this stage.)


I thought that mark/mkmk also generated standalone lookups, but abvm/blwm had lookups inside the feature blocks? But I see that you're right, those lookups are inlined too.

In any case this code is all going to be rewritten in a followup, so we can discuss in more detail there.

Issue #889, opened 2024-07-30
respect 'Propagate Anchors' custom parameter


anthrotype:
A font source can contain a "Propagate Anchors" custom parameter set to false to disable the feature. fontmake respects this, and fontc should too

![image](https://github.com/user-attachments/assets/9679b3a8-e08c-4b7b-8b97-556d6d1b0f88)

anthrotype:
this was fixed by https://github.com/googlefonts/fontc/issues/889

PR #1308, opened 2025-02-28
Respect 'Propagate Anchors' custom parameter

felipesanches:
I've tested this by running ttx_diff on **notofonts/arabic/sources/NotoSansArabic.glyphspackage** and got these improvements:

|   COMPARISON  | Before  |   After   |
| ------------- | ------- | --------- |
| (mark / kern) | 29.249% | Identical |
|     GDEF      | 56.062% | Identical |
|     GPOS      | 87.500% | Identical |
|   ligcaret    | 34.043% | Identical |

(issue #889)

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -68,6 +68,7 @@ pub struct Font {
 /// Custom parameter options that can be set on a glyphs font
 #[derive(Clone, Debug, PartialEq, Hash, Default)]
 pub struct CustomParameters {
+    pub propagate_anchors: Option<bool>,
     pub use_typo_metrics: Option<bool>,
     pub is_fixed_pitch: Option<bool>,
     pub fs_type: Option<u16>,
@@ -598,6 +599,7 @@ impl RawCustomParameters {
                 continue;
             }
             match name.as_str() {
+                "Propagate Anchors" => add_and_report_issues!(propagate_anchors, Plist::as_bool),
                 "Use Typo Metrics" => add_and_report_issues!(use_typo_metrics, Plist::as_bool),
                 "isFixedPitch" => add_and_report_issues!(is_fixed_pitch, Plist::as_bool),
                 "Has WWS Names" => add_and_report_issues!(has_wws_names, Plist::as_bool),
@@ -2584,7 +2586,11 @@ fn preprocess_unparsed_plist(s: &str) -> Cow<str> {
 impl Font {
     pub fn load(glyphs_file: &path::Path) -> Result<Font, Error> {
         let mut font = Self::load_raw(glyphs_file)?;
-        font.propagate_all_anchors();
+
+        // propagate anchors by default unless explicitly set to false
+        if font.custom_parameters.propagate_anchors.unwrap_or(true) {
+            font.propagate_all_anchors();
+        }
         Ok(font)
     }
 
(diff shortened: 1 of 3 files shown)


glyphs-reader/src/font.rs
```diff
@@ -2584,7 +2586,11 @@ fn preprocess_unparsed_plist(s: &str) -> Cow<str> {
 impl Font {
     pub fn load(glyphs_file: &path::Path) -> Result<Font, Error> {
         let mut font = Self::load_raw(glyphs_file)?;
-        font.propagate_all_anchors();
+        if font.custom_parameters.propagate_anchors.is_none()
+            || font.custom_parameters.propagate_anchors == Some(true)
+        {
+            font.propagate_all_anchors();
+        }
```
felipesanches, line 2593: Here I used "is none or true" because these are the two cases where it should propagate anchors:
* When the custom param is missing (defaults to true)
* Or when it is explicitly true

But I think it may be simpler to simply do `if font.custom_parameters.propagate_anchors != Some(false)` instead...

glyphs-reader/src/font.rs
```diff
@@ -2584,7 +2586,11 @@ fn preprocess_unparsed_plist(s: &str) -> Cow<str> {
 impl Font {
     pub fn load(glyphs_file: &path::Path) -> Result<Font, Error> {
         let mut font = Self::load_raw(glyphs_file)?;
-        font.propagate_all_anchors();
+        if font.custom_parameters.propagate_anchors.is_none()
+            || font.custom_parameters.propagate_anchors == Some(true)
```
anthrotype, line 2590: or you could do this even more concisely, while preserving the same logic

```suggestion
        // propagate anchors by default unless explicitly set to false
        if font.custom_parameters.propagate_anchors.unwrap_or(true) {
```

it unwraps the Option to get the bool or returns true if there is None

anthrotype:
thanks. Could you also add a test?

you could create a minimal test font in Glyphs.app with two glyphs, a simple one with an anchor, and a composite glyph using the latter as component. You also add the Propagate Anchors custom param set to false. Pass it through the resources/scripts/tidy_glyphs.py python script before committing to get rid of unwated stuff (timestamps etc.).

then in your test you would simply Font::load the test file and assert that the composite glyph gets no propagated anchors

felipesanches:
@anthrotype, I have attempted to create the test, but I am still not getting good results. Can you help me understand what I'm doing wrong here?

felipesanches:
I'm getting this panic:

```
thread 'propagate_anchors::tests::dont_propagate_anchors' panicked at glyphs-reader/src/propagate_anchors.rs:1038:9:
assertion failed: glyph.layers.first().unwrap().anchors.is_empty()
```

My reasoning here is that:
* I know that the glyph "Aacute" exists
* I know that it has a single layer, so `.layers.first().unwrap()` should be fine
* I expect it not to have anchors and since `.anchors` is a `Vec<Anchor>`, then it should be fine to use `.is_empty()`

But I must have missed something, because it panics... What do you think?

cmyr:
the panic is just saying that the test failed, and it failed because apparently `anchors` isn't empty, so something unexpected is going on..

cmyr:
okay the issue seems to be that the "Propagate Anchors" custom param is set on the _master_, not on the _font_?

felipesanches:
> okay the issue seems to be that the "Propagate Anchors" custom param is set on the _master_, not on the _font_?

Yep! Thanks! Just moved it and the check now passes.

cmyr (approved):
looks good, thanks!

PR #1310, opened 2025-02-28
[fea-rs] Replicate feature writer insert logic

cmyr:
This reworks how we merge external lookups to try and precisely model the logic of the ufo2ft FeatureWriter system.

This tries to simplify the flow by handling the features in groups, reflecting the different feature writers: we do `curs` (which is trivial) and then we do `dist` and `kern`, and then we do `abvm`, `blwm`, `mark`, and `mkmk`. Overall I think this is much easier to follow.

### DIFF
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -33,7 +33,7 @@ use crate::{
 };
 
 use super::{
-    feature_writer::{FeatureBuilder, FeatureProvider},
+    feature_writer::{FeatureBuilder, FeatureProvider, InsertionPoint},
     features::{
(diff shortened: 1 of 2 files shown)


cmyr:
okay so I can reproduce the sanitizer failure locally and am really not sure what's going on there, will need to investigate on monday.

rsheeter (approved):
If I assume that we need to match python this closely, LGTM so approving.

That said, I remain concerned we're leaning too far into matching python here. For example, if our kerning structure is identical other than lookup ids (mine are 4,5,6, yours are 7,8,9, contents are the same, behavior is the same) we should be able to normalize that away. Naively perhaps we would resequence lookups from raw #s to some form of type.nth-of-type pattern.

Issue #780, opened 2024-04-18
Handle production names in glyph-reader::glyphdata


cmyr:
This doesn't feel urgent, but in porting the python [glyphdata](https://github.com/googlefonts/glyphsLib/blob/main/Lib/glyphsLib/glyphdata.py#L199) module I skipped everything related to production glyph names, in particular `_construct_production_name` but also looking up glyph info by production name.

This doesn't feel particularly urgent, but might be worth doing at some point.

anthrotype:
I found one source "Konkhmer Sleokchher.glyphs" from fontc_crater's list that reproduces this (repro command below).
The source contains glyphs named in AGLFN style (uniXXXX).
A couple of glyphs ("uni17BF.b" and "uni17C0.b") are mis-classified by fontc as spacing combining marks, and their advance width is zeroed, leading to diffs in GDEF/hmtx/OS/2.
In glyphsLib/fontmake, these two glyphs get the same category/subCategory as the base glyph without suffix, "uni17BF" and "uni17C0", which are looked up by 'production name' in the GlyphData by glyphsLib and matched with the entries named respectively "yaSign-khmer" and "ieSign-khmer"; these happen to be categorized as "Letter" (not "Mark"), and thus fontmake does not zero-out the advance width.

Because fontc is not looking them up by production name, but only by glyph name or unicode (in this case they are alternate glyphs without a unicode codepoint of their own), and nothing can be found, the fallback code `construct_category_via_agl` gets triggered and ICU gives a general category of "Mc" (Spacing Mark) which in fact translates to Mark/Spacing Combining using Glyphs.app terminology, so in theory these _are_ marks according to their base codepoint, but GlyphData.xml categorises them as Letter... 

```
python resources/scripts/ttx_diff.py 'https://github.com/suonmaysophanith7/KonKhmer_SleokChher#sources/Konkhmer Sleokchher.glyphs'
```

```diff
--- /Users/clupo/oss/fontc/build/default/fontc.GDEF.ttx	2025-02-21 15:18:57
+++ /Users/clupo/oss/fontc/build/default/fontmake.GDEF.ttx	2025-02-21 15:18:57
@@ -270,8 +270,6 @@
       <ClassDef glyph="uni17BD.n" class="3"/>
       <ClassDef glyph="uni17BD.n2" class="3"/>
       <ClassDef glyph="uni17BE" class="3"/>
-      <ClassDef glyph="uni17BF.b" class="3"/>
[...]

PR #1311, opened 2025-02-28
Parse and use production names from GlyphData.xml

anthrotype:
Towards fixing https://github.com/googlefonts/fontc/issues/780 and #248 

Note this only queries the existing production names, doesn't also implement  glyphsLib's `_construct_production_name` (for suffixed alternate glyphs, uncommon ligatures etc.).

But this is already sufficient to make 'output is identical' with Google Sans Code even after disabling `--no-production-names` flag in ttx_diff.py, so we could implement the synthetic prod names later.


rsheeter:
I will land the behavior here bit by bit; closing here so I don't merge it by mistake

PR #1312, opened 2025-02-28
[crater] Document how to run CI locally

cmyr:
JMM

### DIFF
--- a/fontc_crater/README.md
+++ b/fontc_crater/README.md
@@ -39,6 +39,8 @@ $ rm -f resources/scripts/requirements.txt && pip-compile resources/scripts/requ
 # Commit updated file
 ```
 
+To run locally to rebuild the HTML:
+
 ```shell
 # clone git@github.com:googlefonts/fontc_crater.git somewhere, we'll assume at ../fontc_crater
 # CI currently has it's own format for the repo list, use the file from ^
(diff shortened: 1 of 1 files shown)


fontc_crater/README.md
```diff
@@ -48,6 +50,18 @@ $ cargo run --release -p=fontc_crater -- ci ../fontc_crater/gf-repos-2024-09-23.
...
 
+To do a full run locally, such as to test a PR (assuming same fontc_crater
+checkout as above):
+
+```shell
+# make sure the repo is up to date, so you're comparing against the latest run:
+# cd ../fontc_crater && git pull
+
+$ RUST_LOG=debug cargo run --release -p fontc_crater \
+-- ci ../fontc_crater/gf-repos-2025-02-19.json --out ../fontc_crater/results/
```
rsheeter, line 61: The moving target file here makes this not reliably copy/pasteable, could we make the gf-repos file not include the date so the command is copy/pasteable?

rsheeter (approved):

fontc_crater/README.md
cmyr: yea that's something I've been meaning to do anyway, will make updating much simpler.

cmyr: opened #1313  which will let us rename the input file. I'm going to merge this now, and then when that PR lands I can fix all the docs at once.

PR #1317, opened 2025-03-02
Parse .glyphs grayscale colors

rsheeter:
After this and #1316 we compile FoldIt w/o error again.

JMM but only after #1316

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -383,21 +383,30 @@ impl FromPlist for Color {
         let stop_offset = tokenizer.parse::<f64>()?;
         tokenizer.eat(b')')?;
 
-        if colors.len() != 4 {
-            return Err(crate::plist::Error::UnexpectedNumberOfValues {
-                value_type: "rgba values",
-                expected: 4,
+        // See <https://github.com/googlefonts/glyphsLib/blob/c4db6b981d577f456d64ebe9993818770e170454/Lib/glyphsLib/builder/common.py#L41-L50>
+        match *colors.as_slice() {
+            // Grayscale
+            [black, alpha] => Ok(Color {
+                r: black,
+                g: black,
+                b: black,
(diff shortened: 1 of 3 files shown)


glyphs-reader/src/font.rs
```diff
@@ -383,21 +383,31 @@ impl FromPlist for Color {
...
+                stop_offset: stop_offset.into(),
+            }),
+            // RGB
+            4 => Ok(Color {
+                r: colors[0],
+                g: colors[1],
+                b: colors[2],
+                a: colors[3],
+                stop_offset: stop_offset.into(),
+            }),
```
cmyr, line 403: instead of matching on the length, you can match on the actual slice, like:

```suggestion
        match colors.as_slice() {
            // Grayscale
            [black, alpha] => Ok(Color {
                r: black,
                g: black,
                b: black,
                a: alpha,
                stop_offset: stop_offset.into(),
            }),
            // RGB
            [r, g, b, a] => Ok(Color {
                r,
                g,
                b,
                a,
                stop_offset: stop_offset.into(),
            }),
```

etc.

cmyr: The other thing I wonder about, here, is if it means color would prefer to be an enum? Then we could have `Color::RGBA` and `Color::Grayscale` (and in the future, future `Color::Cmyk`?) variants.

cmyr (approved):
a couple questions inline, but nothing blocking.

glyphs-reader/src/font.rs
rsheeter: My rationale to not do this is that grayscale and RGBA aren't actually different so I think this makes sense if/when we support CMYK. Based on python I think we don't plan to do cmyk.

rsheeter: Oh this resolved two questions not one, oops. I like the match on [parts] change, will change.

PR #1318, opened 2025-03-02
[WIP] Make Glyphs production names available

rsheeter:
Split the addition and exposure of new metadata out from #1311. Attempts to capture patterns in prod names and thus avoid enumerating tens of thousands of empty or predictable strings in an array.

TODO: annoyingly although the array of many blanks went away the prod name => idx array remains.


rsheeter:
Closing, will re-raise as an extension of #1352

PR #1321, opened 2025-03-04
Rough in CPAL

rsheeter:
Not by any stretch a full implementation enough I'd like to checkpoint.

We build CPAL with the same set of colors as FontTools for FoldIt, though we order them differently.

Requires https://github.com/googlefonts/fontations/pull/1373 so update write-fonts dep.


fontbe/src/cpal.rs
```diff
@@ -0,0 +1,87 @@
...
+        // Guard clause: no colors
+        if !colors
+            .as_ref()
+            .map(|c| c.palettes.iter().any(|p| !p.is_empty()))
+            .unwrap_or_default()
+        {
+            debug!("Skip cpal; no colors");
+            return Ok(());
+        }
+        let colors = colors.unwrap();
```
cmyr, line 40: I was staring at this for a while trying to figure out how I'd write it and I think I prefer,

```suggestion
        let colors = match context.ir.colors.try_get() {
            Some(c) if !c.palettes.iter().all(Vec::is_empty) => c,
            _ => {
                debug!("Skip cpal; no colors");
                return Ok(());
            }
        };
```

But this makes me wonder why we are even getting here in the case where the palettes are all empty? Why don't we validate that at construction time, and just not bother setting colors at all in that case?

fontbe/src/cpal.rs
```diff
@@ -0,0 +1,87 @@
...
+        let sz0 = colors.palettes[0].len();
+        if colors.palettes.iter().any(|p| p.len() != sz0) {
+            return Err(Error::InconsistentPaletteLength(
+                colors.palettes.iter().map(Vec::len).collect(),
+            ));
+        }
```
cmyr, line 46: ditto here.. I would prefer to enforce our invariants earlier on, when we're constructing the palettes, instead of needing to handle the various edge cases here at the end.

fontbe/src/cpal.rs
```diff
@@ -0,0 +1,87 @@
...
+
+        let color_records = colors
+            .palettes
+            .iter()
+            .flat_map(|p| p.iter())
+            .map(|c| ColorRecord {
+                red: c.r,
+                green: c.g,
+                blue: c.b,
+                alpha: c.a,
```
cmyr, line 56: not necessary but I tend to prefer a `From` impl for these kinds of conversions, it leaves the business logic a cleaner and moves the boilerplate over to somewhere out of sight.

fontbe/src/cpal.rs
```diff
@@ -0,0 +1,87 @@
...
+            })
+            .collect::<Vec<_>>();
+
+        if color_records.len() > u16::MAX.into() {
+            return Err(Error::OutOfBounds {
+                what: "Too many CPAL colorRecords".to_string(),
+                value: format!("{}", color_records.len()),
+            });
+        }
```
cmyr, line 65: There should be an equivalent validation check for this generated by write-fonts, so we could omit this and let it be caught when we try to serialize the table? It's possible that this error message is more helpful though, I don't know how that write-fonts error gets propagated here.

fontbe/src/cpal.rs
```diff
@@ -0,0 +1,87 @@
...
+            .unwrap_or_default()
+        {
+            debug!("Skip cpal; no colors");
+            return Ok(());
+        }
+        let colors = colors.unwrap();
+        let sz0 = colors.palettes[0].len();
+        if colors.palettes.iter().any(|p| p.len() != sz0) {
+            return Err(Error::InconsistentPaletteLength(
+                colors.palettes.iter().map(Vec::len).collect(),
```
cmyr, line 44: A general observation: in encountering various errors in my use of fontc I'm often finding that the additional context in the error isn't that helpful. In investigating the error I'm going to need to reproduce it, and I can see all the values then; if the error just says "inconsistent palette length in CPAL" instead of "inconsistent palette length in CPAL [3, 4, 4, 4]". (it's worse in some other cases where we include large structures in our errors that end up getting dumped to stderr without formatting).

fontbe/src/cpal.rs
```diff
@@ -0,0 +1,87 @@
...
+        debug!(
+            "CPAL has {} color records in {} palette(s) of {}",
+            color_records.len(),
+            colors.palettes.len(),
+            entries_per_palette
+        );
+        context.cpal.set(Cpal {
+            num_palette_entries: entries_per_palette as u16,
+            num_palettes: colors.palettes.len() as u16,
+            num_color_records: color_records.len() as u16,
```
cmyr, line 77: Ah so the write-fonts PR didn't include any codegen changes, and I guess the initial codegen didn't really have any consideration for writing? This field should be getting computed by write-fonts, instead of requiring us to set it correctly, and it looks like the two values above should be computable too?

(this doesn't need to change now but let's open an issue to fixup the codegen)

fontbe/src/cpal.rs
```diff
@@ -0,0 +1,87 @@
...
+            color_records_array: NullableOffsetMarker::new(Some(color_records)),
```
cmyr, line 78: also odd that this is nullable, it isn't described as nullable in the spec (https://learn.microsoft.com/en-us/typography/opentype/spec/cpal#cpal-version-0)

fontbe/src/error.rs
```diff
@@ -95,6 +95,8 @@ pub enum Error {
     CmapConflict(#[from] CmapConflict),
     #[error("Progress stalled computing composite bbox: {0:?}")]
     CompositesStalled(Vec<GlyphName>),
+    #[error("Inconsistent palette lengths observed: {0:?}")]
```
cmyr, line 98: I think it's useful to note that this is a CPAL thing, so that the debugger knows where to start looking

fontir/src/ir.rs
```diff
@@ -1890,6 +1900,43 @@ pub struct Component {
...
 
+/// Data to inform construction of [CPAL](https://learn.microsoft.com/en-us/typography/opentype/spec/cpal#palette-table-header)
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct ColorPalettes {
+    // All palettes should have the same number of colors
+    pub palettes: Vec<Vec<Color>>,
+}
+
+impl ColorPalettes {
+    pub fn new(mut palettes: Vec<Vec<Color>>) -> Self {
```
cmyr, line 1911: mentioned above, but it feels to me like this is an excellent place for us to go and do our validation:

- palettes isn't empty
- palettes all have the same length

fontir/src/ir.rs
```diff
@@ -1890,6 +1900,43 @@ pub struct Component {
...
+        if palettes.len() == 1 {
+            palettes = vec![palettes
+                .into_iter()
+                .next()
+                .unwrap()
+                .into_iter()
+                .collect::<HashSet<_>>()
+                .into_iter()
+                .collect::<Vec<_>>()];
+        }
```
cmyr, line 1922: code golf but I would prefer to write this as,

```suggestion
        if let [one_palette] = palettes.as_mut_slice() {
            one_palette.sort();
            one_palette.dedup();
        }
```

fontir/src/ir.rs
```diff
@@ -1890,6 +1900,43 @@ pub struct Component {
...
+        for p in palettes.iter_mut() {
+            p.sort();
+        }
+        Self { palettes }
+    }
+}
+
+/// We may in time need more than RGBA, but likely not for some years so start simple
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Color {
```
cmyr, line 1933: we could just use the write-fonts `ColorRecord` type here instead of a newtype, if we wanted?

cmyr (approved):
A bunch of little nits and notes inline, I must be particularly well caffeinated this afternoon.

fontbe/src/cpal.rs
rsheeter: That's fair, will migrate it elsewhere.

fontbe/src/cpal.rs
rsheeter: Agreed, moved check

fontbe/src/cpal.rs
rsheeter: I don't think ir should handle production of BE types so I have introduced a free fn rather than a From.

fontbe/src/cpal.rs
rsheeter: I think from here is more helpful than from the final write, we've broadly tried to avoid producing bad final writes.

fontbe/src/cpal.rs
rsheeter: My guess is once cpal and colr are a bit further along is a good time to contemplate helpers.

fontbe/src/error.rs
rsheeter: (removed due to invariant movements)

fontir/src/ir.rs
rsheeter: In the light of a new day I agree, so updated.

fontbe/src/cpal.rs
rsheeter: https://github.com/googlefonts/fontations/issues/1379

PR #1326, opened 2025-03-05
Rough in COLR

rsheeter:
Enough to ttx_diff a font with colors and see a diff. Currently COLR is always empty, but that's helpful as a base to iterate on.


cmyr (approved):
looks good!

PR #1329, opened 2025-03-06
[marks] Fix tricky logic in abvm glyph splitting

cmyr:
This was tricky; in the original code I missed that:
 - we bind the global set of abvm scripts to `scriptsUsingAbvm` (https://github.com/googlefonts/ufo2ft/blob/16ed156bd6a8b9bc/Lib/ufo2ft/featureWriters/markFeatureWriter.py#L1121)
- then we union that with feaScripts, if they exist (https://github.com/googlefonts/ufo2ft/blob/16ed156bd6a8b9bc/Lib/ufo2ft/featureWriters/markFeatureWriter.py#L1127)
- but then later on we use the original unfiltered set to determine the set of _non_ abvm glyphs, i.e. we have both `scriptsUsingAbvm` and `self.scriptsUsingAbvm` in scrope, and they have different values (https://github.com/googlefonts/ufo2ft/blob/16ed156bd6a8b9bc/Lib/ufo2ft/featureWriters/markFeatureWriter.py#L1133)

...so basically  another eventful day in the font mines.

### DIFF
--- a/fontbe/src/features/marks.rs
+++ b/fontbe/src/features/marks.rs
@@ -568,36 +568,38 @@ impl<'a> MarkLookupBuilder<'a> {
         Error::AnchorDeltaError(name, err)
     }
 
-    fn scripts_using_abvm(&self) -> HashSet<UnicodeShortName> {
-        let fea_scripts = super::get_script_language_systems(&self.fea_first_pass.ast)
-            .into_keys()
-            .collect::<HashSet<_>>();
-        INDIC_SCRIPTS
-            .iter()
-            .chain(USE_SCRIPTS)
-            .chain(std::iter::once(&"Khmr"))
-            .filter_map(|s| UnicodeShortName::from_str(s).ok())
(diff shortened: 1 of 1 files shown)


fontbe/src/features/marks.rs
```diff
@@ -622,8 +624,8 @@ impl<'a> MarkLookupBuilder<'a> {
             gsub.as_ref(),
         )?;
         // https://github.com/googlefonts/ufo2ft/blob/5a606b7884bb6da/Lib/ufo2ft/featureWriters/markFeatureWriter.py#L1156
-        // I do not understand why we have all this fancy logic when it feels like
-        // 'not abvm' is implicit?
+        // TK: there's another bug here I think!? we can't trust char map, need
```
cmyr, line 627: In the course of writing this PR I noticed another thing that looks like a bug, will have a patch for this as a followup.

rsheeter (approved):

PR #1331, opened 2025-03-07
[marks] Unreachable glyphs go into non-abvm

cmyr:
It's not totally clear that this is unnecessary, but it matches the current behaviour of fonttools and it isn't _wrong_, so we go with that.

(the existing logic _was_ trying to do this, and was just wrong)


dfrg (approved):
LGTM!

PR #1332, opened 2025-03-07
[crater] Add flag to skip gftools builds

cmyr:
I often like to run crater locally, but runtime is starting to get prohibitive. This option lets us skip all the gftools builds, which cuts the number of targets in half, but means we're at least still trying to build all the fonts we know of.

### DIFF
--- a/fontc_crater/src/args.rs
+++ b/fontc_crater/src/args.rs
@@ -41,6 +41,9 @@ pub(super) struct CiArgs {
     /// This should be consistent between runs.
     #[arg(short = 'o', long = "out")]
     pub(super) out_dir: PathBuf,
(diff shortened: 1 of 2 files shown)


fontc_crater/src/args.rs
```diff
@@ -41,6 +41,9 @@ pub(super) struct CiArgs {
     /// This should be consistent between runs.
     #[arg(short = 'o', long = "out")]
     pub(super) out_dir: PathBuf,
+    /// skip gftools mode (to reduce target count when running locally)
+    #[arg(long)]
+    pub(super) skip_gftools: bool,
```
rsheeter, line 46: nit: google-wide guidance is avoid negative flags. Applied here I would mak the flag just `gftools: bool` and default it to true

PR #1334, opened 2025-03-07
[glyphs] Add 'Composition' subcategory

cmyr:
Missing this was causing us to incorrectly treat some glyphs as ligatures, which was then causing us to not include them in mark-to-base.

The list of categories/subcategories seems to have changed over time and/or not be fully documented; the list that I initially included were those that were available in the version of glyphs I was using at the time, but I've seen some other ones in the wild, and we can add them as they're encountered. (this was encountered in [NotoSansSiddham](https://github.com/notofonts/siddham)).

JMM

### DIFF
--- a/glyphs-reader/src/glyphdata.rs
+++ b/glyphs-reader/src/glyphdata.rs
@@ -69,6 +69,7 @@ pub enum Subcategory {
     SpacingCombining,
     Emoji,
     Enclosing,
+    Composition,
 }
 
 /// A queryable set of glyph data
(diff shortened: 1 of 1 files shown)


rsheeter (approved):

khaledhosny:
GlyphsApp allows any string in Category/subCategory. I sometimes even use it to sort glyphs (where the category does not matter). I’m sure what is the purpose of this part of the code, but I suggest enumerating only categories and sub categories that have significance to the code (Ligature, Mark, etc), and ignoring anything else.

cmyr:
The problem with ignoring this category is that when we don't find a category we try to synthesize one, and that was causing a bunch of glyphs to be categorized as ligatures (based on their names) that would not have been ligatures if we had parsed this category...

cmyr:
In any case good to know that this is freeform, perhaps we will need to rethink our approach at some point.

PR #1336, opened 2025-03-10
[crater] Fixup gftools cli arg

cmyr:
The way I had written this didn't actually let you turn it off.

JMM


rsheeter (approved):

rsheeter (commented):
fontc does this slightly differently, I wonder if we should be consistent: 

```rust
    #[arg(long, default_value = "true", action = ArgAction::Set)]
    pub prefer_simple_glyphs: bool,
```

PR #1354, opened 2025-03-13
Make production name and scripts available, match against production names

### LINKED ISSUE #780: Handle production names in glyph-reader::glyphdata

### LINKED ISSUE #1341: We need the 'script' field from GlyphData.xml
This is the particular motivation: https://github.com/googlefonts/glyphsLib/issues/756

There are other possible ways we could get close, but if we really want to be confident in the implementation we should be using the same data.

rsheeter:
Makes `python resources/scripts/ttx_diff.py 'https://github.com/suonmaysophanith7/KonKhmer_SleokChher#sources/Konkhmer Sleokchher.glyphs'` identical.

Based on @khaledhosny observation that the expected name is predictable from codepoint, store a bitmap of which codepoints have the predicted name plus a list of names that violate the convention. Prod names break down as follows:

* Not present: 45425
* Present, follows uni-prefix pattern: 24776
* Present, follows u-prefix pattern: 5529
* Present, violates pattern: 2490

Fixes #780. Fixes #1341.

Builds on #1361 - unrelated CI fixes - to try to reduce noise.


glyphs-reader/data/update.py
```diff
@@ -9,22 +9,75 @@
...
+    ProdNamePattern(
+        "uni-prefix",
+        lambda gi: gi.has_codepoint() and gi.production_name == f"uni{gi.codepoint:04X}",
+        lambda gi: b"u" + gi.codepoint.to_bytes(3, "little"),
+    ),
+    ProdNamePattern(
+        "u-prefix", 
+        lambda gi: gi.has_codepoint() and gi.production_name == f"u{gi.codepoint:04X}",
+        lambda gi: b"v" + gi.codepoint.to_bytes(3, "little"),
+    ),
```
khaledhosny, line 74: `uni` and `u`-prefixed names names can be generated at runtime from the codepoint, so may be check if the name would match the what would be generated and skip it entirely?

rsheeter: Early hackery ... one day I'll find time to finish it :) We don't know the pattern from the codepoint so we at the absolute minimum needto know that this is a u-prefix style name vs uni-prefix vs some other random thing?

khaledhosny: `uni`-prefix is for BMP codepoints (i.e. <= 0xFFFF), `u`-prefix is for other codepoints.

rsheeter: Oh ty, I hadn't realized they had that pattern. There are also 2490 that deviate from both patterns and 45425 that just don't have a prod name (`{'blank': 45425, 'custom': 2490, 'uni-prefix': 24776, 'u-prefix': 5529}` according to a quick and dirty count)

khaledhosny: I’m fairly sure (almost?) all production glyph names can be algorithmically generated. You basically need to the reverse of https://github.com/adobe-type-tools/agl-specification?tab=readme-ov-file#2-the-mapping, since this is how they are going to be consumed.

rsheeter: I appear to be ending up with 2490 violations of the pattern based on ^. Saves a bunch of space over my original hack, ty for the ping!

glyphs-reader/src/glyphdata.rs
```diff
@@ -73,6 +73,133 @@ pub enum Subcategory {
     Other,
 }
 
+/// The script of a given glyph
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[repr(u8)]
+pub enum Script {
```
cmyr, line 79: This is lovely for being almost but not quite entirely a subset of the scripts defined in ICU 😒

Fun additions here include "Alchemical" and "Dentistry"?


I presume referring to... https://en.wikipedia.org/wiki/Alchemical_Symbols_(Unicode_block) and https://en.wikipedia.org/wiki/Palmer_notation?

cmyr (approved):
great!

glyphs-reader/src/glyphdata.rs
rsheeter: TIL lol

PR #1356, opened 2025-03-13
Fix bug where I cleverly wanted a usize to take on negative value

rsheeter:
Bug courtesy of #1352 

Restores things like `python resources/scripts/ttx_diff.py 'https://github.com/JAMO-TYPEFACE/Orbit#Sources/Orbit.glyphs'` to passing instead of kerploding.

JMM

### DIFF
--- a/glyphs-reader/src/glyphdata_bundled.rs
+++ b/glyphs-reader/src/glyphdata_bundled.rs
@@ -75,7 +75,12 @@ fn name_offset(idx: usize) -> usize {
     if idx == NAME_OFFSETS.len() {
         return NAMES.len();
     }
-    NAME_OFFSETS.get(idx).unwrap()
+    NAME_OFFSETS.get(idx).unwrap_or_else(|| {
+        panic!(
+            "Asked for name {idx} but we only have {} names",
+            NAMES.len()
+        )
+    })
 }
 
 fn name(idx: usize) -> &'static str {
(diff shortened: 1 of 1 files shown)


glyphs-reader/src/glyphdata_bundled.rs
```diff
@@ -75,7 +75,12 @@ fn name_offset(idx: usize) -> usize {
     if idx == NAME_OFFSETS.len() {
         return NAMES.len();
     }
-    NAME_OFFSETS.get(idx).unwrap()
+    NAME_OFFSETS.get(idx).unwrap_or_else(|| {
+        panic!(
+            "Asked for name {idx} but we only have {} names",
+            NAMES.len()
+        )
+    })
```
cmyr, line 83: I'd just make this an assert at the top of the fn, then unwrap down here.

glyphs-reader/src/glyphdata_bundled.rs
```diff
@@ -86,31 +91,31 @@ fn name(idx: usize) -> &'static str {
...
     while lower <= upper {
-        let mid = (lower + upper) / 2;
+        let mid = ((lower + upper) / 2) as usize;
         let (cp, i) = CODEPOINT_TO_INFO_IDX.get(mid).unwrap();
         let c = cp.cmp(&(needle as usize));
         match c {
             Ordering::Equal => return Some(i),
-            Ordering::Less => lower = mid + 1,
-            Ordering::Greater => upper = mid - 1,
+            Ordering::Less => lower = mid as i32 + 1,
```
cmyr, line 117: now that i'm looking at this with fresh eyes, is there some reason we don't want to use the built in binary search? It also seems like these two fns could be sharing mot of their code..

cmyr (approved):
unfortunately this patch has made me look at the code again and find new nits 🤡

glyphs-reader/src/glyphdata_bundled.rs
rsheeter: Unless I'm missing something obvious the built-in is defined on Vec. We don't have a Vec. Code sharing is probably possible indeed, will add.

glyphs-reader/src/glyphdata_bundled.rs
rsheeter: I'm failing to grasp why that's an improvement?

cmyr: I am inferring that it is an invariant of this function that `idx` is in range. If you want to make this invariant explicit, the way I would generally expect that to be done is with an assert at the top of the fn; then the rest of the function can be read knowing the invariants, and it can focus more on the actual logic.

As a secondary point, this is just not what `unwrap_or_else` is really for; the idea is that you're using the closure to compute a fallback value, not to abort.


So `unwrap` is fine, `expect` is fine, but if we're trying to get fancy with a custom panic message, that's what asserts are for.

glyphs-reader/src/glyphdata_bundled.rs
cmyr: the builtins are defined on slice, vec gets them through deref.

rsheeter: It's indirect in some cases, e.g. array of offsets and you get two that define a string. Or a slice of 2-tuples where you are searching .0 and returning .1. I've generalized it to a single bsearch method which is hopefully a good start, we can iterate further over time.

PR #1357, opened 2025-03-14
Touchup name handling

rsheeter:
https://www.youtube.com/watch?v=zMF_jc5BFcs describes my feelings about name handling.

I updated a number of expected results based on ttx diff.

Spot-checking name failures from crater this seems to help with at least the following:

* `python resources/scripts/ttx_diff.py 'https://github.com/batsimadz/Sankofa-Display#sources/Sankofa.glyphs'`
* `python resources/scripts/ttx_diff.py 'https://github.com/CatharsisFonts/Cormorant#sources/generated/CormorantGaramond.glyphs'`

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -948,6 +948,22 @@ fn is_ribbi(style_name: &str) -> bool {
 }
 
 impl NameBuilder {
+    /// If drop_rbbi_suffix remove trailing Regular, Bold, or Italic, e.g. convert Family Regular to just Family
+    pub fn make_family_name(family: &str, subfamily: &str, drop_rbbi_suffix: bool) -> String {
+        let mut family = vec![family];
+        family.extend(subfamily.split_ascii_whitespace());
+        if drop_rbbi_suffix {
+            while let Some(last) = family.last().copied() {
+                if matches!(last, "Regular" | "Bold" | "Italic") {
+                    family.pop();
+                } else {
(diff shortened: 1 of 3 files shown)


fontir/src/ir.rs
```diff
@@ -948,6 +948,27 @@ fn is_ribbi(style_name: &str) -> bool {
 }
 
 impl NameBuilder {
+    /// If drop_style_suffix remove trailing Regular, Bold, or Italic, e.g. convert Family Regular to just Family
```
cmyr, line 951:
```suggestion
    /// If drop_rbbi_suffix remove trailing Regular, Bold, or Italic, e.g. convert Family Regular to just Family
```
?

fontir/src/ir.rs
```diff
@@ -948,6 +948,27 @@ fn is_ribbi(style_name: &str) -> bool {
...
+    pub fn make_family_name(
+        family: &str,
+        subfamily: &str,
+        separator: &str,
```
cmyr, line 955: is `separator` ever not just `" "`? (I did a quick search and didn't see a counter example..)

cmyr (approved):
godspeed 🫡

fontir/src/ir.rs
rsheeter: Oh good catch, maybe I ended up never using anything but space. Will fix.

Issue #1359, opened 2025-03-20
Enable Link-Time Optimization (LTO) and codegen-units = 1


zamazan4ik:
Hi! I found the project after reading an [article](https://developer.chrome.com/blog/memory-safety-fonts) - thank you for your work!

I noticed that in the `Cargo.toml` file Link-Time Optimization (LTO) for the project is not enabled. I suggest switching it on since it will reduce the binary size (always a good thing to have) and will likely improve the application's performance. If you want to read more about LTO and its possible modes, I recommend starting from [this](https://doc.rust-lang.org/rustc/codegen-options/index.html#lto) Rustc documentation.

I think you can enable LTO only for the Release builds so as not to sacrifice the developers' experience while working on the project, since LTO consumes an additional amount of time to finish the compilation routine. In this case, we can create a dedicated `[profile.optimized-dev]` profile where LTO will be disabled (so developers experience will not be affected). If we enable it on the Cargo profile level for the Release profile, users, who install the tools with `cargo install`, will get the LTO-optimized version of the app "automatically". E.g., check `cargo-outdated` Release [profile](https://github.com/kbknapp/cargo-outdated/blob/master/Cargo.toml#L48). You also could be interested in other optimization options like `codegen-units = 1` - it also brings improvements over the current defaults.

Basically, it can be enabled with the following lines:

```
[profile.release]
codegen-units = 1
lto = true
```

I have made quick tests (AMD Ryzen 5900x, Fedora 41, Rust 1.85.1, the latest version of the project at the moment, `cargo build --release` command) - here are the results. I don't know which binaries are more or less important so I report all the results. I didn't do performance benchmark between binaries but I expect at least the same performance level, at most - performance should be improved after enabling LTO for the binaries.

| Binary\Build mode | fontc | fontc_crater | fea-lsp | highlight | otl-normalizer | libascii_plist_derive.so |
| --- | --- | --- | --- | --- | --- | --- |
| Release | 20 Mib | 5.6 Mib | 7.3 Mib | 813 Kib | 1.8 Mib | 3.5 Mib |
| Release + `codegen-units = 1` + ThinLTO | 16 Mib | 4.5 Mib | 5.2 Mib | 708 Kib | 1.4 Mib | 3.3 Mib |
| Release + `codegen-units = 1` + FatLTO | 14 Mib | 4.2 Mib | 4.8 Mib | 636 Kib | 1.3 Mib | 3.3 Mib |

Clean build times:

* Release: 41s
[...]

PR #1360, opened 2025-03-21
Turn on lto and strip

rsheeter:
Exploratory adjustment based on #1359.

Tested locally the impact on size and compile time is significant:

```shell
$ cargo clean && rm -rf build/ target/
$ time cargo build -p fontc --release
# branch real	2m41.823s
# main   real	0m37.618s
$ ls -l target/release/fontc | awk '{print $5" "$9}'
# branch 13048376 target/release/fontc
# main   19544344 target/release/fontc
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,3 +53,8 @@ members = [
     "otl-normalizer",
     "fontc_crater",
 ]
+
+[profile.release]
+codegen-units = 1
+lto = "fat"
+strip = "symbols"


behdad:
What would `lto = true` do?

rsheeter:
> What would `lto = true` do?

It's an alias for fat, https://doc.rust-lang.org/rustc/codegen-options/index.html#lto. true seems less clear to me given that there are multiple significantly different ways for lto to be turned on.

cmyr:
I agree that this would make sense for binaries that we're shipping, but less clear that it's good for all release builds? For instance crater always does a release build, and making that take two extra minutes every time has a real cost (particularly when I'm paying it to run locally on a smaller set of fonts, which might only take a couple of minutes already)

rsheeter:
Good point. Maybe we should make the binary production for releases use a custom profile? Or ... given that the intuitive interpretation of release is that it's the one to release perhaps we should flip that and have crater use a crater-specific profile?

cmyr:
It looks like we can have profiles inherit from other profiles, and their example is a `release-lto` profile: https://rust-lang.github.io/rfcs/2678-named-custom-cargo-profiles.html#guide-level-explanation

PR #1361, opened 2025-03-24
Make CI gods happy

rsheeter:
Fix unrelated issues reported on #1354


cmyr (approved):
Huh, this is very confusing... there was a point release of rust on the 17th, but I wouldn't expect that to change this sort of behaviour.

(edit: and even running that toolchain locally I can't reproduce the failure? But I think the best move is just to merge this and forget it ever happened..)

PR #1366, opened 2025-03-25
Expose and fix off by one in bundled glyphs data

rsheeter:
As is traditional #1357 included an off by one, leading to 108 fontc failures on https://googlefonts.github.io/fontc_crater/

JMM


cmyr (approved):
👍

Issue #1364, opened 2025-03-25
Some tables built by fea-rs are not present in the final font


Hoolean:
e.g. For a source with a `STAT` table defined in its feature file:

**make_ufo.py**
```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "ufolib2",
# ]
# ///
from ufoLib2 import Font

ufo = Font()
ufo.features = """
  table STAT {
        # Elided fallback name
  	ElidedFallbackName {
  		name "Regular";
  	};
  
        # Design axes
  	DesignAxis ital 0 {
  		name "Italic";
  	};
  
        # Axis values
  	AxisValue {
  		flag ElidableAxisValueName;
  		name "Upright";
  		location ital 0;
  	};
  	AxisValue {
  		name "Italic";
  		location ital 1;
  	};
  } STAT;
"""
ufo.save("Stat.ufo", overwrite=True)
```
```bash
uv run make_ufo.py
cargo run Stat.ufo
```

...then the associated `name` entries are present in the final built TTF, but the `STAT` table itself is not.

Potentially a fontbe plumbing issue?

cmyr:
yes we don't currently check all the different tables that can be produced by fea-rs, and ensure those end up in the final font.

cmyr:
@Hoolean do you have a source that displays this? There doesn't seem to currently be any in our test data set (I could make one up but would be useful to see a real-world use)

Hoolean:
`STAT` is working perfectly since #1370, cheers Colin :)

rsheeter:
I suggest we fork this into per-table issues for remaining tables and close this one.

rsheeter:
Took my own advice and so closing this.

PR #1367, opened 2025-03-28
Use BASE table from FEA, if present

cmyr:
We don't currently compile this table ourselves, so the logic here is a bit funny; we just check right at the end of compilation if we FEA generated one, and if so we add it to the final font.

Progress on #1364

### DIFF
--- a/fontbe/src/font.rs
+++ b/fontbe/src/font.rs
@@ -150,6 +150,7 @@ impl Work<Context, AnyWorkId, Error> for FontWork {
             .variant(WorkId::Meta)
             .variant(WorkId::LocaFormat)
             .variant(FeWorkId::StaticMetadata)
+            .variant(WorkId::ExtraFeaTables)
             .build()
     }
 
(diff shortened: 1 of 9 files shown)


rsheeter (approved):
Optional, I feel like it would be much easier to read the test file if it was .glyphs format

cmyr:
I don't think glyphs uses FEA for any of this custom tables business...

rsheeter:
> I don't think glyphs uses FEA for any of this custom tables business...

Do you mean if you handwrite some fea in glyphs that creates an extra table it doesn't work?

cmyr:
you could maybe stick it into one of the "prefix" fields, but that doesn't really seem to be what they're for? Glyphs has structure to its use of FEA, and i have no idea if glyphs.app itself would honour these custom table blocks.


I think it makes sense to test this in the context where we expect it to arise in the wild.

I do incidentally also find it easier to just read a fully formed FEA file, instead of mentally piecing one together out of the various snippets in a glyphs source, but that's sort of a wash. I generally prefer glyphs sources for tests because it's nice to have a single file.

PR #1374, opened 2025-04-01
Reorder `gvar` points to match normalized `glyf`

Hoolean:
This PR extends the normalization in #1107 to `gvar`, which can be reordered in tandem to avoid equivalent diffs.

This patch improved the `gvar` percentage significantly in most crater projects I tested with, but there is still a remaining discrepancy that stems from there being more `gvar` deltas than points in `glyf` in some cases.

I have not had time to investigate further yet – and suspect anything from phantom points to me straight-up introducing bugs ; ) – hence the draft status. Does this difference sound familiar to anyone?

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -446,14 +446,15 @@ def normalize_gvar_contours(ttx: etree.ElementTree, point_orders: dict[str, list
         # apply the same order to every tuple
         for tup in glyph.xpath("./tuple"):
             deltas = tup.xpath("./delta")
-            by_order = {int(delta.attrib["pt"]): delta for delta in deltas}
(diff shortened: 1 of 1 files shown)


Hoolean:
> and suspect anything from phantom points to me straight-up introducing bugs

Arg, it was of course both... the extra deltas I was seeing were the phantom points, which my patch was dropping in error, which I have now corrected.

After this fix I am seeing no more spurious diffs, and some new crater fonts are being reported as identical locally, so undrafting and marking as ready for review :)

cmyr (commented):
oh funny I hadn't noticed that we were normalizing the glyph outlines, and I was wondering why the gvar points were seemingly out of order in some cases.

In any case very good catch, I think this might've been one of the bigger lingering gvar problems :)

PR #1382, opened 2025-04-03
[ttx_diff] Be more defensive when filling deltas

cmyr:
There is some non-deterministic crash on CI where we are out of bounds on an array access; this will check that condition and skip if a delta is out of bounds.

~based on #1381, which should go in first~

JMM


rsheeter (approved):

rsheeter:
> non-deterministic crash on CI

That sounds bad, can we reproduce? Does this just hide some sort of deeper problem? Is the crash consistently with one compilers output, particularly does it suggest some non-determinism or bug in fontc?

PR #1384, opened 2025-04-03
[glyphs] If a metric is defined, it has a value

cmyr:
That is, if a value is not defined for a particular master but is in the metrics list, we treat it as having a value of zero.

This more closely matches fonttools, and better handles the semantics of the glyphs format, which skips serializing `0` in some cases.

### DIFF
--- a/resources/testdata/glyphs3/infinity.glyphs
+++ b/resources/testdata/glyphs3/infinity.glyphs
@@ -1,5 +1,5 @@
 {
-.appVersion = "3219";
+.appVersion = "3343";
 .formatVersion = 3;
 axes = (
 {
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -794,7 +794,7 @@ pub struct AxisMapping {
 #[derive(Default, Debug, Clone, PartialEq, Eq, Hash, FromPlist)]
 struct RawMetric {
     // So named to let FromPlist populate it from a field called "type"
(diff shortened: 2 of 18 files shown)


resources/testdata/glyphs3/infinity.glyphs
```diff
@@ -28,5 +38,19 @@ width = 600;
...
+{
+type = "cap height";
+},
+{
+type = "x-height";
+},
+{
+type = ascender;
+}
+);
```
rsheeter, line 54: Do we now fail unless we have a metrics block? - would be nice to not have to modify infinity since this is not really the point of the test

rsheeter (approved):

resources/testdata/glyphs3/infinity.glyphs
cmyr: because the glyphs2 source includes all of these metrics, we need to either delete them there, or add them here; either way works for me.

What is the point of the test? Is it recorded anywhere?

rsheeter: I fear it's documented only in the tests that use it. It's to test `glyphname = infinity;` which used to parse as a number with suboptimal results. For context, at the time we parsed a plist dag without any knowledge of the desired data type for fields. These days we have better intel about what output type we _want_.

rsheeter: (note the lack of quotes)

Issue #668, opened 2023-12-19
Support global metrics for vertical typesetting


anthrotype:
we currently only set the global metrics for horizontal layout. We should add support for vertical layout fonts by building a `vhea` table, and also store the corresponding variable global vertical metrics tag in the `MVAR` table (e.g. vertical ascender, descender, line gap, vertical caret slope).
A `VORG` table is only needed for CFF and we currently only build TrueType-flavored variable fonts.

PR #1388, opened 2025-04-05
Implement support for vertical typesetting

Hoolean:
Implements #668

This PR begins drafting support for vertically typeset fonts in fontc, parsing glyph heights, vertical origins, and associated global metrics, and building `vhea`, `vmtx`, `VVAR`, and vertical `gvar` phantom points.

The PR is currently sufficient to build identical tables for several fonts on crater, but I am aware that there are blockers to resolve, documentation and tests to add, and drafted changes to iterate into more sensible abstractions before the PR can be considered for merging, and so I am opening a draft now to hash this out collaboratively :)

## Status

Noto Serif Bengali's vertical tables are building identically, and at least Noto Sans Symbols, Noto Serif Ethiopic and Noto Serif Tamil's are building identically aside from a difference in fontmake/fontc's serialisation of empty `HVAR`/`VVAR` tables (see below).

### Interpolated metrics are required for default heights and vertical origins

If a glyph does not provide an explicit advance height or vertical origin, these should fallback to `ascender - descender` and `ascender` respectively. This depends on being able to access values for these global metrics at all positions in the font where there is glyph data, which would require interpolated metrics in the case of e.g. sparse instances.

The PR currently fallsback to the default metrics and, while I did not encounter any diffs from this when testing with the top fonts on crater, this is necessary for fully-accurate `VVAR` tables and `gvar` phantom points.

Shall I spin out a separate issue for this to unblock a fix here?

### Abstracting across `hmtx` and `vmtx`

I am happy with how the abstraction for sharing between `HVAR` and `VVAR` has turned out, but unhappy with my draft for `hmtx` and `vmtx` thus far - in particular some gnarly mutation in the final build - and have not had time to revisit it yet.

I also think that the PR's reuse of fonttool's approach of deriving side bearings from advance, side bearing, and bounds-advance may introduce some under and overflow issues post-oxidisation where fontc was previously more robust... any ideas on what the cleanest approach would be here?

(I've allowed edits on the PR in case there is something obvious and it's easier to just dive in :fist_right: :fist_left:)

### Difference in fontmake/fontc's building of empty `HVAR`/`VVAR`

:ballot_box_with_check: *Update: see below, will be resolved by fonttools/fonttools#3797*

fontmake encodes an empty `VVAR` table with indirect mapping, whereas fontc encodes with direct mapping. fontc's approach is slightly smaller, and so there is either an unwanted semantic difference, or a bug causing sub-optimal serialisation in fontTools.

This can be observed for empty `HVAR` tables too; [crater is reporting the same issue but in `HVAR` for Doto's designspace](https://googlefonts.github.io/fontc_crater/).

My understanding is that empty `HVAR` and `VVAR` tables are semantically meaningful and important to preserve, as they allow the shaper to skip interpolating and interpreting vertical phantom values only to find that no variation in height is required.

Shall I spin out a separate issue for this too?

---

Thanks for your attention :rocket: 


Hoolean:
Update: the differing empty `HVAR`/`VVAR` is caused by fonttools needing to apply location pruning for direct mappings too, which will be resolved by fonttools/fonttools#3797

Hoolean:
(getting a bit off-topic for this PR but the empty `HVAR` diff was more widespread than I realised - after the fontTools PR is merged then 14 new cases on crater will be identical :tada: mostly monospace fonts)

Hoolean:
A thought on interpolated metrics: we can drop the requirement for these *if*:

- `GlyphInstance.height` is made mandatory
    - :alien: UFOs: [height is already mandatory in the UFO spec
](https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#attributes-1)
    - 🇬 Glyphs: default would be derived from the layer's associated master instead of interpolated from instance position; [this is already how glyphsLib (and probably Glyphs) works](https://github.com/googlefonts/glyphsLib/blob/c4db6b981d577f456d64ebe9993818770e170454/Lib/glyphsLib/builder/glyph.py#L359-L389), and so it would be a closer match to the existing stack
- `GlyphInstance.vert_origin` is made mandatory
    - :alien: UFOs: [optional in the UFO spec](https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#publicverticalorigin); does ufo2ft derive the default for brace layers with interpolation or from the owning UFO?
    - 🇬 Glyphs: default would be derived from the layer's associated master instead of interpolated from instance position; [this is already how glyphsLib (and probably Glyphs) works](https://github.com/googlefonts/glyphsLib/blob/c4db6b981d577f456d64ebe9993818770e170454/Lib/glyphsLib/builder/glyph.py#L359-L389), and so it would be a closer match

This would trim a lot of location plumbing at the compilation stage too.

Hoolean:
Definitely possible, definitely sensible - it's end-to-end and there's a lot of interdependency but there are still big slices I can make.

I'll probably start by pulling out the frontend parsing into a separate PR, and then `vhea`/`vmtx` and `VVAR` can follow separately.

Hoolean:
(The fix for differing empty `HVAR`/`VVAR` needed to land in Python land and has now been merged at fonttools/fonttools#3797, so that will arrive on its own when there's a version bump and we update our requirements)

Hoolean:
Closing as obsolete in favour of the smaller PRs this was entirely refactored into:

- #1417
- #1433
- #1489

PR #1396, opened 2025-04-08
Better handling of line-line intersection with low `t` (in open corners)

cmyr:
This is specifically for line-line intersection, where there is some special logic in fonttools that we were not replicating. Specifically this deals with cases where one segment ends either directly on another line, or on one of the end points.

This brings over a few small & specific bits of logic from fonttools to try and match some particular edge cases; I verified that this makes some things on crater improve, and makes nothing worse.


glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -102,22 +102,31 @@ impl<'a> CornerErasureCtx<'a> {
         let Some(maybe_corner) = self.possible_corner(seg_ix) else {
             return false;
         };
+        log::debug!("considering segment starting at {}", maybe_corner.one.end());
         if !maybe_corner.points_are_right_of_line() {
+            log::debug!(
+                "crossing points {} and {} not on same side of line",
+                maybe_corner.point_before_line(),
+                maybe_corner.point_after_line()
+            );
```
rsheeter, line 111: Optional, I think many of the debugs here might be better set to trace.

glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -363,14 +372,48 @@ fn seg_seg_intersection(seg1: PathSeg, seg2: PathSeg) -> Option<Intersection> {
...
+    if let (PathSeg::Line(l1), PathSeg::Line(l2)) = (seg1, seg2) {
+        let pt = l1.eval(hit.t0);
+        // the bezierTools code for line intersections has a bunch of special
+        // cases that were causing us to deviate, so we try to cover those here
+        // as they come up:
+        //
+        // special check for close x coords:
+        // https://github.com/fonttools/fonttools/blob/a6f59a4f87a011/Lib/fontTools/misc/bezierTools.py#L1193-L1212
+
+        if py_isclose(l1.end().x, l1.start().x) || py_isclose(l2.start().x, l2.end().x) {
```
rsheeter, line 386: Optional, this would be easier to read done via trait (so it goes ltr), e.g. `l1.x_almost_unchanged()` or `l1.start().x.approx_eq(l1.end().x)` or some such

glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -363,14 +372,48 @@ fn seg_seg_intersection(seg1: PathSeg, seg2: PathSeg) -> Option<Intersection> {
...
+            return Some(hit);
+        }
+        // final guard statement
+        // https://github.com/fonttools/fonttools/blob/a6f59a4f87a/Lib/fontTools/misc/bezierTools.py#L1221-L1223
+        //eprintln!("POINT {pt} {l1:?}, {l2:?}");
+        if !(both_points_are_on_same_side_of_origin(pt, l1.p1, l1.p0)
+            && both_points_are_on_same_side_of_origin(pt, l2.p0, l2.p1))
```
rsheeter, line 393: `both_points...` followed by 3 point args threw me at first, I suggest a trait here too. `l1.on_one_side_of(pt)` perhaps.

cmyr: this is also me using the same name and arguments as the python fn referenced above. (there's a stray comment here tho, will clean that up)

glyphs2fontir/src/erase_open_corners.rs
cmyr: agree, had chosen debug (and these particular statements) because I wanted our debug output to be the same as python's, so that I could compare the logs and figure out what was going on. trace should be fine for that, though..

glyphs2fontir/src/erase_open_corners.rs
cmyr: I'm really just trying to clearly replicate the code that I've linked to above, which is using the `math.is_close` fn from the python std.

glyphs2fontir/src/erase_open_corners.rs
rsheeter: A trait with a comment would be easier to read and still equally linked to python?

glyphs2fontir/src/erase_open_corners.rs
```diff
@@ -363,14 +372,48 @@ fn seg_seg_intersection(seg1: PathSeg, seg2: PathSeg) -> Option<Intersection> {
...
+    const TOLERANCE: f64 = 1e-09;
+    (a - b).abs() <= (TOLERANCE * a.abs().max(b.abs()))
+}
+
+//https://github.com/fonttools/fonttools/blob/a6f59a4f87a01110/Lib/fontTools/misc/bezierTools.py#L1148
+fn both_points_are_on_same_side_of_origin(a: Point, b: Point, origin: Point) -> bool {
+    let x_diff = (a.x - origin.x) * (b.x - origin.x);
+    let y_diff = (a.y - origin.y) * (b.y - origin.y);
+    !(x_diff <= 0.0 && y_diff <= 0.0)
 }
```
rsheeter, line 411: I couldn't even tell what this was supposed to do without plugging an example through. I think this warrants it's own tests and documentation of what happens when a and b align with origin x or y. I wound up writing a test, apparently lying on the x/y axis through origin is not "sided" and thus false.

I wonder if there is a simpler way to express this.

rsheeter: (also nicer via trait again)

Evan pointed out that by De Morgan's law `!(x_diff <= 0.0 && y_diff <= 0.0)` could become `x_diff > 0.0 || y_diff > 0.0` which is slightly nicer

cmyr: (this convo and the one above are basically the same, so will just reply here)

- I agree that this code is hard to follow, and I definitely wouldn't write it like this if I were starting from scratch,
- but this is _specifically_ about matching the funny edge cases of the python implementation,
- and so the point here isn't to be readable *per se*, but merely to be _easy to compare to the python_. That is, you should be able to open this beside the python, and be confident that they are doing the same thing, without needing to worry about fancy things like "correctness" and "why does my head hurt"

so there are almost certainly better/clearer/simpler ways of expressing this _in isolation_, but those alternatives have the downside of producing code that is structured differently from the python implementation that is our source of truth.


does this reasoning make sense to you?

rsheeter: Other than that I've come up with several ways to make it reach it do more work to reach the same answer lol

rsheeter (approved):

glyphs2fontir/src/erase_open_corners.rs
rsheeter: IMHO if we can write a decent unit test for the exactly-like-python (I sent you mine in chat) and then write a nicer implementation I think that's fine. The python isn't too bad wrt edge cases, the only major novelty I see is that if the line matches the x or y axis of the origin it's not-sided.

I would also argue making it more idiomatic while retaining a clear link to the python impl is entirely a win, it's easier for all future rustaceans and we hope to be rustaceans for an extended period.

PR #1400, opened 2025-04-11
[glyphs] Include names from variable instances

cmyr:
Even if the instance does not have a name, which more closely matches the logic in glyphsLib.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2547,6 +2547,17 @@ impl Instance {
             custom_parameters: value.custom_parameters.to_custom_params()?,
         })
     }
+
+    fn family_name(&self) -> Option<&str> {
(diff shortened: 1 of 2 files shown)


glyphs-reader/src/font.rs
```diff
@@ -2547,6 +2547,16 @@ impl Instance {
...
         })
     }
+
+    fn family_name(&self) -> Option<&str> {
+        self.properties
+            .iter()
+            .find(|raw| raw.key == "familyNames")
+            .and_then(RawName::get_value)
+        // glyphsLib here also checks for 'familyName' in customParams, but we
+        // don't have that key?
```
rsheeter, line 2557: I have the idea we  might have consolidated plural/not-plural

rsheeter (approved):

PR #1401, opened 2025-04-11
[fontir] More closely match ufo2ft when clearing names

cmyr:
Specifically the typographic names are only cleared if _both_ the family and subfamily are identical to their typographic variants.

---

I spent a while trying to get a test written for this but it's oddly tricky to craft a small source file that expresses the problem, and hopefully it's easy enough to check the referenced python and see that we're matching.

JMM

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -1128,15 +1128,14 @@ impl NameBuilder {
 
         // https://github.com/googlefonts/ufo2ft/blob/main/Lib/ufo2ft/outlineCompiler.py#L417-L421
         // Drop typographic names if they match the legacy ones
-        if let Some(family) = self.get(NameId::FAMILY_NAME) {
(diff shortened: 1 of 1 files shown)


rsheeter:
> it's oddly tricky to craft a small source file that expresses the problem

I do think we should add a test. I'm failing to spot why it's a problem from the code, don't you just specify all relevant names in the source?

Or write a unit test that sets the desired names?

rsheeter (approved):
LGTM but I do think we should add a test

cmyr:
> > it's oddly tricky to craft a small source file that expresses the problem
> 
> I do think we should add a test. I'm failing to spot why it's a problem from the code, don't you just specify all relevant names in the source?
> 
> Or write a unit test that sets the desired names?

mm yes adding a test to the builder itself is much easier, good call

PR #1402, opened 2025-04-22
variable notdef glyph

### LINKED ISSUE #1262: generated .notdef glyph isn't variable, should align to each master's ascender/descender
this is a minor one, but I file here nonetheless. 
If you run 

`python resources/scripts/ttx_diff.py resources/testdata/glyphs2/WghtVar.glyphs`

```diff
--- build/default/fontc.gvar.ttx	2025-02-13 11:42:38
+++ build/default/fontmake.gvar.ttx	2025-02-13 11:42:38
@@ -1,6 +1,15 @@
 <gvar>
     <version value="1"/>
     <reserved value="0"/>
+    <glyphVariations glyph=".notdef">
+      <tuple>
+        <coord axis="wght" value="1.0"/>
+        <delta pt="0" x="0" y="-158"/>
+        <delta pt="2" x="0" y="63"/>
+        <delta pt="5" x="0" y="-158"/>
+        <delta pt="7" x="0" y="63"/>
+      </tuple>
+    </glyphVariations>
     <glyphVariations glyph="bracketleft">
       <tuple>
         <coord axis="wght" value="1.0"/>
\ No newline at end of file
```

you'll notice that the only difference between fontc and fontmake is the gvar for the .notdef glyph. When a source, doesn't define a .notdef glyph, we have to create one. In fontc, we only make a static, non variable, .notdef glyph only defined at the default location. In fontmake, each master TTFs get its own .notdef glyph, aligned to the master's ascender/descender. When masters have different ascender/descender, they will thus get different .notdef glyphs; upon varLib-merging the masters into a VF, the gvar will contain some deltas for the .notdef glyph.

Ideally we should match fontmake and generate as many .notdef as there are glyph master locations.

cmyr:
This attempts to generate a variable notdef glyph in cases where metrics differ between masters.

~This includes a test case that crashes as a result of an unexpected location ending up in the variation model.~

- closes #1262

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -14,14 +14,14 @@ use fontdrasil::{
     orchestration::{Access, AccessBuilder, Work},
     types::GlyphName,
 };
-use kurbo::Affine;
+use kurbo::{Affine, BezPath};
 use log::{debug, log_enabled, trace};
 use ordered_float::OrderedFloat;
-use write_fonts::types::GlyphId16;
+use write_fonts::{types::GlyphId16, OtRound};
 
 use crate::{
     error::{BadGlyph, Error},
-    ir::{Component, Glyph, GlyphBuilder, GlyphInstance, GlyphOrder, StaticMetadata},
+    ir::{Component, GlobalMetric, Glyph, GlyphBuilder, GlyphInstance, GlyphOrder, StaticMetadata},
     orchestration::{Context, Flags, IrWork, WorkId},
     variations::VariationModel,
 };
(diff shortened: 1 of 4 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -2052,4 +2052,11 @@ mod tests {
         // this is a spacing-combining mark, so we shouldn't zero it's width
         assert!(glyph.default_instance().width != 0.0);
     }
+
+    // this font causes a crash when we try to generate the notdef glyph, due to
+    // funny phantom locations?
```
Hoolean, line 2057: Chipping in in case this is still causing bother, as I hit something similar with the vertical metrics: this can happen when the font has sparse layers, as there will be outline variations at locations where metrics are not defined in the sources.

[ufo2ft works around this by leaving .notdef undefined at sparse layers](https://github.com/googlefonts/ufo2ft/blob/b3895a96ca910c1764df016bfee4719448cfec4a/Lib/ufo2ft/util.py#L478-L507), but as we are are variable-first here we can go one better and only add .notdef outlines at locations where `Ascender` and `Descender` have entries.

(The intersection of the `Ascender` and `Descender` locations is probably enough to pass everywhere on crater, but the ideal implementation would use the union and interpolate if necessary to fill the gaps)

fontir/src/glyph.rs
```diff
@@ -634,23 +634,96 @@ fn ensure_notdef_exists_and_is_gid_0(
...
+fn synthesize_notdef(context: &Context) -> Result<Glyph, BadGlyph> {
+    let static_metadata = context.static_metadata.get();
+    let global_metrics = context.global_metrics.get();
+
+    let mut builder = GlyphBuilder::new(GlyphName::NOTDEF);
+    let upem = static_metadata.units_per_em;
+    let width = (upem as f64 * 0.5).ot_round();
+    let default = global_metrics.at(static_metadata.default_location());
+    let default_outline = make_notdef_outline(upem, default.ascender.0, default.descender.0);
+    // TODO: Should this be None to match other glyphs' heights? This would deviate from ufo2ft
```
rsheeter: Suggest removing todo and simply noting that this differs from other glyphs but matches ufo2ft

fontir/src/glyph.rs
```diff
@@ -634,23 +634,96 @@ fn ensure_notdef_exists_and_is_gid_0(
...
+            contours: vec![outline],
+            ..Default::default()
+        };
+        builder.try_add_source(location, instance)?;
+    }
+
+    builder.build()
+}
+
+fn make_notdef_outline(upm: u16, ascender: f64, descender: f64) -> BezPath {
```
rsheeter, line 695: Optional, a comment description of what this looks like would be nice. "Draws a box around the outside plus lines between opposite corners" type of deal.

glyphs2fontir/src/source.rs
```diff
@@ -3014,4 +3014,10 @@ mod tests {
             [Tag::new(b"wght")]
         );
     }
+    // this font causes a crash when we try to generate the notdef glyph, due to
+    // funny phantom locations?
```
rsheeter: the test name hmm and the comment that suggests we aren't sure what's happening both inspire doubt in the reader. Can we make it a bit more confident about itself?

cmyr: oops I didn't mean to commit this 😅

Issue #1397, opened 2025-04-09
Don't load unexported glyphs


simoncozens:
Here is a [Glyphs file (gzipped)](https://github.com/user-attachments/files/19665500/IncompatibleUnexported.glyphs.gz) with two glyphs; A is fully compatible; `doesnt_matter` has various incompatibilities but this is fine because it's marked as not exported and nothing should be trying to compile it. Fontmake skips over this glyph.

fontc does not:

```
$ fontc IncompatibleUnexported.glyphs
[2025-04-09T10:57:30.774116Z ThreadId(1) fontc ERROR] Invalid source glyph 'doesnt_matter': 'bad anchor: 'Invalid anchor 'doesntmatter': 'no value at default location'''
```

cmyr:
So I looked into this a bit and it's actually kind of tricky. It isn't as simple as ignoring glyphs that aren't exported, since non-export glyphs are commonly used as components in _other_ glyphs. Perhaps the best route here would be to just warn + skip when we encounter an invalid anchor...

anthrotype:
When a glyph is marked as non exporting, ufo2ft preProcessor will also decompose all components that reference this. Ideally we should attempt to match this behaviour.

cmyr:
Hmm I did think we were doing that but now I'm not sure? But I'm also not sure if that is relevant for this particular crash, which relates to some of these no-export glyphs having inconsistent anchors (as far as I understand it)

anthrotype:
> When a glyph is marked as non exporting, ufo2ft preProcessor will also decompose all components that reference this. Ideally we should attempt to match this behaviour.
>> Hmm I did think we were doing that but now I'm not sure?

actually yes, we are already decomposing composites when any of the components are not in the final font (because they have been marked as non-export, I presume):

https://github.com/googlefonts/fontc/blob/e9b1e97ac58e4ae578ceb01c635b4ca145763e0b/fontir/src/glyph.rs#L519-L530

But the code above only deals with outlines, the anchors are collected elsewhere. We need to replicate the corresponding behavior where we use the AnchorBuilder in the front-ends and stop collecting anchors from non-exporting glyphs.

PR #1403, opened 2025-04-23
[glyphs] Skip inconsitent anchors instead of erroring

cmyr:
That is, if a glyph has an anchor on some layers but not on the default layer we will just skip that anchor.

Does this make sense? I don't think it should hurt, but I'm open to other opinions.

This kind of addresses #1397, in that it fixes the crash. @simoncozens would this address your actual problem, or was your actual issue different from the test?

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -1499,19 +1499,20 @@ impl AnchorBuilder {
     }
 
     pub fn build(self) -> Result<GlyphAnchors, BadGlyph> {
-        // It would be nice if everyone was defined at default
-        for (anchor, positions) in &self.anchors {
-            if !positions.keys().any(|loc| !loc.has_any_non_zero()) {
-                return Err(BadGlyph::new(
-                    self.glyph_name.clone(),
-                    BadAnchor::new(anchor.clone(), BadAnchorReason::NoDefault),
-                ));
-            }
-        }
-
         let anchors = self
             .anchors
             .into_iter()
+            .filter(|(name, positions)| {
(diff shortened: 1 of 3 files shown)


simoncozens:
No, I'm afraid this goes in the wrong direction. If a glyph has inconsistent anchors when you compile it, the user should be made aware of the problem just as if it had inconsistent contours. Dropping the anchor is unhelpful. The problem I experienced was in an unexported glyph, which fontc had no business trying to compile in the first place.

rsheeter:
> The problem I experienced was in an unexported glyph, which fontc had no business trying to compile in the first place

As in the glyph is unexported and not referenced (such as for a component)?

Can we get a minimized example to be sure we're all talking about the same thing?

cmyr:
simon included a test case in #1397.

anthrotype:
FWIW, currently fontmake (ufo2ft actually) would *not* skip over anchors that are present in some masters even though absent from the default master.
Anchors are collected here for VF builds, and I don't see any specific logic to filter these:

https://github.com/googlefonts/ufo2ft/blob/2f11b0ff84ef1f2494e54d1a7a15d92806d8337b/Lib/ufo2ft/featureWriters/baseFeatureWriter.py#L414-L436

We actually let things blow up with "Base master not found" later down the pipeline when the variable FEA gets compiled, see:

https://github.com/googlefonts/fontmake/issues/401#issuecomment-2769955326

anthrotype:
> fonttools compiles masters independently and then merges

not anymore, ufo2ft most of the time (unless each source has different FEA files) will a single variable FEA for the entire VF

PR #1408, opened 2025-04-25
Ignore anchors from non-exportable glyphs

### LINKED ISSUE #1397: Don't load unexported glyphs

anthrotype:
Fixes #1397 

I'll add (failing) tests first to reproduce the issue, then follow up with the simple fix

### DIFF
--- a/fontc/Cargo.toml
+++ b/fontc/Cargo.toml
@@ -50,6 +50,7 @@ skrifa.workspace = true
 kurbo.workspace = true
 chrono.workspace = true
 ordered-float.workspace = true
+rstest.workspace = true
 
 [build-dependencies]
 vergen-gitcl = { version = "1.0", features = ["build", "cargo", "rustc"] }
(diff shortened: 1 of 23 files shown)


anthrotype:
Attempting to build @simoncozens's test font now completes without error (the non exporting glyph is simply ignored when building anchors).

dfrg (approved):

cmyr (approved):
looks good!

PR #1413, opened 2025-05-01
Add IR types for FeatureVarations, parse from dspace

cmyr:
This is the first step towards supporting `rvrn` (and other feature variations features) from designspace files.

### DIFF
--- a/fontir/src/error.rs
+++ b/fontir/src/error.rs
@@ -81,6 +81,11 @@ pub enum Error {
         n: usize,
         size_n: usize,
     },
+    // Look ma, I replace UnknownAxis as well!
+    #[error("Unknown {0}: {1}")]
+    UnknownEntry(&'static str, String),
+    #[error("Invalid {0}: {1}")]
+    InvalidEntry(&'static str, String),
 }
 
 /// An error related to loading source input files
--- a/fontir/src/ir/static_metadata.rs
+++ b/fontir/src/ir/static_metadata.rs
@@ -16,7 +16,7 @@ use write_fonts::{
 };
 
 use fontdrasil::{
-    coords::{NormalizedCoord, NormalizedLocation, UserLocation},
+    coords::{DesignCoord, NormalizedCoord, NormalizedLocation, UserLocation},
     types::{Axis, GlyphName},
 };
 
@@ -73,6 +73,8 @@ pub struct StaticMetadata {
     /// Miscellaneous font-wide data that didn't seem worthy of top billing
     pub misc: MiscMetadata,
     pub gdef_categories: GdefCategories,
+    /// Feature variation rules
+    pub variations: Option<VariableFeature>,
 }
 
 /// IR for a named position in variation space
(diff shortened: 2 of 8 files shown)


fontir/src/error.rs
```diff
@@ -81,6 +81,12 @@ pub enum Error {
         n: usize,
         size_n: usize,
     },
+    #[error("Tag '{0}' is not a known axis")]
+    UnknownAxis(Tag),
+    // this stinks but there are so many one-off error conditions that maybe
+    // only occur in one of the source formats that trying to be exhaustive here feels bad
+    #[error("{0}")]
+    Custom(String),
```
rsheeter, line 89: I suggest two fairly slightly more specific errors instead of just custom, something like:

```rust
// Look ma, I replace UnknownAxis as well!
#[error("Unknown {0}: {1}"]
UnknownEntry(&'static str, String),
#[error("Invalid {0}: {1}"]
InvalidEntry(&'static str, String),
```

fontir/src/ir/static_metadata.rs
```diff
@@ -225,6 +227,60 @@ pub struct Panose {
     pub x_height: u8,
 }
 
+/// A series of substitution rules to be applied to layout features
+/// at specific points in design space.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
```
rsheeter, line 232: Despite being in IR I think it would be reasonable to link to https://fonttools.readthedocs.io/en/stable/designspaceLib/xml.html#rules-element as an example of a likely source.

As it stands I think this is quite designspace specific which makes me wonder if there is a way to generalize the IR representation? - not a blocking issue tho.

rsheeter (approved):

fontir/src/ir/static_metadata.rs
cmyr: The designspace types pretty closely reflect the structure of the FeatureVariations table itself, so I hope that this will generalize; we'll find out how true that is when I move on to glyphs bracket layers :)

anthrotype: Designspace rules only model a particular subset of what FeatureVariations tables can do, specifically single glyph substitutions under specific variation regions. FeatureVariations allow to replace any feature, not only rvrn or rclt or other simple GSUB features for single substitutions. An entire feature record gets replaced under certain conditions (if the current instance falls within a specified region of the variation space), this can be any kind of feature, including GPOS features.

anthrotype: > when I move on to glyphs bracket layers

those are essentially the same thing as Designspace rules, only that instead of there being distinct alternate glyphs that get substituted as is the case with DS rules, there are "alternate layers" (this is the current naming convention for what used to be called 'bracket' layers) associated to some other glyph master layers, and the compiler needs to create glyphs for each of them (GSUB substitutes glyphs)

fontir/src/ir/static_metadata.rs
```diff
@@ -73,6 +73,8 @@ pub struct StaticMetadata {
     /// Miscellaneous font-wide data that didn't seem worthy of top billing
     pub misc: MiscMetadata,
     pub gdef_categories: GdefCategories,
+    /// Feature variation rules
+    pub variations: Option<VariableFeature>,
```
anthrotype, line 77: `variations` seems a bit too vague (glyph variations are also variations), `feature_variations` looks more appropriate here

PR #1414, opened 2025-05-01
Compile FeatureVariations from designspace files

### LINKED ISSUE #412: Generate rvrn/rclt based on conditions in designspace
There is also proposed FEA syntax for specifying conditions, but I do not believe that is really used at all, whereas they are specified in real designspace files.

This was not previously exposed in norad, but it will be as of 0.12.0 (coming soon)

cmyr:
This was a bit hairier than expected, naturally.

The basic idea is pretty simple: we treat this like other 'feature providers', where the state is passed into fea-rs which then handles the compilation. Getting that working was a bit of a project though because we didn't previously support external features/lookups that end up in GSUB.

- closes #412


rsheeter (approved):
Kinda long but nothing obviously popped out as wrong

PR #1416, opened 2025-05-02
Fix typo in normalized space coord maximum

Hoolean:
Hello! Just a measly -1 byte PR to make the constant match 🚀


rsheeter:
Eek, good catch.

PR #1417, opened 2025-05-03
Parse vertical metrics

Hoolean:
As promised, I am splitting #1388 into modular, reviewable chunks that can be merged independently (without the need to read the stream of consciousness in the original PR).

This first part covers the front-end – extracting values from .glyphs and .ufo – and so I've trimmed it to ~250 lines of diffs, which is the fewest I could get to without obfuscating wider design decisions. In total, this PR:

- Populates global and per-glyph vertical layout metrics
- Derives per-source-format default values where they are not available
- Determines whether each sources 'wants' vertical layout tables to be built

All of this re-implements the behaviour of the existing compilers. In case it is helpful for review, the most nuanced part to replicate was populating defaults:

### Existing compilers

- Per-glyph height
  - [Defaults derived from typo metrics at associated master](https://github.com/googlefonts/glyphsLib/blob/c4db6b981d577f456d64ebe9993818770e170454/Lib/glyphsLib/builder/glyph.py#L359-L389)
- Per-glyph vertical origin
  - [Defaults derived from typo metrics at associated master, or otherwise mapped relative to them](https://github.com/googlefonts/glyphsLib/blob/c4db6b981d577f456d64ebe9993818770e170454/Lib/glyphsLib/builder/glyph.py#L359-L389)
- Global vertical layout metrics
  - [Defaults derived from UPM](https://github.com/googlefonts/glyphsLib/blob/c4db6b981d577f456d64ebe9993818770e170454/Lib/glyphsLib/builder/masters.py#L74-L92)

### This PR

:heavy_check_mark: PR follows behaviour exactly

### Existing compilers

- Per-glyph height
    - [No default, mandatory in file format](https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#attributes-1)
- Per-glyph vertical origin
    -  [ufo2ft derives the default from the OS/2 table, or uses 0 (!) if it is not present](https://github.com/googlefonts/ufo2ft/blob/16ed156b/Lib/ufo2ft/outlineCompiler.py#L74-L81)
    - [UFO does not build OS/2 for sparse layers](https://github.com/googlefonts/ufo2ft/blob/16ed156b/Lib/ufo2ft/constants.py#L11-L13)
    - ...and so the default is derived from typo metrics, aside from for sparse layers, where it is 0
- Global vertical layout metrics
    - [No defaults, vertical layout tables are only built if they are provided](https://github.com/googlefonts/ufo2ft/blob/16ed156bd/Lib/ufo2ft/outlineCompiler.py#L154-L163)

### This PR

:chart_with_upwards_trend: PR follows behaviour closely, with caveats:

- Some defaults are to be populated in the backend, although these functions on `Glyph` are provided
- The brace layer vertical origin does not default to 0, and will be interpolated
    - I think this is better behaviour than overriding with 0, which is a bad value and likely a bug in ufo2ft
    - This is very unlikely to cause problems on crater

---

All in all, this should be a close fit and is ready for review - do let me know if there is anything I can do to improve it further (e.g. unit tests, although end-to-end would have to wait for the backend).


ufo2fontir/src/toir.rs
```diff
@@ -83,9 +83,18 @@ fn to_ir_glyph_instance(glyph: &norad::Glyph) -> Result<ir::GlyphInstance, Error
     for contour in glyph.contours.iter() {
         contours.push(to_ir_contour(glyph.name().as_str().into(), contour)?);
     }
+
+    let vertical_origin = glyph.lib.get("public.verticalOrigin").map(|value| {
+        value
+            .as_real()
+            .or(value.as_signed_integer().map(|int| int as f64))
+            .expect("'public.verticalOrigin' must be a number")
```
cmyr, line 91: by policy I would like to not panic on malformed sources, so we should return some error here.

cmyr (approved):
Looks good with one little nit, thanks for taking this on!

ufo2fontir/src/toir.rs
Hoolean: Cheers for the spot, I'd meant to come back to this – I've now furnished it with a proper error

rsheeter (approved):

PR #1418, opened 2025-05-04
Make name reuse deterministic, and use smallest ID as fonttools does

Hoolean:
In the case where multiple existing name records are applicable for reuse, the non-deterministic order of HashMap meant that either could be chosen. This PR explicitly prefers the smallest ID, which makes the process deterministic, and matches [fonttools' behaviour](https://github.com/fonttools/fonttools/blob/d5aec1b9dbcbaa3f97d0d553fe85a92927b48faa/Lib/fontTools/ttLib/tables/_n_a_m_e.py#L326-L329).

(I think this covers everywhere, but is there anywhere else that needs updating?)

### DIFF
--- a/fontbe/src/stat.rs
+++ b/fontbe/src/stat.rs
@@ -1,7 +1,5 @@
 //! Generates a [stat](https://learn.microsoft.com/en-us/typography/opentype/spec/stat) table.
 
-use std::collections::HashMap;
-
 use log::trace;
 
 use fontdrasil::orchestration::{Access, AccessBuilder, Work};
@@ -64,18 +62,10 @@ impl Work<Context, AnyWorkId, Error> for StatWork {
 }
 
 fn make_stat(static_metadata: &StaticMetadata) -> Stat {
-    // Prefer the smallest ID if an existing name can be reused
-    let reverse_names =
-        static_metadata
-            .names
-            .iter()
-            .fold(HashMap::new(), |mut accum, (key, name)| {
-                accum
-                    .entry(name.as_str())
-                    .and_modify(|value| *value = key.name_id.min(*value))
-                    .or_insert(key.name_id);
-                accum
-            });
+    // If possible, reuse a matching name with the smallest ID to follow fonttools:
+    // https://github.com/fonttools/fonttools/blob/d5aec1b9/Lib/fontTools/otlLib/builder.py#L2884-L2886
+    // https://github.com/fonttools/fonttools/blob/d5aec1b9/Lib/fontTools/ttLib/tables/_n_a_m_e.py#L326-L329
+    let reverse_names = static_metadata.reverse_names();
 
     Stat {
         design_axes: static_metadata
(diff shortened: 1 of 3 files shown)


fontbe/src/stat.rs
```diff
@@ -64,11 +64,18 @@ impl Work<Context, AnyWorkId, Error> for StatWork {
 }
 
 fn make_stat(static_metadata: &StaticMetadata) -> Stat {
-    let reverse_names: HashMap<_, _> = static_metadata
-        .names
-        .iter()
-        .map(|(key, name)| (name.as_str(), key.name_id))
-        .collect();
+    // Prefer the smallest ID if an existing name can be reused
```
rsheeter, line 67: specify, with reference if possible, that this is to match FontTools

fontbe/src/stat.rs
```diff
@@ -64,11 +64,18 @@ impl Work<Context, AnyWorkId, Error> for StatWork {
...
+        static_metadata
+            .names
+            .iter()
+            .fold(HashMap::new(), |mut accum, (key, name)| {
+                accum
+                    .entry(name.as_str())
+                    .and_modify(|value| *value = key.name_id.min(*value))
+                    .or_insert(key.name_id);
+                accum
+            });
```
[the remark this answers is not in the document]
Hoolean: Now added as a method with a wee test to ensure it stays consistent in choosing the minimum too :)

Agree that in the long term using a struct for `names` will be the right idea – this will especially help when multilingual names are supported too.

fontbe/src/stat.rs
Hoolean: I've now documented this on the corresponding tables with references to the table ID use and name creation in fonttools

fontbe/src/stat.rs
```diff
@@ -64,11 +62,10 @@ impl Work<Context, AnyWorkId, Error> for StatWork {
...
 
 fn make_stat(static_metadata: &StaticMetadata) -> Stat {
-    let reverse_names: HashMap<_, _> = static_metadata
-        .names
-        .iter()
-        .map(|(key, name)| (name.as_str(), key.name_id))
-        .collect();
+    // If possible, reuse a matching name with the smallest ID to follow fonttools:
+    // https://github.com/fonttools/fonttools/blob/d5aec1b9/Lib/fontTools/otlLib/builder.py#L2884-L2886
+    // https://github.com/fonttools/fonttools/blob/d5aec1b9/Lib/fontTools/ttLib/tables/_n_a_m_e.py#L326-L329
```
rsheeter, line 67: nit: move the comment to the method rather than the call location

Hoolean: I hummed and hawed before homing these originally as each table in fonttools often has its own individual `name` strat depending on when it was implemented. No strong opinion either way tho, so happy to move, and I've pushed a commit now :+1:

cmyr (approved):
Thanks!

PR #1421, opened 2025-05-06
[glyphs] Add FormatVersion type

cmyr:
I found the raw i64 here a bit confusing, (it turns out that our plist derive code uses `Default::default()` to fill in any missing values) and bracket layer parsing depends on the format version.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -422,13 +422,20 @@ impl Shape {
     }
 }
 
+#[derive(Clone, Copy, Debug, Default, PartialEq, PartialOrd)]
+enum FormatVersion {
+    #[default]
+    V2,
+    V3,
+}
+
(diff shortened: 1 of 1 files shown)


glyphs-reader/src/font.rs
```diff
@@ -2873,6 +2889,16 @@ mod tests {
...
 
+    #[test]
+    fn v2_format_version() {
+        let v2_font = glyphs2_dir().join("Mono.glyphs");
+        let as_str = std::fs::read_to_string(&v2_font).unwrap();
+        assert!(!as_str.contains(".formatVersion"), "only exists in v3");
+        let font = RawFont::load(&v2_font).unwrap();
+        // falls back to default
+        assert_eq!(font.format_version, FormatVersion::V2);
+    }
```
rsheeter, line 2900: should be a similar v3 test?

rsheeter (approved):

PR #1424, opened 2025-05-07
[glyphs] read production names from GlyphData to populate glyph postscript names

anthrotype:
Towards fixing https://github.com/googlefonts/fontc/issues/248

After #1354 we have the production names in our generated glyphdata but were using it only for looking up a glyph's category/subCategory using its production name.

This PR uses these same production names for their primary purpose i.e. renaming glyphs in the final font ready for 'production' (which means following the AGLFN rules).

The `--no-production-name` CLI flag is now being used for this when the inputs are .glyphs sources. Previously it was only having an effect when the input was a .designspace or .ufo.

I haven't implemented yet the additional logic in glyphsLib for those glyphs that have suffixes or comprise ligatures but do not have a predefined `production` name set in the GlyphData.xml. I will probably tackle that in a follow up.

I also have _not_ disabled the `--no-production-name` CLI flag in the ttx_diff.py script used by fontc_crater, since the glyph renaming doesn't match 100% fontmake for .glyphs source yet. Which means ttx_diff.py is still comparing fonts where the glyphs kept their original human-readable name (useful for debugging at times, but not reflecting real-world production).

### DIFF
--- a/resources/testdata/glyphs2/ProductionNames.glyphs
+++ b/resources/testdata/glyphs2/ProductionNames.glyphs
@@ -1,0 +1,56 @@
+{
+.appVersion = "3413";
+familyName = "Production Names";
+fontMaster = (
+{
+alignmentZones = (
+"{800, 16}",
+"{700, 16}",
+"{500, 16}",
+"{0, -16}",
+"{-200, -16}"
+);
+ascender = 800;
+capHeight = 700;
+descender = -200;
+id = m01;
+xHeight = 500;
+}
+);
+glyphs = (
+{
+glyphname = A;
+layers = (
+{
+layerId = m01;
+width = 600;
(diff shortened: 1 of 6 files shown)


resources/testdata/glyphs2/ProductionNames.glyphs
```diff
@@ -0,0 +1,56 @@
+{
+.appVersion = "3413";
+familyName = "Production Names";
+fontMaster = (
+{
+alignmentZones = (
```
cmyr, line 6: you could delete `alignmentZones` and all the metrics and this would read a bit more cleanly.

cmyr (approved):
looks good, a few little nits about the test files but happy to see progress on this!

PR #1427, opened 2025-05-07
[glyphs] Remove redundant glyph_names map

cmyr:
This was duplicating the glyph order that already exists in the font object.

JMM

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -42,7 +42,6 @@ use crate::toir::{design_location, to_ir_contours_and_components, to_ir_features
 
 #[derive(Debug, Clone)]
 pub struct GlyphsIrSource {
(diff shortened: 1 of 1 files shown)


anthrotype:
maybe that glyph_names was the set of names that are actually in the font, whereas the font_info.glyph_order may sometimes contain additional glyphs that are in fact missing?

cmyr:
`font.glyph_order` is already processed to only be existing glyphs, and to account for any custom ordering: https://github.com/googlefonts/fontc/blob/78aa47bf0765adbc85dbfd355218b214d230223b/glyphs-reader/src/font.rs#L1941

anthrotype:
ok cool! thanks for confirming

anthrotype (approved):

rsheeter (approved):
I don't remember why this existed, seems nonsensical indeed. Probably by me :)

Issue #1371, opened 2025-04-01
Populate `fvar` instance PostScript names from Glyphs' `variablePostscriptFontName` custom parameter on instances


Hoolean:
[For reference, see Glyphs' handbook](https://handbook.glyphsapp.com/custom-parameter-descriptions/):

> **variablePostscriptFontName**
>
>  PostScript name of the variable font instance. Corresponds to the OpenType name table ID 6 and the postscriptName for NamedInstance entries in the OpenType fvar table. Should be ASCII-only, and no whitespace allowed, e.g., MyFontVar-BoldCondensedItalic. Do not confuse with postscriptFullName.
> 
> The FontName generally consists of a family name (specifically, the one used for FamilyName), followed by a hyphen and style attributes in the same order as in the FullName.
> 
> Must be distinct from the static PostScript Font Name, or may cause a font conflict if both the static and variable font are installed by a user. It is recommended that it be Variations PostScript Name Prefix (ID 25), followed by a hyphen, followed by the style name (ASCII, no spaces). [OpenType spec: name ID 6](https://https//learn.microsoft.com/typography/opentype/spec/name%5C#nid6)

glyphsLib does not consider these right now, but I suspect that compiling with Glyphs directly does.

With this implemented, [the `fvar` table generated programmatically by gftools](https://github.com/googlefonts/axisregistry/blob/6bcca1f8f94b10821e94584d90bbae522f65f13b/Lib/axisregistry/__init__.py#L432-L456) will be representable directly in Glyphs sources.

Hoolean:
(related: [this has a sibling in UFO-land as the `postscriptfontname` attribute on designspace's `<instance>` too](https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#instance-element))

Hoolean:
Implemented by #1428

PR #1428, opened 2025-05-07
Compile per-instance postscript names from Glyphs and designspaces

Hoolean:
Implements https://github.com/googlefonts/fontc/issues/1371

([fonttools already supports this](https://github.com/fonttools/fonttools/blob/cc10bfc7d18e97efe1c5239ce5f47ba73050e3db/Lib/fontTools/varLib/__init__.py#L136-L138), although only from UFOs as glyphsLib does not parse this custom parameter yet)

### DIFF
--- a/fontbe/src/fvar.rs
+++ b/fontbe/src/fvar.rs
@@ -6,7 +6,7 @@ use fontdrasil::orchestration::{Access, Work};
 use fontir::{ir::StaticMetadata, orchestration::WorkId as FeWorkId};
 use write_fonts::{
     tables::fvar::{AxisInstanceArrays, Fvar, InstanceRecord, VariationAxisRecord},
-    types::Fixed,
+    types::{Fixed, NameId},
 };
 
 use crate::{
@@ -15,6 +15,10 @@ use crate::{
 };
 
 const HIDDEN_AXIS: u16 = 0x0001;
+
+// Explicitly specify that an instance has no postscript name.
+// https://learn.microsoft.com/en-us/typography/opentype/spec/fvar#instancerecord
+const NO_POSTSCRIPT_NAME: NameId = NameId::new(0xFFFF);
 
 #[derive(Debug)]
 struct FvarWork {}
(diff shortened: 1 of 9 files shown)


fontbe/src/fvar.rs
```diff
@@ -33,6 +37,15 @@ fn generate_fvar(static_metadata: &StaticMetadata) -> Option<Fvar> {
     // Reuse an existing name record if possible.
     let reverse_names = static_metadata.reverse_names();
 
+    // If a single postscript name is present, we must provide one or explicitly
+    // indicate absence for every instance. fontations only expects None when NO
+    // instances have one.
+    // https://github.com/googlefonts/fontations/blob/b4136692/write-fonts/src/tables/fvar.rs#L13-L21
+    let postscript_names = static_metadata
```
rsheeter, line 44: nit: I would expect `postscript_names` to be the actual names, vs something like `has_postscript_names` that would capture whether there are any.

Hoolean: That makes sense and I prefer how it reads too, updated now

resources/testdata/glyphs3/WghtVar_InstancesPostscript.glyphs
rsheeter, line 1: nit: I think this can be shrunk, likely you only need a space glyph, you can delete metricValues, metrics below, and most/all the properties

rsheeter (approved):
Looks good, ty!

resources/testdata/glyphs3/WghtVar_InstancesPostscript.glyphs
Hoolean: This was to keep a close match to the case without postscript names, but clean sources sounds too tempting and so I've treated us to two smaller dedicated files instead (net -300+150) 🧘

rsheeter: Big fan of dedicated files, ty

Issue #1299, opened 2025-02-26
fontc must strip hyphens and other invalid characters from postscript names


anthrotype:
The [post](https://learn.microsoft.com/en-us/typography/opentype/spec/post) table contains PostScript (glyph) names, which are defined by
https://github.com/adobe-type-tools/agl-specification as only allowing the following characters `[A-Za-z0-9_.]` and must not be longer than 63 characters.

When we do rename glyphs (when --no-production-names flag is _not_ set), we need to do what fontmake (via ufo2ft) does here:

https://github.com/googlefonts/ufo2ft/pull/274
https://github.com/googlefonts/fontmake/issues/455

basically, remove all non PS-valid characters (e.g. hyphens which often Glyphs.app's names contain for script suffixes `-cy`), truncate to max 64 chars and handle any duplicates that may arise as result of this stripping/truncating.

anthrotype:
Fixed by #1430

PR #1430, opened 2025-05-09
glyphdata: support synthetic production names

anthrotype:
Meant to fix https://github.com/googlefonts/fontc/issues/248 and #1299 

by "synthetic" production names I mean production names for glyphs that, even though they have no direct entry in GlyphData, their production name can be derived (e.g. alternate glyphs ending with a suffix or ligatures where each ligature component does have an explicit production name).

I also disabled --no-production-name flag in ttx_diff.py, which means fontc_crater will apply the final glyph renaming after this PR, for both DS and .glyphs sources.

This build GSC and Oswald without ttx diffs (I haven't tried a full local fontc_crater run).

### DIFF
--- a/fontbe/src/post.rs
+++ b/fontbe/src/post.rs
@@ -1,7 +1,8 @@
 //! Generates a [post](https://learn.microsoft.com/en-us/typography/opentype/spec/post) table.
 
 use fontdrasil::orchestration::{Access, AccessBuilder, Work};
-use fontir::orchestration::WorkId as FeWorkId;
+use fontir::orchestration::{Flags, WorkId as FeWorkId};
+use std::collections::HashMap;
 use write_fonts::{
     tables::post::Post,
     types::{FWord, Fixed},
@@ -44,13 +45,42 @@ impl Work<Context, AnyWorkId, Error> for PostWork {
             .global_metrics
             .get()
             .at(static_metadata.default_location());
-        let postscript_names = &static_metadata.postscript_names;
-        let glyph_order = context.ir.glyph_order.get();
-        let mut post = Post::new_v2(
-            glyph_order
-                .names()
-                .map(|g| postscript_names.get(g).unwrap_or(g).as_str()),
-        );
+        // Whether to rename glyphs for 'production' using the provided remame map
+        let do_production_names = context.flags.contains(Flags::PRODUCTION_NAMES);
+        let rename_map = &static_metadata.postscript_names;
+        let mut seen = HashMap::new();
+        let final_glyph_names: Vec<_> = context
+            .ir
+            .glyph_order
+            .get()
+            .names()
+            .map(|g| {
+                if !do_production_names {
+                    // pass through the original name
+                    return g.to_string();
+                }
+                let mut name = rename_map.get(g).unwrap_or(g).to_string();
+                // Adobe Glyph List spec forbids any characters not in [A-Za-z0-9._]; it also say glyphs
+                // must not start with a digit or period (except .notdef) and shouldn't exceed 63 chars,
+                // but ufo2ft only enforces the first rule so we simply follow that.
+                // https://github.com/googlefonts/ufo2ft/blob/2f11b0f/Lib/ufo2ft/postProcessor.py#L220-L233
+                name.retain(|c| c.is_ascii_alphanumeric() || c == '.' || c == '_');
+                // make duplicates unique by adding a number suffix to match fonttools/ufo2ft:
+                // https://github.com/googlefonts/ufo2ft/blob/2f11b0f/Lib/ufo2ft/postProcessor.py#L239-L248
+                if let Some(n) = seen.get(&name) {
+                    let mut n = *n;
+                    while seen.contains_key(&format!("{}.{}", name, n)) {
+                        n += 1;
+                    }
+                    seen.insert(name.clone(), n + 1);
+                    name = format!("{}.{}", name, n);
+                }
+                seen.insert(name.clone(), 1);
+                name
+            })
+            .collect();
+
+        let mut post = Post::new_v2(final_glyph_names.iter().map(|g| g.as_str()));
         post.is_fixed_pitch = static_metadata.misc.is_fixed_pitch.unwrap_or_default() as u32;
         post.italic_angle = Fixed::from_f64(static_metadata.italic_angle.into_inner());
         post.underline_position = FWord::new(metrics.underline_position.ot_round());
--- a/fontbe/src/post.rs
+++ b/fontbe/src/post.rs
@@ -60,10 +60,11 @@ impl Work<Context, AnyWorkId, Error> for PostWork {
                     return g.to_string();
                 }
                 let mut name = rename_map.get(g).unwrap_or(g).to_string();
-                // Adobe Glyph List spec forbids any characters not in [A-Za-z0-9._]; it also say glyphs
+                // Adobe Glyph List spec forbids any characters not in [A-Za-z0-9._]; it also says glyphs
                 // must not start with a digit or period (except .notdef) and shouldn't exceed 63 chars,
                 // but ufo2ft only enforces the first rule so we simply follow that.
                 // https://github.com/googlefonts/ufo2ft/blob/2f11b0f/Lib/ufo2ft/postProcessor.py#L220-L233
+                // https://github.com/adobe-type-tools/agl-specification
                 name.retain(|c| c.is_ascii_alphanumeric() || c == '.' || c == '_');
                 // make duplicates unique by adding a number suffix to match fonttools/ufo2ft:
                 // https://github.com/googlefonts/ufo2ft/blob/2f11b0f/Lib/ufo2ft/postProcessor.py#L239-L248
--- a/fontbe/src/post.rs
+++ b/fontbe/src/post.rs
@@ -45,43 +45,43 @@ impl Work<Context, AnyWorkId, Error> for PostWork {
             .global_metrics
             .get()
             .at(static_metadata.default_location());
-        // Whether to rename glyphs for 'production' using the provided remame map
-        let do_production_names = context.flags.contains(Flags::PRODUCTION_NAMES);
-        let rename_map = &static_metadata.postscript_names;
-        let mut seen = HashMap::new();
-        let final_glyph_names: Vec<_> = context
-            .ir
-            .glyph_order
-            .get()
-            .names()
-            .map(|g| {
-                if !do_production_names {
-                    // pass through the original name
-                    return g.to_string();
-                }
-                let mut name = rename_map.get(g).unwrap_or(g).to_string();
-                // Adobe Glyph List spec forbids any characters not in [A-Za-z0-9._]; it also says glyphs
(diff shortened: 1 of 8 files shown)


rsheeter:
> I also disabled --no-production-name flag in ttx_diff.py, which means fontc_crater will apply the final glyph renaming after this PR, for both DS and .glyphs sources.

Awesome! Worth landing even if it err ... craters crater, then we learn we have something to fix :)

fontbe/src/post.rs
```diff
@@ -44,13 +45,43 @@ impl Work<Context, AnyWorkId, Error> for PostWork {
...
+        let mut seen = HashMap::new();
+        let final_glyph_names: Vec<_> = context
+            .ir
+            .glyph_order
+            .get()
+            .names()
+            .map(|g| {
+                if !do_production_names {
+                    // pass through the original name
+                    return g.to_string();
```
rsheeter, line 60: nit: we're now copying the names when the flag is off where previously we were able to use str and not copy? Maybe it would make sense to branch on the flag and call one of two functions to produce Post?

fontbe/src/post.rs
```diff
@@ -44,13 +45,43 @@ impl Work<Context, AnyWorkId, Error> for PostWork {
...
+                // https://github.com/googlefonts/ufo2ft/blob/2f11b0f/Lib/ufo2ft/postProcessor.py#L239-L248
+                if let Some(n) = seen.get(&name) {
+                    let mut n = *n;
+                    while seen.contains_key(&format!("{}.{}", name, n)) {
+                        n += 1;
+                    }
+                    seen.insert(name.clone(), n + 1);
+                    name = format!("{}.{}", name, n);
+                }
+                seen.insert(name.clone(), 1);
```
rsheeter, line 79: At a glance it looks like if seen stored 2-tuples instead of formatted strings it might save us some string allocations?

fontc/src/lib.rs
```diff
@@ -3117,4 +3118,59 @@ mod tests {
             .iter()
             .all(|instance| instance.unwrap().post_script_name_id.is_none()));
     }
+
+    #[rstest]
+    #[case::yes(true)]
+    #[case::no(false)]
+    fn rename_glyphs_to_production_names(#[case] use_production_names: bool) {
+        let result = TestCompile::compile("glyphs3/ProductionNames.glyphs", |mut args| {
+            args.no_production_names = !use_production_names;
```
rsheeter, line 3127: Neither use_production_names nor do_production_names tell me much about what actually happens when reading. Might we name more informatively? `adobe_style_post_names` or something?

glyphs-reader/src/glyphdata.rs
```diff
@@ -599,9 +605,73 @@ impl GlyphData {
...
+                })
+            })
+            .collect::<Option<_>>()?;
+
+        // only (uniXXXX, uniYYYY, etc.) names with 4 hex digits can be concatenated using the
+        // more compact format uniXXXXYYYY... uXXXXX names for characters beyond BMP are joined
+        // in ligatures using the usual '_'.
+        let any_characters_outside_bmp = prod_names
+            .iter()
+            .any(|name| name.len() > 5 && is_u_name(name.as_ref()));
```
rsheeter, line 645: nit: should be name.len() == 4, to my understanding you'd have to use uniXX if you for some reason decided to write a 2-hex name.

glyphs-reader/src/glyphdata.rs
```diff
@@ -599,9 +605,73 @@ impl GlyphData {
...
+        let any_uni_names = prod_names.iter().any(|name| name.starts_with("uni"));
+
+        if !any_characters_outside_bmp && any_uni_names {
+            let mut uni_names = Vec::new();
+            for part in prod_names {
+                if let Some(stripped) = part.strip_prefix("uni") {
+                    uni_names.push(stripped.to_string());
+                } else if part.len() == 5 && is_u_name(part.as_ref()) {
+                    uni_names.push(part[1..].to_string());
```
rsheeter, line 654: is the to_string required? You can't capture &str's?

rsheeter (approved):

fontbe/src/post.rs
anthrotype: I tried but it makes the code more complicated. Also the collisions should be relatively unfrequent, I think it's ok like this

anthrotype:
I moved some changes to ttx_diff.py to a separate PR https://github.com/googlefonts/fontc/pull/1434, which should be merged before this one

fontbe/src/post.rs
cmyr: yea I agree with something like this; we can just have a `make_prod_names` fn that we call if the flag is set. In general I'm also trying to have less stuff inside `exec`, because it makes it easier to write unit tests.

fontbe/src/post.rs
```diff
@@ -44,13 +45,43 @@ impl Work<Context, AnyWorkId, Error> for PostWork {
...
+                // must not start with a digit or period (except .notdef) and shouldn't exceed 63 chars,
+                // but ufo2ft only enforces the first rule so we simply follow that.
+                // https://github.com/googlefonts/ufo2ft/blob/2f11b0f/Lib/ufo2ft/postProcessor.py#L220-L233
+                // https://github.com/adobe-type-tools/agl-specification
+                name.retain(|c| c.is_ascii_alphanumeric() || c == '.' || c == '_');
+                // make duplicates unique by adding a number suffix to match fonttools/ufo2ft:
+                // https://github.com/googlefonts/ufo2ft/blob/2f11b0f/Lib/ufo2ft/postProcessor.py#L239-L248
+                if let Some(n) = seen.get(&name) {
+                    let mut n = *n;
+                    while seen.contains_key(&format!("{}.{}", name, n)) {
```
cmyr, line 73: huh, why isn't it enough to just have `seen` be a map from `name` to `int`? then we would do,

```rust
let seen_n = seen.entry(&name).or_insert(0);
if *seen_n > 0 {
    name = format!("{name}.{seen_n}");
}
*seen_n += 1;
```

fontbe/src/post.rs
cmyr: heh just noticed this thread, soorrrrry

glyphs-reader/src/glyphdata.rs
```diff
@@ -220,6 +220,16 @@ impl Display for ProductionName {
     }
 }
 
+impl From<ProductionName> for SmolStr {
+    fn from(v: ProductionName) -> SmolStr {
+        match v {
+            ProductionName::Bmp(cp) => format!("uni{:04X}", cp).into(),
+            ProductionName::NonBmp(cp) => format!("u{:X}", cp).into(),
```
cmyr, line 227: use `smol_str::format_smolstr!` instead?

glyphs-reader/src/glyphdata.rs
```diff
@@ -599,9 +605,73 @@ impl GlyphData {
...
+                } else if part.len() == 5 && is_u_name(part.as_ref()) {
+                    uni_names.push(part[1..].to_string());
+                } else if let Some(char) = fontdrasil::agl::char_for_agl_name(part.as_ref()) {
+                    uni_names.push(format!("{:04X}", char as u32));
+                } else {
+                    panic!("Unexpected part while constructing production name: {part}");
+                }
+            }
+            let mut result = String::from("uni");
+            result.push_str(&uni_names.join(""));
```
cmyr, line 662: total nit `join` is also doing an extra allocation here, you could do something like,

`result.extend(uni_names.iter().flat_map(|s| s.chars()))` instead I imagine.

( in general this code could be allocating significantly less, but it's not a big deal)

cmyr (approved):
okay various observations and nits but overall looks good!

glyphs-reader/src/glyphdata.rs
anthrotype: well, in the following else branch we do need to create owned Strings and push to the same vec.. then later on we need to join them all into one big string so I suppose we need.. I could use a Cow<str> perhaps

glyphs-reader/src/glyphdata.rs
anthrotype: I _think_ I have avoided that extra allocation in `join()` that you pointed at in here https://github.com/googlefonts/fontc/pull/1430/commits/250851826b2d0e57ba759af0d9ada4d93c1fb478

anthrotype:
I just realised that gftools is always passing --no-production-name CLI flag to fontc and fontmake whenever it is run in "ttx diff mode":

https://github.com/googlefonts/gftools/blob/ec251ef0908dfb49f34748755ee201270d5cdcb2/Lib/gftools/builder/fontc.py#L70

If we want to have a CLI flag in ttx_diff.py to control wheterh to use the --no-production-name, we need a way from the CLI to tell gftools builder to override that. Currently it is hardcoded.
Or... I could simply punt on the ttx_diff.py flag and change gftools to no longer pass --no-production-names

glyphs-reader/src/glyphdata.rs
rsheeter: It wasn't obvious to me that we'd validated that to be true, that we couldn't possibly get here with an invalid name. If we validated could we add a comment that notes as much? Or, better, have type that proves it because it can't exist unless valid. If we didn't validate then we need to handle malformed ones here.

anthrotype: we don't validate production names at all. We assume the ones from glyphdata are 'valid' AGL names (they must be, that's the whole point of it), but users can set their own in the per-glyph properties. We handle malformed ones in so far as we strip invalid PS characters and prevent name collisions, that's all we care about here, or at least the original code we are trying to port over. This particular snippet of code that you commented on is simpy trying to see if there is any prod name that is in the form u + _more than_ 4 hex digits (beyond BMP) which signals we can't use the uniXXXX form for the compound ligature name.

rsheeter: The glyphdata has to be valid. If users can specify them then we should check them and in some way handle it if they have the wrong number of hex digits, no?

anthrotype: well, we currently don't validate them in glyphsLib/fontmake which this code is meant to match in functionality.. The worst that can happen is that we do as told and write a seeminly invalid production name but garbage in/garbage out?

rsheeter:
Ship it!

glyphs-reader/src/glyphdata.rs
anthrotype: I don't want compilation to fail with a malformed production name, it should simply use that name as is. The whole AGL names reverse-unicode mapping is broken and works on a sort of best-effort basis

anthrotype: but I'll add a test to confirm

anthrotype:
> Ship it!

don't we first want to handle the "Don't use Production Names" custom parameter (see my last comment in #248). I don't know how many sources set that, but they will regress.

rsheeter:
> don't we first want to handle the "Don't use Production Names" custom parameter

I think it's fine if there's a temporary well understood dip.

anthrotype:
> I think it's fine if there's a temporary well understood dip.

ok merging now. Next I'll work on supporting the custom parameters to disable production names from within the sources

PR #1433, opened 2025-05-13
Write static vertical metrics

Hoolean:
This is the second chunk split out of #1388 for reviewability, probably out of a final total of three or four :) This PR covers writing the **default-position** per-glyph metrics for vertical typesetting, i.e., the `vmtx` and `vhea` tables.

I have reused code from `hmtx` and `hhea` where possible, and have improved the abstraction from my original proof-of-concept – mutability is kept to a minimum now.

## Total changes

- Adds an independent job for writing `vhea` and `vmtx`
- Refactors the existing horizontal builder for reusability
    - Splits `maxp` and `head` collation into a separate builder to assist with this
- Makes the dependency of the horizontal and vertical metrics tables on `glyf` for composite bboxes explicit

## Details

In general the additions did not have to get too gnarly, and the most precise code that shows up in the patch diff is just being moved to a different place: this is particularly true for the long metrics serialisation.

The most hands-on change was moving the stats calculation abstraction [closer to what ufo2ft employs](https://github.com/googlefonts/ufo2ft/blob/16ed156bd6a8b9bc035d0aa8045a1271ef79a52e/Lib/ufo2ft/outlineCompiler.py#L797-L816), but I have not seen any horizontal regressions. When testing with the top five vertical fonts, all have identical `hhea`, `vhea`, `hmtx`, and `vmtx`.


rsheeter:
Looks sane at a glance but I don't seem to see new tests to prove it works?

Hoolean:
Indeedy, I've just been `ttx_diff.py`'ing – I'll look at rustling something up. The horizontal refactoring should enjoy the existing coverage, but I'll look at adding something for the vertical.

Hoolean:
I've added tests now, which are end-to-end for UFO and Glyphs to better the coverage for the parsing PR too. This flushed out a small typo in the naming of some custom parameters, which I've addressed in a follow-up commit :)

anthrotype (approved):

PR #1434, opened 2025-05-13
ttx_diff: reorder classdef ids like fontc (using min gid)

anthrotype:
previously (https://github.com/googlefonts/fontc/pull/1254) we were reassigning class ids within ClassDef following fonttools sorting behavior (which uses glyph names). But since the glyph names can change when we apply production names (#1430), and this sorting by name in ttx_diff happens only after glyphs were renamed (whereas in fontTools.otlLib the sorting occurs with the original names), this creates spurious diffs...
So it's preferable that ttx_diff normalization makes the fontmake font match fontc' instead, which means sorting classes of equal length by the minimum glyph index of their members.


cmyr (approved):
looks good!

PR #1437, opened 2025-05-13
Parse bracket layers

cmyr:
This is the first step towards handling glyph variations for glyphs sources.

This patch on its own doesn't do anything; the compilation logic is coming in a separate patch, because it's fiddly and deserves attention.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -258,6 +258,7 @@ pub struct Glyph {
     pub name: SmolStr,
     pub export: bool,
     pub layers: Vec<Layer>,
+    pub bracket_layers: Vec<Layer>,
     pub unicode: BTreeSet<u32>,
     /// The left kerning group
     pub left_kern: Option<SmolStr>,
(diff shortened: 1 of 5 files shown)


glyphs-reader/src/font.rs
```diff
@@ -982,6 +996,17 @@ impl RawLayer {
         self.associated_master_id.is_some() && self.attributes == Default::default()
     }
 
+    fn is_bracket_layer(&self, format_version: FormatVersion) -> bool {
```
rsheeter, line 999: a comment explaining what a bracket layer *is* would be nice, bit of an obtuse concept

rsheeter (approved):
Nit: I think your test files can be further reduced, e.g. do you actually need outlines? If you do can they be reduced?

cmyr:
The test files were both reduced by hand from real fonts, and I want outlines so that in subsequent tests I can check that the right outlines end up in the target glyphs.

For the glyphs3 file I did manually replace the outlines, but for the glyphs2 file the only way to do that would be by hand, and I want to make sure that the test file is something that was actually produced by glyphs2.

PR #1438, opened 2025-05-13
Generate bracket glyphs

cmyr:
On top of #1437, this actually generates the alternate glyphs.

This does _not_ generate the FeatureVariations tables, but it does give us a whole bunch of targets on crater for which those are the only things missing...

### DIFF
--- a/fontc/src/workload.rs
+++ b/fontc/src/workload.rs
@@ -385,6 +385,14 @@ impl Workload {
                 .get()
                 .difference(&preliminary_glyph_order)
             {
+                let id = AnyWorkId::Be(BeWorkIdentifier::GlyfFragment(glyph_name.clone()));
+                // bracket glyphs weren't in the perlim order, but because
+                // they exist in IR they get added along with the main BE glyph
+                // work.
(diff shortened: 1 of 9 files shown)


fontc/src/workload.rs
```diff
@@ -385,6 +385,14 @@ impl Workload {
...
                 .difference(&preliminary_glyph_order)
             {
+                let id = AnyWorkId::Be(BeWorkIdentifier::GlyfFragment(glyph_name.clone()));
+                // bracket glyphs weren't in the perlim order, but because
+                // they exist in IR they get added along with the main BE glyph
+                // work.
+                if self.jobs_pending.contains_key(&id) || self.success.contains(&id) {
+                    log::debug!("job exists for {glyph_name}, skipping");
+                    continue;
+                }
```
rsheeter, line 395: This bit surprises me, I'd have expected the existing logic to suffice.

rsheeter: #1447 proposes doing this w/o modifying the core processing, keeping most of the bracket changes in glyphs2fontir

rsheeter: Resolved by merging ^

PR #1440, opened 2025-05-14
support 'Don't use Production Names' custom param and useProductionNames UFO lib key

### LINKED ISSUE #248: Match fontmake glyph names

anthrotype:
Should fix #248

(unit tests coming after lunch)


anthrotype:
I opened #1442 with the (failing) test which this PR is intended to fix. But I think I just will cherry pick that commit over here

PR #1445, opened 2025-05-15
[fea-rs] Remap ids after inserting rvrn lookups

cmyr:
In the course of the bracket layers work I've been finding a few little bugs in the original designspace-based work.

It turns out we only have one font in the crater collection that both uses designspace feature variation rules, _and_ has other GSUB features (Geologica) so this slipped through the cracks.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -969,6 +969,7 @@ mod tests {
             vec![
                 Tag::new(b"GDEF"),
                 Tag::new(b"GPOS"),
+                Tag::new(b"GSUB"),
                 Tag::new(b"HVAR"),
                 Tag::new(b"MVAR"),
                 Tag::new(b"OS/2"),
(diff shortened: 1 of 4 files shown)


anthrotype (approved):

fontc/src/lib.rs
```diff
@@ -969,6 +969,7 @@ mod tests {
             vec![
                 Tag::new(b"GDEF"),
                 Tag::new(b"GPOS"),
+                Tag::new(b"GSUB"),
```
anthrotype, line 972: was this wanted?

fontc/src/lib.rs
```diff
@@ -3040,14 +3041,23 @@ mod tests {
         let result = TestCompile::compile_source("dspace_rules/Basic.designspace");
         let gsub = result.font().gsub().unwrap();
         let feature_list = gsub.feature_list().unwrap();
-        assert_eq!(feature_list.feature_records().len(), 1);
-        assert_eq!(feature_list.feature_records()[0].feature_tag(), "rvrn");
-        assert!(feature_list.feature_records()[0]
+        assert_eq!(feature_list.feature_records().len(), 2);
```
anthrotype, line 3044: perhaps you forgot to commit the changes to the test source file?

fontc/src/lib.rs
cmyr: yes, I added features.fea to an existing test case and added a `liga` rule, so this now generates GSUB.

fontc/src/lib.rs
cmyr: hm seems.. likely.

PR #1446, opened 2025-05-15
[fea-rs] Match fonttools sorting for feature variations

cmyr:
This is a funny special case because it isn't handled by the usual FEA codepaths; however it is fairly easy to match fonttools and the resulting code is pretty self-contained, so I think it's reasonable.

This doesn't do much now, but becomes important when we land the bracket layers stuff.

### DIFF
--- a/fontbe/src/features/feature_variations.rs
+++ b/fontbe/src/features/feature_variations.rs
@@ -69,15 +69,16 @@ pub(super) fn make_gsub_feature_variations(
         conditional_subs.push((region, substitutions));
     }
     let substitutions = overlay_feature_variations(conditional_subs);
-    let (lookups, lookup_map) = make_substitution_lookups(&substitutions);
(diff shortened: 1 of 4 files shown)


fontbe/src/features/feature_variations.rs
```diff
@@ -111,7 +113,20 @@ fn make_substitution_lookups(
     let mut lookups = Vec::new();
     let mut lookup_map = HashMap::new();
 
-    for sub_rules in subs.iter().flat_map(|(_, subs)| subs.iter()) {
+    // we create an intermediate vec here so we can match fontmake's sorting for
+    // the lookups, which is based on glyph id.
```
anthrotype, line 117: you mean...

```suggestion
    // the lookups, which is based on glyph name.
```

anthrotype (approved):

PR #1447, opened 2025-05-15
Try to implement Colin's original approach to bracket layers

rsheeter:
Tl;dr just do it in glyphs2fontir, don't fiddle with core processing

Reverts changes to core processing.

glyphs2fontir now presents bracket glyphs as normal glyphs it produces in the preliminary glyph order. It then produces two distinct implementations of work for glyphs, one for not-brackets and one for brackets. This enables things like don't run the bracket until the parent (terminology?) is done.

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -8,17 +8,17 @@ use chrono::DateTime;
 use log::{debug, trace, warn};
 
 use fontdrasil::{
-    coords::{DesignCoord, NormalizedCoord, NormalizedLocation},
+    coords::{NormalizedCoord, NormalizedLocation},
     orchestration::{Access, AccessBuilder, Work},
     types::{Axes, GlyphName},
 };
 use fontir::{
     error::{BadGlyph, BadGlyphKind, BadSource, Error},
     ir::{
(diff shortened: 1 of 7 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -912,98 +927,61 @@ fn check_pos(
...
     }
 
     fn read_access(&self) -> Access<WorkId> {
+        // We need access to the glyph whence we emerged (bracket glyphs come from non-bracket glyphs)
+        // and it's anchors
         AccessBuilder::new()
             .variant(WorkId::StaticMetadata)
             .variant(WorkId::GlobalMetrics)
+            .specific_instance(WorkId::Glyph(self.parent_glyph_name().clone()))
+            .specific_instance(WorkId::Anchor(self.parent_glyph_name().clone()))
```
cmyr, line 942: I was thinking about this earlier and I suspect we don't want the parent anchors? we want to take anchors specifically from the relevant layers, right?

(I acknowledge this is my bug)

cmyr (approved):
Looks good with one note, but happy to merge this as-is and open an issue to investigate that?

glyphs2fontir/src/source.rs
rsheeter: Can I merge and leave that to you as follow-on? - I'm strapped for time [and irritable it lol]

PR #1449, opened 2025-05-15
[ttx_diff] Sort feature indices in script table

cmyr:
The order of these doesn't matter, and when adding features for feature variations, the order in fontmake isn't sorted where it is in fontc.


anthrotype (approved):

anthrotype:
just out of curiosity, I know it doesn't matter, but what order is fontmake (feaLib/otlLib) using for feature indices in script tables? The order in which features get assigned to each script in the feature file maybe?

cmyr:
@anthrotype it appears to be sorted except in the case of feature variations, which get appended to the end.

anthrotype:
Hm ok, then maybe we could fix that in fonttools. If feaLib is creating these feature indices in a sorted fashion, then featureVars module which modifies the GSUB table could maintain the same order rather than simply appending

PR #1456, opened 2025-05-20
Generate bracket glyphs, take two

cmyr:
This is a second take at bracket glyphs, based on feedback in #1438.

- the bracket glyphs are now included in the preliminary glyph order, so the impl is much less hacky.
- there are a few additional test cases.

Hopefully this is a good compromise. I'd tried starting this at #1447, but that had introduced some subtle regressions that I had trouble understanding, so I've gone back and just based this on my previous patch since I have a better understanding of what is going on there. Additionally there is a followup patch to this that was going to be difficult to rebase on top of the previous merge.

### DIFF
--- a/fontc/src/workload.rs
+++ b/fontc/src/workload.rs
@@ -437,25 +437,10 @@ impl Workload {
                 .into();
         }
 
-        if let AnyWorkId::Fe(FeWorkIdentifier::Glyph(glyph_name)) = &success {
-            self.update_be_glyph_work(fe_root, glyph_name.to_owned());
-            // we also need to update for any bracket glyphs that were finished
-            // alongside this glyph.
-            let bracket_glyphs = self
-                .also_completes
-                .get(&success)
-                .into_iter()
-                .flat_map(|ids| {
-                    ids.iter().filter_map(|id| match id {
-                        AnyWorkId::Fe(FeWorkIdentifier::Glyph(name)) => Some(name.clone()),
-                        _ => None,
-                    })
-                })
-                .collect::<Vec<_>>();
-            for bracket_glyph in bracket_glyphs {
-                self.update_be_glyph_work(fe_root, bracket_glyph);
-            }
+        if let AnyWorkId::Fe(FeWorkIdentifier::Glyph(glyph_name)) = success {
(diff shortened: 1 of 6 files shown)


fontc/src/workload.rs
```diff
@@ -435,10 +435,25 @@ impl Workload {
...
+                .flat_map(|ids| {
+                    ids.iter().filter_map(|id| match id {
+                        AnyWorkId::Fe(FeWorkIdentifier::Glyph(name)) => Some(name.clone()),
+                        _ => None,
+                    })
+                })
+                .collect::<Vec<_>>();
+            for bracket_glyph in bracket_glyphs {
+                self.update_be_glyph_work(fe_root, bracket_glyph);
+            }
```
rsheeter, line 455: We shouldn't need this change, it suggests we have a bug somewhere else

rsheeter: As per IM I'd prefer we split up the glyph work as in the original, now rolled back, PR. I think that concept was sound.

glyphs2fontir/src/source.rs
```diff
@@ -884,24 +950,39 @@ impl Work<Context, WorkId, Error> for GlyphIrWork {
         let static_metadata = context.static_metadata.get();
         let axes = &static_metadata.all_source_axes;
         let global_metrics = context.global_metrics.get();
+        let is_bracket_layer = self.base_name.is_some();
```
rsheeter, line 953: This feels a bit obtuse, I wonder if we could at least encapsulate this in a method such that this reads self.is_bracket_layer()?

glyphs2fontir/src/source.rs
```diff
@@ -833,6 +897,8 @@ impl Work<Context, WorkId, Error> for KerningInstanceWork {
 #[derive(Debug)]
 struct GlyphIrWork {
     glyph_name: GlyphName,
+    // If present, we are building a bracket glyph.
+    base_name: Option<GlyphName>,
```
rsheeter, line 901: Suggest a name that's a bit clearer, e.g. bracket_glyph_parent or something. base_name is a bit vague.

rsheeter (approved):
Much better, I like that brackets are now isolated to glyphs2fontir

PR #1457, opened 2025-05-20
[glyphs] Bracket glyphs get bracket layer anchors

cmyr:
On top of #1456, this makes sure bracket layers have the correct anchors.


cmyr:
mmm this is failing for a few fonts, will need to investigate

cmyr:
this was resolved in #1456, and the test case is PR'd separately in #1459

PR #1460, opened 2025-05-20
[glyphs] Generate feature variations

cmyr:
This generates the actual lookups, features, and FeatureVariation tables that make use of the bracket glyphs.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -3136,6 +3136,25 @@ mod tests {
         );
     }
 
+    fn get_first_feature_substitution(
+        gsub: write_fonts::read::tables::gsub::Gsub,
+        expected_sub_idx: usize,
+    ) -> write_fonts::read::tables::layout::Feature {
+        let featvars = gsub.feature_variations().unwrap().unwrap();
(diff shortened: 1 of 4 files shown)


fontc/src/lib.rs
```diff
@@ -3174,6 +3174,74 @@ mod tests {
...
+
+        let rlig_sub_record = feature_substitution.substitutions()[0];
+
+        assert_eq!(rlig_sub_record.feature_index() as usize, rlig_idx);
+        let rlig_replacement = rlig_sub_record
+            .alternate_feature(feature_substitution.offset_data())
+            .unwrap();
+
+        assert_eq!(rlig_replacement.lookup_list_indices(), [0]);
+    }
```
rsheeter, line 3214: It's quite hard to read these tests. There's so much unwrapping and record fetching it dominates the code. Idk if we need helper functions or what, but my intuition is it should be possible to make it a bit more readable.

rsheeter (approved):

fontc/src/lib.rs
cmyr: the structure of the featurevariations table is one of the most baroque I've seen... split this out into a helper, at least.

PR #1462, opened 2025-05-21
[fea-rs] Error on conflicting lig sub rules

cmyr:
This matches AFDKO and fonttools.

see https://github.com/fonttools/fonttools/pull/3835

### DIFF
--- a/fea-rs/src/compile/lookups.rs
+++ b/fea-rs/src/compile/lookups.rs
@@ -975,13 +975,24 @@ impl SomeLookup {
         }
     }
 
-    pub(crate) fn add_gsub_type_4(&mut self, target: Vec<GlyphId16>, replacement: GlyphId16) {
+    /// Returns `true` if this replacement shadows an existing rule.
+    ///
+    /// In this case the rule is not added, and the client should report an error.
+    pub(crate) fn add_gsub_type_4(
+        &mut self,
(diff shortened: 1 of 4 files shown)


fea-rs/src/compile/lookups.rs
```diff
@@ -975,13 +975,24 @@ impl SomeLookup {
         }
     }
 
-    pub(crate) fn add_gsub_type_4(&mut self, target: Vec<GlyphId16>, replacement: GlyphId16) {
+    /// Returns `true` if this replacement shadows an existing rule.
+    ///
+    /// In this case the rule is not added, and the client should report an error.
```
anthrotype, line 980: > In this case the rule is not added...

I'd expect then for a method named `add_gsub_type_4` to logically return `false` in this situation, rather than `true`, no?

anthrotype: or is it like with unix commands, zero (false) means success, non-zero means some error?

anthrotype (approved):

fea-rs/src/compile/lookups.rs
cmyr: for me it's like the `insert` method on rust's set types; it returns `true` if the item existed.

This isn't my favourite API in the world but it gets the job done. 🤷

PR #1464, opened 2025-05-21
merge mixed lig and single sub rules

cmyr:
This behaviour was changed in fontools in https://github.com/fonttools/fonttools/pull/3805; this patch gets us caught up.

 This should get us another +something on crater.


anthrotype (approved):

anthrotype:
i'm going to merge this so I can include in the upcoming fontc release

PR #1469, opened 2025-05-22
Improve flattening nested no-export components

khaledhosny:
Currently if a no-export glyph is a component in a composite glyph, which is also a component in another composite glyph, all the composite glyphs will be decomposed into simple glyphs. This is because no-export components are flattened before decomposing no-export glyphs.

Decomposing no-export glyphs first then flattening, does not decompose the composite glyph with nested components, which keeps the file size smaller and matches what ufo2ft does.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -2303,6 +2303,39 @@ mod tests {
         assert_noexport("designspace_from_glyphs/WghtVar_NoExport.designspace");
     }
 
+    #[test]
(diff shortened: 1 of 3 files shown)


anthrotype (approved):
LGTM, thanks!

indeed ufo2ft converts no-export components to simple outlines at the very beginning, before applying other filters like FlattenComponents.

rsheeter:
Nice, ty!

PR #1474, opened 2025-05-22
[glyphs] Feature variations go under rvrn by default

cmyr:
see https://github.com/googlefonts/glyphsLib/issues/1081; glyphs.app puts these under rlig, but glyphsLib does not, so let's match glyphsLib for now.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -3188,22 +3188,22 @@ mod tests {
         let lookup_list = gsub.lookup_list().unwrap();
         assert_eq!(lookup_list.lookup_count(), 1);
         let feature_list = gsub.feature_list().unwrap();
-        // by default, glyphs puts feature variations in rlig
-        let rlig_idx = feature_list
+        // by default, glyphsLib puts feature variations in rvrn
+        let rvrn_idx = feature_list
             .feature_records()
             .iter()
-            .position(|rec| rec.feature_tag() == "rlig")
+            .position(|rec| rec.feature_tag() == "rvrn")
             .unwrap();
-        let rlig = feature_list.feature_records()[rlig_idx];
-        // base rlig feature has no lookups
-        assert!(rlig
+        let rvrn = feature_list.feature_records()[rvrn_idx];
+        // base rvrn feature has no lookups
+        assert!(rvrn
             .feature(feature_list.offset_data())
             .unwrap()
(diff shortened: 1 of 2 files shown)


rsheeter (approved):

schriftgestalt:
Please NEVER us `rvrn` if you don’t have a VERY good reason. 
I have explained that a few time so here is the short version:
- try to have a alternates on unencoded glyphs (e.g. on `.sc` variants). When rvrn is applied there are no .sc glyphs, yet.
- try to have OpenType substitutions on glyphs that have feature variations. By the time e.g. the smcp lookup is applied there is only a ".var001" (or whatever suffix you choose). But those are not considered in the lookup by the designer because he has never seen those glyphs.

rvrn should be removed from OpenType. I just recently learned why it was introduced. And it was to patch over an implementation detail in MS Word. Never fix issues in apps with the font.

twardoch:
I totally agree with Georg. Feature variations should simply work in all features. I myself lamented that you can only assign one feature to the rules element in DesignSpace. 

If I want swash glyphs to appear/disappear depending on some axis position, I should put feature variations into "swsh". If I want ligatures to turn on or off depending on some axis, feature variations should be done in the "liga" (or "dlig", if this behavior is off by default). 

The "rvrn" feature was an unnecessary addition.

anthrotype:
> you can only assign one feature to the rules element in DesignSpace

actually you can set whatever feature tag (or even list of tags) using `com.github.fonttools.varLib.featureVarsFeatureTag` DS lib key, see https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#rules-element

the question here is not whether rvrn or rlib or whatever is better than the other one, is what the default should be. The default originally when we first implemented this was rvrn, and it's too late to change the default in my opinion.

The user can set rules `processing="last"` attribute if they want to append the conditional lookups at the end of the lookup list (we use `rclt` instead of `rlig` but it's basically the same), otherwise the default behavior has always been and IMO should continue to be that rvrn lookups are processed before the ones from other features (thus inserted at the beginning of the lookup list).

twardoch:
> This can also take a comma-separated list of feature tags, e.g. `salt,ss01`, if you wish the same rules to be applied with several features.

Oh, this is great news. I remember once asking for it, but I didn’t follow-up and realize that you’ve implemented it!

PR #1476, opened 2025-05-22
Move some feature variations logic to fontir

cmyr:
We want to be able to call the 'overlay_feature_variations' method from both fontbe and glyphs2fontir, which means it can't be in `fontbe`.

This is just moving things around, no functional change.

JMM


rsheeter (approved):
LGTM with mild discomfort because this feels like it _should_ be in fontbe, it's the specific type of work that is supposed to live in fontbe.

PR #1477, opened 2025-05-22
[glyphs] More preciesly match glyphsLib bracket logic 

cmyr:
_this is on top of #1476_

Specifically this means running the 'overlay_feature_variations' method
twice, once in glyphs2ir and then finally when generating the actual
feature variations tables.

To make this possible we had to refactor that code to work on glyph
names, instead of on GIDs.

With this patch we are showing identical for a number of new fonts on crater.


rsheeter:
> a number of new fonts

"a number"? What number!! :D

rsheeter (approved):

PR #1478, opened 2025-05-23
Allow building on WASM

simoncozens:
There are three blockers to running fontc over WASM:

* The pervasive use of timers, which can be a pain when `Instant::now()` causes a crash; we either make them optional (`JobTimer`) or we conditionally hide the actual `instant` fields.
* The use of file paths, which can be a pain when you don't have a filesystem; we provide ways of getting pre-loaded binary data into the Args struct and using that instead, and provide an alternate `run` method which directly emits the binary file instead of writing it to disk.
* The use of threads in the workload orchestrator; we don't use a separate ThreadPool and use a nightly version of crossbeam while on WASM to make it all compatible with [wasm-bindgen-rayon](https://github.com/RReverser/wasm-bindgen-rayon).

The result is a web-based font compiler; code available on request, demo at http://www.corvelsoftware.co.uk/fontc (works best on Chrome).

### DIFF
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -7,6 +7,8 @@ use fontir::orchestration::Flags;
 use regex::Regex;
 use serde::{Deserialize, Deserializer, Serialize};
 
+use crate::InMemorySource;
+
 /// What font can we build for you today?
 #[derive(Serialize, Deserialize, Parser, Debug, Clone, PartialEq)]
 #[command(version)]
@@ -22,6 +24,11 @@ pub struct Args {
     /// DEPRECATED: old name for positional input file
     #[arg(short, long)]
     source: Option<PathBuf>,
+
+    /// Used by WASM and other library consumers to pass in a source directly.
+    #[clap(skip)]
+    #[serde(skip)]
+    pub input_binary: Option<InMemorySource>,
 
     /// Whether to write IR to disk.
     #[arg(short, long, default_value = "false")]
@@ -133,6 +140,7 @@ impl Args {
         Args {
             glyph_name_filter: None,
             input_source: Some(input_source),
+            input_binary: None,
             source: None,
             emit_ir: false,
             output_file: None,
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -28,8 +28,7 @@ use fontir::{
     source::Source,
 };
 
-use fontbe::orchestration::Context as BeContext;
-use fontbe::paths::Paths as BePaths;
+use fontbe::{orchestration::Context as BeContext, paths::Paths as BePaths};
 use fontir::paths::Paths as IrPaths;
 
 use log::debug;
@@ -83,6 +82,7 @@ pub fn run(args: Args, mut timer: JobTimer) -> Result<(), Error> {
     let be_root = BeContext::new_root(args.flags(), be_paths, &fe_root);
     let mut timing = workload.exec(&fe_root, &be_root)?;
 
+    #[cfg(not(target_family = "wasm"))]
     if args.flags().contains(Flags::EMIT_TIMING) {
         let path = args.build_dir.join("threads.svg");
         let out_file = OpenOptions::new()
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -53,6 +53,22 @@ fn create_source(source: &Path) -> Result<Box<dyn Source>, Error> {
     }
 }
 
+/// Represents a source that is loaded in memory, e.g. for testing or
+/// when the source is not a file on disk (WASM or library use).
+///
+/// Currently only supports Glyphs sources, but could be extended to e.g. UFOZ or other formats.
+#[derive(Debug, Clone, PartialEq)]
+pub enum InMemorySource {
+    Glyphs(String),
+}
+
+/// Creates a [`Source`] from a source file loaded in memory
+fn create_in_memory_source(source: &InMemorySource) -> Result<Box<dyn Source>, Error> {
+    match source {
+        InMemorySource::Glyphs(source) => Ok(Box::new(GlyphsIrSource::new_from_memory(source)?)),
+    }
+}
+
 /// Run the compiler with the provided arguments
 pub fn run(args: Args, mut timer: JobTimer) -> Result<(), Error> {
     let time = timer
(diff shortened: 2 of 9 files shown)


cmyr:
Very cool!

I think for the timer stuff the best solution is to actually have some sort of `Timings` struct that we pass around, and then on wasm we can just `cfg` out the actual logic in one place? I've wanted to rework that for a while anyway so this sounds like a good opportunity, let me take a stab at that now...

simoncozens:
I thought about having a DummyTimer object which just does nothing, but this seemed less invasive. Reworking the timings code would make this a lot easier, though.

cmyr (commented):
okay yes I think we can simplify the timing stuff quite a bit, will have a PR shortly.

rsheeter:
Oh we're just never submitting that back to the timing mechanism? I guess that's OK to remove for now.

fontc/src/args.rs
```diff
@@ -23,6 +25,10 @@ pub struct Args {
     #[arg(short, long)]
     source: Option<PathBuf>,
 
+    #[clap(skip)]
+    #[serde(skip)]
+    pub input_binary: Option<InMemorySource>,
```
cmyr, line 31: can we add a line of documentation noting that this is for use by wasm, (or something else, if I have misunderstood)?

fontc/src/lib.rs
```diff
@@ -53,6 +52,18 @@ fn create_source(source: &Path) -> Result<Box<dyn Source>, Error> {
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum InMemorySource {
```
cmyr, line 60: ditto, a quick line of explanation would be helpful here.

fontc/src/lib.rs
```diff
@@ -88,6 +100,16 @@ pub fn run(args: Args, mut timer: JobTimer) -> Result<(), Error> {
     write_font_file(&args, &be_root)
 }
 
+/// Run and return a binary
+pub fn run_binary(args: Args) -> Result<Vec<u8>, Error> {
```
cmyr: this name makes me think that we're running a binary executable or something, maybe something like `generate_binary`?

glyphs-reader/src/font.rs
```diff
@@ -3068,6 +3072,15 @@ impl Font {
         RawFont::load(glyphs_file.as_ref()).and_then(Font::try_from)
     }
 
+    pub fn load_from_string(data: &str) -> Result<Font, Error> {
```
cmyr: can we move this up so that it is right below `pub fn load`? Will make it easier to ensure that we update both at the same time if we ever make any changes.

fontc/src/args.rs
```diff
@@ -23,6 +25,11 @@ pub struct Args {
     #[arg(short, long)]
     source: Option<PathBuf>,
 
+    /// Used by WASM and other library consumers to pass in a source directly.
+    #[clap(skip)]
+    #[serde(skip)]
+    pub input_binary: Option<InMemorySource>,
```
rsheeter, line 31: I dislike having multiple input_ fields here where only one should be set. To me these are specifically cli args. wasm probably shouldn't need to build with clap at all.

We should instead have both cli launch and wasm launch call a common function.

fontc/src/lib.rs
```diff
@@ -53,6 +52,22 @@ fn create_source(source: &Path) -> Result<Box<dyn Source>, Error> {
     }
 }
 
+/// Represents a source that is loaded in memory, e.g. for testing or
+/// when the source is not a file on disk (WASM or library use).
+///
+/// Currently only supports Glyphs sources, but could be extended to e.g. UFOZ or other formats.
+#[derive(Debug, Clone, PartialEq)]
+pub enum InMemorySource {
```
rsheeter, line 60: I wonder if we should instead have something like `FontSource { DesignSpacePath, GlyphsPath, GlyphsMemory }`? - that would be very natural to hang create_source off, thereby keep whether it's in memory or not a little more isolated.

fontc/src/lib.rs
```diff
@@ -88,6 +104,18 @@ pub fn run(args: Args, mut timer: JobTimer) -> Result<(), Error> {
     write_font_file(&args, &be_root)
 }
 
+/// Run and return a binary
+///
+/// This is the library entry point to fontc.
+pub fn generate_binary(args: Args) -> Result<Vec<u8>, Error> {
```
rsheeter, line 110: Nit: it's an OpenType Font, not just any binary. Suggest ...return an opentype font and generate_font

fontc/src/timing.rs
```diff
@@ -264,7 +267,10 @@ impl JobTime {
         match self {
             // you can call 'run' without queuing, if needed
             JobTime::Ready(mut state) | JobTime::Queued(mut state) => {
-                state.run = Instant::now();
+                #[cfg(not(target_family = "wasm"))]
```
rsheeter, line 270: I don't love having wasm cfg's strewn everywhere, lets try to isolate that more. I wonder if we should consider something like passing in a Clock where Clock is an enum that can by either system time (e.g. Instant's) or Fake, just returns some madeup time.

fontir/Cargo.toml
```diff
@@ -26,13 +26,17 @@ indexmap.workspace = true
...
 log.workspace = true
 env_logger.workspace = true
-parking_lot.workspace = true
 
 write-fonts.workspace = true  # for pens
 
 chrono.workspace = true
 smol_str.workspace = true
 
+[target.'cfg(not(platform_family = "wasm"))'.dependencies]
```
rsheeter, line 35: This should have a comment explaining why and what the condition to move off nightly is. Also ... why can't we use workspace still and put the cfg's there?

glyphs2fontir/src/source.rs
```diff
@@ -845,12 +860,13 @@ impl Work<Context, WorkId, Error> for FeatureWork {
...
-                .canonicalize() // if we got this far, the path must exist
-                .unwrap()
-                .parent()
-                .unwrap() // the path must be in a directory
-                .to_path_buf(),
+            self.font_file_path.as_ref().map(|path| {
+                path.canonicalize()
+                    .unwrap()
+                    .parent()
+                    .unwrap() // the path must be in a directory
```
rsheeter, line 867: I know this preexists but since we're here ... let's expect rather than unwrap so if this does ever go off we have more clues.

fontir/Cargo.toml
simoncozens: I don't know that we can put target-specific dependencies in the workspace Cargo.toml.

PR #1482, opened 2025-05-23
Simplify timing more

cmyr:
Motivated by #1478, this will make it less invasive to use `cfg` to get us building on wasm.

### DIFF
--- a/fontc/src/timing.rs
+++ b/fontc/src/timing.rs
@@ -17,7 +17,7 @@ use std::{collections::HashMap, io, thread::ThreadId, time::Instant};
 pub struct JobTimer {
     /// The beginning of time
     t0: Instant,
-    job_times: HashMap<ThreadId, Vec<JobTime>>,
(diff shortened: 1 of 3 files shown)


fontc/src/timing.rs
```diff
@@ -206,101 +218,73 @@ fn color(id: &AnyWorkId) -> &'static str {
...
+#[derive(Debug, Clone)]
+pub enum JobTime {
+    /// Initial state of a runnable job.
+    Runnable(JobTimeState),
+    /// State after submission to an execution queue.
+    Queued(JobTimeState),
+    /// State after the job starts running
+    Running(JobTimeState),
+    /// State when the job has finished.
+    Done(JobTimeState),
```
rsheeter, line 242: These comments just duplicate what the type tells you. "Running(State) is the state when running" is ... noise. I suggest we instead migrate some of the old comments from each state, e.g. queued has notes about threads, etc.

PR #1487, opened 2025-05-23
Ensure normalized coords are clamped during overlay

cmyr:
It's possible for the design coordinates to be out of bounds, in which case the normalized coordinates can be out of bounds.

JMM


rsheeter (approved):
Derp!

PR #1489, opened 2025-05-25
Write variable vertical metrics

Hoolean:
This is the final puzzle piece from #1388 (🎊): writing **variable** heights to eliminate the remaining diffs for vertical fonts on crater. In particular, the main changes in this PR are to:

- Refactor `HVAR` for reuse to write `VVAR`
- Populate vertical `gvar` phantom points

## Question before merging

This PR will need furnishing with tests. However, before doing so, there is one issue I would like to address:

When a glyph instance's height or vertical origin is absent, the default value must be derived from the global metrics _at that location_. This can sometimes crash fontc, as outlines may be at positions where global metrics are undefined, e.g. for designspace sparse layers or Glyphsapp brace layers, and [fontc does not implement metrics interpolation yet](https://github.com/googlefonts/fontc/blob/ec53b31eac40a9f5e5977817d8918e8aa2d04b75/fontir/src/ir.rs#L504).

What solution do we want to get vertical metrics over the line?

1. If we fetch metrics as lazily as possible, we can limit present-day panics to a much rarer and avoidable case that is unlikely to appear on crater
    - i.e. by only getting metrics when a default is really needed, only vertical fonts that define sparse glyphs but do not give explicit heights or origins are affected
2. Alternatively, we  wait for interpolated metrics
3. ...or is there a third option?

### DIFF
--- a/fontbe/src/font.rs
+++ b/fontbe/src/font.rs
@@ -9,7 +9,7 @@ use write_fonts::{
         avar::Avar, cmap::Cmap, colr::Colr, cpal::Cpal, fvar::Fvar, gasp::Gasp, gdef::Gdef,
         glyf::Glyf, gpos::Gpos, gsub::Gsub, gvar::Gvar, head::Head, hhea::Hhea, hmtx::Hmtx,
         hvar::Hvar, loca::Loca, maxp::Maxp, meta::Meta, mvar::Mvar, name::Name, os2::Os2,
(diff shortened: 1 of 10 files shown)


Hoolean:
(the failing tests are from the issue described above)

Hoolean:
A small update: this can be rebased on `main` to resolve the Qs in the PR after #1518 is merged

Issue #1020, opened 2024-10-10
fonttools' special handling of standard axis names leads to name table diffs vs fontc


anthrotype:
fontTools.varLib (used by fontmake to build VFs) contains a hardcoded set of axis names that receive special treatment, see: 

https://github.com/fonttools/fonttools/blob/afceebcda515b779faabe0acdec13018bab0c246/Lib/fontTools/varLib/__init__.py#L890-L921

this is a remnant of the days when Designspace document did not yet have an explicit `<axes>` element defining the set of axes used therein, with attributes such as the `name` used internally to reference them from locations, the four-letter `tag` to use in fvar, or more recently `labelname` elements with human-readable localized strings for UI use.

There were designspaces containing axes with the name attribute set to "weight", "width", "slant", "optical" and "italic" (in lowercase following a convention probably coming from MutatorMath), and varLib would assign these special group of axes a four-letter tag and UI label to store in the name table. Note that any other axis was not even supported at that stage.

```python
    standard_axis_map = OrderedDict(
        [
            ("weight", ("wght", {"en": "Weight"})),
            ("width", ("wdth", {"en": "Width"})),
            ("slant", ("slnt", {"en": "Slant"})),
            ("optical", ("opsz", {"en": "Optical Size"})),
            ("italic", ("ital", {"en": "Italic"})),
        ]
    )
```

Even after `<axes>` element was spec'ed and implemented (at some point it became even required, forgot exactly when), these "standard" axes continued to be handled specially as you can see from the above linked code.
Probably this was done to continue supporting old files with fewer or no modifications, I don't exactly remember.

Note in particular this bit below.

```python
        if axis_name in standard_axis_map:
            if axis.tag is None:
                axis.tag = standard_axis_map[axis_name][0]
            if not axis.labelNames:
                axis.labelNames.update(standard_axis_map[axis_name][1])
```

So, if an axis name was "weight" (lowercase) - thus one of the standard axis names - and it didn't also define explicit `labelname` sub-elements (these are usually omitted unless non-english localised strings for axis names are desired), the required English string used to show the axis in UI would be "Weight" (title case); "width" => "Width", ..., "optical" => "Optical Size", etc.
Whereas for axes whose name is not in that standard set, the English UI label name would simply be the same as the `name` attribute, so no special handling.

(But what if I really want to name my axis "optical" all lowercase?! Well you can't, unless you add a `labelname` element with xml:lang="en" attribute, which most don't (glyphsLib does not)).

[...]

PR #1490, opened 2025-05-27
Prefer axis (English) `<labelname>` over axis.name like fonttools does

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/1020

Some .designspace files contain axis `<labelname>` elements with an English string "en" which may be different from the default axis name attribute (there was a convention to have axis names all lowercase in the old MutatorMath days, which some tools have kept).

fonttools varLib prefers to use the "en" labelname when present over the axis.name. This PR is meant to match this.

Note this requires a change in norad which hasn't been released yet.

I don't plan to add support in here for non-English localised names proper. We will do that later at some point.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -1290,6 +1290,47 @@ mod tests {
         )
     }
 
+    #[test]
+    fn custom_english_ui_label_name() {
+        // The 'weight' axis has a custom `<labelname xml:lang="en">Heaviness</labelname>`;
+        // we expect this to be used for the axis name for UI purposes (fvar, STAT).
+        let result = TestCompile::compile_source("wght_var_axis_label_names.designspace");
+        let static_metadata = result.fe_context.static_metadata.get();
+
+        assert_eq!(
+            vec!["weight".to_string()],
+            static_metadata
+                .axes
+                .iter()
+                .map(|axis| axis.name.clone())
+                .collect::<Vec<_>>()
+        );
+
+        let font = result.font();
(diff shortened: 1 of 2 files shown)


fontc/src/lib.rs
```diff
@@ -1290,6 +1290,47 @@ mod tests {
...
+        let stat = font.stat().unwrap();
+
+        assert_eq!(
+            vec!["Heaviness".to_string()],
+            stat.design_axes()
+                .unwrap()
+                .iter()
+                .map(|axis| resolve_name(&name, axis.axis_name_id()).unwrap())
+                .collect::<Vec<_>>()
+        )
```
rsheeter, line 1331: nit: this test would be easier to read with helpers, right now theres a lot of x.unwrap.iter.map.collect noise

glyphs2fontir/src/toir.rs
```diff
@@ -219,6 +219,8 @@ fn to_ir_axis(
         default: default.to_user(&converter),
         max: max.to_user(&converter),
         converter,
+        // TODO: support localized axis names from .glyphs sources
```
rsheeter, line 222: Please file and cite an issue, as TODO it' sjust lost

rsheeter (approved):

fontc/src/lib.rs
anthrotype: fvar.axes() and stat.design_axes() are different type, i'd need to implement a trait and am feeling lazy just for this test

PR #1492, opened 2025-05-27
[fea-rs] Disable optimization for empty ConditionSets

cmyr:
It is more efficient to write null offsets in this case, but fonttools does not perform this optimization and it is a minor optimization that applies to a small number of fonts, so it feels fine to disable for now.

This gives us a cheap +1 on crater.

see https://github.com/fonttools/fonttools/issues/3844 for discussion on adding this optimization in fonttools.

JMM


rsheeter:
This sounds like an unfortunate change to me. Will ping on IM.

PR #1498, opened 2025-05-28
update fonttools in requirements.txt

anthrotype:
https://github.com/fonttools/fonttools/releases/tag/4.58.1


anthrotype:
clippy failure is unrelated

PR #1499, opened 2025-05-28
[fea] Handle nested contextual lookups in aalt

cmyr:
If an aalt feature had a contextual lookup that referenced a contextual lookup we were not adding rules from any of _those_ lookups referenced lookups to aalt.

Now we do!

gives us at least a +1 on crater.

JMM


dfrg (approved):
LGTM 👍

PR #1501, opened 2025-05-29
Determine GDEF categories for bracket glyphs too

Hoolean:
fontmake includes these for e.g. Noto Sans Thai Looped, and as [glyphsLib calculates them for each glyph in the generated UFO](https://github.com/googlefonts/glyphsLib/blob/bd8276222afad582c030f839d90271bc0ee70d0b/Lib/glyphsLib/builder/features.py#L213-L272) then we can follow this behaviour by also including the new bracket glyphs when deriving our own.

This PR assumes that bracket glyphs have the same derived category as their parent. This should generally be a safe assumption, as the category is only derived from anchors and the parent glyph's categories. On the other hand, with some light refactoring we could derive it from the bracket glyphs' layers directly if we believe that the bracket anchor names could be different; happy to make that change if that is the case :)

(test to follow once we have settled on the approach)

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -310,7 +310,7 @@ impl Work<Context, WorkId, Error> for StaticMetadataWork {
             selection_flags |= SelectionFlags::REGULAR;
         }
 
-        let categories = make_glyph_categories(font);
+        let categories = make_glyph_categories(font_info);
 
         // Only build vertical metrics if at least one glyph defines a vertical
         // attribute.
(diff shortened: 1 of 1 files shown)


anthrotype:
> This PR assumes that bracket glyphs have the same derived category as their parent.

yeah, I think that's a safe assumption to make. If the categories do end up being different in glyphsLib because bracket glyphs happen to have different anchors from their associated master glyphs, it's probably just a bug.

cmyr (approved):
looks good!

Hoolean:
Ta both :rocket: I've added a unit test for good measure

PR #1504, opened 2025-05-29
[glyphs] Propagate component bracket layers

cmyr:
One of the larger remaining todos for bracket layers: if a glyph does not have bracket layers but it has components that do, we need to create dummy bracket layers in order for the components to be set correctly.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -323,6 +323,18 @@ impl Layer {
         })
     }
 
+    /// NOTE: panics if called on a non-bracket layer
+    fn bracket_info(&self, axes: &[Axis]) -> BTreeMap<String, (Option<i64>, Option<i64>)> {
+        assert!(
+            !self.attributes.axis_rules.is_empty(),
+            "all bracket layers have axis rules"
+        );
+        axes.iter()
+            .zip(&self.attributes.axis_rules)
+            .map(|(axis, rule)| (axis.tag.clone(), (rule.min, rule.max)))
+            .collect()
+    }
+
     // TODO add is_alternate, is_color, etc.
 }
 
@@ -2898,6 +2910,10 @@ impl Font {
     pub fn load(glyphs_file: &path::Path) -> Result<Font, Error> {
         let mut font = Self::load_raw(glyphs_file)?;
 
+        // ensure that glyphs with components that have bracket layers
+        // also have bracket layers.
+        font.align_bracket_layers();
+
         // propagate anchors by default unless explicitly set to false
         if font.custom_parameters.propagate_anchors.unwrap_or(true) {
             font.propagate_all_anchors();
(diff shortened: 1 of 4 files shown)


glyphs-reader/src/font.rs
```diff
@@ -323,6 +323,18 @@ impl Layer {
         })
     }
 
+    /// NOTE: panics if called on a non-bracket layer
```
rsheeter, line 326: why is that desirable? Just return None or an empty map or something? Or return an error?

glyphs-reader/src/font.rs
```diff
@@ -2898,14 +2910,91 @@ impl Font {
...
-    // load without propagating anchors
+    /// if a glyph has components that have alternate layers, copy the layer
+    /// locations into the glyph.
+    ///
+    /// This is required to correctly handle bracket glyphs.
+    // https://github.com/googlefonts/glyphsLib/blob/c4db6b981d/Lib/glyphsLib/builder/bracket_layers.py#L176
+    fn align_bracket_layers(&mut self) {
+        // python does this recursively but we'll presort instead
+        let todo = self.depth_sorted_composite_glyphs();
+        //let mut problematic_glyphs = HashMap::new();
```
rsheeter, line 2932: delete me?

glyphs-reader/src/font.rs
```diff
@@ -2898,14 +2910,91 @@ impl Font {
...
+                            .or_insert(HashSet::new())
+                            .extend(missing.cloned());
+                    }
+                }
+            }
+
+            if needed.is_empty() {
+                // all good, next glyph
+                continue;
+            }
```
rsheeter, line 2964: This is a bit long and seems to have several distinct tasks that would benefit from becoming fns.

Suggest hoisting out a fn so you can have loop body only process things that need work, e.g.

```rust
for (name, (master, missing_brackets)) in todo.iter().filter_map(...compute Option<master, missing_brackets>...) {
   ...do the fixup work...
```

glyphs-reader/src/propagate_anchors.rs
```diff
@@ -22,14 +22,24 @@ impl Font {
     pub fn propagate_all_anchors(&mut self) {
         propagate_all_anchors_impl(&mut self.glyphs);
     }
+
+    /// returns a list of all glyphs, sorted by component depth.
```
rsheeter, line 26: all glyphs or only composites? - the commen tsays all, the fn name suggests only composites

resources/testdata/glyphs3/glyph-with-bracket-component.glyphs
```diff
@@ -0,0 +1,187 @@
...
+"258ED206-D907-4350-8EBA-CBDDA1C16BF7" = 1;
+};
+name = Black;
+weightClass = 900;
+}
+);
+unitsPerEm = 1000;
+versionMajor = 3;
+versionMinor = 0;
+}
```
rsheeter, line 187: This looks very likely to contain data you don't need? For example, do you require instance interpolations, instances, or non-empty shapes?

glyphs-reader/src/font.rs
cmyr: it's internal API, and should only ever be called when working with things we know are bracket layers, which are already sorted (as in they're in `glyph.bracket_layers` instead of `glyph.layers`).

this is basically enforcing an invariant.

glyphs-reader/src/font.rs
cmyr: this is another case where I am directly transcribing the python, and trying to match the structure as closely as possible because it makes it much easier to identify possible transcription errors.... several times in this kind of code I've used the wrong glyph name as a key or failed to invert a signed number or similar, and I find it very valuable to be able to look at the two impls side-by-side and and line-by-line.

rsheeter: I don't think extracting functions should prevent alignment with python?

Tests - which could be finegrained if you broke up small functions - should help as well.

Finally, I expect to own/operate the Rust code for an extended period so I think a little more pain now pays off for the entire future.

PR #1511, opened 2025-06-04
pin cattrs<25.1 in requirements.txt to workaround bug

anthrotype:
... affecting UFO json [de]serialization in ufoLib2 (as used by gftools builder when compiling static instances with fontmake)

We shall remove this once my fix gets published upstream.


anthrotype:
I bumped cattrs alongside everything else with https://github.com/googlefonts/fontc/pull/1509, but just realised that v25.1.0 has contains a regression which impacts ufoLib2 and consequently gftools.builder.
Keeping it temporarily pinned should avoid a bunch of new fontmake-only failures in fontc_crater

PR #1513, opened 2025-06-05
[glyphs] Expand kerning to bracket glyphs

cmyr:
This is the last remaining thing I'm aware of, related to bracket glyphs, and should fix a few outstanding mark/kern diffs on crater.

The test + test file was taken from glyphsLib and cleaned up a little bit.

(edit: this should actually give us a healthy +identical on crater)

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -905,27 +905,30 @@ impl Work<Context, WorkId, Error> for KerningGroupWork {
 
         let mut groups = KerningGroups::default();
 
-        // If glyph uses a group for either side it goes in that group
-        font.glyphs
+        // build up the kern groups; a glyph may belong to a group on either or
+        // both 'side'.
+        for (name, glyph) in font
+            .glyphs
             .iter()
             // ignore non-export glyphs
-            .filter(|(_, glyph)| glyph.export)
-            .flat_map(|(glyph_name, glyph)| {
-                glyph
-                    .right_kern
-                    .iter()
-                    .cloned()
-                    .map(KernGroup::Side1)
-                    .chain(glyph.left_kern.iter().cloned().map(KernGroup::Side2))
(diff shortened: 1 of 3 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -905,27 +905,21 @@ impl Work<Context, WorkId, Error> for KerningGroupWork {
...
-                    .map(|group| (group, GlyphName::from(glyph_name.as_str())))
-            })
-            .for_each(|(group_name, glyph_name)| {
-                groups
-                    .groups
-                    .entry(group_name)
-                    .or_default()
-                    .insert(glyph_name);
-            });
+        for (name, glyph) in font.glyphs.iter().filter(|x| x.1.export) {
```
rsheeter: I don't think deleting the comments is an improvement, perhaps restore similar?

glyphs2fontir/src/source.rs
```diff
@@ -2057,6 +2114,71 @@ mod tests {
...
+        ) -> BTreeMap<(KernSide, KernSide), OrderedFloat<f64>> {
+            items
+                .iter()
+                .map(|(side1, side2, val)| {
+                    ((parse_side(side1), parse_side(side2)), (*val as f64).into())
+                })
+                .collect()
+        }
+
+        // actual test:
```
rsheeter, line 2150: this would be obvious if the helpers weren't inline

resources/testdata/glyphs2/BracketTestFontKerning.glyphs
```diff
@@ -0,0 +1,397 @@
...
+{
+name = "Disable Last Change";
+value = 1;
+}
+);
+familyName = "New Font";
+fontMaster = (
+{
+ascender = 800;
+capHeight = 700;
```
rsheeter: Looks possible to minimize the test input, for example do you actually care about these for this test?  What about disable last change, etc?

rsheeter (approved):

PR #1514, opened 2025-06-05
[glyphs] Use axis defaults when generating ConditionSet

cmyr:
Instead of just assuming that all axes are going to have a normalized minimum value of -1.0, which is not the case when min == default.


rsheeter (approved):
Oops, I think I missed this in an earlier review :(

PR #1518, opened 2025-06-07
Implement interpolation of global metrics

Hoolean:
This PR moves the representation of metrics as deltas earlier, to allow us to calculate interpolated metrics.

This has a few use cases; mostly it should help with #1402 and #1489.

Most of the diff is from moving the conversion earlier, and updating tests to use the new builder pattern. Aside from this there are very few changes, aside from a few `(ufo|glyphs)2ir` test constants that needed updating to expect the IR to contain rounded metrics values already.

I anticipate that this PR will need a little more tidying and documentation, and so am submitting it now to get feedback on the general direction first :)

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -25,7 +25,7 @@ use fontdrasil::{
 use crate::{
     error::{BadAnchor, BadAnchorReason, BadGlyph, BadGlyphKind, Error},
     orchestration::{IdAware, Persistable, WorkId},
-    variations::{VariationModel, VariationRegion},
+    variations::{ModelDeltas, VariationModel},
 };
 
 mod path_builder;
@@ -302,7 +302,7 @@ pub struct GlobalMetricsBuilder(
 ///
 /// This type is constructed by [GlobalMetricsBuilder].
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
-pub struct GlobalMetrics(HashMap<GlobalMetric, GlobalMetricDeltas>);
+pub struct GlobalMetrics(HashMap<GlobalMetric, ModelDeltas<f64>>);
 
 #[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub enum GlobalMetric {
@@ -397,9 +397,6 @@ fn adjust_offset(offset: f64, angle: f64) -> f64 {
         0.0
     }
 }
-
-/// Map of deltas associated with a given global metric across regions.
-pub type GlobalMetricDeltas = Vec<(VariationRegion, Vec<f64>)>;
 
 impl GlobalMetricsBuilder {
     /// Creates an empty [GlobalMetricsBuilder], to furnish with metrics data
@@ -615,7 +612,7 @@ impl GlobalMetrics {
     ///
     /// This lower level of access is useful for interpolating the value at an
     /// invididual positition manually, and testing.
-    pub fn deltas(&self, metric: GlobalMetric) -> &GlobalMetricDeltas {
+    pub fn deltas(&self, metric: GlobalMetric) -> &ModelDeltas<f64> {
         // We presume that ctor initializes for every GlobalMetric
         self.0.get(&metric).unwrap()
     }
@@ -676,7 +673,7 @@ impl GlobalMetrics {
     ///
     /// This lower level of access is most useful for serialising the variation
     /// spaces.
-    pub fn iter(&self) -> impl Iterator<Item = (&GlobalMetric, &GlobalMetricDeltas)> + '_ {
+    pub fn iter(&self) -> impl Iterator<Item = (&GlobalMetric, &ModelDeltas<f64>)> + '_ {
         self.0.iter()
     }
 }
--- a/fontbe/src/mvar.rs
+++ b/fontbe/src/mvar.rs
@@ -8,9 +8,8 @@ use fontdrasil::{
     orchestration::{Access, Work},
     types::Axes,
 };
-use fontir::{
-    ir::GlobalMetricDeltas, orchestration::WorkId as FeWorkId, variations::VariationModel,
-};
+use fontir::variations::ModelDeltas;
+use fontir::{orchestration::WorkId as FeWorkId, variations::VariationModel};
 use write_fonts::types::MajorMinor;
 use write_fonts::{
     tables::{
@@ -50,7 +49,7 @@ impl MvarBuilder {
         }
     }
 
-    fn add_deltas(&mut self, mvar_tag: Tag, deltas: &GlobalMetricDeltas) {
+    fn add_deltas(&mut self, mvar_tag: Tag, deltas: &ModelDeltas<f64>) {
         if deltas.len() == 1 {
             let (region, _) = deltas.first().unwrap();
             assert!(region.is_default());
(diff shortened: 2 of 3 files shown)


fontir/src/ir.rs
```diff
@@ -274,14 +278,17 @@ impl From<KernGroup> for KernSide {
...
 /// Global metrics. Ascender/descender, cap height, etc.
 ///
 /// Represents the values of these metrics at a specific position in design space.
 /// At a minimum should be defined at the default location.
-#[derive(Serialize, Deserialize, Default, Debug, Clone, PartialEq, Eq)]
-pub struct GlobalMetrics(
-    pub(crate) HashMap<GlobalMetric, HashMap<NormalizedLocation, OrderedFloat<f64>>>,
-);
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct GlobalMetrics(HashMap<GlobalMetric, GlobalMetricValues>);
```
cmyr: Can we add docs to the effect that "This type is constructed with a [`GlobalMetricsBuilder`]"?

fontir/src/ir.rs
```diff
@@ -274,14 +278,17 @@ impl From<KernGroup> for KernSide {
     }
 }
 
+#[derive(Default, Debug)]
+pub struct GlobalMetricsBuilder(
```
cmyr, line 296: In general I encourage docs for all public types.

fontir/src/ir.rs
```diff
@@ -553,6 +551,48 @@ impl GlobalMetrics {
         }
     }
 
+    pub fn build(self, axes: &Axes) -> Result<GlobalMetrics, Error> {
```
cmyr, line 583: - I would probably rename `set` above to something more builder-y, like 'add'?
- I would document the `build` method, and in particular I would describe the circumstances in which it will return an error.

cmyr (approved):
Sorry to let this sit, it I think this is a very good idea!

Hoolean:
Sorry for the delay, not had much personal time in the past few weeks to push this forward but back and at 'em now :rocket:

I've rebased, updated the documentation with the review in mind, and fixed a few outstanding renames or TODOs that were owed after the refactor. If the tests all pass, then we may be ready to go :)

fontir/src/ir.rs
Hoolean: Updated the docs for the struct, and those in the its enclosed functions – especially where their exact meaning has changed after the refactor

fontir/src/ir.rs
Hoolean: Documented everything here too now

fontir/src/ir.rs
Hoolean: For the former: I've now documented the `build` function, and the main error cases to consider.

For the latter: I've left this as `set` to reflect that you can clobber existing values with it, but updated the docs in kind; I tried out the alternative name as normally `add` shouts build to me more too, but along with the previous concern it also made the patch a lot heavier, and so I think we should be fine to keep it as-is if we're happy to.

fontir/src/ir.rs
```diff
@@ -284,15 +288,22 @@ impl From<KernGroup> for KernSide {
     }
 }
 
-/// Global metrics. Ascender/descender, cap height, etc.
+/// Used to construct a global metrics variation space [GlobalMetrics], from
```
cmyr:
```suggestion
/// Used to construct a [GlobalMetrics] variation space, from
```

fontir/src/ir.rs
```diff
@@ -387,11 +398,12 @@ fn adjust_offset(offset: f64, angle: f64) -> f64 {
     }
 }
 
-/// Map of values associated with a given global metric at various locations.
-pub type GlobalMetricValues = HashMap<NormalizedLocation, OrderedFloat<f64>>;
+/// Map of deltas associated with a given global metric across regions.
+pub type GlobalMetricDeltas = Vec<(VariationRegion, Vec<f64>)>;
```
cmyr: this is so slightly confusing... presumably the values at the default location are not deltas? And the values at other locations are relative to the default? This might need to be spelled out more explicitly.

fontir/src/ir.rs
```diff
@@ -543,6 +548,8 @@ impl GlobalMetrics {
         self.values_mut(metric).insert(pos, value.into());
     }
 
+    /// Specify the value of a metric at a specific location, but only if there
+    /// is not one already provided. Returns the winning value.
```
cmyr: Treat docs like git commits: one short summary line, then an empty line, then any elaboration as required, e.g.:

```suggestion
    /// Set the value of a metric at a specific location, if none exists.
    ///
    /// If a value already exists at this location, it is not changed.
    /// 
    /// Returns the (potentially new, potentially preexisting) value.
```

fontir/src/ir.rs
```diff
@@ -552,6 +559,9 @@ impl GlobalMetrics {
         *self.values_mut(metric).entry(pos).or_insert(value.into())
     }
 
+    /// Specify the value of a metric at a specific location, if we have one;
+    /// this is a helper build method to reduce boilerplate. If there is an
+    /// existing value, it will be overwritten.
```
cmyr:
```suggestion
    /// A helper method for setting possibly missing values.
    ///
    ///
    /// This is a helper build method to reduce boilerplate. If there is an
    /// existing value, it will be overwritten.
```

fontir/src/ir.rs
```diff
@@ -563,6 +573,60 @@ impl GlobalMetrics {
...
+            .collect::<Result<HashMap<_, _>, _>>()?;
+
+        Ok(GlobalMetrics(deltas))
+    }
+}
+
+impl GlobalMetrics {
+    /// Access the raw deltas for a particular metric. This lower level of
+    /// access is useful for interpolating the value at an invididual positition
+    /// manually, and testing.
```
cmyr:
```suggestion
    /// Access the raw deltas for a particular metric.
    ///
    /// This lower level of access is useful for interpolating the value at an
    /// invididual positition manually, and testing.
```

fontir/src/ir.rs
```diff
@@ -602,7 +666,9 @@ impl GlobalMetrics {
         }
     }
 
-    pub fn iter(&self) -> impl Iterator<Item = (&GlobalMetric, &GlobalMetricValues)> + '_ {
+    /// Access the raw deltas for every metric. This lower level of access is
+    /// most useful for serialising the variation spaces.
```
cmyr:
```suggestion
    /// Access the raw deltas for every metric.
    ///
    /// This lower level of access is most useful for serialising the variation spaces.
```

cmyr (approved):
Looks great! Some little nits about docs inline, but happy to merge this :)

fontir/src/ir.rs
Hoolean: Re-wrapped locally and applied

fontir/src/ir.rs
Hoolean: Ta Colin, I've adopted these suggestions and some others locally, and will follow this for future PRs too

fontir/src/ir.rs
Hoolean: Trimmed extra line locally and applied

fontir/src/ir.rs
Hoolean: Applied locally

fontir/src/ir.rs
Hoolean: Applied locally

fontir/src/ir.rs
Hoolean: Agreed; this type was a leftover from the previous approach, but now that it's the baby of `VariationModel` it's not so distinct. With that in mind, I've moved it there and fleshed out the documentation to cover those special cases, its expected construction and use, and relation to OpenType in general.

Hoolean:
Addressed feedback – I've left the suggestions unresolved in case you want to double-check any

PR #1520, opened 2025-06-09
[ttx_diff] Normalize name table indentation

cmyr:
Some of the modifications we make to the name table can change the indentation. With this patch we normalize indentation after the last modification to this table.

JMM


anthrotype:
just out of curiosity (don't mean to block/revert), can you tell me a font that reproduces this particular issue?

cmyr:
I believe it was Fustat.

PR #1521, opened 2025-06-09
[glyphs] Process bracket rules in glyph-order-order

cmyr:
Instead of in alphabetical order by glyph name, which is what we get when we just iterate the map of glyphs.

This should give us a +small bump on crater.


resources/testdata/glyphs3/bracket-rules-processed-in-glyph-order.glyphs
rsheeter, line 1: as ever ... do you actually need all of this or can teh file be reduced?

rsheeter (approved):

anthrotype (approved):

PR #1525, opened 2025-06-10
Make handling Glyphs Instance postscript name closer to glyphsLib

khaledhosny:
GlyphsLib checks for `postscriptFontName` in instance properties and custom parameters, while fontc checks for `variablePostscriptFontName` property only. GlyphsLib propably should be updated to check for `variablePostscriptFontName` as well.


glyphs-reader/src/font.rs
```diff
@@ -2703,10 +2708,15 @@ impl Instance {
     /// Get the optional postscript name to use for the `fvar` named instance.
     pub fn postscript_name(&self) -> Option<&str> {
         // https://handbook.glyphsapp.com/custom-parameter-descriptions/
-        self.properties
+        // https://github.com/googlefonts/glyphsLib/blob/bd8276222afad582c030f839d90271bc0ee70d0b/Lib/glyphsLib/builder/instances.py#L118
+        ["variablePostscriptFontName", "postscriptFontName"]
```
cmyr, line 2712: Is there any concern about precedence?

cmyr (approved):
Looks good, I'm going to do a quick crater run locally and then happy to see this merged :)

glyphs-reader/src/font.rs
anthrotype: > Is there any concern about precedence?

in what sense?

cmyr: this will stop at the first thing it finds, so we will always prefer the first key to the second one.

anthrotype: well, that's also what glyphsLib does, so yes

https://github.com/googlefonts/glyphsLib/blob/7a8f643a711ffb9d351d037425a7eb60759e6219/Lib/glyphsLib/builder/instances.py#L118-L121

(that's the whole point of the PR if I understand correctly)

khaledhosny: Yes, this was the intention.

Issue #918, opened 2024-08-27
Glyphs instances 'Axis Location' custom parameters are ignored


anthrotype:
Currently we only search for 'Axis Location' custom parameters in the Glyphs masters. However, the instances custom parameters can also define 'Axis Location', which complement the 'Axis Location's  defined in the masters, and contribute to define the global axis mappings for avar.

PR #1528, opened 2025-06-11
Allow duplicate axis mappings values; support instance 'Axis Location' parameters

anthrotype:
duplicate keys (equivalent to avar's `fromCoordinate`) are bad, but duplicate values (avar's `toCoordinate`) are fine and should not be skipped.

This bug can be reproduced with [FamiljenGrotesk-Italic.glyphs](https://github.com/Familjen-Sthlm/Familjen-Grotesk/blob/master/sources/FamiljenGrotesk-Italic.glyphs)

```
python resources/scripts/ttx_diff.py 'https://github.com/Familjen-Sthlm/Familjen-Grotesk#sources/FamiljenGrotesk-Italic.glyphs'
```

```diff
--- /Users/clupo/oss/fontc/build/gftools/fontc.avar.ttx	2025-06-11 16:05:37
+++ /Users/clupo/oss/fontc/build/gftools/fontmake.avar.ttx	2025-06-11 16:05:37
@@ -4,6 +4,7 @@
       <mapping from="-1.0" to="-1.0"/>
       <mapping from="0.0" to="0.0"/>
       <mapping from="0.32" to="0.5"/>
+      <mapping from="0.3333" to="0.5"/>
       <mapping from="0.6667" to="0.75"/>
       <mapping from="1.0" to="1.0"/>
     </segment>
```

I first push an updated (failing) test and then follow up with the actual fix

https://github.com/googlefonts/fontc/actions/runs/15588614719/job/43901447309?pr=1528#step:6:371


anthrotype:
I took the opportunity to also fix #918 since I was touching the very same code

rsheeter (approved):
Good catch on the or test for duplication!

PR #1538, opened 2025-06-17
Propagate anchors to bracket layers

cmyr:
Based on https://github.com/googlefonts/glyphsLib/pull/1088, with this PR we will propagate anchors from components to bracket layers.

This does not handle computing GDEF categories for bracket glyphs, which I will PR separately.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -312,6 +312,12 @@ impl Layer {
             .unwrap_or(&self.layer_id)
     }
 
+    /// A key used to identify a bracket layer during anchor propagation
+    pub(crate) fn axis_rules_key(&self) -> Option<String> {
+        (!self.attributes.axis_rules.is_empty())
+            .then(|| format!("{:?} {}", self.attributes.axis_rules, self.master_id()))
+    }
+
     pub fn is_intermediate(&self) -> bool {
         self.associated_master_id.is_some() && !self.attributes.coordinates.is_empty()
     }
--- a/glyphs-reader/src/propagate_anchors.rs
+++ b/glyphs-reader/src/propagate_anchors.rs
@@ -49,25 +49,40 @@ fn propagate_all_anchors_impl(glyphs: &mut BTreeMap<SmolStr, Glyph>) {
     let mut all_anchors = HashMap::new();
     for name in todo {
         let glyph = glyphs.get(&name).unwrap();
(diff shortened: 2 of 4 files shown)


glyphs-reader/src/font.rs
```diff
@@ -312,6 +312,12 @@ impl Layer {
             .unwrap_or(&self.layer_id)
     }
 
+    /// A key used to identify a bracket layer during anchor propagation
+    pub(crate) fn axis_rules_key(&self) -> Option<String> {
```
rsheeter, line 316: Would SmolStr make sense?

rsheeter: Or a wee type?

glyphs-reader/src/propagate_anchors.rs
```diff
@@ -49,26 +49,41 @@ fn propagate_all_anchors_impl(glyphs: &mut BTreeMap<SmolStr, Glyph>) {
...
-                &mut num_base_glyphs,
-            );
+        for layer in glyph.layers.iter().chain(glyph.bracket_layers.iter()) {
+            let anchors =
+                anchors_traversing_components(glyph, layer, &all_anchors, &mut num_base_glyphs);
             maybe_log_new_anchors(&anchors, glyph, layer);
-            all_anchors.entry(name.clone()).or_default().push(anchors);
+            // if this is a bracket layer we use the actual axis values as the key,
+            // since it's possible that layers with the same axis values do not share
+            // the same layer id.
```
rsheeter, line 58: This comment would go great on a key type!

cmyr: agree this feels a bit hacky, but a key type is also a challange: the issue is that we want the map to store both these custom keys (for bracket layers) as well as just master ids (uuid-type strings) and  then when we start mixing types like that it turns into a PITA because in the ID case you can't just borrow the name anymore, i.e. instead of going,

```rust
map.get(&layer.name)
```

you need to do,

```rust
map.get(&MyKey::JustAnId(layer.name.clone()))
```

and personally it doesn't quite feel worth it for a fairly self-contained internal API?

glyphs-reader/src/font.rs
cmyr: It's not a SmolStr simply because these aren't small enough, and they'd always end up in the allocation path anyway. :(

PR #1540, opened 2025-06-18
Library cleanliness

simoncozens:
Following Rod's comments on #1478, this cleans up the library interface to fontc by:

* Reducing WASM cfg ifdefs by choosing once between real and fake clocks for timing
* Separating out CLI-specific functions such argument parsing and moving the Clap dependency to the binary
* Merging "on disk" and "in memory" sources into a single `Input` type
* Sharing common code between the library and binary entry points

The clocks stuff turned out to be a little gnarly because of the use of `Instant` in the timing structs. I broke it down into fairly small atomic commits for ease of reviewing.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -81,12 +81,7 @@ pub fn run(args: Args, mut timer: JobTimer) -> Result<(), Error> {
     )?;
     timer.add(time.complete());
 
-    let workload = Workload::new(
-        args.source(),
-        args.input_binary.as_ref(),
-        JobTimer::default(),
-        args.skip_features,
-    )?;
+    let workload = Workload::new(args.source(), None, JobTimer::default(), args.skip_features)?;
 
     let fe_root = FeContext::new_root(args.flags(), ir_paths);
     let be_root = BeContext::new_root(args.flags(), be_paths, &fe_root);
@@ -334,13 +329,8 @@ mod tests {
                 raw_font: Vec::new(),
             };
 
-            let mut workload = Workload::new(
-                args.source(),
-                args.input_binary.as_ref(),
-                timer,
-                args.skip_features,
-            )
-            .unwrap();
+            let mut workload =
+                Workload::new(args.source(), None, timer, args.skip_features).unwrap();
             let completed = workload.run_for_test(&result.fe_context, &result.be_context);
 
             result.work_executed = completed;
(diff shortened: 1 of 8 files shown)


fontc/src/lib.rs
```diff
@@ -20,69 +23,86 @@ use std::{
...
-/// when the source is not a file on disk (WASM or library use).
-///
-/// Currently only supports Glyphs sources, but could be extended to e.g. UFOZ or other formats.
-#[derive(Debug, Clone, PartialEq)]
-pub enum InMemorySource {
-    Glyphs(String),
+#[cfg(feature = "cli")]
+pub use crate::timing::RealClock;
+
+pub enum Input {
```
cmyr, line 40: I'd still give this a brief bit of documentation explaining the rationale.

fontc/src/lib.rs
```diff
@@ -20,69 +23,86 @@ use std::{
...
-    let workload = Workload::new(args.clone(), timer)?;
+    fn try_from(path: &Path) -> Result<Self, Self::Error> {
+        Input::new(path)
+    }
+}
 
-    let fe_root = FeContext::new_root(args.flags(), ir_paths);
-    let be_root = BeContext::new_root(args.flags(), be_paths, &fe_root);
-    let mut timing = workload.exec(&fe_root, &be_root)?;
+#[cfg(feature = "cli")]
```
cmyr: no change required but generally i'm used to seeing any `#[attributes]` after any doc comments.

fontc/src/timing.rs
```diff
@@ -4,7 +4,44 @@
...
+use std::{
+    collections::HashMap,
+    io,
+    thread::ThreadId,
+    time::{Duration, Instant},
+};
+
+/// A trait for clock implementations, allowing for different time sources
+/// such as real time or fake time for testing.
+pub trait Clock: std::fmt::Debug + Clone + 'static {
```
cmyr: I really don't love using a trait for this, since it leaks this implementation detail out into the rest of the code... but I guess rod didn't like the other way? not a hill I need to die on.

fontc/src/timing.rs
```diff
@@ -4,7 +4,49 @@
...
-use std::{collections::HashMap, io, thread::ThreadId, time::Instant};
+use std::{collections::HashMap, io, thread::ThreadId};
+
+#[cfg(target_family = "wasm")]
+use dummy_instant::Instant;
+#[cfg(not(target_family = "wasm"))]
+use std::time::Instant;
+
+#[cfg(target_family = "wasm")]
+mod dummy_instant {
```
cmyr, line 17: I would add just a little summary doc on this module explaining how Instant is not available on wasm, and this is a workaround.

PR #1541, opened 2025-06-18
[glyphs] Match glyphsLib for component anchor propagation

cmyr:
I think glyphsLib is wrong on this count but it will be useful to match for the time being, and then we can easily revert this if we decide to change glyphsLib.

See https://github.com/googlefonts/glyphsLib/issues/1090

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2993,20 +2993,26 @@ impl Font {
         if font.custom_parameters.propagate_anchors.unwrap_or(true) {
             font.propagate_all_anchors();
         }
+
+        // ensure that glyphs with components that have bracket layers
+        // also have bracket layers.
+        font.align_bracket_layers();
+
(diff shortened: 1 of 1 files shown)


dfrg (approved):

anthrotype:
we should revert this once https://github.com/googlefonts/glyphsLib/pull/1092 is merged

glyphs-reader/src/font.rs
```diff
@@ -2993,20 +2993,26 @@ impl Font {
         if font.custom_parameters.propagate_anchors.unwrap_or(true) {
             font.propagate_all_anchors();
         }
+
+        // ensure that glyphs with components that have bracket layers
+        // also have bracket layers.
+        font.align_bracket_layers();
```
anthrotype, line 2999: (this call was missing before from load_from_string, so not just a simple git revert)

PR #1542, opened 2025-06-18
[glyphs] Prune default axes of unused

cmyr:
In some cases the default 'wdth' and 'XXXX' axes are referenced in instances, which is causing some miscompilations.

This fixes a number of fvar+name diffs on crater.


dfrg (approved):
Seems reasonable!

PR #1560, opened 2025-07-14
[crater] Update python deps for 3.13

cmyr:
This will need to correspond to a manual update of the python version on the CI machine.


rsheeter (approved):
LGTM but I think this might be the same as #1559, just without updating .github/workflows/release.yml?

PR #1561, opened 2025-07-14
[crater] Explicitly call python3 instead of python

cmyr:


cmyr:
just aiming for maximal portability? but 🤷

PR #1564, opened 2025-07-17
[crater] Rework representation of targets

cmyr:
This patch is kind of doing a bunch of stuff and I apologize for getting it all tied up together.

I. Preamble

Crater, like much software, was accreted more than it was designed. This has started to pose challanges.

The main problem is that historically we were only concerned with font source files; repositories and config files were just a way for us to discover sources.

Later on, we started to also care about config files, so that we could do gftools builds; but this work brought in a bunch of assumptions about where config files could be found, and how they related to sources.

II. This Commit

- The main purpose of this commit is to move us from being focused on sources to being focused on configs. We will only build something if we have a config, and for a given config we will only build a source that is named by that config.

- In addition to that, this commit tries to generally be more consistent about how we construct paths, and how we report errors if we cannot compile a given target.

- Finally, this commit adds preliminary support for 'virtual configs', which occur when we want to provide a config file for a repository that we do not otherwise want to (or are unable to) modify. These config files live in the google/fonts repository.

  More work will be required before we can actually build these virtual configs, but at least now we understand that they exist.


anthrotype:
> We will only build something if we have a config...

are there font source repositories without any configs in which we were previously discovering and building some sources whereas now we will no longer? Or is that covered by the 'virtual configs'?

cmyr:
No, previously we would rely on configs to _find_ sources, but we would then forget about them.

fontc_crater/src/ci.rs
```diff
@@ -242,44 +242,64 @@ fn make_targets(cache_dir: &Path, repos: &[FontSource]) -> ResolvedTargets {
...
         };
-        let relative_config_path = config_path
-            .strip_prefix(cache_dir)
-            .expect("config always in cache dir");
-        let sources_dir = config_path
-            .parent()
-            .expect("config path always in sources dir");
-        // config is always in sources, sources is always in org/repo
-        let repo_dir = relative_config_path.parent().unwrap().parent().unwrap();
+        // we map repos to URLS in a special sources.json file
```
rsheeter: I don't think I understand what "special" is trying to tell me? I wonder if we could describe what is in sources.json (or point to the description) more directly? - it's not entirely clear to me from this what I'm going to find there

fontc_crater/src/ci.rs
```diff
@@ -242,44 +242,64 @@ fn make_targets(cache_dir: &Path, repos: &[FontSource]) -> ResolvedTargets {
...
+        let repo_dir = repo.repo_path(Path::new(""));
```
rsheeter, line 260: nit: would read nicer without Path::new, could we modify repo_path to allow &str input?

fontc_crater/src/ci.rs
```diff
@@ -242,44 +242,64 @@ fn make_targets(cache_dir: &Path, repos: &[FontSource]) -> ResolvedTargets {
...
         result
             .source_repos
-            .insert(repo_dir.to_owned(), repo.repo_url.clone());
+            .insert(repo_dir.clone(), repo.repo_url.clone());
+
+        let sources_dir = if repo.config_is_virtual() {
```
rsheeter: "virtual" for a file that exists feels a touch off to me. Might "external" or even out_of_repo or some siuch, aiming to convey that the config doesn't live in this repo (but does live somewhere!) convey the concept more clearly?

rsheeter: `config_in_repo` seems like it would work? If the config is not in the repo { do something else } type of deal.

fontc_crater/src/target.rs
```diff
@@ -8,15 +8,20 @@ use std::{
...
 use serde::{Deserialize, Serialize};
-use thiserror::Error;
+
+static VIRTUAL_CONFIG_DIR: &str = "sources";
 
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub(crate) struct Target {
-    /// path to the source dir for this target (relative to the git cache root)
-    source_dir: PathBuf,
+    /// path to the source repo, from the cache dir root
```
rsheeter: Suggest "relative to cache root" instead of "from..."

fontc_crater/src/target.rs
```diff
@@ -8,15 +8,20 @@ use std::{
...
+    repo_dir: PathBuf,
     sha: String,
-    /// Filename of config file, in the source directory.
-    pub(crate) config: Option<PathBuf>,
+    /// Path to the config file.
+    ///
+    /// - From the cache root if it is virtual;
+    /// - From the repo root otherwise
+    pub(crate) config: PathBuf,
```
rsheeter, line 23: Similarly suggest "relative to..." instead of "from", clearer to at least this reader and consistent with how source is described below.

fontc_crater/src/target.rs
```diff
@@ -28,112 +33,77 @@ pub(crate) enum BuildType {
...
-    loop {
-        match result.file_name() {
-            Some(name) if name == "sources" || name == "Sources" => return Ok(result),
-            Some(_) | None => {
-                if !result.pop() {
-                    break;
-                }
-            }
+impl Target {
+    pub(crate) fn new2(
```
rsheeter: is new2 temporary? If not perhaps we could use names that convey something? The purpose or when it shoudl be used or some such?

fontc_crater/src/target.rs
```diff
@@ -239,48 +215,88 @@ impl<'de> Deserialize<'de> for Target {
     }
 }
 
+/// in the format,
+///
+/// $ORG/$REPO/$CONFIG_PATH?$SHA $SRC_PATH ($BUILD_TYPE)
+///
+/// where a virtual config's $CONFIG_PATH starts with the literal path element
```
rsheeter, line 222: I would find EXTERNAL or OUT_OF_REPO or similar more informative than VIRTUAL, virtual makes me think you synthesized it, not that it exists somewhere else.

rsheeter (approved):
LGTM subject to "virtual" changing, that term leads me (and I suspect others) to assume things that are entirely wrong

fontc_crater/src/ci.rs
cmyr: we could, but it isn't defined in this crate so would be a bit of a hassle.

fontc_crater/src/ci.rs
cmyr: 1) This method is also not defined here, we're just calling it.
2) I'm not sure I understand, are you proposing a new method called `config_in_repo`?

fontc_crater/src/target.rs
cmyr: oops yes this was leftover from earlier in the refactor, when I hadn't removed the previous constructor.

fontc_crater/src/target.rs
cmyr: So if this is a serious concern I'll need to go and make a separate PR + release in google_fonts_sources, which is where I first picked this terminology. Would you prefer that?

fontc_crater/src/ci.rs
felipesanches: > Finally, this commit adds preliminary support for 'virtual configs', which occur when we want to provide a config file for a repository that we do not otherwise want to (or are unable to) modify. These config files live in the google/fonts repository.

To me this sounds like a `config_override`.

PR #1567, opened 2025-07-18
fix non-deterministic panic in GlyfLocaWork

### LINKED ISSUE #1436: the check-ots job sometimes fails randomly?
this CI job sometimes fails at random:

https://github.com/googlefonts/fontc/blob/8ffba4caa50c68b4fc9916d5e77461e98b9e6f83/.github/workflows/rust.yml#L136-L145

For example see this run (from my #1430 PR):

https://github.com/googlefonts/fontc/actions/runs/14998438182/job/42138571730?pr=1430

re-running the job usually fixes it. I have no idea what's going on, but I file this lest I forget.

This has happened to me before but I was too lazy to delve deeper

anthrotype:
whereby 'Be(GlyfFragment(.notdef)) is not available' for a generated .notdef

(Courtesy of Gemini)

Fixes #1436

I actually don't know how to write a unit test this...


fontc/src/lib.rs
```diff
@@ -3703,4 +3718,31 @@ mod tests {
...
+        // Check that GlyfLocaWork starts with an unknown dependency, blocking it from running.
+        let glyf_loca_job = test
+            .workload
+            .jobs_pending
+            .get(&AnyWorkId::Be(BeWorkIdentifier::Glyf))
+            .expect("GlyfLocaWork should exist");
+        assert_eq!(
+            glyf_loca_job.read_access,
+            AnyAccess::Be(Access::Unknown),
+            "GlyfLocaWorka.read_access should start with Access::Unknown"
```
rsheeter: nit: typo GlyfLocaWork_a_

fontc/src/workload.rs
```diff
@@ -381,17 +381,28 @@ impl Workload {
...
                 // Glyph order is done so all IR must be done. Copy dependencies from the IR for the same name.
                 self.update_be_glyph_work(fe_root, glyph_name.clone());
             }
+
+            // Now that we have a final glyph order we can resolve the Access::Unknown for glyf/loca
+            let mut glyf_loca_deps = AccessBuilder::<AnyWorkId>::new()
+                .variant(FeWorkIdentifier::StaticMetadata)
+                .variant(FeWorkIdentifier::GlyphOrder);
+            for glyph_name in final_glyph_order.names() {
+                glyf_loca_deps = glyf_loca_deps
```
rsheeter, line 398: .specific_instance is primarily intended for construction of a fixed list in read/write access fns. In this context I find it a bit hard to read. I wonder if making `add_access` public and using that directly would read more clearly here...?

fontc/src/workload.rs
```diff
@@ -381,17 +381,28 @@ impl Workload {
...
+                    .specific_instance(BeWorkIdentifier::GlyfFragment(glyph_name.clone()));
+            }
+            let glyf_loca_job = self
+                .jobs_pending
+                .get_mut(&BeWorkIdentifier::Glyf.into())
+                .unwrap();
```
rsheeter: I suggest using expect and noting that if glyf isn't pending we have a bug in the message

fontc/src/workload.rs
anthrotype: hm I dunno, it reads fine to me. using add_access would simply be more verbose..

fontc/src/workload.rs
anthrotype: done

PR #1570, opened 2025-07-18
[glyphs] Handle RTL kerning

cmyr:
This matches glyphsLib, which just converts these into LTR kerning rules.

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -953,6 +953,8 @@ impl Work<Context, WorkId, Error> for KerningGroupWork {
 
         let mut groups = KerningGroups::default();
 
+        //https://github.com/googlefonts/glyphsLib/blob/682ff4b177/Lib/glyphsLib/builder/groups.py#L114
+        let rtl_glyphs = get_glyphs_with_rtl_kerning(font);
         // build up the kern groups; a glyph may belong to a group on either or
         // both 'side'.
         for (name, glyph) in font
@@ -961,6 +963,7 @@ impl Work<Context, WorkId, Error> for KerningGroupWork {
             // ignore non-export glyphs
             .filter(|x| x.1.export)
         {
+            let is_rtl = rtl_glyphs.contains(name.as_str());
             // if there are bracket layers for this glyph, make sure the
             // generated glyphs are assigned the same groups as the parent
             let bracket_names = bracket_glyph_names(glyph, &font_info.axes)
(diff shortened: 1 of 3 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -1047,16 +1048,43 @@ impl Work<Context, WorkId, Error> for KerningInstanceWork {
     }
 }
 
+#[allow(clippy::type_complexity)] // it's not _that_ bad
```
rsheeter: We have generally chosen to fix these. A type alias for the rather ugly map type seems like it would do it.

I would expect get_kerning to merely fetch, it's a bit weird for it to take a Cow and modify things. It's also deviating from the intended behavior that kerning is done, we're just fetching it at this point. Is there a reason we can't do this earlier such that get_kerning remains just a lookup?

glyphs2fontir/src/source.rs
```diff
@@ -2176,30 +2204,31 @@ mod tests {
         assert!(bad_kerns.is_empty(), "{bad_kerns:#?}");
     }
 
+    // helpers:
```
rsheeter, line 2275: This comment isn't helpful.

Why isn't parse_side `impl From<&str> for KernSide` (since it's infallible)?

glyphs2fontir/src/source.rs
```diff
@@ -1047,16 +1048,43 @@ impl Work<Context, WorkId, Error> for KerningInstanceWork {
...
+        let kern1 = flip_class_side(kern1);
+        let kern2 = flip_class_side(kern2);
+        combined_kerns.insert((kern1, kern2), *value);
+    }
+
+    Some(Cow::Owned(combined_kerns))
+}
 
-    font_info.font.kerning_ltr.get(our_id)
+fn flip_class_side(s: &SmolStr) -> SmolStr {
```
rsheeter: I wonder if introducing a type (enum? there seem to be distinct cases) for kerning identifier and hanging some of this stuff off it would work out nicely?

glyphs2fontir/src/source.rs
```diff
@@ -2241,6 +2270,21 @@ mod tests {
         );
     }
 
+    #[test]
+    fn combine_kerning() {
```
rsheeter: Suggest adding a comment explaining what this is testing or introducing a paragraph_length_method_name_explaining_whats_going_on

rsheeter: (at a glance I can't tell why this is the right answer)

glyphs2fontir/src/source.rs
```diff
@@ -2241,6 +2270,21 @@ mod tests {
         );
     }
```
rsheeter, line 2340: It would be nice to add unit tests for new helpers, parse side and flip side seem independently testable.

rsheeter (approved):
Broadly LGTM, though I hope we can avoid the get_kerning makes changes aspect by altering where the work is done.

glyphs2fontir/src/source.rs
cmyr: Added some docs explaining what's going on.

glyphs2fontir/src/source.rs
cmyr: We turn this into an enum later, when we're actually generating the IR; this is about making the pre-IR RTL kerns look like the pre-IR RTL kerns.

glyphs2fontir/src/source.rs
cmyr: this is just a helper used in tests, not sure there's any clear advantage to having a `From` impl since it wouldn't ever be used anywhere else.

Issue #1568, opened 2025-07-18
non-deterministic GSUB compilation


cmyr:
Not sure what this is but just noticed some jitter in crater results, which I can reproduce on at least one font just be rerunning ttx_diff:

```
python3 resources/scripts/ttx_diff.py 'https://github.com/calcom/font?b833fb1129#sources/Cal-Sans.glyphs'
```

Not going to dig into this right now but opening to track.

anthrotype:
oh actually there _is_ something elsewhich has to do with bracket layers (GSUB Feature Variations), notably if you build Playfair (which has a bunch of them) multiple times, and compare the resulting GSUB...

```
python3 resources/scripts/ttx_diff.py 'https://github.com/googlefonts/Playfair?a49f9f9dc9#sources/Playfair.glyphspackage'
```

PR #1572, opened 2025-07-22
fix non-determinism in FeatureParamsCharacterVariants nameIDs

anthrotype:
Part of fixing #1568

We were iterating over HashMap's items while incrementally assigning new nameIDs for the Feature Parameters Character Variants, but the iteration order changes each time, so best to use a BTreeMap in this case

https://github.com/googlefonts/fontc/blob/00c61129fee5f42b1d9a92633429fed3404e28f5/fea-rs/src/compile/features.rs#L256-L259

After applying this patch, running the following command multiple times produces the same GSUB table:

```
python3 resources/scripts/ttx_diff.py 'https://github.com/calcom/font?b833fb1129#sources/Cal-Sans.glyphs'
```

JMM


cmyr (approved):
yep, good call :)

PR #1573, opened 2025-07-23
Attempt to fix non-deterministic GSUB compilation

### LINKED ISSUE #1568: non-deterministic GSUB compilation

anthrotype:
Trying to fix #1568 

not 100% sure of these changes yet, but at least I can get fontc to produce the same GSUB when run multiple times.

I'm using Playfair.glyphspackage for testing and comparing against glyphsLib.


cmyr (approved):
looks good!

PR #1583, opened 2025-08-05
Split cursive lookups into RTL/LTR

cmyr:
This matches fonttools.


fontbe/src/features/marks.rs
```diff
@@ -569,47 +570,61 @@ impl<'a> MarkLookupBuilder<'a> {
...
+            // LTR only if explicit member of group, else RTL:
+            // https://github.com/googlefonts/ufo2ft/blob/98e8916a8/Lib/ufo2ft/featureWriters/cursFeatureWriter.py#L76
+            if ltr_glyphs
+                .map(|glyphs| glyphs.contains(*gid))
+                .unwrap_or(false)
+            {
+                ltr_builder.insert(*gid, entry, exit);
+            } else {
+                rtl_builder.insert(*gid, entry, exit);
+            }
```
rsheeter, line 601: This looks like it doesn't need an if? Something like:

```rust
ltr_glyphs.filter(|glyphs| glyphs.contains(*gid))
  .map_or(rtl_builder, |_| ltr_builder)
  .insert(*gid, entry, exit);
```

Edit: tried it, wasn't quite right, corrected

fontbe/src/features/marks.rs
```diff
@@ -569,47 +570,61 @@ impl<'a> MarkLookupBuilder<'a> {
...
+                LookupFlag::IGNORE_MARKS | LookupFlag::RIGHT_TO_LEFT,
+                None,
+            ));
         }
-        // In the future we might to do an LTR/RTL split, but for now return a
-        // vector with one element.
-        Ok(vec![PendingLookup::new(vec![builder], flags, None)])
+
+        Ok(result)
     }
```
rsheeter, line 643: A lot of this function feels like the same stuff written twice. https://github.com/googlefonts/fontc/pull/1585 (PR to this branch) tries to deduplicate. I find the resulting make_cursive_lookups easier to read but ymmv, feel free to take or ignore.

fontbe/src/features/properties.rs
rsheeter, line 1: These fns seem testable?

rsheeter (approved):

fontbe/src/features/marks.rs
cmyr: This is a case where I think the `if` does a better job of actually expressing the logic, but I do enjoy the code golf 🤓

fontbe/src/features/marks.rs
cmyr: cool yea there are some good ideas in there, I've expanded on it a bit in a new commit.

fontbe/src/features/properties.rs
cmyr: these are mostly just direct lookups from the bundled unicode data, so there isn't toooo much of our own logic to test. Added a simple sanity test for the slightly weird `unicode_script_direction` fn.

PR #1588, opened 2025-08-07
Truncate axis location values like glyphsLib

rsheeter:
`python3 resources/scripts/ttx_diff.py 'https://github.com/usted/Albert-Sans?f7f4608223#sources/AlbertSans-Italic.glyphs' --compare gftools --config ~/.fontc_crater_cache/usted/Albert-Sans/sources/config.yaml` is identical with this change.

JMM.


anthrotype:
glyphsLib no longer truncates, we should revert this once we update our requirements, see https://github.com/googlefonts/glyphsLib/releases/tag/v6.11.5

PR #1589, opened 2025-08-08
[fontir] Tweak glyph ir logging

cmyr:
Tiny fixups to try and make some of these things more visible, as I try to debug a funny component flattening issue.

JMM

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -265,6 +265,8 @@ fn flatten_all_non_export_components(context: &Context) {
 fn flatten_non_export_components_for_glyph(context: &Context, glyph: &Glyph) -> Glyph {
     let mut builder = GlyphBuilder::from(glyph.clone());
     builder.clear_components();
+
+    log::debug!("flattening non-export components of '{}'", glyph.name);
     for (loc, instance) in glyph.sources() {
         let mut new_instance = instance.clone();
         new_instance.components.clear();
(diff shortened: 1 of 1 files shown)


fontir/src/glyph.rs
```diff
@@ -416,8 +417,8 @@ fn flatten_glyph(context: &Context, glyph: &Glyph) -> Result<(), BadGlyph> {
     if glyph.default_instance().components.is_empty() {
         return Ok(());
     }
-    trace!(
-        "Flatten {} {:?}",
+    log::debug!(
```
rsheeter: I think this should stay trace, and probably the flattening non-export components of should also be trace?

fontir/src/glyph.rs
```diff
@@ -416,8 +417,8 @@ fn flatten_glyph(context: &Context, glyph: &Glyph) -> Result<(), BadGlyph> {
...
+        "flattening {} (components: {:?})",
```
rsheeter: Nit: capitalization of the first word of the log message is inconsistent. I think mostly we capitalize so Flattening

rsheeter (approved):

fontir/src/glyph.rs
cmyr: What is your heuristic for making that determination?

my issue with trace is that if you turn it on our logging is incredibly verbose, and it's very hard to find anything. These operations don't happen that often, and I often want to see if they happen because they end up being a common source of glyf/gvar difference.

rsheeter: It's not a hard rule but broadly I would expect:

1. Debug is safe to turn on globally. It'll be noisy but not completely unusable
1. Trace is too noisy to turn on globally, it's very detailed and you shoudl turn it on only for the module you are troubleshooting (e.g. `export RUST_LOG=warn,fontir::glyph=debug`)
    * Aside: Invalid kern side seems to be very common, to the point I wonder if it's _really_ a warning

Are you turning trace on globally or for a specific module?

cmyr: I'm basically not turning trace on at _all_ 😅

PR #1590, opened 2025-08-08
Ignore fontc only name 25

rsheeter:
fontmake doesn't support this name. It's not an error that fontc does.

Makes `python3 resources/scripts/ttx_diff.py 'https://github.com/googlefonts/googlesans-code?0713088a41#sources/GoogleSansCode-Italic.glyphspackage'` identical.


rsheeter:
I don't want to endlessly invest in improving glyphsLib to close gaps. This one struck me as low ROI to make py changes.

PR #1593, opened 2025-08-11
[crater] Support annotations

cmyr:
This allows users to add 'annotations' to targets, which can be used to explain known problems and link to issues.

Annotations are added by manually editing an `annotations.json` file stored alongside the output.

Annotations are maximally dumb: they are associated with a target and will be visible for as long as that target produces a diff, unless they are manually removed.

(This seems more and more useful as we get into the long tail, and is particularly motivated by the fact that I'm currently focused on investigating diffs and opening issues.)

annotations are displayed like:

Screenshot 2025-08-11 at 2 48 33 PM

and if no annotations exist, a little button will give you the following:

Screenshot 2025-08-11 at 2 44 44 PM

### DIFF
--- a/fontc_crater/resources/style.css
+++ b/fontc_crater/resources/style.css
@@ -98,3 +98,40 @@ td.rev {
 .hidden_row {
   display: none;
 }
+
+/* Modal overlay - covers entire screen */
+.modal-overlay {
+  display: none;
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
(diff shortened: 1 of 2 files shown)


cmyr:
I realized i could make this process a bit easier by providing a way for the user to paste a github link, and then generating the annotation for them (since most annotations should just be links to issues?) but I'll merge this as-is and start using it and then we can see if it's worth more fanciness.

PR #1594, opened 2025-08-12
[fontbe] Round component pts before generating deltas

cmyr:
This matches fontmake (and also our own simple glyphs) where rounding occurs before the final variation data is generated.

gives us another +small identical.

JMM

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -213,7 +213,7 @@ fn add_phantom_points(advance: u16, points: &mut Vec<Point>) {
     points.push(Point::new(0.0, 0.0));
 }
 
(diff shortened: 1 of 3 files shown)


anthrotype (approved):
LGTM

you may want to cite where fontmake does this rounding of components x/y coordinates, for posterity:

https://github.com/fonttools/fonttools/blob/5ae2943a4334d5bbaef6d7984c131479e6759f96/Lib/fontTools/pens/ttGlyphPen.py#L110

ufo2ft uses fonttools TTGlyphPointPen to 'draw' UFO glyphs (which may contain floats) to TTGlyphs and component offsets are always rounded in there.

PR #1595, opened 2025-08-13
[fea-rs] Better match fonttools handling of script stmt

cmyr:
This is kind of interesting; I'm not really sure that fonttools is correct, but it matches afdko and this is a corner case that is unlikely to matter in a real source.

According to the spec, any lookupflags should be cleared whenever we encounter a script statement.

fonttools follows this, _unless_ the current language system is already (script, dflt), where 'script' is the script set by the new statement.

We _kind of_ matched this, previously, except that we did not take the language into account; we would skip clearing the flags if the new script was the same as the old one.

The fonttools logic here is definitely better. That said, the behaviour is unintuitive in some cases, for instance:

```fea
feature derp {
    lookup W {
        lookupflag IgnoreMarks;
        # this script statement clears the preceding IgnoreMarks flag.
        script latn;
        sub one by two;
    } W;

    lookup X {
        lookupflag IgnoreMarks;
        # but both fonttools and afdko ignore _this_ script statement,
        # and this lookup ends up with IgnoreMarks set.
        script latn;
        sub three by four;
    } X;
} derp;
```

Per the spec, it seems like _neither_ lookup should have `IgnoreMarks` set, but in practice both fonttools & afdko set it for the second lookup, so we will match that.

### DIFF
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -446,7 +446,17 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
 
     fn set_script(&mut self, stmt: typed::Script) {
         let script = stmt.tag().to_raw();
-        if Some(script) == self.script {
+
+        // fonttools logic here is kind of particular, so let's match it literally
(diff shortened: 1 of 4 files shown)


anthrotype (approved):
LGTM 😖

PR #1596, opened 2025-08-14
ttx_diff: use more specific xpath to select nameID=25

anthrotype:
I am not entirely sure why, but after #1590 running ttx_diff.py on Kablammo.glyphs gives a weird "lxml.etree.XPathEvalError: unknown error".  Notice the +2 in red under https://googlefonts.github.io/fontc_crater/#other-failures

Maybe the xml file is to big and the xpath expression too broad? i am not entirely sure. But making the xpath expression more selective and explicit seems to fix the error.

JMM


anthrotype:
repro here: https://github.com/googlefonts/glyphsLib/actions/runs/16962826725/job/48079483461?pr=1101#step:5:158

fix to follow

rsheeter:
That ... is super weird. Seems like a bug. ty for fixing it.

PR #1597, opened 2025-08-14
[glyphs-reader] Don't truncate Axis Location values as in latest glyphsLib

anthrotype:
Reverts #1588

Requires https://github.com/googlefonts/glyphsLib/releases/tag/v6.11.5

JMM


cmyr (approved):
👍

PR #1598, opened 2025-08-14
update requirements.txt

anthrotype:
notably bumps glyphsLib which should give us more greens on crater.

ufo2ft is pinned to 3.5.1 until I figure out why 3.6.0 was yielding -114 identical fonts a couple weeks ago: https://github.com/googlefonts/fontc/pull/1576#issuecomment-3127922846

JMM


anthrotype:
> There was also a patch to fonttools that should give us a small bump

i'll get to it now, and bump it here with the rest

Issue #1600, opened 2025-08-14
The long fascinating story of NotoSerifTangunt `u17EF8`


cmyr:
_I've spent about a day on this so far and it does not feel hugely meaningful, so I'm going to dump context into this issue and go think about something else._

This is the story of a glyph that produces a diff, and the various bugs that seem to contribute to this diff, none of which (possibly) matter.

The glyph in question is `u17EF8`, in [Noto Serif Tangut](https://github.com/notofonts/tangut). (As we all doubtless know, Tangut is an extinct tibeto-burman language, and its script is known only through a small number of extent fragments.)

Anyway, the interesting part of the glyph in question looks like this:

<img width="590" height="522" alt="Image" src="https://github.com/user-attachments/assets/06eb0e02-cc25-448f-bccf-6f01fae74687" />

### bug one
Specifically, both fontc and fontmake are treating that little circled corner as an 'open corner', and erasing it. Interestingly glyphs.app is _not_ erasing this corner, even though I believe the corner erasure code in glyphsLib was borrowed directly from glyphs.app.

So this is _problem no 1_; our corner erasure code has deviated somewhat from glyphs'.

### bug two
Ultimately, though, both fontc and fontmake are doing the corner erasure here, so _that_ isn't the problem. The problem is that apparently there is some very minor difference in some function that is called by our corner erasing code.

~Specifically, both compilers end up in the situation where they are trying to take the subsegment of the curved part of that triangle, right near an extrema. Specifically the cubic coords are `(384.0, 208.), (364.91, 186.86), (346.24, 168.53), (328.0, 153.)`, and `t` is `0.9999999999999881`.~

(Okay after more investigation during writing, it looks like the actual problem here is that we are returning different values of `t` when trying to determine the intersection; python is giving us 1.0, and rust is giving us `0.9999999999999881`. I haven't dug into why this is.)

So anyway: bug two is that there remains some minor difference in our intersection/splitting code.

### bug three

In any case, this floating point business shouldn't really matter, because these numbers would all have the same rounding behaviour. (incidentally the actual _diff_ that brought this all to my attention is simply that for this glyph, `fontmake` has an extra duplicate offcurve point in its outline.)

So anyway: what happens here is that fontmake ends up with a cubic segment where all the points are equal, and fontc ends up with a cubic segment where all the points are very minorly different.

[...]

PR #1601, opened 2025-08-15
[glyphs] Update corner erasure code

cmyr:
The glyphs.app implementation has changed slightly from when this code was first ported.

see https://github.com/googlefonts/fontc/issues/1600#issuecomment-3190896627

I'll have a glyphsLib patch up that matches this shortly.


rsheeter (approved):
LGTM, I assume we should time landing so this goes in alongside the glyphsLib update to avoid a negative delta on crater?

PR #1606, opened 2025-08-19
[glyphs] Match glyphsLib for curve-curve intersections

cmyr:
Specficially in the case where there are multiple intersections, the curve-curve implementation does not behave the same as the line/curve or line/line implementations.

In the latter, the first hit is always the one with the lowest `t` on the first segment; but in the case of segment/segment intersection, the order is based on the order in which recursion occurs.

The key here is to not sort the results, but to retain the order we get after recursion.

This should be worth another +2-4 on crater.

### DIFF
--- a/glyphs2fontir/src/erase_open_corners.rs
+++ b/glyphs2fontir/src/erase_open_corners.rs
@@ -357,6 +357,17 @@ struct Intersection {
     t1: f64,
 }
 
+impl Intersection {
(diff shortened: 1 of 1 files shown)


anthrotype:
> in the case of segment/segment intersection, the order is based on the order in which recursion occurs.

and that order is stable, right?

anthrotype (approved):

Issue #1614, opened 2025-08-25
Variable fonts: never set USE_MY_METRICS flag on component in `glyf` table


madig:
To work around the issue described in https://github.com/fonttools/fonttools/pull/3912. Even setting the flag correctly might result in different shapers disagreeing about what should be done.

anthrotype:
We can keep the old behavior of automatically setting this flag for static fonts only, where there's no ambiguity and the flag does essentially nothing.

PR #1615, opened 2025-08-26
[glyf] Only set USE_MY_METRICS flag for static fonts, not variable fonts

anthrotype:
The USE_MY_METRICS flag is primarily useful for hinted fonts where composite glyphs should reuse the hinted metrics of one of their components.

Since fontc doesn't support hinting, and the flag can be dangerous in variable fonts where composite and component metrics may differ across masters, we now only set the flag for static fonts to match fontmake's legacy behavior.

We plan to change fontmake to do the same for variable fonts as well.

Fixes https://github.com/googlefonts/fontc/issues/1614

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -3791,37 +3791,39 @@ mod tests {
         )
     }
 
-    #[rstest]
-    #[case::static_font("glyphs3/StaticComposite.glyphs")]
-    #[case::variable_font("glyphs3/WghtVarComposite.glyphs")]
-    fn use_my_metrics_behavior_by_font_type(#[case] font_file: &str) {
+    fn has_use_my_metrics_flag(font_file: &str, composite_glyph_name: &str) -> bool {
         let result = TestCompile::compile_source(font_file);
-        let static_metadata = result.fe_context.static_metadata.get();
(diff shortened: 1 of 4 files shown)


fontc/src/lib.rs
```diff
@@ -3790,4 +3790,38 @@ mod tests {
...
+        let Some(glyf::Glyph::Composite(glyph)) = &glyphs[gid as usize] else {
+            panic!("Expected 'equal' to be a composite glyph");
+        };
+
+        let has_use_my_metrics = glyph.components().any(|comp| {
+            comp.flags
+                .contains(write_fonts::read::tables::glyf::CompositeGlyphFlags::USE_MY_METRICS)
+        });
+
+        if is_variable {
```
rsheeter: This branch makes me think this should just be two separate tests, it's not winning much from being an rstest? - can still call a common helper for the duplicate part

anthrotype: Done https://github.com/googlefonts/fontc/pull/1615/commits/75134a71acd06fba8edc69a18adc088053895ded

anthrotype:
https://github.com/googlefonts/fontmake/releases/tag/v3.11.0

PR #1628, opened 2025-09-11
[ufo] Merge libs from designspace and default master

cmyr:
This is a conservative merge; we only update dictionary objects, and we only add items from the child that are not present in the base.

It is possible for there to be keys in the default master's lib that are not in the designspace lib, and which we want to have access to (the motivating example being the presence of the 'eraseOpenCorners' filter).

In an earlier patch I had added special handling for the case where we were compiling a single ufo; this new approach lets us delete that special case.

The decision to just look at the default master was arrived in conversation with cosimo. In general the keys that we care about are autogenerated by glyphsLib, and should be consistent between masters. We could try and verify that this is true but that feels like  very low ROI at this point.

### DIFF
--- a/ufo2fontir/src/source.rs
+++ b/ufo2fontir/src/source.rs
@@ -270,9 +270,15 @@ impl Source for DesignSpaceIrSource {
             })
             .collect();
 
+        // if source was a designspace we don't want to copy over keys like
+        // public.skipExportGlyphs, but we do if it was a UFO. (we assume that
+        // these keys are correct, if we were passed a designspace directly)
+        let skip_public_keys =
+            designspace_or_ufo_file.extension().and_then(OsStr::to_str) == Some("designspace");
         merge_default_master_lib_into_designspace_lib(
             &mut designspace.lib,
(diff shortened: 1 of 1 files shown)


ufo2fontir/src/source.rs
```diff
@@ -366,6 +369,33 @@ impl Source for DesignSpaceIrSource {
...
+        *base = child;
+        return;
+    }
+    for (key, value) in child {
+        match (base.get_mut(&key), value) {
+            (None, value) => {
+                base.insert(key, value);
+            }
+            (Some(Value::Dictionary(dict)), Value::Dictionary(child)) => {
+                merge_default_master_lib_into_designspace_lib(dict, child);
```
anthrotype: you're merging recursively? that's fancier than what I had in mind, I thought it'd be sufficient to simply merge keys at the top level only. I haven't fully thought about implications of merging it this way.

cmyr: fair, maybe this is unnecessarily fancy? I'll do top-level only for now, it will reduce the potential surface area for unexpected side effects.

PR #1631, opened 2025-09-12
[fea] Allow '.null' as glyph name

cmyr:
The spec is very clear that '.' is not valid in a glyph name with the sole exception of '.notdef', and that if you are using 'null' as a glyphname you should escape it with a backslash.

Ultimately, though, a bunch of fonts _do_ use '.null' as a glyph name, and it feels pointless to try and update all of these sources.

As a compromise we will now accept these glyph names, but print a warning, as our last concession to principle.


florianpircher:
Minor: perhaps the warning should suggest `NULL` instead of `null`. I generally see this name with the dot and lowercase or all-caps.

PR #1636, opened 2025-09-16
[fontir] Fix crash when component is missing

cmyr:
This matches fonttools, which just ignores references to components that do not exist.

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -333,7 +333,14 @@ fn convert_components_to_contours(context: &Context, original: &Glyph) -> Result
             continue;
         }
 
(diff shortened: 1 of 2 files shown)


fontir/src/glyph.rs
```diff
@@ -333,7 +333,13 @@ fn convert_components_to_contours(context: &Context, original: &Glyph) -> Result
             continue;
         }
 
-        let referenced_glyph = context.get_glyph(component_base.clone());
+        let Some(referenced_glyph) = context.try_get_glyph(component_base.clone()) else {
```
rsheeter, line 336: Nit: a citation for python doing the same, or at least a comment, might be nice

rsheeter (approved):

PR #1639, opened 2025-09-17
[glyphs] Treat the font axes as source of truth

cmyr:
That is, when building the axis mapping only consider axes that are defined in the font and then look for them in the master, instead of looking at the axes defined in the master and assuming that they exist in the font.

Gives us another +2-4.

JMM

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2152,20 +2152,17 @@ fn user_to_design_from_axis_location(
         return None;
     }
 
-    if !master_locations.is_empty() && from.axes.is_empty() {
-        log::warn!("masters define locations but font has no axes");
-        return None;
-    }
-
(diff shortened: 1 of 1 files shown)


glyphs-reader/src/font.rs
```diff
@@ -2152,20 +2152,17 @@ fn user_to_design_from_axis_location(
...
-    if !master_locations.is_empty() && from.axes.is_empty() {
-        log::warn!("masters define locations but font has no axes");
-        return None;
-    }
-
     let mut axis_mappings: BTreeMap<String, AxisUserToDesignMap> = BTreeMap::new();
     for (master, axis_locations) in from.font_master.iter().zip(master_locations) {
-        for axis_location in axis_locations {
-            let Some(axis_index) = axis_index(&from.axes, |a| a.name == axis_location.axis_name)
+        // masters can reference axes that aren't defined in the font,
```
```suggestion
        // masters' Axis Locations custom parameters can reference axes that aren't defined in the font,
```

anthrotype (approved):

Issue #1638, opened 2025-09-17
fontc should decompose components if transform values exceed F2Dot14 limits


anthrotype:
Sometimes sources (e.g. VendSans.glyphs below) contain components with > 200% scale (emdash glyph is 250% wider than hyphen in VendSans) which don't fit within the limits of F2Dot14, the fixed type used inside glyf table for component transforms:

```
$ python3 resources/scripts/ttx_diff.py 'https://github.com/ktkm/Vend-Sans?2663f1c7da#sources/VendSans.glyphs'
```

fontmake decomposes this (the fonttools TTGlyphPen handles this automatically) whereas fontc keeps the component but clamps transform values to the max allowed, but this would look different than intended so it is incorrect:

```diff
$ diff -u /Users/clupo/oss/fontc/build/default/fontc.glyf.ttx /Users/clupo/oss/fontc/build/default/fontmake.glyf.ttx
--- /Users/clupo/oss/fontc/build/default/fontc.glyf.ttx	2025-09-17 14:35:53
+++ /Users/clupo/oss/fontc/build/default/fontmake.glyf.ttx	2025-09-17 14:35:53
@@ -3096,8 +3096,14 @@
       <component glyphName="uni0304" x="-27" y="0" flags="0x4"/>
     </TTGlyph>

-    <TTGlyph name="emdash" xMin="47" yMin="254" xMax="627" yMax="314">
-      <component glyphName="hyphen" x="-143" y="0" scalex="1.99994" scaley="1.0" flags="0x4"/>
+    <TTGlyph name="emdash" xMin="95" yMin="254" xMax="820" yMax="314">
+      <contour>
+        <pt x="95" y="254" on="1"/>
+        <pt x="95" y="314" on="1"/>
+        <pt x="820" y="314" on="1"/>
+        <pt x="820" y="254" on="1"/>
+      </contour>
+      <instructions/>
     </TTGlyph>

     <TTGlyph name="endash" xMin="95" yMin="254" xMax="530" yMax="314">
```

This is the original fonttools issue from 7 years ago:

https://github.com/fonttools/fonttools/issues/1200

> The UFO GLIF spec allows any numbers for xScale, xyScale, yxScale, yScale, xOffset, yOffset, however the OpenType glyf spec uses F2Dot14 numbers, which are encoded as a signed 16-bit integer and therefore can only contain values from -32768 (-0x8000, or -2.0) to +32767 included (0x7FFF, or +1.99993896484375...).
[...]

PR #1642, opened 2025-09-18
Handle components with overflowing 2x2 transform exceeding F2Dot14 limits

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/1638

Makes VendSans.glyphs among others "output is identical"

```
$ python3 resources/scripts/ttx_diff.py 'https://github.com/ktkm/Vend-Sans?2663f1c7da#sources/VendSans.glyphs'
```

(Will add some tests after lunch)

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -93,6 +93,12 @@ fn create_component_ref_gid(
         round_xy_to_grid: true, // ufo2ft defaults to this, match it
         ..Default::default()
     };
+
+    // By this point, fontir should have decomposed any components with transforms
+    // outside the -2.0 to +2.0 range. For values between MAX_F2DOT14 and 2.0,
+    // F2Dot14::from_f32() will saturate to MAX_F2DOT14, matching fonttools behavior:
+    // https://github.com/googlefonts/fontc/issues/1638
+
     let component = Component::new(
         gid,
         Anchor::Offset {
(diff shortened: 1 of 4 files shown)


anthrotype:
ci failures are unrelated to this PR (see https://github.com/googlefonts/fontc/pull/1643)

fontbe/src/glyphs.rs
```diff
@@ -1044,4 +1050,31 @@ mod tests {
         };
         assert_eq!((0, 1), (x, y));
     }
+
+    /// Minimum and maximum values that can fit in an F2Dot14: [-2.0, 1.99993896484375]
+    const MAX_F2DOT14: f32 = (0x7FFF as f32) / ((1 << 14) as f32);
+    const MIN_F2DOT14: f32 = -2.0;
```
rsheeter: Shouldn't this be in fonttypes? F2Dot14::MIN?

anthrotype: i'm only using these for this particular test...

rsheeter: I also wonder if we should test it, e.g. min -2, max > 1.999 and < 2 or some such?

anthrotype: ah I see what you mean! yes

anthrotype: > I also wonder if we should test it, e.g. min -2, max > 1.999 and < 2 or some such?

do you mean in here in fontbe? Or an integration test in fontc/lib?

rsheeter: I was imagining maybe a PR to add min/max to font-types, but that could be done as incremental follow-ons

anthrotype: > a PR to add min/max to font-types

it already does have F2Dot14::MIN and MAX, I'm not sure what you mean

rsheeter: Oh I thought we didn't because I saw consts here.

anthrotype: I had forgotten, but you reminded me and now I use them https://github.com/googlefonts/fontc/pull/1642/commits/3a7fb15eff3e01a875ee19a9f8b5aa2903abbebe

I added more tests to fontir as well
https://github.com/googlefonts/fontc/pull/1642/commits/d792288656afe4ae23d7ee8d5145b75eb3f5ac75

let me know if you want a full integration test in fontc/lib as well

rsheeter: > let me know if you want a full integration test in fontc/lib as well

Not specifically, what you have seems fine to me

PR #1643, opened 2025-09-18
fix new clippy errors with Rust 1.90

anthrotype:
I fixed the trivial unnecessary parentheses etc. but I wasn't sure about fontc_crater's `SkipReason` enum being unused so I kept it with `#[allow(dead_code)]`

JMM


cmyr (approved):
Thanks! I'll add a commit to remove SkipReason.

Issue #1592, opened 2025-08-08
flattening components can drop intermediate layers


cmyr:
Ran into the following bug, which is evidenced in 'Bitter' (for the `Aogenek` glyph).

The font has 'Light' and 'Bold' masters.

- `Aogonek` is composed of `A.ogonekAccent` and `ogonekcomb.case`.
- `ogonekcomb.case` has a single component, `ogonek`.
- `ogonek` has an intermediate (brace) layer, `{100}`.
- `A.ogonekAccent` is marked non-export.

Because `A.ogonekAccent` is non export, we end up inlining it into `Aogonek`; that glyph then has mixed contours & components, and so we convert components to contours. In this process, we drop the brace layer.

This case is at least _somewhat_ tricky, because we need all the components to have the same locations.

I guess the solution (hinted at in a comment) is that when we flatten the components, if there are any present intermediate layers we need to synthesize relevant intermediate values for the other components at that position, by interpolating?

Test case attached:

[Bitter-component-gvar.glyphs.zip](https://github.com/user-attachments/files/21689560/Bitter-component-gvar.glyphs.zip)

anthrotype:
note that proper composite decomposition (I prefer this term to "flattening") in situations where the locations of composite glyph and any of its components do not align involves _two_ cases (quoting myself from the above):

> in the first case (composite with more masters than its components), we need to interpolate the component glyphs on-the-fly at the missing locations while decomposing the composite glyph.
> For the second case (composite with fewer masters than its component[s]), we first traverse all the components across all masters and collect the set of locations in which they may be defined; then we interpolate our composite glyph at the new locations that it doesn't already define, and finally we proceed with decomposing it.

I believe in #1644 you only cover the first case. The second case is equally important, because if a composite contains a component that is defined at additional intermediate locations that aren't defined for the parent glyph, decomposing the latter without taking those into account will make it look different from how it was supposed to appear before decomposition.

PR #1644, opened 2025-09-18
[ir] Instantiate missing intermediates when flattening

cmyr:
This fixes an issue where a composite glyph with an intermediate layer would cause a crash during flattening if any of its components did not have a layer at the same location.

With this patch, we will interpolate a new instance at the missing location for the component, and use that.

This is related to (basically the inverse of) #1592, which I'll look at next.

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -350,19 +350,15 @@ fn convert_components_to_contours(context: &Context, original: &Glyph) -> Result
                 .cloned(),
         );
 
-        // Any contours of the referenced glyph at this location should be kept
-        // For now just fail if the component source locations don't match ours, don't try to interpolate
         trace!(
             "'{}' retains {} {component_affine:?} at {loc:?}",
             original.name,
             referenced_glyph.name
         );
-        let Some(inst) = simple.sources.get_mut(&loc) else {
-            return Err(BadGlyph::new(
-                simple.name.clone(),
-                BadGlyphKind::UndefinedAtNormalizedLocation(loc.clone()),
-            ));
-        };
+        let inst = simple
+            .sources
+            .get_mut(&loc)
+            .expect("only instances at known locations are added to queue above");
         let ref_inst = get_or_instantiate_instance(&referenced_glyph, &loc, context)?;
 
         for contour in ref_inst.contours.iter() {
@@ -379,12 +375,7 @@ fn convert_components_to_contours(context: &Context, original: &Glyph) -> Result
     }
 
     let simple: Glyph = simple.build()?;
-    debug_assert!(
-        simple.name == original.name,
-        "{} != {}",
-        simple.name,
-        original.name
-    );
+    assert_eq!(simple.name, original.name,);
     context.glyphs.set(simple);
     Ok(())
 }
@@ -769,14 +760,20 @@ impl Work<Context, WorkId, Error> for GlyphOrderWork {
 
 #[cfg(test)]
 mod tests {
-    use std::{collections::HashSet, path::Path};
+    use std::{
+        collections::{BTreeSet, HashSet},
+        path::Path,
+    };
 
-    use fontdrasil::{orchestration::Access, types::GlyphName};
+    use fontdrasil::{
+        orchestration::Access,
+        types::{Axis, GlyphName},
+    };
     use kurbo::{Affine, BezPath, Rect, Shape};
     use rstest::rstest;
 
     use crate::{
-        ir::{Component, Glyph, GlyphBuilder, GlyphInstance, GlyphOrder},
+        ir::{Component, Glyph, GlyphBuilder, GlyphInstance, GlyphOrder, StaticMetadata},
         orchestration::{Context, Flags, WorkId},
         paths::Paths,
     };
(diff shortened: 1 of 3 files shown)


fontir/src/glyph.rs
```diff
@@ -361,12 +363,7 @@ fn convert_components_to_contours(context: &Context, original: &Glyph) -> Result
                 BadGlyphKind::UndefinedAtNormalizedLocation(loc.clone()),
             ));
         };
-        let Some(ref_inst) = referenced_glyph.sources().get(&loc) else {
-            return Err(BadGlyph::new(
-                referenced_glyph.name.clone(),
-                BadGlyphKind::UndefinedAtNormalizedLocation(loc.clone()),
-            ));
-        };
+        let ref_inst = get_or_instantiate_instance(&referenced_glyph, &loc, context)?;
```
anthrotype, line 362: there's a comment higher up on line 354 which no longer applies

> // For now just fail if the component source locations don't match ours, don't try to interpolate

anthrotype:
i'm reviewing this now, just wanted to note that when I read "flattening" I got confused initially, because we have another filter operation that in the fontmake pipeline we call flattenComponents which removes nested composite glyphs (with components of components etc.) which depth greater > 1, so "flattening" the component tree, but it still keeps components as components (e.g. Aacute -> A + acutecomb.case, acutecomb.case -> acutecomb, acutecomb -> acute => Aacute -> A + acute).
Whereas here in the PR title and description when you say "flattening" you are referring to what we usually call "decomposing" of composite glyphs, i.e. removing the components altogether and turning components into simple contours.

fontir/src/glyph.rs
```diff
@@ -392,6 +389,106 @@ fn convert_components_to_contours(context: &Context, original: &Glyph) -> Result
...
+fn get_or_instantiate_instance<'a>(
+    glyph: &'a Glyph,
+    loc: &NormalizedLocation,
+    context: &Context,
+) -> Result<Cow<'a, GlyphInstance>, BadGlyph> {
+    if let Some(instance) = glyph.sources().get(loc) {
+        return Ok(Cow::Borrowed(instance));
+    }
+    log::debug!("instantiating instance of '{}' at {loc:?}", glyph.name);
+    let meta = context.static_metadata.get();
```
anthrotype, line 393: using the global VariationModel from the static_metadata will not work in this case where you want to interpolate a instance on the fly. You have to construct new "sparse" submodels for each glyph (or rather for each unique set of source locations). If all glyphs are defined along the same set of locations, then you don't have this particular problem (i.e. a composite missing a component at given location); only when glyphs contain sources at different locations from one another (and they reference each other via components) this problem arises. And that's when a single global VariationModel doesn't work.

anthrotype: We do something similar in fontbe/glyphs.rs or fontbe/hvar.rs where we keep a hashmap cache of models we lookup up by a given glyph's set of locations, and reuse or make new VariationModels as needed

fontir/src/glyph.rs
```diff
@@ -392,6 +389,106 @@ fn convert_components_to_contours(context: &Context, original: &Glyph) -> Result
...
+                    .flat_map(iter_pathel_points)
+                    .collect(),
+            )
+        })
+        .collect();
+    let deltas = meta
+        .variation_model
+        .deltas(&point_seqs)
+        .map_err(|e| BadGlyph::new(&glyph.name, e))?;
+    let points = VariationModel::interpolate_from_deltas(loc, &deltas);
```
anthrotype, line 425: hm why this `interpolate_from_deltas` return Vec2? It should return kurbo::Points. A delta is a Vec2, i.e. a displacement with a direction and magnitude, whereas "point" is a "whole" or fixed location in coordinate space.

fontir/src/glyph.rs
cmyr: hm, I'm not sure I totally understand.

The case that we're dealing with is something like this (to take a simple but also real case):
- we have a variable font with one axis. There are masters at locations 0.0 and 1.0 (normalized)
- we have a composite glyph that has an intermediate layer at location 0.5.
- this composite has a component that only has layers at 0.0 and 1.0.

In this case, why wouldn't the global variation model be suitable for interpolating a new layer at location 0.5 for the component? Or am I missing something, and there are more complicated cases where this doesn't work?

fontir/src/glyph.rs
cmyr: it's because this fn is generic over scalars and 2-tuples, and Vec2 is the only 2-tuple type that supports addition.

fontir/src/glyph.rs
anthrotype: I don't remember wehter the global VariationModel in static metadata contains only the "master" locations or the union of all the glyph layers locations (probably the former). In either case, it's wrong to assume you can use that to interpolate a particular glyph instance, because each glyph has or can have its own variation sub-model, independent of the others. Interpolation should be done with that specific model, otherwise it'll be incorrect.

In your particular example, if the global model was initialized with locations 0.0 and 1.0 and your composite has locations 0.0 and 1.0, then you can reuse that. But if, say, the particular glyph you want to interpolate has three source locations 0.0, 0.3, and 1.0, and you want to interpolate at 0.5, you need to construct a new model just for that glyph (and similar glyphs that are defined along the same set of locations).

fontir/src/glyph.rs
anthrotype: there's a TODO:

https://github.com/googlefonts/fontc/blob/85613b12e9dedb819d709e70cc57c3d38bc6160b/fontir/src/variations.rs#L303-L305

can't a generic function have a specific return type parameter given the type of the input parameter?

anthrotype: anyway, nevermind, this is a separate issue

fontir/src/glyph.rs
anthrotype: @cmyr take a look at https://github.com/googlefonts/fontc/issues/400

cmyr: okay gotcha. I don't think I understand all of the possible complexities of sparse masters, and my understanding is that this particular issue (composite has more masters than a component) only occurs (at least in our current sources) in the situation described, where the composite has an intermediate layer? I could certainly add a check to make sure that the variation model is appropriate, to be a bit safer.

PR #1651, opened 2025-09-24
[ir] Make has_components_and_contours a method

cmyr:
Just a little bit of cleanup.

I'm mostly opening this to ask a question: what if a glyph has an instance with contours and another instance with components? Is this checked before now?


anthrotype:
> what if a glyph has an instance with contours and another instance with components? Is this checked before now?

it should crash, it means the outlines are not interpolatable. fontmake checks interpolation compatibility which should fail if this happens.
see https://github.com/googlefonts/fontmake/blob/main/Lib/fontmake/compatibility.py
What does fontc do?

PR #1652, opened 2025-09-24
Also interpolate components

cmyr:
On top of #1650, this adds the ability to interpolate components with simple transforms.

This doesn't handle complex transformations, because I'm not sure how we should be doing that.

More generally, I have some concerns with the overall control flow in this file. There's a lot going on, and there are lots of moving pieces.

- it's hard to figure out the order in which things should be called. Some stuff, like inlining non-export glyphs, has to happen all the time.
- other stuff happens based on flags or other conditions.
- it's not clear what conditions are expected to be true of an IR glyph. For instance, is it okay for an IR glyph to have instances with a different 'shape', e.g. where one instance has a contour and the other has a component? Or for contours to have different numbers of points? If we knew this was already validated, other work would be simpler.
- it would be nice if we had clear expectations about things like intermediate layers, so that we could get all those things normalized ahead of time. For instance: can we figure out all of the glyphs that are going to get decomposed, and ensure ahead of time that the composite & component instances are compatible? (I guess not, thinking out loud, since we might not want to actually modify the component glyph?)


anthrotype:
>  For instance, is it okay for an IR glyph to have instances with a different 'shape', e.g. where one instance has a contour and the other has a component? Or for contours to have different numbers of points? 

No, and no. I think we have a CheckedGlyph somewhere, but maybe that was in fontbe..

anthrotype:
> can we figure out all of the glyphs that are going to get decomposed, and ensure ahead of time that the composite & component instances are compatible? (I guess not, thinking out loud, since we might not want to actually modify the component glyph?)

if we are going to decompose a glyph in the end, then yes. But if a glyph stays composite then it's ok for it be 'sparse', i.e. have more or fewer locations than any of its component glyphs, so you don't want to do it upfront regardless

fontir/src/glyph.rs
```diff
@@ -284,9 +284,26 @@ fn flatten_non_export_components_for_glyph(context: &Context, glyph: &Glyph) ->
             // okay so now we have a component that is not going to be exported,
             // and we need to flatten.
             let xform = component.transform;
-            let Some(referenced_instance) = referenced_glyph.sources().get(loc) else {
+            let Ok(referenced_instance) =
+                get_or_instantiate_instance(&referenced_glyph, loc, context)
+            else {
+                log::debug!(
+                    "component {} of glyph {} not defined at {loc:?}",
```
anthrotype: why is this not an error?

anthrotype:
oh right. Hm, ufo2ft preProcessor similarly decomposes those skipExportGlyphs before doing any subsequent filter, including the decomposition of components with non-matching 2x2...

its SkipExportGlyphsIFilter (where "I" stands for "interpolatable" filter, zipping same-named glyph across all sources) uses the ufo2ft.instantiator to generate instances for missing components so... yeah what I said earlier about fontMath applies here too for VFs, not just to the static build pipeline...

fontir/src/glyph.rs
cmyr: good call!

fontir/src/glyph.rs
```diff
@@ -474,6 +511,54 @@ fn contours_from_deltas(originals: &[BezPath], mut new_points: &[Vec2]) -> Vec<B
     new_contours
 }
 
+/// Interpolate components (specifically their transforms)
+///
+/// Matching varLib, we will just blindly interpolate the scalar values, which
```
anthrotype:
```suggestion
/// Matching ufo2ft, we will just blindly interpolate the scalar values, which
```

varLib errors out if component don't have matching 2x2, they are already decomposed by the time they reach varLib. It's in ufo2ft.instantiator that this blind element-wise interpolation of 2x3 affine transformations happens, through fontMath technically.

you could link https://github.com/googlefonts/ufo2ft/issues/949

Issue #1653, opened 2025-09-25
different starting points for all-offcurve TrueType contours


anthrotype:
Reproducer:

```sh
$ python3 resources/scripts/ttx_diff.py 'https://github.com/appajid/Sirivennela?1e9e8dbbc7#sources/Sirivennela-Regular.ufo'
```

```diff
--- /Users/clupo/oss/fontc/build/default/fontc.glyf.ttx	2025-09-25 09:50:17
+++ /Users/clupo/oss/fontc/build/default/fontmake.glyf.ttx	2025-09-25 09:50:17
@@ -3206,16 +3206,7 @@

     <TTGlyph name="ellipsis" xMin="12" yMin="0" xMax="388" yMax="66">
       <contour>
-        <pt x="12" y="20" on="0"/>
-        <pt x="12" y="46" on="0"/>
-        <pt x="32" y="66" on="0"/>
-        <pt x="58" y="66" on="0"/>
-        <pt x="78" y="46" on="0"/>
-        <pt x="78" y="20" on="0"/>
-        <pt x="58" y="0" on="0"/>
-        <pt x="32" y="0" on="0"/>
-      </contour>
-      <contour>
+        <pt x="188" y="0" on="0"/>
         <pt x="169" y="20" on="0"/>
         <pt x="169" y="46" on="0"/>
         <pt x="188" y="66" on="0"/>
@@ -3223,9 +3214,19 @@
         <pt x="235" y="46" on="0"/>
         <pt x="235" y="20" on="0"/>
         <pt x="216" y="0" on="0"/>
-        <pt x="188" y="0" on="0"/>
+      </contour>
+      <contour>
+        <pt x="32" y="0" on="0"/>
+        <pt x="12" y="20" on="0"/>
+        <pt x="12" y="46" on="0"/>
+        <pt x="32" y="66" on="0"/>
+        <pt x="58" y="66" on="0"/>
+        <pt x="78" y="46" on="0"/>
+        <pt x="78" y="20" on="0"/>
+        <pt x="58" y="0" on="0"/>
       </contour>
       <contour>
+        <pt x="342" y="0" on="0"/>
         <pt x="322" y="20" on="0"/>
         <pt x="322" y="46" on="0"/>
         <pt x="342" y="66" on="0"/>
@@ -3233,7 +3234,6 @@
         <pt x="388" y="46" on="0"/>
         <pt x="388" y="20" on="0"/>
         <pt x="368" y="0" on="0"/>
[...]

PR #1654, opened 2025-09-25
[ttx_diff.py] only normalize contour order when contours perfectly match between fontc vs fontmake

anthrotype:
Fixes part of https://github.com/googlefonts/fontc/issues/1653

Previously, ttx_diff.py would normalize contour order within each compiler's output independently, which could create misleading diffs when the contours actually differed (e.g., different starting points).

This change compares contours between fontc and fontmake first, and only normalizes the order of whole contours when the set of contors serialized as xml strings is identical. When contours differ, they are left in their original order to make the actual differences more apparent.

The ttx diff for Sirivennela-Regular.ufo now looks more sane and makes the different starting points more evident:

```diff
--- /Users/clupo/oss/fontc/build/default/fontc.glyf.ttx	2025-09-25 11:14:20
+++ /Users/clupo/oss/fontc/build/default/fontmake.glyf.ttx	2025-09-25 11:14:20
@@ -3206,6 +3206,7 @@

     <TTGlyph name="ellipsis" xMin="12" yMin="0" xMax="388" yMax="66">
       <contour>
+        <pt x="342" y="0" on="0"/>
         <pt x="322" y="20" on="0"/>
         <pt x="322" y="46" on="0"/>
         <pt x="342" y="66" on="0"/>
@@ -3213,9 +3214,9 @@
         <pt x="388" y="46" on="0"/>
         <pt x="388" y="20" on="0"/>
         <pt x="368" y="0" on="0"/>
-        <pt x="342" y="0" on="0"/>
       </contour>
       <contour>
+        <pt x="188" y="0" on="0"/>
         <pt x="169" y="20" on="0"/>
         <pt x="169" y="46" on="0"/>
         <pt x="188" y="66" on="0"/>
@@ -3223,9 +3224,9 @@
         <pt x="235" y="46" on="0"/>
         <pt x="235" y="20" on="0"/>
         <pt x="216" y="0" on="0"/>
-        <pt x="188" y="0" on="0"/>
       </contour>
       <contour>
+        <pt x="32" y="0" on="0"/>
         <pt x="12" y="20" on="0"/>
         <pt x="12" y="46" on="0"/>
         <pt x="32" y="66" on="0"/>
@@ -3233,7 +3234,6 @@
         <pt x="78" y="46" on="0"/>
         <pt x="78" y="20" on="0"/>
         <pt x="58" y="0" on="0"/>
-        <pt x="32" y="0" on="0"/>
       </contour>
       <instructions/>
     </TTGlyph>
```

I also exposed the `--keep-direction` flag to ttx_diff.py, so it gets forwarded to fontc and fontmake. I used underscore instead of hyphen because absl.flags wants that.

JMM

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -521,42 +521,92 @@ def stat_like_fontmake(ttx):
 
 
 # https://github.com/googlefonts/fontc/issues/1107
-def normalize_glyf_contours(ttx: etree.ElementTree) -> dict[str, list[int]]:
-    # store new order of prior point indices for each glyph
-    point_orders: dict[str, list[int]] = {}
+def normalize_glyf_contours(
+    fontc_ttx: etree.ElementTree, fontmake_ttx: etree.ElementTree
+) -> tuple[dict[str, list[int]], dict[str, list[int]]]:
+    """Reorders contours when they are identical between fontc and fontmake.
 
-    for glyph in ttx.xpath("//glyf/TTGlyph"):
-        contours = glyph.xpath("./contour")
-        if len(contours) < 2:
+    If contours differ (e.g., different starting points), leaves
+    them in their original order to avoid misleading diffs.
+
+    Returns a tuple of two dicts, one for fontc and one for fontmake, containing
+    the new order of prior point indices for each glyph, later used for sorting
+    gvar contours.
+    """
+    fontc_point_orders: dict[str, list[int]] = {}
+    fontmake_point_orders: dict[str, list[int]] = {}
+
+    # Get glyphs from both TTX trees
+    fontc_glyphs = {g.attrib["name"]: g for g in fontc_ttx.xpath("//glyf/TTGlyph")}
+    fontmake_glyphs = {
+        g.attrib["name"]: g for g in fontmake_ttx.xpath("//glyf/TTGlyph")
+    }
+
+    # Only process glyphs that exist in both outputs
+    for glyph_name in fontc_glyphs.keys() & fontmake_glyphs.keys():
+        fontc_glyph = fontc_glyphs[glyph_name]
+        fontmake_glyph = fontmake_glyphs[glyph_name]
+
+        fontc_contours = fontc_glyph.xpath("./contour")
+        fontmake_contours = fontmake_glyph.xpath("./contour")
+
+        # Skip glyphs with mismatched contour counts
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -517,42 +521,92 @@ def stat_like_fontmake(ttx):
...
-        for el in non_contours:
-            glyph.remove(el)
-        glyph.extend(normalized + non_contours)
 
-        # store new indices order
-        name = glyph.attrib["name"]
-        point_orders[name] = [idx for indices, _ in annotated for idx in indices]
+        # Compare contours as sets to see if they're identical (ignoring order)
+        fontc_strings = {to_xml_string(c) for c in fontc_contours}
+        fontmake_strings = {to_xml_string(c) for c in fontmake_contours}
```
cmyr, line 559: it feels like this could be a source of bugs?

like the paths, `(0,0), (4,0), (4,4), (0,4), (0,0)` and `(0,0), (4,4), (4,0), (0,4), (0,0)` would be considered equal even though the points are in a meaningfully different order? (this is not counting the case where the paths have different lengths but still produce a unique set of glyphs, which also isn't checked?)


In any case I don't know how 'real' this situation is, but it does make me a bit uneasy.

anthrotype: > like the paths, (0,0), (4,0), (4,4), (0,4), (0,0) and (0,0), (4,4), (4,0), (0,4), (0,0) would be considered equal even though the points are in a meaningfully different order? 

no! `c` is a whole contour (list of points), not single point.. We already `continue` if the contour count doesn't match, before proceeding to compare contours as xml strings here.

anthrotype: > the case where the paths have different lengths but still produce a unique set of glyphs

I don't understand what you mean. If two contours have different lengths (different number of points in each), their xml string will be different..

cmyr: yes ignore me I was thinking of normalizing point order _within_ a contour 🙄

anthrotype: yeah, no this PR is only about NOT normalizing the order of whole contours when their respective points within them do not match exactly (the contour's xml string being different)

anthrotype: But I do admit that the wording of the PR description could be confusing

Issue #552, opened 2023-11-08
[Glyphs] Issue with components from intermediate layers


anthrotype:
exactly the same as https://github.com/googlefonts/glyphsLib/issues/954

cmyr:
This is very similar to #1592, but for decomposition instead of flattening.

PR #1657, opened 2025-09-25
[ir] Interpolate component layers during flattening

### LINKED ISSUE #1592: flattening components can drop intermediate layers

cmyr:
That is, if a component has a location that is missing from the composite being flattened, add a new interpolated layer to the composite.

- closes #1592 

Note that this patch only addresses the issue during flattening, not during decomposition. Decomposition is tracked in #552.


fontir/src/glyph.rs
```diff
@@ -424,6 +458,14 @@ fn get_or_instantiate_instance<'a>(
         return Ok(Cow::Borrowed(instance));
     }
     log::debug!("instantiating instance of '{}' at {loc:?}", glyph.name);
+    instantiate_instance(glyph, loc, context).map(Cow::Owned)
+}
+
+fn instantiate_instance(
```
anthrotype, line 464: how about instantiate_glyph?

fontir/src/glyph.rs
```diff
@@ -1688,6 +1730,57 @@ mod tests {
...
+            ),
+        ]);
+
+        let context = test_context_with_locations(vec![loc1.clone(), loc2.clone()]);
+        context.glyphs.set(aogonek.0.clone());
+        context.glyphs.set(aogonek_accent.0);
+        context.glyphs.set(ogonekcomb_case.0);
+        context.glyphs.set(ogonek.0);
+
+        flatten_all_non_export_components(&context).unwrap();
```
anthrotype, line 1779: btw I think this function should really be renamed decompose_all_non_export_components because flatten is used in fontmake and elsewhere to mean making a nested component tree flat (with depth=1)

anthrotype (approved):

fontir/src/glyph.rs
cmyr: you mean as an alternative name?

cmyr: going to keep this, we aren't making a full glyph, we're making an instance?

PR #1658, opened 2025-09-25
[ir] Also interpolate width/height etc in instances

cmyr:
This also more generally reworks the interpolation code so that it all lives in one place (methods on `GlyphInstance`).

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -14,7 +14,7 @@ use fontdrasil::{
     orchestration::{Access, AccessBuilder, Work},
     types::GlyphName,
 };
-use kurbo::{Affine, BezPath, PathEl, Point, Vec2};
+use kurbo::Affine;
 use log::{debug, log_enabled, trace};
 use ordered_float::OrderedFloat;
 use write_fonts::types::GlyphId16;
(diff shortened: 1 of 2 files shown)


fontir/src/glyph.rs
```diff
@@ -466,19 +465,21 @@ fn instantiate_instance(
     loc: &NormalizedLocation,
     context: &Context,
 ) -> Result<GlyphInstance, BadGlyph> {
+    log::debug!("instantiating instance of '{}' at {loc:?}", glyph.name);
```
anthrotype:
```suggestion
    log::debug!("instantiating '{}' at {loc:?}", glyph.name);
```

fontir/src/ir.rs
```diff
@@ -1805,6 +1805,136 @@ impl GlyphInstance {
...
+
+    /// Return the values in this instance used for interpolation.
+    ///
+    /// This is used for instantiating interpolated instances.
+    ///
+    /// The returned values must be provided in the following order:
+    ///
+    /// - the (x,y) values of points in any contours, in order
+    /// - the decomposed transform of any components, in order
+    /// - the width, height, and vertical origin (if present).
```
anthrotype, line 1817: anchor positions also interpolate... in fact Glyphs.app anchor propagation works even when a composite glyph inherit anchors from a component that doesn't have a given location, which gets interpolated on the fly

cmyr: hm yea anchors aren't part of `ir::Glyph`, so we'll need to do that separately (opened #1661)

PR #1664, opened 2025-09-29
ttx_diff: normalize all-offcurve starting point

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/1653

```
$ python3 resources/scripts/ttx_diff.py 'https://github.com/appajid/Sirivennela?1e9e8dbbc7#sources/Sirivennela-Regular.ufo
[...]
output is identical
```

JMM

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -520,6 +520,37 @@ def stat_like_fontmake(ttx):
     ver.attrib["value"] = "0x00010001"
 
 
+def normalize_all_offcurve_starting_point(contour):
+    """Rotate an all-offcurve contour to start at point closest to origin.
+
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -520,6 +520,37 @@ def stat_like_fontmake(ttx):
     ver.attrib["value"] = "0x00010001"
 
 
+def normalize_all_offcurve_starting_point(contour):
+    """Rotate an all-offcurve contour to start at point closest to origin.
```
rsheeter, line 524: Maybe clarify how ties are resolved?

rsheeter (approved):

resources/scripts/ttx_diff.py
anthrotype: done

PR #1666, opened 2025-09-29
Add VVAR for variable fonts that support vertical

rsheeter:
@cmyr pointed out we have a bunch of `VVAR	fontmake only` diffs in crater, first swing at adding it.

Makes _at least_ the following identical:

* `python3 resources/scripts/ttx_diff.py 'https://github.com/notofonts/ethiopic?1cc6933c58#sources/NotoSerifEthiopic.glyphs'`
* `python3 resources/scripts/ttx_diff.py 'https://github.com/notofonts/tamil?626575fa0c#sources/NotoSerifTamil.glyphs'`

Hopefully helps with others whose diff is `(VVAR)` as well.

Mostly just boilerplate and shuffling things to not be HVAR-only. Note that `resources/testdata/HVVAR/SingleModel_Direct/SingleModelDirect-Regular.ufo/fontinfo.plist` now specifies the magic keys to trigger build vertical.


fontbe/src/vvar.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn read_access(&self) -> Access<AnyWorkId> {
+        AccessBuilder::new()
+            .variant(FeWorkId::StaticMetadata)
+            .variant(FeWorkId::GlobalMetrics)
+            .variant(FeWorkId::GlyphOrder)
+            .variant(FeWorkId::Glyph(GlyphName::NOTDEF))
+            .build()
+    }
+
+    /// Generate [HVAR](https://learn.microsoft.com/en-us/typography/opentype/spec/HVAR)
```
anthrotype:
```suggestion
    /// Generate [VVAR](https://learn.microsoft.com/en-us/typography/opentype/spec/VVAR)
```

fontbe/src/vvar.rs
```diff
@@ -0,0 +1,127 @@
+//! Generates an [HVAR](https://learn.microsoft.com/en-us/typography/opentype/spec/HVAR) table.
```
anthrotype:
```suggestion
//! Generates an [VVAR](https://learn.microsoft.com/en-us/typography/opentype/spec/VVAR) table.
```

anthrotype:
btw, both NotoSerifEthiopic.glyphs and NotoSerifTamil.glyphs may now be identical in both fontc and fontmake, but they actually produce a no-op VVAR with _empty_ VarData items, so they may not actually be the best candidates for testing:

```xml
  <VVAR>
   <Version value="0x00010000"/>
    <VarStore Format="1">
      <Format value="1"/>
      <VarRegionList>
        <!-- RegionAxisCount=2 -->
        <!-- RegionCount=0 -->
      </VarRegionList>
      <!-- VarDataCount=1 -->
      <VarData index="0">
        <!-- ItemCount=222 -->
        <NumShorts value="0"/>
        <!-- VarRegionCount=0 -->
        <Item index="0" value="[]"/>
        <Item index="1" value="[]"/>
        <Item index="2" value="[]"/>
        <Item index="3" value="[]"/>
        <!-- empty [] all the way down to the end -->
```

I wonder why do we even bother compiling such a VVAR...

fontbe/src/error.rs
```diff
@@ -97,6 +97,8 @@ pub enum Error {
     CompositesStalled(Vec<GlyphName>),
     #[error("Inconsistent palette lengths observed: {0:?}")]
     InconsistentPaletteLength(Vec<usize>),
+    #[error("No metrics instance for '{0:?}'")]
+    NoGlobalMetricsInstance(NormalizedLocation),
```
anthrotype, line 101: actually, the correct thing to do is _not_ to error when we don't have a GlobalMetricsInstance at the same location as a given glyph instance, but we should interpolate one! https://github.com/googlefonts/fontc/pull/1518 was added by @Hoolean exactly for this purpose.

rsheeter:
> @rsheeter did you see this https://github.com/googlefonts/fontc/pull/1489 ?

Nope, just noticed in conversation with Colin and filled in some of the blanks! If that is better happy to land that and close this.

anthrotype:
> If that is better happy to land that and close this.

I prefer this one because it extracts the common code in a shared metric_variations.rs and has tests. But we can incorporate the missing bits from that one here. I'm already working on that actually

fontc/src/lib.rs
```diff
@@ -2073,44 +2074,135 @@ mod tests {
...
-                ]
-            })
-            .collect::<Vec<_>>();
-        assert_eq!(region_coords, vec![[0.0, 1.0, 1.0]]);
+        // so we expect a single region with a single wght axis for both hvar and vvar
+        assert_var_regions(&varstore, &vec![vec![[0.0, 1.0, 1.0]]]);
+
         // we expect one ItemVariationData and one delta set per glyph
-        assert_eq!(varstore.item_variation_data_count(), 1, "{varstore:#?}");
+        // assert_eq!(varstore.item_variation_data_count(), 1, "{varstore:#?}");
```
anthrotype: why is this commented out?

rsheeter: Human error :) I think I commented out part of the test while shuffling code.

fontbe/src/error.rs
rsheeter: Should that just be how accessing metrics works? Is there a case where we *don't* want that?

anthrotype: you'll actually never hit this NoGlobalMetricsInstance error because when you call GlobalMetrics::at below inside AdvanceDeltas::new, new GlobalMetricsInstances will be interpolated automatically if missing:

```rust
        let metrics = glyph_locations
            .into_iter()
            .map(|loc| {
                let loc = loc.subset_axes(&axes);
                let metrics = global_metrics.at(&loc);
                (loc, metrics)
            })
            .collect();
```

we can probably get rid of it altogether and simply assert/unwrap

anthrotype: > Is there a case where we don't want that?

no, we do want to interpolate all the time

PR #1668, opened 2025-09-30
Match fontmake name nuance

rsheeter:
Makes `python3 resources/scripts/ttx_diff.py 'https://github.com/ktkm/Vend-Sans?2663f1c7da#sources/VendSans-Italic.glyphs'` identical

JMM

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -940,14 +940,17 @@ impl RawName {
         // See <https://github.com/googlefonts/fontc/issues/1011>
         // In order of preference: dflt, default, ENG, whatever is first
         // <https://github.com/googlefonts/glyphsLib/blob/1cb4fc5ae2cf385df95d2b7768e7ab4eb60a5ac3/Lib/glyphsLib/classes.py#L3155-L3161>
+        // If the same name is defined repeatedly, e.g. dflt has 2 values, pick the last occurrence.
+
+        let scale: i32 = 1000;
         self.values
             .iter()
             .enumerate()
             // (score [lower better], index)
             .map(|(i, raw)| match raw.language.as_str() {
-                "dflt" => (-3, raw.value.as_str()),
-                "default" => (-2, raw.value.as_str()),
-                "ENG" => (-1, raw.value.as_str()),
+                "dflt" => (-scale * 3 - i as i32, raw.value.as_str()),
+                "default" => (-scale * 2 - i as i32, raw.value.as_str()),
+                "ENG" => (-scale - i as i32, raw.value.as_str()),
                 _ => (i as i32, raw.value.as_str()),
             })
             .min()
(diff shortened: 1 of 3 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -2631,6 +2632,23 @@ mod tests {
         );
     }
 
+    #[test]
+    fn prefers_first_if_defined_repeatedly() {
```
anthrotype: did you mean prefer "last"?

```suggestion
    fn prefers_last_if_defined_repeatedly() {
```

glyphs-reader/src/font.rs
```diff
@@ -940,14 +940,18 @@ impl RawName {
...
             // (score [lower better], index)
             .map(|(i, raw)| match raw.language.as_str() {
-                "dflt" => (-3, raw.value.as_str()),
-                "default" => (-2, raw.value.as_str()),
-                "ENG" => (-1, raw.value.as_str()),
+                "dflt" => (-scale * 3 - i as i32, raw.value.as_str()),
+                "default" => (-scale * 2 - i as i32, raw.value.as_str()),
+                "ENG" => (-scale - i as i32, raw.value.as_str()),
+                // fontmake probably picks the last value for the first surprise name
+                // Lets leave that gem until an actual source exhibits reliance on that behavior.
```
anthrotype: I don't understand this inline comment

glyphs-reader/src/font.rs
```diff
@@ -940,14 +940,18 @@ impl RawName {
         // See <https://github.com/googlefonts/fontc/issues/1011>
         // In order of preference: dflt, default, ENG, whatever is first
         // <https://github.com/googlefonts/glyphsLib/blob/1cb4fc5ae2cf385df95d2b7768e7ab4eb60a5ac3/Lib/glyphsLib/classes.py#L3155-L3161>
+        // If the same name is defined repeatedly, e.g. dflt has 2 values, pick the last occurrence.
```
anthrotype, line 943: effectively fontmake (glyphsLib actually) prefers the last of any duplicates simply because at parsing time it populates a `_localized_names` dictionary in a for loop where the last item with a given language will be used as key overwriting any previous ones:

https://github.com/googlefonts/glyphsLib/blob/1cb4fc5ae2cf385df95d2b7768e7ab4eb60a5ac3/Lib/glyphsLib/classes.py#L3126-L3131

glyphs2fontir/src/source.rs
rsheeter: ha yes! Initially I thought it would prefer first it seems lol

glyphs-reader/src/font.rs
rsheeter: If we had multiple unknown (not dflt DFLT etc) names each with multiple definitions it appears to me we'd pick the last definition of the first one. I didn't implement that aspect because it seems both accidental and unlikely to matter in reality. Let me tweak the comment.

rsheeter: ...I think it's just making things confusing, the comment above is much more helpful. Removed.

glyphs-reader/src/font.rs
rsheeter: I initially naively assumed that wouldn't matter, obviously there would be one value per language. And then ofc sources popped up with many.

rsheeter:
> technically this duplicated entries are a source font bug but LGTM if it matches fontmake

lol I share both of these thoughts

PR #1669, opened 2025-09-30
Implement vertical gvar phantom points

anthrotype:
Adapted @Hoolean's https://github.com/googlefonts/fontc/pull/1489 on top of https://github.com/googlefonts/fontc/pull/1666

Added tests covering both VVAR and gvar vertical phantom points.

### DIFF
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -14,7 +14,7 @@ use fontdrasil::{
     types::GlyphName,
 };
 use fontir::{
-    ir::{self, GlobalMetrics, GlobalMetricsInstance, GlyphInstance, GlyphOrder},
+    ir::{self, GlobalMetrics, GlyphOrder},
     orchestration::{Flags, WorkId as FeWorkId},
     variations::{VariationModel, VariationRegion},
 };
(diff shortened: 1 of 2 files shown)


fontbe/src/glyphs.rs
```diff
@@ -221,23 +221,42 @@ fn create_composite(
 
 /// * <https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L335-L367>
 /// * <https://docs.microsoft.com/en-us/typography/opentype/spec/tt_instructing_glyphs#phantoms>
-fn add_phantom_points(advance: u16, points: &mut Vec<Point>) {
+fn add_phantom_points(
+    instance: &GlyphInstance,
```
cmyr: maybe doesn't need to happen now, but in my own work I made this into a method on `GlyphInstance`, and I think that made sense. Maybe as followup?

cmyr (commented):
This is causing some regressions on crater, here are some examples:

```sh
python3 resources/scripts/ttx_diff.py 'https://github.com/googlefonts/buenard?cf400a29bb#sources/Buenard.glyphs'
python3 resources/scripts/ttx_diff.py 'https://github.com/weiweihuanghuang/Work-Sans?b35c810861#sources/WorkSans-Italic.glyphs'
python3 resources/scripts/ttx_diff.py 'https://github.com/m4rc1e/Commissioner?7f7dc8e9ed#sources/Commissioner.glyphs'
python3 resources/scripts/ttx_diff.py 'https://github.com/SUSE/suse-font?7159afb255#sources/SUSE.glyphs'
```

fontbe/src/glyphs.rs
anthrotype: Thanks for the review. I'll look at the regressions and move that into a method, I'll do it here no need for follow up

anthrotype:
> This is causing some regressions on crater, here are some examples:

@cmyr I rebased this PR on main and all the diffs you reported went away, so they weren't caused by this PR

anthrotype:
> then I'm happy to merge.

i'm gonna move add_phantom_points to a GlyphInstance method first

PR #1670, opened 2025-09-30
[ir] Don't round when computing deltas for intermediates

cmyr:
This was a sneaky little bug: the rounding of intermediate values caused us to incorrectly interpolate affine transformations that contained fractional values.

This matches ufo2ft, which sets [`round_geometry=False` on the instantiator](https://github.com/googlefonts/ufo2ft/blob/01d3faee6bd42eee2874f715c2aab2754849b955/Lib/ufo2ft/_compilers/baseCompiler.py#L266).

This resolves a number of gvar diffs, and Should give us at least a +4 identical.


fontir/src/glyph.rs
```diff
@@ -1639,6 +1643,39 @@ mod tests {
         assert!(inst.height.is_none());
         assert_eq!(inst.vertical_origin, Some(150.));
     }
+
+    fn interpolate_transform(one: Affine, two: Affine) -> Affine {
```
rsheeter: I was confused by interpolate with A, B and no t (how far between a and be the result should be). Maybe we should either add t or name in a way that indicates where our interpolation lands?

rsheeter: e.g. interpolate_midpoint(a, b) I would get

rsheeter (approved):

anthrotype (approved):

PR #1671, opened 2025-09-30
[glyphs] Fix logic bug around bracket glyph components

cmyr:
We were deciding whether to replace a bracket glyph's component with a an appropriate bracket glyph for that component based on whether the two alternates had equivalent names, instead of whether the replacements applied in equivalent regions.

### DIFF
--- a/resources/testdata/glyphs3/manual_bracket_layer_and_component_has_one_too.glyphs
+++ b/resources/testdata/glyphs3/manual_bracket_layer_and_component_has_one_too.glyphs
@@ -1,0 +1,253 @@
+{
+.appVersion = "3343";
+.formatVersion = 3;
+axes = (
+{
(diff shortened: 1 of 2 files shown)


resources/testdata/glyphs3/manual_bracket_layer_and_component_has_one_too.glyphs
```diff
@@ -0,0 +1,270 @@
...
+}
+);
+metrics = (
+);
+properties = (
+);
+unitsPerEm = 1000;
+versionMajor = 1;
+versionMinor = 701;
+}
```
rsheeter, line 253: As ever, please minimize if possible. I imagine things like a blank metrics and perhaps some of the properties are not used?

PR #1673, opened 2025-10-02
Show % identical on crater

rsheeter:
Show % identical, combine identical/target into one column. Makes the number driving toward 100 a little easier to see, particularly in the face of a volatile denominator.

image

JMM

### DIFF
--- a/fontc_crater/resources/style.css
+++ b/fontc_crater/resources/style.css
@@ -51,11 +51,19 @@ td.rev {
 }
 
 #results_head th:nth-child(5) {
-  width: 9%
+  width: 8%
+}
+
+#results_head th:nth-child(6) {
+  width: 8%
 }
 
 #results_head th:nth-child(7) {
-  width: 9%
+  width: 8%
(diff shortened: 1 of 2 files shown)


cmyr:
Is your main interest in wanting to just see the identical %? I do find the combined column a bit information dense, and it does make it harder to see the changes in total targets.

Does #1674 give you the information you need?

rsheeter:
Either works, perhaps unsurprisingly I prefer my own formatting. I think you should pick a winner, merge one and close the other. Both are an improvement over what we have.

> I do find the combined column a bit information dense, and it does make it harder to see the changes in total targets.

fwiw I have almost exactly the reverse reaction, I can get the information much faster in the denser presentation. That said, I don't know if that generalizes.

> Does https://github.com/googlefonts/fontc/pull/1674 give you the information you need?

Strictly speaking yes, but at least for me it's harder to read and also isn't in the format I want to copy/paste into things like status updates.

PR #1675, opened 2025-10-02
Fix BASE table parsing infinite loop memory crash

RickyDaMa:
When parsing MinMax records, the token declaring the record isn't consumed, and so the parser immediately starts creating another MinMax record, but the token declaring the record isn't consumed, and so the parser immediately starts creating another MinMax record, but the token declaring the record isn't consumed, and so the parser immediately starts creating another MinMax record, but the token declaring the record isn't consumed, and so the parser immediately starts creating another MinMax record, but the token declaring the record isn't consumed, and so the parser immediately starts creating another MinMax record, but the token declaring the record isn't consumed, and so the parser immediately starts creating another MinMax record—

_Killed out of memory._

Instead, we now consume the MinMax token so the parser actually progresses. Also speculatively fixes this same pattern in some GDEF table parsing code


Hoolean:
(We've made an issue for full compilation in the future, too: #1676 😃)

PR #1678, opened 2025-10-03
Use ~ not absolute for display of paths

rsheeter:
ttx_diff output is made longer and harder to read by use of absolute paths. Try to print something closer to what a user would type, e.g. `~/.fontc_crater_cache` not `/usr/ate/a/donkey/near/rsheeter/.fontc_crater_cache`

### DIFF
--- a/resources/scripts/ttx_diff.py
+++ b/resources/scripts/ttx_diff.py
@@ -37,6 +37,7 @@ JSON:
 from collections import defaultdict
 from absl import app
 from absl import flags
+from functools import cache
 from lxml import etree
 from pathlib import Path
 import json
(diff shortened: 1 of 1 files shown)


resources/scripts/ttx_diff.py
```diff
@@ -146,14 +147,38 @@ def to_xml_string(e) -> str:
...
 
 
+@cache
+def home() -> Path:
+    return Path("~").expanduser()
+
+
+@cache
+def home_str() -> str:
+    return str(home())
```
cmyr: is there any advantage to having two fns here instead of just inlining?

rsheeter: Nope! Will collapse

PR #1683, opened 2025-10-07
[glyphs-reader] handle special custom parameters before to_custom_params() to avoid 'unknown' warnings

### LINKED ISSUE #1682: [glyhps-reader] 'unknown custom parameters' false warnings for known parameters
The `RawCustomParameters::to_custom_params()` method (called to convert 'raw' parameters to CustomParameters struct for font, masters and instances) iterates through all the raw custom parameters and processes them through a large match statement; the ones that aren't matched fall through to the default case which emits a logging warning 'unknown custom parameter {name}'.

However, there are parameters like "Axis Mappings", "Axis Location" or "Variable Font Origin" that are NOT listed in that match statement but are accessed via separate helper methods (e.g. axis_mappings() or get() or string() methods of RawCustomParameters), thus they will trigger a "unknown custom parameter" warning even though they are actually "know" and handled separately.

We could add them to the match so they get silently skipped instead of logged. But then we would have to maintain a duplicate list of parameter names that are handled specially.

Thinking about it, I realise these special parameters are all read BEFORE the respecite `to_custom_params()` calls and then never accessed afterwards. They are consumed for things like v2->v3 conversion or the axis mappings setup and never needed again. So we could change it so that the helper methods that currently read these parameters will pop them out and consume them, so that later `to_custom_params()` can process what's left and issue warnings for the parameters that are actually unknown.

anthrotype:
Previously, some special custom parameters like "Axis Mappings", "Axis Location", and "Variable Font Origin" were triggering false "unknown custom parameter" logging warnings because they would fall through the match statement in RawCustomParams::to_custom_params() method. Here we make so that these special parameters get extracted and parsed ('consumed') before the generic to_custom_params() conversion runs, which prevents them from being flagged as 'unknown'.

Fixes #1682

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -570,6 +570,16 @@ struct RawCustomParameterValue {
     disabled: Option<bool>,
 }
 
+impl RawCustomParameterValue {
+    /// Returns true if this parameter has been taken
+    ///
+    /// We use a special sentinel value to indicate that a parameter has been handled
+    /// specially and shouldn't be parsed as/stored in CustomParameters.
+    fn taken(&self) -> bool {
+        self == &Self::default()
+    }
+}
+
 impl FromPlist for FormatVersion {
     fn parse(tokenizer: &mut Tokenizer) -> Result<Self, crate::plist::Error> {
         let raw: i64 = FromPlist::parse(tokenizer)?;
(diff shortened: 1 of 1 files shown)


cmyr (approved):
Okay this takes a minor efficiency hit (lots of resizing of the array) but I think it's worth it for the knowledge that we're exhaustively handling these params.

anthrotype:
shall we instead replace them with a sentinel value that we can then check in to_custom_params to avoid popping them and resizing?

cmyr:
i was considering something like that, but I'd be worried about unintended side effects at some point 😒

anthrotype:
classic premature optimization, but too late, I committed

PR #1684, opened 2025-10-07
Compile anything which implements Source

simoncozens:
Current, the input to the compiler API is `Input` - which admittedly makes sense, given the name. But it means that inputs are restricted to the set of source types enumerated in the `Input` enum. If we have people implementing their own source types (which we do!) then it makes sense for allow compilation of *any* object which `impl Source`. This pushes the creation of a `Source` from an `Input` higher up the pipeline, so that any `Box<dyn Source>` can be compiled.

Also improves the lifetime story of `Workload` to avoid messing with `'static` - `Workload`s only need hang around as long as the `Source`s that they are compiling.

### DIFF
--- a/fontc/src/workload.rs
+++ b/fontc/src/workload.rs
@@ -52,7 +52,7 @@ use log::{debug, trace, warn};
 use crate::{
     timing::{JobTime, JobTimer},
     work::{AnyAccess, AnyContext, AnyWork},
-    Error, Input,
+    Error,
 };
 
 /// A set of interdependent jobs to execute.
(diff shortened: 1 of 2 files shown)


fontc/src/workload.rs
```diff
@@ -115,16 +115,13 @@ fn priority(id: &AnyWorkId) -> u32 {
     }
 }
 
-impl Workload {
+impl<'a> Workload<'a> {
```
cmyr: where does this lifetime parameter come from? More generally I don't think any of the lifetime additions in this patch are necessary? (if I just delete them all it still compiles)

cmyr (approved):
This looks good to me, modulo me not being sure the lifetimes are necessary :)

simoncozens:
(I removed the lifetime business.)

PR #1686, opened 2025-10-07
VariationModel only needs to know the axis order

cmyr:
Instead of needing to fully own a copy of the `Axes` struct.

- this will simplify using the variation model for the mini design spaces used by smart components
- this means there is only one copy of 'Axes', which lives in StaticMetadata
- overall this makes it easier to construct a VariationModel.

### DIFF
--- a/fontdrasil/src/variations.rs
+++ b/fontdrasil/src/variations.rs
@@ -8,7 +8,7 @@ use std::{
 
 use crate::{
     coords::{NormalizedCoord, NormalizedLocation},
-    types::{Axes, Axis},
+    types::Axes,
 };
 use log::{log_enabled, trace};
 use ordered_float::OrderedFloat;
@@ -112,7 +112,7 @@ pub struct VariationModel {
     pub default: NormalizedLocation,
 
     /// Non-point axes
-    axes: Axes,
+    axis_order: Vec<Tag>,
 
     // TODO: why isn't this a Map<Loc, Region>
     // All Vec's have same length and items at the same index refer to the same master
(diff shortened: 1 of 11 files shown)


fontdrasil/src/variations.rs
```diff
@@ -130,39 +130,32 @@ impl VariationModel {
...
-        axes: Axes,
-    ) -> Result<Self, VariationModelError> {
-        for axis in axes.iter() {
-            if axis.is_point() {
-                return Err(VariationModelError::PointAxis(axis.tag));
-            }
-        }
-
-        let default = axes
+    /// Axis order should reflect the importance of the axis, and should not include
```
anthrotype: what do you mean by "the importance of the axis"? As in, the so-called "point" axes aren't important and should be left out from axis order?

anthrotype: oh I see that sentence was already there before this. It doesn't make any sense to me, the axis order is simply that, the order in which axes appear in the fvar, I don't think it means that the first axis is more important than the second one, just the way they are listed.

fontdrasil/src/variations.rs
```diff
@@ -130,39 +130,32 @@ impl VariationModel {
...
+                    location.insert(*axis, NormalizedCoord::new(0.0));
                 }
             }
 
             expanded_locations.insert(location);
         }
 
         // sort locations such that [i..N] cannot influence [0..i-1]
-        let axis_order = axes.iter().map(|a| a.tag).collect();
+        //let axis_order = axes.iter().map(|a| a.tag).collect();
```
anthrotype: why did you keep this?

fontdrasil/src/variations.rs
cmyr: yea, I've kept the existing comment, which I'm also not clear about.

PR #1689, opened 2025-10-09
Glyphs smart components

cmyr:
This is a checkpoint, and only glyphs3 sources are supported. (I believe that smart components might also exist in glyphs2 sources)

### DIFF
--- a/glyphs-reader/src/smart_components.rs
+++ b/glyphs-reader/src/smart_components.rs
@@ -25,6 +25,12 @@ pub enum BadSmartComponent {
         child: AxisPole,
         axis: SmolStr,
     },
+    #[error("Smart component layers are not interpolation compatible")]
+    IncoherentLayers,
+    #[error("Axis '{0}' is not defined for the default layer")]
+    AxisUndefinedForDefaultLayer(SmolStr),
+    #[error("No layers exist with associated id '{0}'")]
+    NoLayer(String),
 }
 
 /// Instantiate an instance of a smart component.
@@ -55,6 +61,9 @@ pub(crate) fn instantiate_for_layer(
         })
         .collect::<Vec<_>>();
 
+    if relevant_layers.is_empty() {
+        return Err(BadSmartComponent::NoLayer(layer_master_id.to_string()));
+    }
     if relevant_layers.len() == 1 {
         log::debug!("smart component {} only has one layer?", component.name);
         let mut shapes = relevant_layers[0].shapes.clone();
@@ -62,6 +71,16 @@ pub(crate) fn instantiate_for_layer(
             .iter_mut()
             .for_each(|shape| shape.apply_affine(component.transform));
         return Ok(shapes);
+    }
+
+    validate_relevant_layers(&relevant_layers)?;
+    for axis in ref_glyph.smart_component_axes.keys() {
+        if !relevant_layers[0]
+            .smart_component_positions
+            .contains_key(axis)
+        {
+            return Err(BadSmartComponent::UnknownAxis(axis.clone()));
+        }
     }
 
     let locations = relevant_layers
@@ -75,11 +94,11 @@ pub(crate) fn instantiate_for_layer(
         .smart_component_axes
         .iter()
         .map(|(name, range)| {
-            let default_value = if relevant_layers[0]
+            let default_value = if *relevant_layers[0]
                 .smart_component_positions
                 .get(name)
-                .copied()
-                == Some(AxisPole::Min)
+                .unwrap()
+                == AxisPole::Min
             {
                 *range.start()
             } else {
@@ -129,6 +148,29 @@ pub(crate) fn instantiate_for_layer(
     });
 
     Ok(shapes)
+}
+
+fn validate_relevant_layers(layers: &[&Layer]) -> Result<(), BadSmartComponent> {
+    let Some((head, tail)) = layers.split_first() else {
+        return Ok(());
+    };
+
+    for layer in tail {
+        if layer.shapes.len() != head.shapes.len()
+            || layer
+                .shapes
+                .iter()
+                .zip(&head.shapes)
+                .any(|pair| match pair {
+                    (Shape::Component(_), Shape::Component(_)) => false,
+                    (Shape::Path(one), Shape::Path(two)) => one.nodes.len() != two.nodes.len(),
+                    _ => true,
+                })
+        {
+            return Err(BadSmartComponent::IncoherentLayers);
+        }
+    }
+    Ok(())
 }
 
 // component parts just have names, not tags, but VariationModel needs tags;


fontdrasil/src/coords.rs
```diff
@@ -315,6 +315,11 @@ impl<Space> Location<Space> {
             .collect()
     }
 
+    /// `true` if this location contains only and exactly the axes provided
+    pub fn has_exact_axes(&self, axes: &[Tag]) -> bool {
+        axes.len() == self.0.len() && axes.iter().all(|tag| self.0.contains_key(tag))
```
rsheeter, line 325: Nit: len == len seems invalid given that `&[Tag]` is not necessarily unique, just the all are in self.0 alone might make sense?

fontdrasil/src/coords.rs
```diff
@@ -373,6 +378,16 @@ impl<Space> Location<Space> {
...
+    pub fn fit_to_axes(&mut self, axes: &[Tag]) {
+        self.0.retain(|k, _| axes.contains(k));
+        for ax in axes {
+            self.0.entry(*ax).or_default();
+        }
+    }
+
     pub fn has_non_zero(&self, tag: Tag) -> bool {
         self.get(tag).unwrap_or_default().to_f64() != 0.0
     }
```
rsheeter, line 398: New behavior seems testable but no new tests

fontdrasil/src/variations.rs
```diff
@@ -333,6 +321,34 @@ impl VariationModel {
         Ok(result)
     }
 
+    /// Remove unknown axes and add missing axes to locations, if needed
+    fn normalize_point_seqs<'a, T: Clone>(
```
rsheeter: Could we call this fit_to_axes for symmetry with the location method?

glyphs-reader/src/error.rs
```diff
@@ -22,6 +25,14 @@ pub enum Error {
     WorstPlistEver(#[from] crate::plist::Error),
     #[error("Invalid code page {0}")]
     InvalidCodePage(u32),
+    #[error("Invalid value: {0}")]
+    BadValue(String),
+    #[error("Bad smart component '{component}' in glyph '{glyph}': {issue}")]
+    SmartComponent {
```
rsheeter: Nit: other errors have a prefix, e.g. BadSmartComponent, InvalidSmartComponent, or w/e

glyphs-reader/src/font.rs
```diff
@@ -290,6 +291,9 @@ pub struct Glyph {
     pub category: Option<Category>,
     pub sub_category: Option<Subcategory>,
     pub production_name: Option<SmolStr>,
+    /// If this is a smart component, these are the variable values.
+    // should this just be an 'axis'?
+    pub smart_component_axes: BTreeMap<SmolStr, RangeInclusive<i64>>,
```
rsheeter, line 295: It's isn't obvious if the key is a tag or a name, suggest adding a type or comment

glyphs-reader/src/font.rs
```diff
@@ -320,6 +332,7 @@ pub struct Layer {
     pub shapes: Vec<Shape>,
     pub anchors: Vec<Anchor>,
     pub attributes: LayerAttributes,
+    pub smart_component_positions: BTreeMap<SmolStr, SmartComponentValue>,
```
rsheeter: This feels obtuse enough to merit a comment about what it's telling us

glyphs-reader/src/font.rs
```diff
@@ -1066,6 +1096,8 @@ struct RawLayer {
     anchors: Vec<RawAnchor>,
     #[fromplist(alt_name = "attr")]
     attributes: LayerAttributes,
+    // if this layer is part of a smart component
```
rsheeter: Tell me what the key and value are, what would one do with the i164. It's an index into ... something?

glyphs-reader/src/font.rs
```diff
@@ -1132,6 +1166,8 @@ struct RawShape {
     // ref is reserved so take advantage of alt names
     #[fromplist(alt_name = "ref", alt_name = "name")]
     glyph_name: Option<SmolStr>,
+    // if this is a reference to a smart component, these are the values for interpolation
+    piece: BTreeMap<SmolStr, f64>,
```
rsheeter, line 1188: Can't tell what this holds, what are the key and value?

glyphs-reader/src/font.rs
```diff
@@ -1165,6 +1201,8 @@ pub struct Component {
     /// For instance, if an acute accent is a component of a ligature glyph,
     /// we might rename its 'top' anchor to 'top_2'
     pub anchor: Option<SmolStr>,
+    /// If this references a smart component, these are the values for interpolation.
+    pub smart_component_values: BTreeMap<SmolStr, f64>,
```
rsheeter, line 1226: Here too "the values for interpolation" is a bit opaque, what IS this?

There seem to be a bunch of these BTreeMap some identifier => something or other, should we add type?

glyphs-reader/src/font.rs
```diff
@@ -2455,6 +2494,17 @@ impl RawLayer {
...
+            .part_selection
+            .into_iter()
+            .map(|(k, v)| match v {
+                1 => Ok((k, SmartComponentValue::Min)),
+                2 => Ok((k, SmartComponentValue::Max)),
+                other => Err(Error::BadValue(format!(
+                    "expected only 1 or 2 in partSelection, found '{other}'"
+                ))),
+            })
+            .collect::<Result<_, _>>()?;
```
rsheeter, line 2527: Is it valid to have 2 entries that both have the same SmartComponentValue, e.g. 2 Min's?

glyphs-reader/src/font.rs
```diff
@@ -3085,6 +3143,9 @@ impl Font {
         if self.custom_parameters.propagate_anchors.unwrap_or(true) {
             self.propagate_all_anchors();
         }
+        // if any glyphs reference smart components, convert them to outlines.
+        // (see https://glyphsapp.com/learn/smart-components)
+        self.inline_all_smart_components()
```
rsheeter: Optional, for at least me `instantiate_smart_components` would more be a more intuitive name

glyphs-reader/src/font.rs
```diff
@@ -3159,6 +3220,80 @@ impl Font {
...
+                                    })?,
+                                );
+                            } else {
+                                layer.shapes.push(shape);
+                            }
+                        }
+                        _ => layer.shapes.push(shape),
+                    };
+                }
+            }
```
rsheeter, line 3308: This feels more gross than it perhaps has to be?

glyphs-reader/src/font.rs
```diff
@@ -3159,6 +3220,80 @@ impl Font {
...
+
+        if glyphs_with_smart_components.is_empty() {
+            return Ok(());
+        }
+
+        // convert the smart components to normal outlines, per-glyph
+        for mut glyph in glyphs_with_smart_components {
+            for layer in glyph.layers.iter_mut() {
+                let old_shapes = std::mem::take(&mut layer.shapes);
+                for shape in old_shapes {
```
rsheeter, line 3277: I haven't landed on a version I love but I felt this tweak to be at least a little nicer:

```rust
impl Shape {
    pub(crate) fn as_smart_component(&self) -> Option<&Component> {
        match self {
            Shape::Component(comp) if !comp.smart_component_values.is_empty() => Some(comp),
            _ => None,
        }
    }
}

...

                    // Be ye smart component?
                    let Some(comp) = shape.as_smart_component() else {
                        layer.shapes.push(shape);
                        continue;
                    };

                    // Is your referant usable?
                    let Some(ref_glyph) = self
                        .glyphs
                        .get(&comp.name)
                        .filter(|g| !g.smart_component_axes.is_empty())
                    else {
                        log::warn!(
                            "{} references {} but {} has no smart component axes, skipping",
                            glyph.name,
                            comp.name,
                            comp.name
                        );
                        // Is this right? We just ... keep the uninstantiated smart component?
                        layer.shapes.push(shape);
                        continue;
                    };
```

Just trying to cut indentation and take simple steps

glyphs-reader/src/font.rs
```diff
@@ -4846,4 +4981,95 @@ etc;
...
+        };
+        // the first point is at its lowest value (crotch -100) (first point is at end!)
+        assert_eq!(path.nodes.last().unwrap().pt.round().y, 267.0);
+        // the shoulder width means the max x is also at its lowest value
+        assert_eq!(
+            path.nodes.iter().map(|nd| nd.pt.x.round() as i64).max(),
+            Some(335)
+        );
+    }
 }
```
rsheeter, line 5106: Should we test some defective smart components?

glyphs-reader/src/smart_components.rs
```diff
@@ -0,0 +1,234 @@
...
+            let default_value = if relevant_layers[0]
+                .smart_component_positions
+                .get(name)
+                .copied()
+                == Some(SmartComponentValue::Min)
+            {
+                *range.start()
+            } else {
+                *range.end()
+            };
```
rsheeter, line 106: Is the assumption that there are exactly two and they must be min and max? We don't seem to actually check this?

glyphs-reader/src/smart_components.rs
rsheeter, line 1: The python refs are much appreciated but ... this is fairly hard to follow. Lots of noise for iters and maps and such that make it hard to see the logic. Maybe moving some of that into helper methods and/or adding comments might help?

glyphs-reader/src/smart_components.rs
```diff
@@ -0,0 +1,234 @@
...
+                points = &points[path.nodes.len()..];
+                Shape::Path(path)
+            }
+            // we just skip components, which matches fonttools. Should we error instead?
+            Shape::Component(_) => shape.clone(),
+        })
+        .collect()
+}
+
+//https://github.com/fonttools/fonttools/blob/03a3c8ed9e/Lib/fontTools/varLib/models.py#L47
```
rsheeter, line 215: The citation makes me wonder if this belongs somewhere in the var model code?

glyphs-reader/src/font.rs
cmyr: a given layer could be at the min of two different axes, if that's what you mean?

glyphs-reader/src/font.rs
cmyr: This is much easier to express after updating to rust 2024, and i like the `as_smart_component` method :)

glyphs-reader/src/font.rs
cmyr: I've ported the tests from fonttools, although that doesn't provide a ton of coverage of malformed inputs. It doesn't seem like this stuff is really used much in the wild after all, so I'm inclined to just get this in as a checkpoint, and if we run into problems we can flesh it out more.

PR #1693, opened 2025-10-15
Update gftools to latest v0.9.92 in requirements.txt

anthrotype:
We hadn't updated it since v0.9.85 which was released back in May 2025 apparently, I guess it's time to compare with the latest?

https://pypi.org/project/gftools/#history

JMM


cmyr (approved):
thanks!

PR #1698, opened 2025-10-16
[glyphs] Consistent smart comp layer order

### LINKED ISSUE #1694: smart components differences vs fontmake in Murecho.glyphs
```
$ python3 resources/scripts/ttx_diff.py 'https://github.com/positype/Murecho-Project?0efba44c1c#sources/Murecho.glyphs' --compare gftools --config ~/.fontc_crater_cache/positype/Murecho-Project/sources/config.yaml
```

gives a bunch of diffs in the glyf and gvar tables.

The diffs involve glyphs that use smart components: in particular uni624E, uni6253 and uni6255

All the three glyphs use the same smart component glyph "_smart.NSradical003'

cmyr:
We expect the master layer to always be at the start of the list, but this is not enforced in the source file.

gives us a +1?

closes #1694


glyphs-reader/src/smart_components.rs
```diff
@@ -53,14 +53,23 @@ pub(crate) fn instantiate_for_layer(
...
+    let mut relevant_layers = ref_glyph
         .layers
         .iter()
         .filter(|layer| {
             !layer.smart_component_positions.is_empty() && layer.master_id() == layer_master_id
         })
         .collect::<Vec<_>>();
 
+    // make sure default layer is first:
+    if let Some(idx) = relevant_layers
```
rsheeter, line 66: Does Python do this sort of reordering? - if so a citation would be nice. If not, why do we have to do it and python doesn't?

rsheeter (approved):

glyphs-reader/src/smart_components.rs
cmyr: added!

PR #1699, opened 2025-10-17
[pyproject.toml] fix dynamic version warning when running `uv venv`

anthrotype:
pyproject.toml is used to set up maturin to build the fontc wheel that we publish to PyPI.
We omit the version field from `[project]` because we want the same version as fontc's Cargo.toml to be used so they are in sycn. Maturin merges Cargo.toml metadata with pyproject.toml, with the latter overriding the former.

Now, version is a required field in pyproject.toml and tools such as `uv venv` would issue a warning when that's absent.

We can still omit it but we have to declare that version field as "dynamic" (i.e. to be defined at build time)

JMM

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,9 +24,12 @@ exclude = [
 [project]
 # 'name' key is required for [project]
 name = "fontc"
-# comment out 'version' to use the same as Cargo.toml, only override if needed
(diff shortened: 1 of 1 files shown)


anthrotype:
I tested locally by running `uv venv` (no more warnings) and maturin (wheels built successfully). 
The CI only runs maturin with tagged commit so it's not worth waiting on it to complete.

PR #1700, opened 2025-10-17
Glyphs with ufo filters

### LINKED ISSUE #1665: .glyphs sources converted from DS+UFO with glyphsLib may contain ufo2ft-specific filters in userData
which fontmake happily applies, whereas fontc doesn't know about

E.g. 

```
python3 resources/scripts/ttx_diff.py 'https://github.com/NaN-xyz/Rubik-Filtered?c89f25696f#glitchpop/sources/RubikGlitchPop.glyphs'
```

Will show a bunch of diffs in the component names because it actually runs the `flattenComponents` filter from ufo2ft, because it is defined inside the Glyphs' masters's `userData` (the equivalent of the `lib.plist` from UFO-land, a bag of residual private app-specific stuff that's carried over, often without the font developer even knowing or remembering...)

fontc should probably look for these when compiling from .glyphs inputs.

cmyr:
- closes #1665
- gets much of the rubik family matching (+20?)

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -568,6 +568,7 @@ fn apply_optional_transformations(
     context: &Context,
     glyph_order: &GlyphOrder,
 ) -> Result<(), BadGlyph> {
+    let source_args = context.static_metadata.get().args;
     // If we are decomposing all components, the rest of the flags can be ignored
     if context.flags.contains(Flags::DECOMPOSE_COMPONENTS) {
         for glyph_name in glyph_order.names() {
@@ -596,7 +597,7 @@ fn apply_optional_transformations(
         }
     }
 
-    if context.flags.contains(Flags::FLATTEN_COMPONENTS) {
+    if context.flags.contains(Flags::FLATTEN_COMPONENTS) || source_args.flatten_components {
         for glyph_name in glyph_order.names() {
             let glyph = context.get_glyph(glyph_name.clone());
             flatten_glyph(context, &glyph)?;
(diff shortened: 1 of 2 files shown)


fontir/src/glyph.rs
```diff
@@ -596,7 +597,7 @@ fn apply_optional_transformations(
         }
     }
 
-    if context.flags.contains(Flags::FLATTEN_COMPONENTS) {
+    if context.flags.contains(Flags::FLATTEN_COMPONENTS) || source_args.flatten_components {
```
rsheeter, line 600: Having multiple places to store the same flag is a pattern I've been burned by, and seen others be burned by, many times. How might we avoid this? Can we perhaps, maybe in a follow-on, get rid of context.flags and *only* have fields on static metadata?

cmyr: yea I think unifying this somehow would make a lot of sense.

fontir/src/ir/static_metadata.rs
```diff
@@ -78,6 +78,8 @@ pub struct StaticMetadata {
 
     /// Miscellaneous font-wide data that didn't seem worthy of top billing
     pub misc: MiscMetadata,
+    /// Additional compilation arguments derived from the source
+    pub args: SourceArgs,
```
rsheeter, line 82: I don't think we're benefitting very much from an extra struct over direct fields?

I also don't think "source args" is informative, better to describe whats here, which is basically instructions on how to process components?

PR #1705, opened 2025-10-20
[glyphs] Add support for localized font properties

### LINKED ISSUE #1695: new localized properties support in glyphsLib 6.12 causes name table diffs in crater
after bumping glyphsLib to 6.12.0 in requirements.txt with #1691, we got -28 no-longer identicals, seemingly all related to the `name` table.

I suspect it has to do with the new support for localized font info properties added with https://github.com/googlefonts/glyphsLib/pull/1108

(in turn meant to fix https://github.com/googlefonts/glyphsLib/issues/860)

we should support localized Glyphs font properties as well

anthrotype:
meant to match https://github.com/googlefonts/glyphsLib/pull/1108

Fixes #1695

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -74,8 +74,9 @@ pub struct Font {
     pub axis_mappings: UserToDesignMapping,
     pub virtual_masters: Vec<BTreeMap<String, OrderedFloat<f64>>>,
     pub features: Vec<FeatureSnippet>,
-    pub names: BTreeMap<String, String>,
-    pub localized_names: BTreeMap<String, Vec<(String, String)>>,
+    // Incliudes both default (English) and localized property names.
+    // Maps property key (e.g., "familyNames") to list of (language code, value)
+    pub all_names: BTreeMap<String, Vec<(String, String)>>,
     pub instances: Vec<Instance>,
     pub version_major: i32,
     pub version_minor: u32,
@@ -1010,6 +1011,19 @@ struct RawName {
     values: Vec<RawNameValue>,
 }
 
+/// Compute a priority score for a language code when selecting default names.
+///
+/// Lower scores have higher priority.
+/// Priority order: dflt > default > ENG > others (by insertion order)
+fn default_language_score(lang: &str, index: usize, scale: i32) -> i32 {
+    match lang {
+        "dflt" => -scale * 3 - index as i32,
+        "default" => -scale * 2 - index as i32,
+        "ENG" => -scale - index as i32,
+        _ => index as i32,
+    }
+}
+
 impl RawName {
     fn is_empty(&self) -> bool {
         self.value.is_none() && self.values.is_empty()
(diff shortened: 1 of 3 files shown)


glyphs-reader/src/font.rs
```diff
@@ -75,6 +75,7 @@ pub struct Font {
     pub virtual_masters: Vec<BTreeMap<String, OrderedFloat<f64>>>,
     pub features: Vec<FeatureSnippet>,
     pub names: BTreeMap<String, String>,
+    pub localized_names: BTreeMap<String, Vec<(String, String)>>,
```
rsheeter: Suggest a comment or two about the relationship between names and localized names. Immediate questions that pop into mind for me:

* is `names` the english subset of `localized names` (equivalent to localized_names['some key']`?)
* Why is the type of `names` not the same as the value type for `localized names`?

glyphs-reader/src/font.rs
```diff
@@ -3544,6 +3564,37 @@ mod tests {
         GlyphsAndPackage,
     }
 
+    /// Asserts that two Font structs are equal, excluding the localized_names field.
+    /// This is useful for comparing v2 and v3 Glyphs files where localized_names
+    /// may differ due to format changes.
+    fn assert_fonts_equal_sans_localized_names(f1: &Font, f2: &Font, context: &str) {
```
rsheeter, line 3650: This has several properties that seem undesirable:

1. Have to update every time you add a field, easy to forget
1. Reports only the first field that differs, not all of them

I wonder if we might clone, erase the localized_names field, and diff? - that would be simpler and addresses both concerns

resources/testdata/glyphs3/LocalizedNames.glyphs
```diff
@@ -0,0 +1,187 @@
...
+},
+{
+language = FRA;
+value = "Sous licence SIL Open Font License";
+},
+{
+language = DEU;
+value = "Lizenziert unter der SIL Open Font License";
+}
+);
```
rsheeter, line 130: Suggest adding at least one case where there are multiple definitions for the same language to demonstrate we intentionally pick a winner

glyphs-reader/src/font.rs
anthrotype: i'm not super happy about the distinction between bare `names` and `localized_names` myself, I did that initially to keep diffs to a minumum but I'll see if we can have one map to for all the names, English or not

rsheeter: Naively, perhaps it becomes `.names()` that just returns a value from `localized_names`? Or does it not fit together that way?

anthrotype: yeah, I think that should work. thanks

anthrotype: I unified the two maps in https://github.com/googlefonts/fontc/pull/1705/commits/5542ca367136c2ed4dc98e11217436114f25b73a

PR #1708, opened 2025-10-21
Fix HVAR/VVAR performance regression from eager metric computation

### LINKED ISSUE #1706: Performance regression for compiling from Designspace in commit 20eb45d39debb699abf5126db175ef66a5d0ad73
When comparing compilation performance from Designspaces, I noticed a performance regression compared to fontc v0.3.2. Compilation of a big font that you'll see as a backlink to this issue went from 12s on my machine to 60s. Bisecting it is pointing to commit 20eb45d39debb699abf5126db175ef66a5d0ad73.

Command to repro, using [hyperfine](https://github.com/sharkdp/hyperfine) to benchmark and [uvx](https://github.com/astral-sh/uv) to get v0.3.2.:

`hyperfine -w 1 -m 3 -n v0.3.2 "uvx fontc -b /tmp/from_ufo Font.designspace" -n git
 "./target/release/fontc -b /tmp/from_ufo Font.designspace"`
```
Benchmark 1: v0.3.2
  Time (mean ± σ):     29.858 s ±  0.933 s    [User: 42.125 s, System: 84.724 s]
  Range (min … max):   29.245 s … 30.932 s    3 runs

Benchmark 2: git
  Time (mean ± σ):     61.290 s ±  0.466 s    [User: 187.107 s, System: 4.605 s]
  Range (min … max):   60.908 s … 61.810 s    3 runs

Summary
  v0.3.2 ran
    2.05 ± 0.07 times faster than git
```

(Note: the Pythonized fontc is slower than raw fontc-v0.3.2 but you get the point.)

anthrotype:
The VVAR implementation in #1666 pre-computed GlobalMetricsInstance for all glyph locations upfront, causing thousands of unnecessary interpolations for sparse fonts. HVAR doesn't need metrics at all (only width), and VVAR was even recomputing the same location's metrics once per glyph, instead of once per unique glyph location.

We now compute metrics on-demand for VVAR only and cache per unique location.

Fixes #1706

### DIFF
--- a/fontbe/src/metric_variations.rs
+++ b/fontbe/src/metric_variations.rs
@@ -3,14 +3,14 @@
 //! [VVAR](https://learn.microsoft.com/en-us/typography/opentype/spec/VVAR) tables
 
 use std::any::type_name;
-use std::collections::{BTreeSet, HashMap};
+use std::collections::{BTreeSet, HashMap, HashSet};
 
 use fontdrasil::{
     coords::NormalizedLocation,
     types::{Axes, GlyphName},
     variations::VariationModel,
 };
-use fontir::ir::{GlobalMetrics, GlobalMetricsInstance, Glyph, GlyphInstance, StaticMetadata};
+use fontir::ir::{GlobalMetrics, GlobalMetricsInstance, Glyph, StaticMetadata};
 use write_fonts::{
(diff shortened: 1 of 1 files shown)


cmyr:
Will reiterate that it will be nice to set up some ongoing performance tracking at some point, would be great to catch this sort of thing in CI.

PR #1711, opened 2025-10-21
[glyphs] Instantiate smart components depth-first

cmyr:
That is, ensure that a smart component that has its own smart components will get them instantiated before it gets instantiated itself.

This improves our own behaviour but doesn't get us any green numbers because glyphsLib also doesn't handle this correctly, see https://github.com/googlefonts/glyphsLib/issues/1111


anthrotype (approved):
LGTM but let's wait to merge before glyphsLib catches up?

anthrotype:
https://github.com/googlefonts/glyphsLib/pull/1113 is merged, we can merge this as soon as we bump glyphsLib 6.12.2

PR #1714, opened 2025-10-22
[glyphs] Fix logic bug in category construction

cmyr:
A misplaced 'filter_map' was causing us to skip the (null) attributes of the first glyph in a ligature sequence if that glyph had a name we could not recognize. This caused us to treat the second glyph as the first, and so miscategorize various glyphs as non-spacing marks, which was causing various little diffs.

This is worth at least a +4, and improves the diff on a number of other targets.

### DIFF
--- a/glyphs-reader/src/glyphdata.rs
+++ b/glyphs-reader/src/glyphdata.rs
@@ -586,9 +586,12 @@ impl GlyphData {
         if let Some(base_names) = self.split_ligature_glyph_name(base_name) {
             let base_names_attributes: Vec<_> = base_names
                 .iter()
-                .filter_map(|name| self.query_no_synthesis(name, None))
+                .map(|name| self.query_no_synthesis(name, None))
                 .collect();
-            if let Some(first_attr) = base_names_attributes.first() {
+            if let Some(first_attr) = base_names_attributes
+                .first()
+                .expect("if we have base_names it is non-empty")
+            {
                 // if first is mark, we're a mark
                 if first_attr.category == Category::Mark {
                     return Some((Category::Mark, first_attr.subcategory));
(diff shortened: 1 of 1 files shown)


glyphs-reader/src/glyphdata.rs
```diff
@@ -1219,6 +1219,22 @@ mod tests {
         assert_eq!(g, Some((Category::Mark, Some(Subcategory::Nonspacing))));
     }
 
+    #[test]
+    fn unknown_name_combined_with_mark() {
+        // if first part is unknown, we don't assign a category
+        assert_eq!(get_category("Whata-WEIRDNameLOL_brevecomb", &[]), None)
```
anthrotype, line 1228:
```suggestion
        assert_eq!(get_category("Whata-WEIRDNameLOL_brevecomb", &[]), None);
```

glyphs-reader/src/glyphdata.rs
```diff
@@ -586,9 +586,9 @@ impl GlyphData {
         if let Some(base_names) = self.split_ligature_glyph_name(base_name) {
             let base_names_attributes: Vec<_> = base_names
                 .iter()
-                .filter_map(|name| self.query_no_synthesis(name, None))
+                .map(|name| self.query_no_synthesis(name, None))
                 .collect();
-            if let Some(first_attr) = base_names_attributes.first() {
+            if let Some(first_attr) = base_names_attributes.first().unwrap() {
```
rsheeter: Maybe a comment wrt unwrap safety?

anthrotype: is this unwrap safe? base_names_attributes can't ever be empty because split_ligature_glyph_name always returns something? maybe write a comment

glyphs-reader/src/glyphdata.rs
cmyr: the assert returns `()` which is the implied return value of the fn, so this is allowed 🤷

anthrotype: so you did it on purpose? naughty boy

PR #1717, opened 2025-10-24
Package ttx-diff as a standalone PyPI package

anthrotype:
This migrates ttx_diff.py from resources/scripts/ into a standalone Python package under ttx_diff/ that can be published to PyPI.

Making ttx-diff available as a standalone package enables easier installation for external users (pip install ttx-diff), better dependency management and reusability outside the fontc repository, e.g. to aid projects transitioning from fontmake to fontc.

The `ttx_diff/src/ttx_diff/core.py` module is the same old `resources/scripts/ttx_diff.py` (moved via `git mv` to preserve git history), with additional logic to be able to run from anywhere (not just fontc repo root as it was currently). It also accepts pre-compile `fontc` and `otl-normalizer` binaries via CLI flags, or falls back to searching in `$PATH` for globally installed binaries.

When running `ttx-diff` script from inside the fontc workspace, it will continue to automatically build fontc and otl-normalizer from source like it currently does, unless provided with explicit binary paths.

The ttx-diff dependencies are now stored in its pyproject.toml but the old `requirements.in` and `requirements.txt` are kept, so fontc_crater runs use pinned versions  and are reproducible.

We can finally add tests to ttx-diff, no more excuses...

I added a new `python.yml` workflow (to mimic the existing `rust.yml` one) that tests ttx-diff on linux and mac (no windows 🤷 ) and on Python 3.10 (the minimum nowadays) up to 3.14 (but skipped a few in between because if it works on 3.10 and 3.14, really *ought to* work on 3.11...)

The version is determined dynamically from git tags formatted as `ttx-diff-v*` (it's a monorepo and we have other tags like `fontc-v*`). It works.

### DIFF
--- a/resources/scripts/requirements.in
+++ b/resources/scripts/requirements.in
@@ -1,5 +1,5 @@
 # ttx-diff package with all its dependencies (single source of truth)
--e ./ttx_diff
+-e file:./ttx_diff
 
 # Pin versions for CI stability (overrides ttx-diff's minimum versions)
 # keep fontmake version pinned to ensure output from ttx-diff is stable;
--- a/resources/scripts/requirements.txt
+++ b/resources/scripts/requirements.txt
@@ -4,7 +4,7 @@
 #
 #    pip-compile resources/scripts/requirements.in
 #
--e file:///Users/clupo/oss/fontc/ttx_diff
+-e file:./ttx_diff
     # via -r resources/scripts/requirements.in
 absl-py==2.3.1
     # via
@@ -182,7 +182,7 @@ openstep-plist==0.5.1
     #   glyphslib
 opentype-sanitizer==9.2.0
     # via gftools
-orjson==3.11.3
+orjson==3.11.4
     # via
     #   babelfont
     #   ufolib2
@@ -238,7 +238,7 @@ resvg-cli==0.44.0
     # via nanoemoji
 rich==14.2.0
     # via gftools
-ruamel-yaml==0.18.15
+ruamel-yaml==0.18.16
     # via gftools
 ruamel-yaml-clib==0.2.14
     # via ruamel-yaml


fontc_crater/src/target.rs
```diff
@@ -145,7 +145,7 @@ impl Target {
             Default::default()
         };
         let mut cmd = format!(
-            "python3 resources/scripts/ttx_diff.py '{repo_url}{sha_part}#{}'",
+            "python3 -m ttx_diff '{repo_url}{sha_part}#{}'",
```
rsheeter, line 148: why not just ttx-diff, people should install it?

ttx_diff/README.md
```diff
@@ -0,0 +1,107 @@
...
+```bash
+pip install ttx-diff
+```
+
+### From source
+
+```bash
+git clone https://github.com/googlefonts/fontc.git
+cd fontc/ttx_diff
+pip install -e .
```
rsheeter, line 29: nit: pip install -e ./fontc/ttx_diff/ saves a line?

fontc_crater/src/target.rs
anthrotype: i wanted to avoid fontc_crater falling into a trap of invoking a random ttx-diff on PATH. The previous code was calling python3 with the script path. This is the equivalent, to ensure it runs the ttx_diff installed within the same environment as `python3` itself.

ttx_diff/README.md
anthrotype: well usually one `cd` into a git repository they have just cloned. They may as well cd into the ttx_diff directory and do the canonical `-e .` for clarity's sake

anthrotype:
amazing.. we haven't even merged this and pygit2 already broke us 🤣 
https://github.com/googlefonts/fontc/actions/runs/18788046631/job/53611237030?pr=1717

PR #1721, opened 2025-10-31
Corner components

cmyr:
This implements (partial) handling of glyphs [corner components](https://handbook.glyphsapp.com/components/corner/), based on the implementation in glyphsLib.

This passes the test cases from glyphsLib, with the exception of an off-by-one related to slightly different bezier math; in that case we seem to be matching the outline actually expected by glyphs.app, and I've modified our version of the test data so that we pass.

This doesn't really move the needle much on crater, unfortunately; there is one font that should now be matching but there's a lingering issue I need to track down.

### DIFF
--- a/glyphs-reader/src/corner_components.rs
+++ b/glyphs-reader/src/corner_components.rs
@@ -1,14 +1,11 @@
 //! Corner components support
 //!
 //! Implements corner component insertion for Glyphs fonts.
-//! Based on: https://github.com/googlefonts/glyphsLib/blob/main/Lib/glyphsLib/filters/cornerComponents.py
+//! Based on: <https://github.com/googlefonts/glyphsLib/blob/main/Lib/glyphsLib/filters/cornerComponents.py>
 
 use std::collections::BTreeMap;
 
-use kurbo::{
-    Affine, CubicBez, Line, ParamCurve, ParamCurveArclen, ParamCurveNearest, PathSeg, Point,
-    QuadBez,
-};
+use kurbo::{Affine, Line, ParamCurve, ParamCurveArclen, ParamCurveNearest, PathSeg, Point};
 use smol_str::SmolStr;
 use thiserror::Error;
 use write_fonts::OtRound;
@@ -56,33 +53,9 @@ impl BadCornerComponentReason {
 
 /// Find the t parameter on a segment at a given distance along it
 ///
-/// <https://github.com/googlefont/glyphsLib/blob/f90e4060/Lib/glyphsLib/filters/cornerComponents.py#L151>
+/// <https://github.com/googlefonts/glyphsLib/blob/f90e4060/Lib/glyphsLib/filters/cornerComponents.py#L151>
 fn point_on_seg_at_distance(seg: PathSeg, distance: f64) -> f64 {
     seg.inv_arclen(distance, 1e-6)
-    // below is an LLM impl
-    //if let PathSeg::Line(line) = &seg {
-    //let length = line.length();
-    //if length == 0.0 {
-    //return 0.0;
-    //}
-    //return (distance / length).clamp(0.0, 1.0);
-    //}
-
-    //// Simple approximation: find t where arc length ≈ distance
-    //let mut accumulated = 0.0;
-    //let steps = 100;
-    //let mut last_pt = seg.start();
-
-    //for i in 1..=steps {
-    //let t = i as f64 / steps as f64;
-    //let pt = seg.eval(t);
-    //accumulated += (pt - last_pt).hypot();
-    //if accumulated >= distance {
-    //return t;
-    //}
-    //last_pt = pt;
-    //}
-    //1.0
 }
 
 /// Insert all corner components for a layer
@@ -104,6 +77,8 @@ pub(crate) fn insert_corner_components_for_layer(
         return Ok(());
     }
 
+    // if we insert points for one corner, it will change the index of
+    // a subsequent corner, so we track how many points we've inserted.
     let mut inserted_pts = 0;
     let mut current_shape = 0;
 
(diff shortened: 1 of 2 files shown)


glyphs-reader/src/corner_components.rs
```diff
@@ -0,0 +1,602 @@
...
+            }
+        }
+    }
+}
+
+struct CornerComponent {
+    corner_path: Path,
+    other_paths: Vec<Path>,
+    // the 'left' anchor of the component, (0,0) by default
+    #[expect(dead_code)] // used for alignment, not handled yet
```
anthrotype, line 258: > not handled yet

does glyphsLib support these left/right anchors for aligning corner components?

https://handbook.glyphsapp.com/components/corner/ says these are relevant for diagonal stems

in general, would be nice if we can document which corner component features are NOT yet supported and maybe file a follow-up issue

glyphs-reader/src/corner_components.rs
```diff
@@ -0,0 +1,602 @@
...
+
+    for hint in corner_hints {
+        if hint.shape_index != current_shape {
+            current_shape = hint.shape_index;
+            inserted_pts = 0;
+        }
+
+        let Some(corner_glyph) = glyphs.get(&hint.name) else {
+            log::warn!("corner component '{}' not found", hint.name);
+            return Ok(());
```
anthrotype: glyphsLib does `continue` here instead of returning early like you do: 

https://github.com/googlefonts/glyphsLib/blob/bb17c74be63e1541a711aac230da4caf762c1920/Lib/glyphsLib/filters/cornerComponents.py#L503-L509

which means if one component is missing, subsequent ones will not be inserted... was that intentional?

glyphs-reader/src/corner_components.rs
```diff
@@ -0,0 +1,602 @@
...
+///
+/// <https://github.com/googlefonts/glyphsLib/blob/f90e4060b/Lib/glyphsLib/filters/cornerComponents.py#L127>
+fn unbounded_seg_seg_intersection(seg1: PathSeg, seg2: PathSeg) -> Option<Point> {
+    // Line-line intersection
+    match (seg1, seg2) {
+        (PathSeg::Line(one), PathSeg::Line(two)) => one.crossing_point(two),
+        (seg, PathSeg::Line(line)) | (PathSeg::Line(line), seg) => {
+            // a value by which we extend our line, to find the crossing point.
+            // should be enough for anybody!
+            const LITERALLY_UNBOUNDED: f64 = 10000.;
```
anthrotype: depends on the font's units per EM... Hopefully this won't bite us later.
The python code doesn't need this magic constant. Why do we?

glyphs-reader/src/corner_components.rs
```diff
@@ -0,0 +1,602 @@
...
+            let direction = (line.p1 - line.p0).normalize();
+            let extended_line = Line::new(
+                line.p0 - direction * LITERALLY_UNBOUNDED,
+                line.p1 + direction * LITERALLY_UNBOUNDED,
+            );
+            seg.intersect_line(extended_line)
+                .first()
+                .map(|hit| seg.eval(hit.segment_t))
+        }
+        _ => None,
```
anthrotype, line 449: curve-curve intersection would silently return None, is that how it's supposed to work? just asking

glyphs-reader/src/corner_components.rs
cmyr: these are not supported in glyphsLib, which is why I left them out here; this is strongly based on the glyphsLib impl, and so I don't fully know what other features aren't implemented?

glyphs-reader/src/corner_components.rs
cmyr: good catch!

glyphs-reader/src/corner_components.rs
anthrotype: glyphsLib apparently parses these left/right anchors (as `effective_start` and `effective_end`) even though it logs "not currently supported"

https://github.com/googlefonts/glyphsLib/blob/bb17c74be63e1541a711aac230da4caf762c1920/Lib/glyphsLib/filters/cornerComponents.py#L315-L318

... then when doing the translation it only uses the x coordinate from effective_start, ignoring the y coordinate; and the effective_end (from right anchor) is not even used anywhere.

anthrotype: > these are not supported in glyphsLib

ok. The test is even marked xfail

https://github.com/googlefonts/glyphsLib/blob/f90e4060bafbb62ea76d269e77ef2823cb57190a/tests/corner_components_test.py#L16-L17

I'm ok to leave this for later, if it ever becomes necessary

glyphs-reader/src/corner_components.rs
cmyr: the python version of 'unbounded_seg_seg_intersection' seemed unnecessarily complicated, and this captures the same logic. In practice this is about finding the crossing point of segments that are quite close to each other, so 10k units seemed reasonable. Making this constant bigger doesn't have any downside, though, so I'm happy to do that..

glyphs-reader/src/corner_components.rs
cmyr: so good question: that had been my reading of https://github.com/googlefonts/glyphsLib/blob/f90e4060bafbb62ea76d269e77ef2823cb57190a/Lib/glyphsLib/filters/cornerComponents.py#L142-L148, but it looks like the python would actually crash in that case. In practice curved corner segments meeting curved glyph segments is uncommon, but it is supported in the glyphs sources.

jenskutilek: Is it actually uncommon? In a normal sans serif, all these glyphs would have curved corner components meeting curved glyph segments:
Bildschirmfoto 2025-11-28 um 09 11 33

(If all corners are rounded using a quarter-circle corner component or similar)

anthrotype: glyphsLib doesn't support them either. We have this issue tracking this

https://github.com/googlefonts/fontc/issues/1784

PR #1727, opened 2025-11-04
[crater] Fixup cache invalidation logic

cmyr:
Previously we would use the python dependencies (the output of `pip freeze`) as well as the presence of any changes to ttx_diff since the previously run commit to invalidate caches.

As ttx_diff is now a python package, and is installed from a local path, it ins included in the output of pip freeze with a sha appeneded, and this sha changes on each commit to the fontc repo.

The solution here is just to manually remove ttx_diff from the output of pip freeze, since changes are still tracked independently.

### DIFF
--- a/fontc_crater/src/main.rs
+++ b/fontc_crater/src/main.rs
@@ -3,6 +3,7 @@
 use std::{
     collections::BTreeMap,
     fmt::Display,
+    io::Write,
     path::Path,
     process::{Command, Stdio},
     sync::atomic::{AtomicUsize, Ordering},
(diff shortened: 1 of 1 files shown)


fontc_crater/src/main.rs
```diff
@@ -110,18 +111,40 @@ fn get_git_rev(repo_path: Option<&Path>) -> Option<String> {
...
+                stdin
+                    .write_all(line.as_bytes())
+                    .expect("failed to write to stdin");
+            }
+        }
+    });
+
+    let output = sha1sum.wait_with_output().expect("Failed to read stdout");
+    assert!(output.status.success());
+    eprintln!("hmm {}", std::str::from_utf8(&output.stderr).unwrap());
```
anthrotype: hmm

PR #1730, opened 2025-11-05
[fea-rs] Publicize .edit(), compute ranges

simoncozens:
As discussed with @cmyr yesterday - .edit() is a useful thing, but currently the ranges produced in the new tree are inaccurate and need to be recompute.

### DIFF
--- a/fea-rs/src/token_tree/edit.rs
+++ b/fea-rs/src/token_tree/edit.rs
@@ -17,7 +17,9 @@ pub(crate) fn apply_edits(
     let mut builder = TreeBuilder::default();
     let mut cursor = base.cursor();
     apply_edits_recurse(&mut cursor, &mut builder, &mut edits, skip_parent);
-    builder.finish()
+    let mut new_tree = builder.finish();
+    new_tree.update_positions_from_root();
+    new_tree
 }
 
 fn apply_edits_recurse(


fea-rs/src/token_tree/edit.rs
```diff
@@ -1,6 +1,4 @@
 //! facilities for editing the AST.
-//!
-//! This is currently unused.
 
 #![allow(dead_code)]
```
cmyr, line 3: could you also remove this `#![allow]` please?

fea-rs/src/token_tree.rs
```diff
@@ -309,10 +309,10 @@ impl Node {
     /// nodes.
     ///
     /// if skip_parent is true, children of inserted nodes are added directly,
-    /// without their parent nodes. (we use this when resolving includes
+    /// without their parent nodes. (we use this when resolving includes.)
     ///
-    /// range start/end just fall on token boundaries.
-    pub(crate) fn edit(&self, edits: Vec<(Range<usize>, Node)>, skip_parent: bool) -> Node {
+    /// Assumes that the range start/end just fall on token boundaries.
```
cmyr, line 314: I think that 'just' was a typo for 'must', but if this is public API now I would be a bit more formal:

```suggestion
    /// # Panics
    ///
    /// Panics if the provided range is not the range of an existing node or token.
```

Some things I noticed just now, reading through the old code:

- there's no concept of an insertion that does not replace _something_. (Of course you could fake this by prepending a copy of the node-to-be-replaced at the front of the replacement) This would be easy enough to add, it just wasn't needed originally.
- the edit cannot occur unless at a 'node boundary'. For instance for the statement `sub a by b;` we could replace the whole statement, or we could replace any individual token, but we could not replace `sub a` or `by b`.

PR #1731, opened 2025-11-05
[glyphs] Ensure brackets have axis rules for each axis

cmyr:
In glyphs2, a bracket layer can only reference a value on a single axis (from experimentation, it seems to only work with the first axis) and so in the case where multiple axes are defined, we need to add empty axis rules for the additional axes.

This fixes one crater diff.


dfrg (approved):
👍

PR #1733, opened 2025-11-06
[glyphs] Ensure non-bracket layers dont have axis rules

cmyr:
This showed up in one font (likely the result of glyphs.app leaving some stuff in place for a layer that has been set as an alternate and then later turned back to a master?) but it was later impacting our logic for propagating anchors.

Gives us a +small identical.


glyphs-reader/src/font.rs
```diff
@@ -4985,6 +4988,14 @@ etc;
         )
     }
 
+    #[test]
+    fn drop_axis_layers_if_no_assoc_master() {
```
```suggestion
    fn drop_axis_rules_if_no_assoc_master() {
```

anthrotype: the test name is still a bit cryptic..

PR #1735, opened 2025-11-10
Create gradients from .glyphs; fields not yet correct

rsheeter:


fontir/src/ir.rs
```diff
@@ -2030,6 +2032,23 @@ pub struct PaintSolid {
     pub color: Color,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct PaintLinearGradient {
+    pub color_line: Vec<ColorStop>,
```
anthrotype, line 2037: at some point we'll need a separate ColorLine struct, since a color line is not just a vector of color stops but may also define extend modes (pad, repeat, reflect)

rsheeter: Yes, but I'd like to punt that to future.

PR #1737, opened 2025-11-11
Parse fillColor, as seen in CoralPixels.glyphs

rsheeter:


glyphs-reader/src/font.rs
```diff
@@ -4861,18 +4894,22 @@ etc;
         let font = Font::load(&glyphs3_dir().join("COLRv1-grayscale.glyphs")).unwrap();
         assert_eq!(
             vec![
-                Color {
-                    r: 64,
-                    g: 64,
-                    b: 64,
-                    a: 255,
+                ColorStop {
```
cmyr, line 4878: Adding constructors would make this much less noisy. I'd probably go with `Color::rgba` over `Color::new` just for the self-documenting aspect, but 🤷

cmyr (approved):
I find the plist parsing stuff to be such a papercut here, but never enough to justify trying to improve it...

glyphs-reader/src/font.rs
rsheeter: > I find the plist parsing stuff to be such a papercut here

Same!

> Adding constructors would make this much less noisy

Good catch, making it so

PR #1741, opened 2025-11-11
Create base glyphs and clip list

rsheeter:
Always solid color. Baby steps :)

### DIFF
--- a/fontbe/src/colr.rs
+++ b/fontbe/src/colr.rs
@@ -1,11 +1,25 @@
 //! Generates a [COLR](https://learn.microsoft.com/en-us/typography/opentype/spec/colr) table.
 
 use crate::{
-    error::Error,
+    error::{Error, GlyphProblem},
     orchestration::{AnyWorkId, BeWork, Context, WorkId},
 };
-use fontdrasil::orchestration::{Access, Work};
-use fontir::orchestration::WorkId as FeWorkId;
+use fontdrasil::{
+    orchestration::{Access, AccessBuilder, Work},
+    types::GlyphName,
+};
+use fontir::{
+    ir::{self, ColorPalettes, GlyphOrder},
+    orchestration::WorkId as FeWorkId,
+};
+use write_fonts::{
+    tables::colr::{
+        BaseGlyphList, BaseGlyphPaint, Clip, ClipBox, ClipList, Colr, Paint, PaintGlyph, PaintSolid,
+    },
+    types::F2Dot14,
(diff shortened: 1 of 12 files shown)


fontbe/src/colr.rs
```diff
@@ -14,21 +26,87 @@ pub fn create_colr_work() -> Box<BeWork> {
...
+        )
+        .into();
+
+        let mut clips = Vec::<Clip>::new();
+        for glyph_name in paint_graph.base_glyphs.iter().map(|(g, _)| g) {
+            let next_gid = glyph_order.glyph_id(glyph_name).expect("Validated earlier");
+            let next_glyph = context
+                .glyphs
+                .get(&WorkId::GlyfFragment(glyph_name.clone()).into());
+            let next_clip = next_glyph.data.bbox().unwrap_or_default();
```
anthrotype, line 95: ufo2ft quantizes clipBoxes to 1/10th of the upem (e.g. 100 units for 1000 UPEM) to increase the chance of them getting reused. Search for "colrClipBoxQuantization" in ufo2ft.
Maybe for later

fontbe/src/colr.rs
```diff
@@ -14,21 +26,87 @@ pub fn create_colr_work() -> Box<BeWork> {
...
+fn to_colr_paint(glyph_order: &GlyphOrder, palette: &ColorPalettes, ir_paint: &ir::Paint) -> Paint {
+    match ir_paint {
+        ir::Paint::Glyph(paint) => Paint::Glyph(PaintGlyph {
+            paint: to_colr_paint(glyph_order, palette, &paint.paint).into(),
+            glyph_id: glyph_order
+                .glyph_id(&paint.name)
+                .expect("Validated earlier"),
+        }),
+        ir::Paint::Solid(paint) => Paint::Solid(PaintSolid {
+            palette_index: palette.index_of(paint.color).unwrap_or(FOREGROUND_COLOR) as u16,
```
anthrotype: `.unwrap_or(FOREGROUND_COLOR)` ...  I wonder shouldn't this be an error?

fontbe/src/colr.rs
rsheeter: I think I'll save that for now, but good to know about

fontbe/src/colr.rs
rsheeter: Hm, I think you might be right, its an error unless it turns out half our sources have bad palette indices

PR #1745, opened 2025-11-13
[ttx_diff] Add support for comparing precompiled fonts

anthrotype:
Adds --fontc_font and --fontmake_font flags to allow comparing precompiled fonts directly without building from source. This is useful for comparing font differences when you already have both builds or one was compiled using some special build setup.

The flags require both fonts to be provided together and make the source argument optional. Precompiled fonts are copied to the build directory as fontc.ttf/fontmake.ttf so the rest of the comparison logic works unchanged.

### DIFF
--- a/ttx_diff/src/ttx_diff/core.py
+++ b/ttx_diff/src/ttx_diff/core.py
@@ -1436,25 +1436,24 @@ def get_crate_path(
 
 
 def main(argv):
-    has_precompiled_fonts = (
-        FLAGS.fontc_font is not None or FLAGS.fontmake_font is not None
-    )
     has_source = len(argv) == 2
 
-    if has_precompiled_fonts:
-        if FLAGS.fontc_font is None or FLAGS.fontmake_font is None:
-            sys.exit(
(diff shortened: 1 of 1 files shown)


ttx_diff/src/ttx_diff/core.py
```diff
@@ -1417,10 +1436,25 @@ def get_crate_path(
...
+    has_precompiled_fonts = (
+        FLAGS.fontc_font is not None or FLAGS.fontmake_font is not None
+    )
+    has_source = len(argv) == 2
 
-    source = resolve_source(argv[1]).resolve()
+    if has_precompiled_fonts:
+        if FLAGS.fontc_font is None or FLAGS.fontmake_font is None:
+            sys.exit(
+                "When using precompiled fonts, both --fontc_font and --fontmake_font must be provided"
```
rsheeter: Hm, I wonder if it would be simpler to check if fontc_font is None == fontmake_font is None then proceed able to assume neither or both is set? (optional)

PR #1746, opened 2025-11-13
Produce color layers for split glyphs

rsheeter:
`python3 -m ttx_diff resources/testdata/glyphs3/COLRv1-manyshapes-per-glyph.glyphs` is now 82% matching for COLR, the remaining diffs are individual coords in clip boxes and gradients

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -1633,6 +1633,38 @@ struct ColorGlyphsWork {
     font_info: Arc<FontInfo>,
 }
 
+impl ColorGlyphsWork {
+    fn get_glyph(&self, glyph_name: &SmolStr) -> &glyphs_reader::Glyph {
+        self.font_info
+            .font
+            .glyphs
+            .get(glyph_name)
+            .unwrap_or_else(|| panic!("No glyph for name {glyph_name}"))
+    }
+}
+
+/// Creates a [PaintGlyph] for a split glyph, that is one that is known to have 1..N shapes with the same paint
+fn create_paint_glyph(
+    default_master_id: &str,
+    color_glyph: &glyphs_reader::Glyph,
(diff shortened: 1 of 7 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -1633,6 +1633,37 @@ struct ColorGlyphsWork {
     font_info: Arc<FontInfo>,
 }
 
+impl ColorGlyphsWork {
+    fn color_glyph(&self, glyph_name: &SmolStr) -> &glyphs_reader::Glyph {
```
anthrotype: there isn't anything specifically color-releated in this method, it's simply getting the named glyph

glyphs2fontir/src/source.rs
```diff
@@ -1633,6 +1633,37 @@ struct ColorGlyphsWork {
...
+    else {
+        return Err(Error::BadGlyph(BadGlyph::new(
+            color_glyph.name.clone(),
+            BadGlyphKind::MissingMaster(default_master_id.to_string()),
+        )));
+    };
+    Ok(PaintGlyph {
+        name: color_glyph.name.clone().into(),
+        paint: to_ir_paint(color_glyph.name.clone(), &default_layer.shapes[0])?,
+    })
```
anthrotype, line 1665: I presume it's just temporary that you are getting the PaintGlyph shape from the default layer, until proper variable support?

anthrotype (approved):

glyphs2fontir/src/source.rs
rsheeter: Yes. Will add a TODO.

glyphs2fontir/src/source.rs
rsheeter: Updated to get_glyph

PR #1748, opened 2025-11-13
Also build and release wheels for otl-normalizer

anthrotype:
using maturin, similarly to how we already do for fontc itself.

### DIFF
--- a/otl-normalizer/pyproject.toml
+++ b/otl-normalizer/pyproject.toml
@@ -23,6 +23,7 @@ exclude = [
 [project]
 # 'name' key is required for [project]
 name = "otl-normalizer"
+description = "Normalized text representation of OpenType layout subtables"
 # We want to use the same 'version' as Cargo.toml; only override if needed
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -3,54 +3,101 @@
...
+          if [[ -n "${{ inputs.tool }}" ]]; then
+            TOOL="${{ inputs.tool }}"
+          # Otherwise detect from tag
+          elif [[ "${{ github.ref }}" == refs/tags/fontc-v* ]]; then
+            TOOL="fontc"
+          elif [[ "${{ github.ref }}" == refs/tags/otl-normalizer-v* ]]; then
+            TOOL="otl-normalizer"
+          else
+            # Fallback (shouldn't happen)
+            TOOL="fontc"
```
rsheeter: maybe we should just fail?

anthrotype: Done

PR #1754, opened 2025-11-14
Build directly declared (typically for COLRv0) palettes

rsheeter:
`CPAL` now identical in `python3 -m ttx_diff 'https://github.com/fridamedrano/Kalnia-Glaze?3338853e09#sources/KalniaGlaze.glyphs'` (alas only fontmake produced COLR tho, coming soon!)

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -17,7 +17,7 @@ use fontir::{
     error::{BadGlyph, BadGlyphKind, BadSource, Error},
     feature_variations::{NBox, overlay_feature_variations},
     ir::{
-        self, AnchorBuilder, Color, ColorGlyphs, ColorPalettes, Condition, ConditionSet,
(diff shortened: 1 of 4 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -1603,28 +1604,52 @@ impl Work<Context, WorkId, Error> for ColorPaletteWork {
...
+        } else {
+            Vec::new()
+        };
+
+        if palettes.is_empty() {
+            palettes.push(Vec::new());
+        }
+
+        // Plus any colors declared directly on glyphs
+        palettes[0].extend(
```
anthrotype: this is only modifying the first, adding to all would make them same length at least

anthrotype (approved):

glyphs2fontir/src/source.rs
rsheeter: Done

PR #1756, opened 2025-11-14
test colr gradients' scaled geometry and actual clip boxes in integration tests

anthrotype:
follow up from #1755

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -3855,12 +3855,12 @@ mod tests {
             panic!("Expected LinearGradient");
         };
 
-        assert_eq!(grad.x0().to_i16(), 111, "p0.x");
-        assert_eq!(grad.y0().to_i16(), 50, "p0.y");
-        assert_eq!(grad.x1().to_i16(), 494, "p1.x");
-        assert_eq!(grad.y1().to_i16(), 450, "p1.y");
-        assert_eq!(grad.x2().to_i16(), 511, "p2.x");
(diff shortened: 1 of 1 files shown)


fontc/src/lib.rs
```diff
@@ -3841,20 +3841,127 @@ mod tests {
...
+            panic!("Expected LinearGradient");
+        };
+
+        assert_eq!(grad.x0().to_i16(), 111, "p0.x");
+        assert_eq!(grad.y0().to_i16(), 50, "p0.y");
+        assert_eq!(grad.x1().to_i16(), 494, "p1.x");
+        assert_eq!(grad.y1().to_i16(), 450, "p1.y");
+        assert_eq!(grad.x2().to_i16(), 511, "p2.x");
+        assert_eq!(grad.y2().to_i16(), -333, "p2.y");
     }
```
rsheeter, line 3864: Nit: assert all the values at once, e.g. assert_eq!((1,2,3,4), (grad.x0, grad.x1, etc)). It sucks to only be informed of one value that isn't right out of many.

rsheeter: Here and below.

rsheeter (approved):

fontc/src/lib.rs
anthrotype: Done https://github.com/googlefonts/fontc/pull/1756/commits/3b3921f802f43507710bb6d60c3c1b0b61c0911b

PR #1757, opened 2025-11-14
Rough in static COLRv0

rsheeter:
`python3 -m ttx_diff 'https://github.com/fridamedrano/Kalnia-Glaze?3338853e09#sources/KalniaGlaze.glyphs'` now has matching CPAL and COLR instead of only fontmake produced for both.

### DIFF
--- a/fontbe/src/colr.rs
+++ b/fontbe/src/colr.rs
@@ -2,7 +2,7 @@
 
 use crate::{
     error::{Error, GlyphProblem},
-    orchestration::{AnyWorkId, BeWork, Context, WorkId},
+    orchestration::{AnyWorkId, BeWork, Context, Glyph, WorkId},
 };
 use fontdrasil::{
     orchestration::{Access, AccessBuilder, Work},
@@ -15,13 +15,13 @@ use fontir::{
 use write_fonts::{
     tables::{
         colr::{
-            BaseGlyphList, BaseGlyphPaint, Clip, ClipBox, ClipList, ColorLine, ColorStop, Colr,
-            Extend, LayerList, Paint, PaintColrLayers, PaintGlyph, PaintLinearGradient,
-            PaintRadialGradient, PaintSolid,
+            BaseGlyph, BaseGlyphList, BaseGlyphPaint, Clip, ClipBox, ClipList, ColorLine,
+            ColorStop, Colr, Extend, Layer, LayerList, Paint, PaintColrLayers, PaintGlyph,
+            PaintLinearGradient, PaintRadialGradient, PaintSolid,
         },
         glyf::Bbox,
     },
-    types::F2Dot14,
+    types::{F2Dot14, GlyphId16},
 };
 
 static OPAQUE: F2Dot14 = F2Dot14::ONE;
@@ -240,6 +240,89 @@ fn to_colr_paint(
     }
 }
 
+fn add_or_extend_clip(clips: &mut Vec<Clip>, quantization: i16, gid: GlyphId16, glyph: &Glyph) {
+    let bbox = glyph.data.bbox().unwrap_or_default();
+    // Quantize the bbox to maximize clipbox reuse
+    let next_clip = quantize_bbox(&bbox, quantization);
+    let next_clip = ClipBox::format_1(
+        next_clip.x_min.into(),
+        next_clip.y_min.into(),
+        next_clip.x_max.into(),
+        next_clip.y_max.into(),
+    );
+    if let Some(curr) = clips.last_mut()
+        && curr.end_glyph_id.to_u32() + 1 == gid.to_u32()
+        && *curr.clip_box == next_clip
+    {
+        // wow wow wow, a run!
+        curr.end_glyph_id = gid;
+        return; // DONE
+    }
+
+    // Evidently we didn't make a run
+    clips.push(Clip::new(gid, gid, next_clip));
+}
+
+fn is_paint_glyph_solid(paint: &ir::Paint) -> bool {
+    if let ir::Paint::Glyph(paint_glyph) = paint
+        && let ir::Paint::Solid(_) = &paint_glyph.paint
+    {
+        true
+    } else {
+        false
+    }
+}
+
+fn colr_v0_compatible(paint: &ir::Paint) -> bool {
+    if is_paint_glyph_solid(paint) {
+        return true;
+    }
+    if let ir::Paint::Layers(layers) = paint
+        && layers.iter().all(is_paint_glyph_solid)
+    {
+        return true;
+    }
+    false
+}
+
+/// Invariants:
+///
+/// * All glyph names are valid
+/// * All paint are PaintGlyph whose nested paint is PaintSolid (e.g. are COLRv0 compatible)
+fn new_colr0(
+    palette: &ColorPalettes,
+    glyph_order: &GlyphOrder,
+    glyph_name: &GlyphName,
+    paint: &[&ir::PaintGlyph],
+    layers: &mut Vec<Layer>,
+) -> Result<BaseGlyph, Error> {
+    let gid = glyph_order.glyph_id(glyph_name).expect("Prevalidated");
+    let candidate_layers = paint
+        .iter()
+        .map(|p| {
+            let ir::Paint::Solid(solid) = &p.paint else {
+                unreachable!("We prevalidated, what is {p:?}?!");
+            };
+            let palette_idx = palette.index_of(solid.color).expect("Prevalidated");
+            Layer::new(
+                glyph_order.glyph_id(&p.name).expect("Prevalidated"),
+                palette_idx as u16,
+            )
+        })
+        .collect::<Vec<_>>();
+    let first_idx = if let Some(pos) = layers
+        .windows(candidate_layers.len())
+        .position(|window| candidate_layers.as_slice() == window)
+    {
+        pos
+    } else {
+        let pos = layers.len();
+        layers.extend(candidate_layers);
+        pos
+    };
+    Ok(BaseGlyph::new(gid, first_idx as u16, paint.len() as u16))
+}
+
 impl Work<Context, AnyWorkId, Error> for ColrWork {
     fn id(&self) -> AnyWorkId {
         WorkId::Colr.into()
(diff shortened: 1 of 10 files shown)


glyphs2fontir/src/toir.rs
```diff
@@ -346,51 +346,180 @@ impl TryFrom<Font> for FontInfo {
...
         !matches!(self, Self::NonColor)
     }
 }
 
+fn new_color_glyph(original: &Glyph, nth: &mut usize) -> Glyph {
+    let new_glyph_name: SmolStr = format!("{}.color{nth}", original.name).into();
+    let new_glyph = Glyph {
+        name: new_glyph_name.clone(),
+        export: original.export,
+        ..Default::default()
```
anthrotype, line 400: here I think we need to also copy the original glyph's `production_name` if it has one, and append the .color suffix to that, so that it gets the expected final postscript name..
Unless it is too late here and the StaticMetadata.postcriptNames has been already populated with the original glyphs only 🤔

anthrotype: like this:

```diff
diff --git a/glyphs2fontir/src/toir.rs b/glyphs2fontir/src/toir.rs
index 38c32b39..de3ec293 100644
--- a/glyphs2fontir/src/toir.rs
+++ b/glyphs2fontir/src/toir.rs
@@ -380,8 +380,13 @@ impl Colrv1Group {
 
 fn new_color_glyph(original: &Glyph, nth: &mut usize) -> Glyph {
     let new_glyph_name: SmolStr = format!("{}.color{nth}", original.name).into();
+    let new_production_name = original
+        .production_name
+        .as_ref()
+        .map(|production_name| format!("{}.color{nth}", production_name).into());
     let new_glyph = Glyph {
         name: new_glyph_name.clone(),
+        production_name: new_production_name,
         export: original.export,
         ..Default::default()
     };
```

this should be the correct way... but then I see we still get other diffs in the post names. I believe it's actually a bug in glyphsLib, which is not keeping the .colorN suffix for the color layer glyphs it's creating, but assigining (some of them, not all?!) the same postscript name as their master glyph. I'll investigate

anthrotype: yes, it was a bug in glyphsLib. After applying https://github.com/googlefonts/glyphsLib/pull/1117 locally, I can confirm with your PR (_and_ my own patch in the comment above), I get identical COLR table for KalniaGlaze.glyphs. Good job!

anthrotype: applied my own patch [4c23e72](https://github.com/googlefonts/fontc/commit/4c23e728185c2fd2654dd7999176bbd0c6573ddd)

anthrotype:
`colr_cliprun` test fails because uses "COLRv1-solid.glyphs" which contains simple solid color glyphs. These are now classified as COLRv0-compatible (via colr_v0_compatible()) and take the COLRv0 path, so clips are never added. The test should use a font that requires COLRv1 features.

`cpal_from_declared_palettes` test fails with "No such file or directory", it's looking for COLRv0-simple.glyphs. Was it renamed to COLRv0-1layer.glyphs?

glyphs2fontir/src/toir.rs
```diff
@@ -346,51 +346,185 @@ impl TryFrom<Font> for FontInfo {
...
+            let mut new_layer = old_layer.clone();
+            new_layer.attributes.color = paint_group.color();
+            if let Some(split_at) = old_layer
+                .shapes
+                .iter()
+                .position(|s| Colrv1Group::key_for(&new_layer, s) != paint_group)
+            {
+                // Only some of the shapes fit the current run take them
+                // Amusingly this results in the opposite arrangement than we want so swap
+                new_layer.shapes = old_layer.shapes[0..split_at].to_vec();
```
anthrotype: I think this change is what's causing the failures in fontc/lib.rs `tests::colr_split_*` with 

```
BadGlyph { name: A.color1, kind: FrontendSpecific("All layers are empty?!") }
```

previously we were doing:

```rust
                    // Only some of the shapes fit the current run take them
                    // Amusingly this results in the opposite arrangement than we want so swap
                    new_layer.shapes = old_layer.shapes.split_off(split_at);
                    std::mem::swap(&mut new_layer.shapes, &mut old_layer.shapes);
```

The old code mutated old_layer.shapes by consuming shapes as it processed each paint group. The new code copies shapes without mutation, so:

- First paint group (A.color0) takes all matching shapes
- Second paint group (A.color1) tries to process the same old_layer.shapes again, but those shapes don't match the second paint group anymore, resulting in split_at = 0 and empty layers, which triggers the error above.

I believe you should restore the old code.

glyphs2fontir/src/toir.rs
rsheeter: Ty!

glyphs2fontir/src/toir.rs
rsheeter: Hm. Why am I not just remembering what slice to take when I bulid the runs to begin with... I'm a bit meeting slammed today but wtal when I have time.

rsheeter: I tried building the runs up front and prefer it, uploaded that

rsheeter: Resolving as I thik the new run structure does this better than both old split off and doing it completely wrong :)

glyphs2fontir/src/toir.rs
```diff
@@ -383,14 +529,8 @@ fn split_color_glyphs(font: Font) -> Result<(Font, IndexMap<SmolStr, Vec<SmolStr
...
-    let mut additions = Vec::new();
-    for (glyph_name, glyph) in font.glyphs.iter_mut() {
-        // Normal path: nop, not a color glyph
-        if glyph.layers.iter().all(|l| !l.attributes.color) {
-            continue;
-        }
-        // Remember the name!
-        color_glyphs.entry(glyph_name.clone()).or_default();
+    let mut additions: Vec<(SmolStr, Glyph)> = Vec::new();
+    for glyph in font.glyphs.values_mut() {
```
anthrotype, line 533: font.glyphs is a `BTreeMap` (sorted alphabetically by glyph name), we are iterating on it here and then below in `split_colrv0_glyph` and `split_colrv1_glyph` we are populating `&mut color_glyphs` with the new layer .colorN  glyphs, but this color_glyphs is an IndexMap (preserving _insertion_ order, so again alphabetical).

... but then fontbe/colr.rs is iterating over this paint_graph.base_glyphs IndexMap to build things like the BaseGlyphList and the ClipList which the spec says should be sorted by glyph_id (to allow binary search).

I didn't have time to confirm this suspicion with an actual test, but it's worth double checking that we aren't building COLR tables with invalid orderings.

rsheeter: Based on ttx_diff being identical I think we are ok on ordering but I do agree it might be more accidental than on purpose. Will check.

anthrotype: you could make a .glyphs file with two glyphs "A" and "B" and a glyphOrder custom parameter that makes them order "B" and "A" and assert that the COLR base glyph's glyph_id we produce is [1, 2] and not [2, 1]... if it doesn't make sense, it's because kids

rsheeter: Confirmed broken, fixing. Good catch.

fontbe/src/colr.rs
```diff
@@ -240,6 +240,85 @@ fn to_colr_paint(
...
+    glyph_name: &GlyphName,
+    paint: &[&ir::PaintGlyph],
+    layers: &mut Vec<Layer>,
+) -> Result<BaseGlyph, Error> {
+    let gid = glyph_order.glyph_id(glyph_name).expect("Prevalidated");
+    let candidate_layers = paint
+        .iter()
+        .map(|p| {
+            let ir::Paint::Solid(solid) = &p.paint else {
+                panic!("Prevalidated");
```
cmyr: for code that should be genuinely unreachable `unreachable!()`, which is equivalent to `panic!` but better communicates your intention. (there's also the unsafe `unreachable_unchecked`, for when you're actually telling the compiler that a given branch can be removed.

If we _do_ ever think it's possible to hit this we might want to include the gid or glyph name in the message?

fontbe/src/colr.rs
```diff
@@ -240,6 +240,85 @@ fn to_colr_paint(
...
+    }
+    if let ir::Paint::Layers(layers) = paint
+        && layers.iter().all(is_paint_glyph_solid)
+    {
+        return true;
+    }
+    false
+}
+
+fn new_colr0(
```
cmyr, line 292: all of the 'expects' in this fn imply that it is only called under specific conditions, would it be worth including a comment that states this explicitly?

fontbe/src/colr.rs
```diff
@@ -240,6 +240,85 @@ fn to_colr_paint(
...
+) -> Result<BaseGlyph, Error> {
+    let gid = glyph_order.glyph_id(glyph_name).expect("Prevalidated");
+    let candidate_layers = paint
+        .iter()
+        .map(|p| {
+            let ir::Paint::Solid(solid) = &p.paint else {
+                panic!("Prevalidated");
+            };
+            let palette_idx = palette.index_of(solid.color).expect("Prevalidated");
+            Layer {
```
cmyr: just as a general point, write-fonts always generates `new` constructor methods for its types, which can be a bit less verbose than using the struct initializer syntax directly (not much of an advantage in this case)

fontbe/src/colr.rs
```diff
@@ -264,67 +343,103 @@ impl Work<Context, AnyWorkId, Error> for ColrWork {
...
+            } else {
+                None
+            },
+            if !colr_v0_layers.is_empty() {
+                Some(colr_v0_layers)
+            } else {
+                None
+            },
+            num_colr_v0_layers,
+        );
```
cmyr, line 430: obv just a personal preference, but I find code harder to read when there is logic inside of a method call. I'd personally just write this as,
```suggestion
        let num_colr_v0_layers = colr_v0_layers.len() as u16;
        let base_glyph_records = (!colr_v0_glyphs.is_empty()).then_some(colr_v0_glyphs);
        let layer_records = (!colr_v0_layers.is_empty()).then_some(colr_v0_layers);
        let mut colr = Colr::new(
            num_v0_glyphs,
            base_glyph_records,
            layer_records,
            num_colr_v0_layers,
        );
```

(also it looks like we could improve codegen here, we shouldn't need to manually provide the lengths of the records/layers, I don't think?)

fontc/src/lib.rs
```diff
@@ -4067,6 +4121,60 @@ mod tests {
         );
     }
 
+    fn assert_colr0(compile: &TestCompile, expected: Vec<(String, Vec<(String, u16)>)>) {
```
cmyr, line 4124: looking at where this is called, maybe `&str` would be better than `String`?

glyphs2fontir/src/toir.rs
```diff
@@ -346,35 +346,181 @@ impl TryFrom<Font> for FontInfo {
...
+    // The original glyph becomes uncolored,
+    // each color run becomes a new glyph named [original].color[i]
+    let mut nth = 0;
+    for layer in original.layers.iter() {
+        if layer.shapes.is_empty() || layer.attributes.color_palette.is_none() {
+            continue;
+        }
+        if layer.associated_master_id.as_deref() != Some(default_master_layer.layer_id.as_str()) {
+            continue;
+        }
```
cmyr: I'd probably collapse these into one big check?

fontbe/src/colr.rs
rsheeter: So updated

fontbe/src/colr.rs
rsheeter: added

fontbe/src/colr.rs
rsheeter: It's at least a little nicer, so updated

fontbe/src/colr.rs
rsheeter: I agree that's nicer, made it so.

fontc/src/lib.rs
rsheeter: I believe that created ref to temp issues.

glyphs2fontir/src/toir.rs
rsheeter: Hm, agreed.

PR #1760, opened 2025-11-18
But what if a master had a palette?!

rsheeter:
Coral Pixels evidently does this.

`python3 -m ttx_diff resources/testdata/glyphs3/COLRv0-palette-redefinition.glyphs` is now identical

`python3 -m ttx_diff 'https://github.com/tanukifont/Coral-Pixels?2817da74e3#sources/CoralPixels.glyphs'` is now identical for CPAL, 99.984% for COLR. Amusingly the glyph order is quite different, in a way that makes me think fontmake sorts (left fontc, right fontmake):

image

Builds on #1757

### DIFF
--- a/fontbe/src/colr.rs
+++ b/fontbe/src/colr.rs
@@ -293,11 +293,11 @@ fn new_colr0(
     palette: &ColorPalettes,
     glyph_order: &GlyphOrder,
     glyph_name: &GlyphName,
-    paint: &[&ir::PaintGlyph],
+    paints: &[&ir::PaintGlyph],
     layers: &mut Vec<Layer>,
 ) -> Result<BaseGlyph, Error> {
     let gid = glyph_order.glyph_id(glyph_name).expect("Prevalidated");
-    let candidate_layers = paint
+    let candidate_layers = paints
         .iter()
         .map(|p| {
             let ir::Paint::Solid(solid) = &p.paint else {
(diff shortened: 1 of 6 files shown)


fontbe/src/colr.rs
```diff
@@ -289,23 +289,23 @@ fn new_colr0(
...
-        })
-        .collect::<Vec<_>>();
+    let mut candidate_layers = Vec::with_capacity(paints.len());
+    for paint in paints {
+        let ir::Paint::Solid(solid) = &paint.paint else {
+            panic!("{:?} should have been pre-validated!", paint.paint);
+        };
+        let Some(palette_idx) = palette.index_of(solid.color) else {
+            return Err(Error::MissingPaletteEntry(solid.color));
+        };
```
cmyr: instead of an explicit return I generally would write this as,


```suggestion
        let palette_idx = palette.index_of(solid.color).ok_or(Error::MissingPaletteEntry(solid.color))?;
```

fontc/src/lib.rs
```diff
@@ -3694,6 +3694,50 @@ mod tests {
...
+    #[test]
+    fn cpal_prefer_master() {
+        let result = TestCompile::compile_source("glyphs3/COLRv0-palette-redefinition.glyphs");
+        let cpal = result.font().cpal().unwrap();
+        assert_eq!(
+            (
+                2,
+                2,
+                [
+                    ColorRecord {
```
cmyr, line 3706: there is a generated constructor for this, which takes `bgra`; so you could write,

`ColorRecord::new(22, 21, 20, 23)` instead.

I do preemptively agree that this auto-generated constructor could have a name that made the argument order clear, at least.

rsheeter: bgra is ... special. argb or rgba would be far more typical. I suppose that would require a codegen hack to produce, for now I do agree ColorRecord::ctor is nicer ... unless that means bgra.

fontbe/src/colr.rs
rsheeter: Ha, agreed. Not sure why I wrote it that way tbh.

PR #1763, opened 2025-11-18
Compile MinMax BASE table statements in fea

### LINKED ISSUE #1676: Compile `MinMax` statements in `BASE`
See https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#9a-base-table

Support for parsing and compiling these statements has been added to fonttools since fea-rs began development (fonttools/fonttools#3783), and similar support would be good to have in fontc too in order to facilitate the FEA output of [autobase](https://github.com/simoncozens/autobase), and to keep feature parity.

cmyr:
closes #1676

### DIFF
--- a/fea-rs/src/compile/tables/base.rs
+++ b/fea-rs/src/compile/tables/base.rs
@@ -113,6 +113,7 @@ impl BaseAxisBuilder {
                     rest.push(BaseLangSysRecord::new(*tag, table.compile()));
                 }
             }
+            rest.sort_unstable_by_key(|rec| rec.base_lang_sys_tag);
 
             (dflt, rest)
         }
--- a/fea-rs/test-data/fonttools-tests/spec9a2.fea
+++ b/fea-rs/test-data/fonttools-tests/spec9a2.fea
@@ -3,5 +3,6 @@ table BASE {
    HorizAxis.BaseScriptList latn romn -120 0, cyrl romn -120 0, grek romn -120 0, hani ideo -120 0, kana ideo -120 0, hang ideo -120 0;
    HorizAxis.MinMax cyrl dflt -800, 1200;
    # Some comment
+   HorizAxis.MinMax cyrl SRB  -999, 999;
    HorizAxis.MinMax cyrl RUS  -1000, 1000;
 } BASE;
--- a/fea-rs/test-data/fonttools-tests/spec9a2.ttx
+++ b/fea-rs/test-data/fonttools-tests/spec9a2.ttx
@@ -33,7 +33,7 @@
               </MaxCoord>
               <!-- FeatMinMaxCount=0 -->
             </DefaultMinMax>
-            <!-- BaseLangSysCount=1 -->
+            <!-- BaseLangSysCount=2 -->
             <BaseLangSysRecord index="0">
               <BaseLangSysTag value="RUS "/>
               <MinMax>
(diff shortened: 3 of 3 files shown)


fea-rs/src/token_tree/typed.rs
```diff
@@ -1257,6 +1270,32 @@ impl ScriptRecord {
     }
 }
 
+impl BaseMinMax {
+    pub(crate) fn is_horiz(&self) -> bool {
+        match self.iter().next().map(|t| t.kind()) {
+            Some(Kind::HorizAxisMinMaxKw) => true,
+            Some(Kind::VertAxisMinMaxKw) => false,
+            other => panic!("unexpected token in BaseTagList {other:?}"),
```
anthrotype: copy/pasta? the panic message mentions BaseTagList but we're in BaseMinMax

fea-rs/src/compile/compile_ctx.rs
```diff
@@ -1604,6 +1604,15 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
             base.horiz_script_list
                 .sort_unstable_by_key(|rec| rec.script);
         }
+        for minmax in table.iter_horiz_min_max() {
+            let script = minmax.script();
+            let lang = minmax.language();
+            let (min, max) = minmax.minmax();
+            base.horiz_min_max
+                .entry(script.to_raw())
+                .or_insert(Vec::new())
```
anthrotype:
```suggestion
                .or_default()
```

fea-rs/src/compile/validate.rs
```diff
@@ -301,8 +301,60 @@ impl<'a, V: VariationInfo> ValidationCtx<'a, V> {
...
+                node.iter_vert_min_max(),
+            );
+        }
+    }
+
+    fn validate_base_axis(
+        &mut self,
+        taglist: &typed::BaseTagList,
+        script_list: Option<typed::BaseScriptList>,
+        _minmax: impl Iterator<Item = typed::BaseMinMax>,
```
anthrotype, line 325: this is currently unused. potentially this method could check e.g. a script listed in MinMax exists in the script list, that the min/max values fit the signed i16 range, or duplicate entries etc.

fea-rs/src/compile/tables/base.rs
```diff
@@ -17,24 +53,73 @@ pub(crate) struct ScriptRecord {
...
+
+            for (tag, table) in minmax.into_iter().flatten() {
+                if *tag == Tag::new(b"dflt") {
+                    dflt = Some(table.compile());
+                } else {
+                    rest.push(BaseLangSysRecord::new(*tag, table.compile()));
+                }
+            }
+
+            (dflt, rest)
```
anthrotype, line 118: BaseLangSysRecords should be sorted "in alphabetical order by BaseLangSysTag"

https://learn.microsoft.com/en-us/typography/opentype/spec/base

I think we aren't sorting them here (and write-fonts isn't either?) so if fonts have multiple language specific minmax entries for the same scripts we'd keep them in insertion order rather than alphabetical

cmyr: good catch!

fea-rs/src/compile/validate.rs
cmyr: thought about this a bit... we don't have access to the root script list here, so we can't check that (at least here) and the other bits of validation feel unnecessary, and likely cause us to deviate from fontmake in any case.

The stuff that it's important to validate here is stuff that we will later want to assume is true. (For instance, we will later want to find the index of a given baseline tag for items in the script list, so we need to ensure that those tags exist).

PR #1765, opened 2025-11-19
Fix panic with empty color glyphs

anthrotype:
This test reproduces a panic when compiling 

python3 -m ttx_diff 'https://github.com/EkType/Honk?99094d4f6e#sources/HonkExportFile.glyphs'

whereby an empty glyph 'CR' has palette layers that trigger the COLRv0 code path, however those layers have no shapes.

These empty glyphs are being unconditionally added to color_glyphs despite actually having no shapes.

### DIFF
--- a/glyphs2fontir/src/toir.rs
+++ b/glyphs2fontir/src/toir.rs
@@ -563,8 +563,12 @@ fn split_color_glyphs(font: Font) -> Result<(Font, IndexMap<SmolStr, Vec<SmolStr
             continue;
         }
 
-        // Remember the name so it gets added to COLR
-        color_glyphs.entry(glyph.name.clone()).or_default();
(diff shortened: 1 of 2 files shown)


glyphs2fontir/src/toir.rs
```diff
@@ -799,4 +799,47 @@ mod tests {
...
+                    anchors: vec![],
+                    attributes: LayerAttributes {
+                        color: true, // This triggers COLRv1 path
+                        ..Default::default()
+                    },
+                    ..Default::default()
+                },
+            ],
+            ..Default::default()
+        };
```
rsheeter, line 835: Optional: would it be easier to just add a new source?

anthrotype: maybe.. but I have already done it

PR #1775, opened 2025-11-21
Remove Persistent/PersistableStorage

cmyr:
This is a relic from back when we were thinking about incremental compilation, and has been causing me headaches recently. I was curious how hard it would be to remove and it turns out to not be that hard?

The specific motivation here is that I have twice this week wanted to move `GlyphOrder` into `fontdrasil` but have been prevented because of the implementation of `Persistable` for that type; the trait is defined in `fontir`, which also has a blanket impl for write-fonts types, and that means I can't have another custom impl for a foreign type, and the solution would require either moving the whole trait into `fontdrasil`, having some ugly wrapper type, or having fontir be a dep of fontdrasil, which feels backwards.

In any case, this was actually quite clean, and might unlock additional cleanups.


rsheeter (approved):
LGTM subject to confirmation --emit-ir still works, that is a useful capability. If it breaks it ... would have to think just a little about whether thats desirable.

cmyr:
Okay I'm resurrecting this PR, and believe it's time has come.

- this is now a more thorough root-and-branch removal of all the persistence related code, including all of the various hacks we had implemented to get persistence working (the `Bytes` type that just wrapped `Vec<u8>`, the `LocaFormatWrapper` type, etc.)
- this removes the --emit-ir flag, and removes support for dumping IR. If there are particular bits of state that are useful for debugging, we can add support for those to the --emit-debug flag, on a case-by-case basis, motivated by our actual needs; this feels better than me doing a bunch of additional work to support dumping IR that may never be looked at.
- In addition to removing 1k LOC, this has a few other measurable benefits: it shaved 8% (~2MB) off our binary size and ~6%
(2s) off of a clean compile on my m4 macbook pro.

rsheeter:
Land it!

PR #1778, opened 2025-11-25
Add support for "Link Metrics..." custom parameters

### LINKED ISSUE #1716: Support "Link Metrics ..." custom parameters in Glyphs format
The [custom parameter "Link Metrics With Master", and its syntax-sugar sibling "Link Metrics With First Master"](https://handbook.glyphsapp.com/custom-parameter-descriptions/), allows the glyph widths and kerning of one Glyphs master to be overwritten with those of another. This is typically used to enforce identical metrics across duplex axes such as Grade.

This is implemented in glyphsLib through the [`metricsSource`](https://github.com/googlefonts/glyphsLib/blob/f90e4060bafbb62ea76d269e77ef2823cb57190a/Lib/glyphsLib/classes.py#L1640-L1665) abstraction, which is then referenced when building [widths](https://github.com/googlefonts/glyphsLib/blob/f90e4060bafbb62ea76d269e77ef2823cb57190a/Lib/glyphsLib/builder/glyph.py#L256-L273) and [kerning](https://github.com/googlefonts/glyphsLib/blob/f90e4060bafbb62ea76d269e77ef2823cb57190a/Lib/glyphsLib/builder/kerning.py#L31-L35) respectively to make the metrics identical by construction. This matches the behaviour in Glyphs.app, and its compiler, too.

anthrotype:
Fixes #1716 

the implementation is intended to match fontmake's (inline comments point to the relevant lines in glyphsLib).

I still haven't had the chance to test this with GSF but will do tomorrow

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -5914,6 +5914,27 @@ name = _corner.hi;
     }
 
     #[test]
+    fn metrics_source_id_with_master_by_name() {
+        // Glyphs 2 format uses master names instead of IDs for "Link Metrics With Master"
+        let font = Font::load(&glyphs2_dir().join("LinkMetricsWithMasterByName.glyphs")).unwrap();
+
+        assert_eq!(font.masters[0].metrics_source_id, None);
+        assert_eq!(font.masters[1].metrics_source_id, None);
+        // Third master (Black) links by name "Bold", should resolve to second master's ID
+        assert_eq!(
+            font.masters[2]
+                .custom_parameters
+                .link_metrics_with_master
+                .as_deref(),
+            Some("Bold")
+        );
+        assert_eq!(
+            font.masters[2].metrics_source_id.as_ref(),
+            Some(&font.masters[1].id)
+        );
+    }
+
+    #[test]
     fn metrics_source_id_with_missing_master() {
         let font = Font::load(&glyphs3_dir().join("LinkMetricsWithMissingMaster.glyphs")).unwrap();
 
--- a/resources/testdata/glyphs2/LinkMetricsWithMasterByName.glyphs
+++ b/resources/testdata/glyphs2/LinkMetricsWithMasterByName.glyphs
@@ -1,0 +1,85 @@
+{
+.appVersion = "3506";
+customParameters = (
+{
+name = Axes;
+value = (
+{
+Name = Weight;
+Tag = wght;
+}
+);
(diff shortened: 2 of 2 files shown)


glyphs-reader/src/font.rs
```diff
@@ -1449,6 +1458,34 @@ impl FontMaster {
...
     }
+
+    /// Returns the master ID to use for glyph metrics and kerning, if this master
+    /// has linked metrics via the "Link Metrics With First Master" or
+    /// "Link Metrics With Master" custom parameters.
+    ///
+    /// Returns None if no linking is configured, meaning this master uses its own metrics.
+    ///
+    /// See <https://github.com/googlefonts/glyphsLib/blob/a045b48/Lib/glyphsLib/classes.py#L1641-L1665>
+    pub fn metrics_source_id<'a>(&'a self, font: &'a Font) -> Option<&'a str> {
```
cmyr: Thought: instead of having this kind of annoying method call, could we just compute this when we convert from `RawFont` to `Font`, and store it in the master directly?

glyphs2fontir/src/source.rs
```diff
@@ -1796,6 +1827,16 @@ mod tests {
 
     use super::*;
 
+    // TODO: Consider moving to fontdrasil if this becomes useful elsewhere
+    /// Create a [`NormalizedLocation`] with a single tag/coord pair.
+    macro_rules! norm_loc {
```
cmyr: There's a hidden method on `Location` intended for use in tests, `Location::for_pos`. It lets you do, `NormalizedLocation::for_pos(&[("wdth", 0.5), ("wght", 1.0)])`. Maybe that would suit your needs?

More generally it isn't clear that a macro makes sense here; this isn't doing anything that couldn't be achieved with a method. (except for maybe existing outside of the crate defining the type?)

cmyr (approved):
A couple little notes inline, but looks good!

glyphs2fontir/src/source.rs
anthrotype: oh I had forgotten about that, thanks! I'll use that

glyphs-reader/src/font.rs
anthrotype: is that what you had in mind? https://github.com/googlefonts/fontc/pull/1778/commits/209b58c728088cdd282430f9e22cd9c241e4af9d

PR #1789, opened 2025-12-02
Allow building without rayon

simoncozens:
Currently we can build fontc on WASM, but it's a miserable experience because you have to build it in certain ways, with atomics and bulk memory, and on certain compilers, and it only works on certain browsers that support multi-threaded WASM. (Chrome does, Safari doesn't.)

This PR replaces the current single-threaded implementation with a sequential scope. This allows building without rayon altogether, which is a much cleaner solution for WASM builds, and of course for those times when you want to compile fonts on ASICs and embedded hardware.

### DIFF
--- a/fontc/src/workload.rs
+++ b/fontc/src/workload.rs
@@ -578,6 +578,7 @@ impl Workload {
             Vec<Arc<AtomicUsize>>,
         )>::with_capacity(512)));
 
+        // In non-rayon mode, calling the runner needs a mutable borrow for update_launchable.
         #[cfg_attr(feature = "rayon", allow(unused_mut))]
         let mut runner = |scope: &Scope| {
             // Whenever a task completes see if it was the last incomplete dependency of other task(s)


fontc/src/workload.rs
```diff
@@ -573,7 +578,8 @@ impl Workload {
             Vec<Arc<AtomicUsize>>,
         )>::with_capacity(512)));
 
-        let runner = |scope: &rayon::Scope| {
+        #[cfg_attr(feature = "rayon", allow(unused_mut))]
```
anthrotype, line 582: maybe add a comment why allow(unused_mut) is needed (for the sequential path I presume)

anthrotype (approved):
It's worth mentioning the existing test infrastructure uses [`run_for_test`](https://github.com/googlefonts/fontc/blob/45890d9e2cc10e7f715689e7413686e8624739a0/fontc/src/workload.rs#L865) which doesn't use the Scope mechanism at all but executes jobs directly and synchronously. So tests pass regardless of rayon feature. Though I realise this predates this PR, maybe it could be tackled as a follow up.

LGTM

PR #1792, opened 2025-12-02
[fontir] 'Fix' overflow in overlay_feature_variations

cmyr:
This was ported directly from the python,  and the python uses an
(arbitrary precision) python integer to track the rank. We were using a
u64 here, and there exists at least one font (playfair) where there end
up being more than 64 unique regions for feature variations.

This new solution tries to balance performance and correctness: we can
handle up to 511 unique regions, will panic in debug if this is
exceeded, and will log if we exceed in release.

I am not 100% confident in this approach, or in 511 as a reasonable
maximum, but it's a starting point.

### DIFF
--- a/fontir/src/feature_variations.rs
+++ b/fontir/src/feature_variations.rs
@@ -11,6 +11,7 @@ use std::{
 use crate::ir::StaticMetadata;
 use fontdrasil::{coords::NormalizedCoord, types::GlyphName};
 use indexmap::IndexMap;
+use smallvec::SmallVec;
 use write_fonts::{
     tables::layout::{ConditionFormat1, ConditionSet},
     types::{F2Dot14, Tag},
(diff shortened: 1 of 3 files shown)


fontir/src/feature_variations.rs
```diff
@@ -186,6 +187,70 @@ impl Region {
...
+        let [a1, b1, c1, d1] = self.0;
+        let [a2, b2, c2, d2] = rhs.0;
+        Self([a1 | a2, b1 | b2, c1 | c2, d1 | d2])
+    }
+}
+
+impl BitOrAssign for Rank {
+    fn bitor_assign(&mut self, rhs: Self) {
+        let [a2, b2, c2, d2] = rhs.0;
+        self.0[0] |= a2;
```
behdad: Can you use loops, and make this generic the whole implementation generic to the number of items? Sounds easier to me.

rsheeter:
> use SmallVec and arbitrary size

+1, is there a reason this is infeasible?

PR #1794, opened 2025-12-03
[name] Skip empty strings in NameBuilder

cmyr:
This was easy to diagnose, easy to track down, and easy to fix 🥲.

Gives us a +small on crater.


rsheeter:
> This was easy to diagnose, easy to track down, and easy to fix 🥲.

Is that even allowed?

PR #1796, opened 2025-12-03
Improve support for in-memory compilation, helps wasm use

rsheeter:
This PR aspires toward #1791's goal of making in-memory compiles for wasm happier by making it clearer how to say don't touch the disk

1. Split cli args and compile args so the latter can be more interestingly typed
   * cli args just turn `--flag --here` into a compile arg
1. Make all paths distinct and optional
   * Emit x is removed from Flags
   * build_dir is removed, each distinct file we write gets an Option for a path
   * It falls out of ^ that to write no files you set all output paths to None

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -438,8 +438,8 @@ fn write_debug_glyph_order(debug_dir: &Path, glyphs: &GlyphOrder) {
     let glyph_order_file = debug_dir.join("glyph_order.txt");
     let glyph_order = glyphs.names().map(|g| g.as_str()).collect::<Vec<_>>();
     let glyph_order = glyph_order.join("\n");
-    if let Err(e) = fs::write(glyph_order_file, glyph_order) {
-        log::error!("failed to write glyph order to debug/glyph_order.txt: '{e}'");
+    if let Err(e) = fs::write(&glyph_order_file, glyph_order) {
+        log::error!("failed to write glyph order to {glyph_order_file:?}: '{e}'");
     }
 }
 
--- a/fontbe/src/orchestration.rs
+++ b/fontbe/src/orchestration.rs
@@ -837,15 +837,16 @@ impl PersistentStorage<AnyWorkId> for BePersistentStorage {
             return None;
         }
         let raw_file = File::open(file.clone())
-            .map_err(|e| panic!("Unable to write {file:?} {e}"))
+            .map_err(|e| panic!("Unable to read {file:?} {e}"))
             .unwrap();
         Some(Box::from(BufReader::new(raw_file)))
     }
 
     fn writer(&self, id: &AnyWorkId) -> Box<dyn io::Write> {
-        let Some(dir) = self.dir.as_ref() else {
-            panic!("Write requested with no output dir");
-        };
+        let dir = self
+            .dir
+            .as_ref()
+            .expect("Write requested with no output dir");
         let file = Paths::target_file(dir, id.unwrap_be());
         let raw_file = File::create(file.clone())
             .map_err(|e| panic!("Unable to write {file:?} {e}"))


simoncozens:
In #1684 I made it possible to pass a `Source` in; (useful if you are generating your own IR, not just one of the defined Inputs) this removes that ability, unfortunately.

fontc/src/args.rs
```diff
@@ -227,12 +186,36 @@ impl Serialize for ValidatedRegex {
...
+        let input = self.source()?;
+        let timing_file = self.emit_timing.then(|| self.build_dir.join("threads.svg"));
+        let debug_dir = self.emit_debug.then(|| self.build_dir.clone());
+        let ir_dir = self.emit_ir.then(|| self.build_dir.clone());
+        Ok(fontc::Args {
+            flags,
+            skip_features: self.skip_features,
+            output_file: self
+                .output_file
+                .or_else(|| Some(self.build_dir.join("font.ttf"))),
```
anthrotype, line 203: there's the same problem I identified in Simon's PR when --emit-ir is used in combination with a custom --output-file.
When IR is enabled, ir_dir is set to build_dir; the font gets persisted (via BePersistentStorage) to build_dir/font.ttf regardless of output_file. But `write_font_file` function checks if `!output_file.exists()` and fails if user passed --output-file to something that != build_dir/font.ttf.

Maybe we should simply copy/rename font from ir_dir/font.ttf to wherever output_file is when ir_dir is Some.

fontc/src/lib.rs
```diff
@@ -191,67 +165,57 @@ pub fn require_dir(dir: &Path) -> Result<(), Error> {
...
+    if let Some(output_file) = args.output_file.as_ref()
         && let Some(parent) = output_file.parent()
     {
         require_dir(parent)?;
     }
 
-    // the build dir stores the IR (for incremental builds) and the default output
-    // file ('font.ttf') so we don't need to create one unless we're writing to it
-    if output_file.is_none() || flags.contains(Flags::EMIT_IR) {
-        require_dir(build_dir)?;
```
anthrotype, line 224: you still need to require_dir(build_dir)?, because if it does not exist, when --emit-ir is set alongside --output-file pointing to a _different_ directory (where none of the parents happens to be the build dir), you get a crash later when creating the  build/anchor_ir directory:

```
$ rm -rf build && ./target/release/fontc --emit-ir --output-file /tmp/fontc-test/custom.ttf resources/testdata/static.designspace
[2025-12-04T11:10:37.315180Z ThreadId(1) fontc ERROR] io failed for 'build/anchor_ir/': 'No such file or directory (os error 2)'
```

I think we may need to test the CLI as well

fontc/src/args.rs
anthrotype: reproducer below (note that I make the `build/` directory myself otherwise [the other bug](https://github.com/googlefonts/fontc/pull/1796/files#r2588618689) shows up)

```
$ cargo build --release
$ mkdir -p build/
$ ./target/release/fontc --emit-ir --output-file /tmp/fontc-test/custom.ttf resources/testdata/static.designspace
[2025-12-04T11:13:40.885535Z ThreadId(1) fontc ERROR] Missing file '/tmp/fontc-test/custom.ttf'
```

fontbe/src/orchestration.rs
```diff
@@ -843,7 +843,10 @@ impl PersistentStorage<AnyWorkId> for BePersistentStorage {
     }
 
     fn writer(&self, id: &AnyWorkId) -> Box<dyn io::Write> {
-        let file = self.paths.target_file(id.unwrap_be());
+        let Some(dir) = self.dir.as_ref() else {
+            panic!("Write requested with no output dir");
```
simoncozens: This is just `.expect()`?

fontc/src/compile_args.rs
```diff
@@ -0,0 +1,36 @@
...
+        let input_source = crate::testdata_dir().join(source).canonicalize().unwrap();
+        Self {
+            flags: Flags::default(),
+            skip_features: false,
+            output_file: Some(build_dir.join("font.ttf")),
+            ir_dir: Some(build_dir.to_path_buf()),
+            debug_dir: None,
+            timing_file: None,
+            input: Input::new(&input_source)
+                .unwrap_or_else(|e| panic!("Unable to create input for {input_source:?}: {e}")),
```
simoncozens: Also a funny way to spell `expect`?

anthrotype: I believe it's in order to format the string. If you do `.expect(format!("..."))` clippy will complain and suggest to fix it like that

fontc/src/lib.rs
```diff
@@ -191,67 +165,57 @@ pub fn require_dir(dir: &Path) -> Result<(), Error> {
...
+        require_dir(&fontir::paths::Paths::anchor_ir_dir(ir_dir))?;
+        require_dir(&fontir::paths::Paths::glyph_ir_dir(ir_dir))?;
+        require_dir(&fontbe::paths::Paths::glyph_dir(ir_dir))?;
     }
     // It's confusing to have leftover debug files
-    if be_paths.debug_dir().is_dir() {
-        fs::remove_dir_all(be_paths.debug_dir()).map_err(|source| Error::FileIo {
-            path: be_paths.debug_dir().to_owned(),
+    if let Some(debug_dir) = args.debug_dir.as_ref() {
+        fs::remove_dir_all(debug_dir).map_err(|source| Error::FileIo {
```
anthrotype: This one is similar, but distict. When --emit-debug is set and the debug_dir (which equals build_dir) doesn't exist, `remove_dir_all` fails with "No such file or directory".

```
$ rm -rf build && ./target/release/fontc --emit-debug --output-file /tmp/out.ttf resources/testdata/static.designspace
[2025-12-04T11:49:16.207521Z ThreadId(1) fontc ERROR] io failed for 'build': 'No such file or directory (os error 2)'
```

you need to restore the old `if be_paths.debug_dir.is_dir()` which you have deleted here

fontbe/src/orchestration.rs
anthrotype: pre-existing, but since I stumbled on it we might as well fix

```suggestion
            .map_err(|e| panic!("Unable to read {file:?} {e}"))
```

fontbe/src/features.rs
anthrotype: "debug/glyph_order.txt" may be incorrect, the error should use the actual `debug_dir`

fontir/src/orchestration.rs
```diff
@@ -373,7 +368,10 @@ impl PersistentStorage<WorkId> for IrPersistentStorage {
     }
 
     fn writer(&self, id: &WorkId) -> Box<dyn Write> {
-        let file = self.paths.target_file(id);
+        let Some(ir_dir) = self.ir_dir.as_ref() else {
+            panic!("Write requested while inactive");
```
cmyr, line 372: If the intention is to communicate that this path should be genuinely unreachable I would prefer the `unreachable!` macro.

cmyr (commented):
I've been holding off on merging #1775 because I wanted to discuss some of our options in a bit more depth, but I think it also provides an opportunity to simplify some of this logic.

fontir/src/orchestration.rs
rsheeter: Ty. I consistently forget that one exists.

fontbe/src/features.rs
rsheeter: updated to emit glyph_order_file

fontbe/src/orchestration.rs
rsheeter: Ty

fontc/src/args.rs
anthrotype: Fixed in #1797

fontc/src/lib.rs
anthrotype: Fixed in #1797

fontc/src/lib.rs
anthrotype: Fixed in #1797

fontbe/src/orchestration.rs
rsheeter: It is! I must have thought I was going to include format tokens and then not done so.

rsheeter: Fixed.

rsheeter:
> please take a look at that before duplicating work, thanks!

I feel seen :D

PR #1797, opened 2025-12-04
Remove input from fontc::Args, rename as Options; restore `Box<dyn Source>` support

anthrotype:
Refactor the compile arguments API to:
- Rename `fontc::Args` to `fontc::Options` for clarity (avoids confusion with the clap-based CLI `args::Args`)
- Remove `input` field from Options - source is now passed separately
- Update `generate_font()` to accept `Box<dyn Source>` + `Options`
- Update `run()` to accept `Input` + `Options`

This restores the ability to pass custom `Box<dyn Source>` implementations that was added in #1684, which library users need for generating their own IR programmatically.

Also:
- Add `Default` derive to `Options` for ergonomic in-memory/WASM usage
- Rename `cli_args.rs` back to `args.rs` and delete `compile_args.rs` (Options inlined in lib.rs since it's short enough)

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -356,12 +356,9 @@ mod tests {
         /// Create test options with default settings
         fn for_test(build_dir: &Path) -> Options {
             Options {
-                flags: Flags::default(),
-                skip_features: false,
                 output_file: Some(build_dir.join("font.ttf")),
                 ir_dir: Some(build_dir.to_path_buf()),
-                debug_dir: None,
-                timing_file: None,
+                ..Default::default()
             }
         }
     }


anthrotype:
before y'all start bikeshedding, here's why I think the name Options beats Args.

It's more semantically accured: `Args` is "what I was given" and close to CLI terminology; `Options` is "how to behave" and closer to configuration terminology. After removing the input from the old Args, we're left with _how_ to compile, not _what_, and that sounds more like options.
Plus we don't need two structs both confusingly named "Args", one in cli_args and the other in compile_args module.
I can go on...
Library users don't usually pass "arguments", they configure options. If someone creates Options programmatically (not from CLI), calling it "Args" is awkward.

```rust
let args = Args { flags, skip_features: true, ... };
generate_font(source, args)  // args to a function that's not main()?
```

fontc/src/lib.rs
```diff
@@ -86,20 +84,42 @@ impl TryFrom<&Path> for Input {
...
+///
+/// # Errors
+///
+/// Returns [`Error::NoOutputFile`] if `options.output_file` is `None`.
 #[cfg(feature = "cli")]
-pub fn run(args: Args, mut timer: JobTimer) -> Result<(), Error> {
+pub fn run(input: Input, options: Options, mut timer: JobTimer) -> Result<(), Error> {
+    if options.output_file.is_none() {
+        return Err(Error::NoOutputFile);
+    }
```
cmyr, line 112: Couldn't this be just a required field on the `Args` struct, and then `Args::parse()` will fail immediately?

anthrotype: the check exists because `Options.output_file` must be `Option<PathBuf>` for library/WASM users who don't want file output. Our own main.rs always sets it because it uses `Args::try_into()` which fills it with ` .or_else(|| Some(build_dir.join("font.ttf")))`, so you may argue it's redundant. However the `run()` function is pub and someone might call it directly with hand-built `Options` so the check is a safety net for that case.

anthrotype: --output-file has an implicit default (`"{build_dir}/font.ttf"`) and the NoOutputFile error only guards against direct `run()` callers, not CLI users. And we want `fontc source.glyphs` to continue working as before, without requiring to pass --output-file.

PR #1799, opened 2025-12-04
[glyphs] Handle master names like 'ThinItalic'

cmyr:
That is, we shouldn't be appending 'Italic' to these names.

Gives us another little +small. JMM

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2367,7 +2367,8 @@ impl RawFont {
                 && (names.is_empty()
                     || !names
                         .iter()
-                        .any(|name| *name == "Italic" || *name == "Oblique"))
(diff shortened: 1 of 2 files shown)


glyphs-reader/src/font.rs
```diff
@@ -2367,7 +2367,7 @@ impl RawFont {
                 && (names.is_empty()
                     || !names
                         .iter()
-                        .any(|name| *name == "Italic" || *name == "Oblique"))
+                        .any(|name| name.contains("Italic") || name.contains("Oblique")))
```
rsheeter, line 2371: This should either be ends with or we should add a test case where it's not at the end. I don't _think_ I've seen it anywhere but at the end given it's typical to do MS style weight width slope?

cmyr: this is just matching the logic in glyphsLib, I've added a reference

PR #1800, opened 2025-12-04
Destroying directories is bad mmkay

rsheeter:
`--build-dir .` deemed harmful, make it less so

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -216,14 +216,7 @@ pub fn init_paths(options: &Options) -> Result<(), Error> {
         require_dir(&fontir::paths::Paths::glyph_ir_dir(ir_dir))?;
         require_dir(&fontbe::paths::Paths::glyph_dir(ir_dir))?;
     }
-    // It's confusing to have leftover debug files
     if let Some(debug_dir) = options.debug_dir.as_ref() {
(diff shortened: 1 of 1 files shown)


fontc/src/lib.rs
```diff
@@ -216,14 +216,7 @@ pub fn init_paths(options: &Options) -> Result<(), Error> {
...
         require_dir(&fontbe::paths::Paths::glyph_dir(ir_dir))?;
     }
-    // It's confusing to have leftover debug files
     if let Some(debug_dir) = options.debug_dir.as_ref() {
-        if debug_dir.is_dir() {
-            fs::remove_dir_all(debug_dir).map_err(|source| Error::FileIo {
-                path: debug_dir.clone(),
-                source,
-            })?;
-        }
```
cmyr, line 226: is `debug_dir` now also always the same as `build_dir`? if so maybe we nuke it?

In any case happy to see this get merged, obviously..

cmyr (approved):
oops

fontc/src/lib.rs
rsheeter: No, by default they are build/ and build/debug respectively. I think it's a bug that setting build-dir breaks that.

Issue #1704, opened 2025-10-20
should optionally propagate anchors for designspace+UFOs


anthrotype:
currently we only propagate anchors for .glyphs sources as a preprocess step in Font::load, before conversion to IR (a custom parameter can optionally disable the anchor propagation).

There are designspace+UFOs project that would also need the anchor propagation (e.g. GoogleSans).

To support this we need to move anchor propagation from glyphs-reader to fontir, and enable it for both Glyphs and UFO/DesignSpace sources, controlled via a compilation flag such as #1701

PR #1802, opened 2025-12-08
Support --flag=false to force-disable source-enabled flags

anthrotype:
Some compilation flags such as ERASE_OPEN_CORNERS and FLATTEN_COMPONENTS (and the future PROPAGATE_ANCHORS #1704) are either enabled by default for specific source types or the source may contains compilation flags that enable them.
Currently the CLI can only enable these flags, but not force-disable them, because bitflags can only be true or false and source compilation flags are OR'ed intlo the CLI flags so they are enabled whenever either is set.

In this PR, I changed --flatten-components and --erase-open-corners to tri-state args whereby:
- when omitted, we use source default or source-level compilation flag (e.g. custom parameter or lib filter)
- if --flag or --flag=true: we enable them, irrespective of source
- if --flag=false: force disable

To accomplish this I added a new `flags_to_disable: Flags` field to  the `Option` struct and updated `merge_compilation_flags()` to apply the disable mask _after_ merging CLI and source compilation flags.

(I had a version of this commit in my WIP local work on anchor-propagation-in-fontir but figured that it'd be cleaner to push it seprately)

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -84,34 +84,58 @@ impl TryFrom<&Path> for Input {
     }
 }
 
+/// Flags to explicitly disable, overriding source defaults.
+///
+/// Defaults to empty (no flags disabled).
+#[derive(Debug, Clone, Copy)]
+pub struct DisableFlags(Flags);
+
+impl Default for DisableFlags {
+    fn default() -> Self {
+        DisableFlags(Flags::empty())
+    }
+}
+
+impl DisableFlags {
+    /// Returns an empty set of flags to disable.
+    pub fn empty() -> Self {
+        DisableFlags(Flags::empty())
+    }
+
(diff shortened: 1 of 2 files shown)


fontc/src/lib.rs
```diff
@@ -100,6 +100,20 @@ pub struct Options {
     pub debug_dir: Option<PathBuf>,
 }
 
+impl Default for Options {
+    fn default() -> Self {
+        Self {
+            flags: Flags::default(),
+            flags_to_disable: Flags::empty(), // Must be empty, not default!
```
rsheeter: it would be useful to say why

anthrotype: I ended up making a `DisableFlags` newtype to make intent clearer, avoiding the footgun of one constructing Option manually with flags_to_disable: Flags::default() (which != Flags::empty())

PR #1803, opened 2025-12-08
Use the physical number of CPU instead of logical

wmedrano:
TLDR: 13% improvement on AMD/Intel CPUs.

Rayon uses the number of logical cores by default. On most AMD and Intel CPUs, the number of logical cores is double the physical due to [SMT](https://en.wikipedia.org/wiki/Simultaneous_multithreading). However, scheduling more cores seems to add more overhead than benefit.

`num_cpus` is used to get the number of physical cores. This is already a transitive dependency through:

```mermaid
graph LR
    ufo2fontir --> norad --> close_already --> threadpool --> num_cpus
```

# Results

Tested on:
- GoogleSans Flex
-  `AMD Ryzen Threadripper PRO 3945WX 12-Cores`
    - 12 physical cores
    - 24 logical cores

## Command

```
time poop "/tmp/fontc-before ../googlesans-flex/sources/GoogleSansFlex.designspace --flatten-components --decompose-transformed-components" "/tmp/fontc-after  ../googlesans-flex/sources/GoogleSansFlex.designspace --flatten-components --decompose-transformed-components"
```

## Stats

Measurement | Benchmark 1 (Before) Mean | Benchmark 2 (After) Mean | Delta (Improvement)
-- | -- | -- | --
Wall Time | 14.3s $\pm$ 140ms | 12.3s $\pm$ 154ms | ⚡-13.7% $\pm$ 2.3%
CPU Cycles | 568G $\pm$ 640M | 349G $\pm$ 871M | ⚡-38.5% $\pm$ 0.3%
Branch Misses | 2.21G $\pm$ 12.6M | 1.36G $\pm$ 2.44M | ⚡-38.4% $\pm$ 0.9%
Peak RSS (Memory) | 2.03GB $\pm$ 93.2MB | 2.03GB $\pm$ 103MB | -0.2% $\pm$ 11.0%
Instructions | 1.12T $\pm$ 30.6M | 1.12T $\pm$ 45.1M | +0.0% $\pm$ 0.0%

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -20,6 +20,7 @@ use fontbe::orchestration::AnyWorkId;
 use std::{
     ffi::OsStr,
     fs,
+    num::NonZeroUsize,
     path::{Path, PathBuf},
 };
 
@@ -96,6 +97,7 @@ pub struct Options {
     pub timing_file: Option<PathBuf>,
     pub ir_dir: Option<PathBuf>,
     pub debug_dir: Option<PathBuf>,
+    pub num_threads: Option<NonZeroUsize>,
 }
 
 /// Run the compiler with the provided input and options.
@@ -179,7 +181,7 @@ fn generate_font_internal(
         options.debug_dir.clone(),
         &fe_root,
     );
-    let timer = workload.exec(&fe_root, &be_root)?;
+    let timer = workload.exec(options.num_threads, &fe_root, &be_root)?;
     Ok((fe_root, be_root, timer))
 }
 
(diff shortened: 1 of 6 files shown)


fontc/src/lib.rs
```diff
@@ -20,6 +20,7 @@ use fontbe::orchestration::AnyWorkId;
 use std::{
     ffi::OsStr,
     fs,
+    num::NonZeroUsize,
```
cmyr, line 23: Fancy! I suppose the rationale here is that you're using this to enforce the invariant, and this isn't about trying to save a word in a struct representation or something...

rsheeter: Reminds me of my frequent wish for a NotEmptyVec :)

wmedrano:
After #1812 the win is pretty negligible.

Closing for now to focus on the other bottlenecks.

PR #1804, opened 2025-12-09
CI for unicode script property

cmyr:
This adds a little util binary that checks that python and rust are computing the same set of unicode script extensions for the full set of codepoints, and then adds a CI step to run that.

This job will fail if ever we are running different versions of unicode.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -174,6 +174,24 @@ jobs:
       - name: cargo test fea-rs
         run: cargo test -p fea-rs
 
+  # run a script that sanity checks that we're on the same version of unicode
+  py-rs-unicode-properties:
+    name: check python vs rust unicode properties
+    runs-on: ubuntu-latest
+    steps:
+      - name: install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - uses: actions/checkout@v4
+      - uses: getsentry/action-setup-venv@v3.0.0
+        id: venv
+        with:
+          python-version: 3.13
+          cache-dependency-path: resources/scripts/requirements.txt
+          install-cmd: pip install -r resources/scripts/requirements.txt
+      - name: run check_unicode_props
+        run: cargo run -p fontbe --bin check_unicode_props
+
+
   # Check if GS builds, builds repeatably, and passes OTS
   # https://github.com/googlefonts/fontc/issues/647
   build-gs-repeatably:


.github/workflows/rust.yml
```diff
@@ -174,6 +174,24 @@ jobs:
...
+    name: check python vs rust unicode properties
+    runs-on: ubuntu-latest
+    steps:
+      - name: install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - uses: actions/checkout@v4
+      - uses: getsentry/action-setup-venv@v1.0.0
+        id: venv
+        with:
+          python-version: 3.13.9
```
rsheeter: we seem to have multiple inconsistent python-version in this file. Can we use the same one? Do we have to be so specific or can we say e.g. 3.13 and leave .x to setup?

fontbe/src/bin/check_unicode_props.rs
```diff
@@ -0,0 +1,135 @@
...
+fn py_script_extensions_batch(start: u32, end: u32) -> Vec<(u32, BTreeSet<TinyAsciiStr<4>>)> {
+    let python_code = format!(
+        r#"
+from ufo2ft.util import unicodeScriptExtensions
+
+for cp in range({start}, {end} + 1):
+    scripts = list(unicodeScriptExtensions(cp))
+    print(f"{{cp}}: {{','.join(scripts)}}")
+"#,
+    );
```
rsheeter, line 26: is there an advantage to doing this rather than having a .py file that can be run independently if one is so inclined?

rsheeter: For example, perhaps `include_str a file in resources/scripts` ?

fontbe/src/bin/check_unicode_props.rs
```diff
@@ -0,0 +1,135 @@
...
+        .expect("failed to execute python");
+
+    if !output.status.success() {
+        panic!(
+            "python failed, stderr: '{}'",
+            String::from_utf8_lossy(&output.stderr)
+        )
+    }
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
```
rsheeter: suggest from_utf8.expect, it shouldn't be bad utf8 but if it is we shouldn't ignore it

fontbe/src/bin/check_unicode_props.rs
```diff
@@ -0,0 +1,135 @@
...
+    })
+}
+
+fn main() {
+    const BATCH_SIZE: u32 = 200000;
+
+    // Test ALL valid Unicode codepoints (0x0000 to 0x10FFFF, excluding surrogates)
+    //the surrogates are  0xD800..=0xDFFF
+    let rangea = 0x0..0xD7FF;
+    let rangeb = 0xF000..0x10FFFF;
```
rsheeter, line 88: Feels harder to follow than 0x0..0x10FFFF.skip(if surrogate)?

.github/workflows/rust.yml
cmyr: will open a separate issue to update the fea-rs test code/tests more generally.

fontbe/src/bin/check_unicode_props.rs
cmyr: this was the simplest approach I could think of, felt nice to just have a single file here. Can you think of a reason you might want to run this independently? It's easy enough to split out if a reason arises...

fontbe/src/bin/check_unicode_props.rs
cmyr: good call!

fontbe/src/bin/check_unicode_props.rs
cmyr: this is about splitting up the work into chunks to send to python, and we only want to send the start/end bound, so we don't need to write every codepoint through stdout.

.github/workflows/rust.yml
```diff
@@ -174,6 +174,24 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - uses: actions/checkout@v4
+      - uses: getsentry/action-setup-venv@v3.0.0
+        id: venv
+        with:
+          python-version: 3.13
+          cache-dependency-path: resour/scripts/requirements.txt
```
anthrotype: is this intentional or a typo?

```suggestion
          cache-dependency-path: resources/scripts/requirements.txt
```

cmyr: good catch!

PR #1812, opened 2025-12-09
Cache GlobalMetrics::at

### LINKED ISSUE #1810: GlobalMetrics::at
If I run Google Sans Flex in samply with one rayon thread so it's easier to read the overall time breakdown a truly amazing amount of worker time goes to `GlobalMetrics::at`:

```shell
$ export RAYON_NUM_THREADS=1
$ ../samply/target/release/samply record target/release/fontc ../googlesans-flex/sources/GoogleSansFlex.designspace
```

Image

If I modify ir.rs so GlobalMetrics::at records the position:

```rust
        let mut file = OpenOptions::new()
            .append(true)
            .create(true)
            .open("/tmp/global_metrics_at.txt")
            .unwrap();
        file.write_all(&format!("{pos:?}\n").as_bytes()).unwrap();
```

And then check how many distinct positions we call at for it would appear there are 145,947 calls to request 360 distinct global metrics instances:

```shell
$ cargo run ../googlesans-flex/sources/GoogleSansFlex.designspace
$ wc -l /tmp/global_metrics_at.txt 
145947 /tmp/global_metrics_at.txt
$ sort /tmp/global_metrics_at.txt | uniq | wc -l
360
````

One tends to suspect we could precompute, or cache, at and profit :)

### PR DESCRIPTION
**Pull request by rsheeter**
Based on #1810, try adding a cache for global metrics. Fixes #1810.

```shell
$ hyperfine \
  --parameter-list branch main,$(git rev-parse --abbrev-ref HEAD) \
  --setup "git checkout {branch} && cargo build --release" \
   --warmup 5 --runs 10 \
   --prepare 'rm -rf build/' \
   "target/release/fontc --source ../googlesans-flex/sources/GoogleSansFlex.designspace --flatten-components --decompose-transformed-components"

    1.09 ± 0.03 times faster than target/release/fontc --source ../googlesans-flex/sources/GoogleSansFlex.designspace --flatten-components --decompose-transformed-components (branch = main)
```

Note that I tried storing an Arc to avoid copying but hyperfine couldn't tell the difference (1.08 ± 0.03 faster) so I left it copying. If I repeat the profiling from #1810 I no longer see GlobalMetrics::at prominantely reported.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -10,6 +10,7 @@ use indexmap::{IndexMap, IndexSet};
 use kurbo::{Affine, BezPath, PathEl, Point};
 use log::{log_enabled, trace, warn};
 use ordered_float::OrderedFloat;
+use parking_lot::RwLock;
 use serde::{Deserialize, Serialize, de::Error as _};
 use smol_str::SmolStr;
 use write_fonts::{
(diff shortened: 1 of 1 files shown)


fontir/src/ir.rs
```diff
@@ -304,8 +305,18 @@ pub struct GlobalMetricsBuilder(
 /// cap height, etc.
 ///
 /// This type is constructed by [GlobalMetricsBuilder].
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
-pub struct GlobalMetrics(HashMap<GlobalMetric, ModelDeltas<f64>>);
+#[derive(Serialize, Deserialize, Debug)]
```
anthrotype, line 308: why is GlobalMetrics no longer `Clone`?

anthrotype: oh is it because RwLock doesn't let you derive Clone any longer, and we weren't actually cloning it anyway. Alternatively we could manually implement Clone (like you did for PartialEq), but we don't need since we get it as `Arc<GlobalMetrics>`, nevermind

anthrotype (approved):
LGTM

PR #1823, opened 2025-12-11
Fixup sub/superscript metric fallback logic

cmyr:
If there are explicit values in a font for the 'Os2Subscript___' metrics, we want to reuse those for the corresponding superscript metrics.

This means that we need to query metrics in this routine, which ran into borrowck issues due to the long-lived closure that had a mutable ref to self. To work around that, I've replaced that closure with a macro.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -437,13 +437,11 @@ impl GlobalMetricsBuilder {
     ) {
         let units_per_em = units_per_em as f64;
 
-        let mut set_if_absent = |metric, value| self.set_if_absent(metric, pos.clone(), value);
-
         // https://github.com/googlefonts/ufo2ft/blob/fca66fe3ea1ea88ffb36f8264b21ce042d3afd05/Lib/ufo2ft/fontInfoData.py#L38-L45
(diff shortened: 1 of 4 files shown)


fontir/src/ir.rs
```diff
@@ -426,13 +426,18 @@ impl GlobalMetricsBuilder {
     ) {
         let units_per_em = units_per_em as f64;
 
-        let mut set_if_absent = |metric, value| self.set_if_absent(metric, pos.clone(), value);
+        // a macro instead of a closure to avoid borrowck
```
rsheeter: This smells, I didn't spot it at a glance, why can't this be normal code?

wmedrano: Personally, I think calling `self.set_if_absent(metric, pos.clone(), value);` directly is fine.

- set_if_absent can also take an `Into<Cow<NormalizedLocation>>` if you want to avoid the explicit clone

cmyr: The whole reason this exists afaict is because always calling `self.set` ends up just... being a bunch of extra lines? And the closure doesn't work with this patch because you can't call `self.get` because the closure has an `&mut` ref to self.

PR #1831, opened 2025-12-12
Update requirements.txt

anthrotype:
Notably includes updates fonttools, ufo2ft and glyphsLib which should give us some more greens on crater


cmyr (approved):
thanks!

Issue #1661, opened 2025-09-26
Anchor positions need to be interpolated during propagation


cmyr:
anchor positions also interpolate... in fact Glyphs.app anchor propagation works even when a composite glyph inherit anchors from a component that doesn't have a given location, which gets interpolated on the fly

_Originally posted by @anthrotype in https://github.com/googlefonts/fontc/pull/1658#discussion_r2382831577_

anthrotype:
while investigating the remaining ttx diffs in GSF, I noticed that in glypys-reader `propagate_anchors.rs`, `get_component_layer_anchors()` looks up component anchors using `layer.layer_id` (except for bracket layers where it matches them by their axis rules and associated master id):

https://github.com/googlefonts/fontc/blob/40ba7f6b27b74606d6b5828311ef5d77e2c655b7/glyphs-reader/src/propagate_anchors.rs#L395-L400

That in theory is supposed to only ever match _master_ layers, which by definition share the same layer id (== master.id) across all the glyphs. Any other non-master layer is supposed to get its own unique layerId (a UUID) so even if multiple glyphs have intermediate layers at the same axis coordinates, the corresponding layers' ids in each will not be equal.

However it can happen (and in fact it happens in GSF's "eth" composite glyph and its components  "comp_eth" and "comp_eth_bar") that multiple glyphs share the same layer id even for layers which are not master layers, such as intermediate ones. It could be that a glyph was duplicated without the layer ids not being regenerated, or layers copied across glyphs through python scripts, etc.

I'm just noting this here, because this happenstance of shared intermediate layer ids for the composite glyph "eth" was exactly what was causing the remaining 1.8% diff in the markkern.txt: fontc was propagating the "center" anchor to the "eth" composite glyph from the "comp_eth" component for the intermediate layers that happened to share the same layerId, and this meant additional regions added to the GDEF VarStore, which fontmake would instead not add. GlyphsLib only processes master or alternate layers in its anchor propagation preprocessing transformation, [skipping intermediate layers altogether](https://github.com/googlefonts/glyphsLib/issues/1017).

In the end, the GSF team wants the "Propagate Anchors" to be off (via custom parameter) and thus we get 100% output identical.

But this is something to bear in mind which may be affecting other fonts in the [fontc_crater](https://googlefonts.github.io/fontc_crater)

PR #1832, opened 2025-12-12
Move anchor propagation from glyphs-reader to fontir

### LINKED ISSUE #1704: should optionally propagate anchors for designspace+UFOs

anthrotype:
Fixes #1704

This huge PR refactors anchor propagation to be source-agnostic by moving it from `glyphs-reader` (Glyphs-specific preprocessing) to `fontir` so that it works with not only for Glyphs sources but also for DS + UFOs or any other frontends.

GoogleSans.designspace markkern.txt goes from the current 82.760% to 97.262% (the remaining diffs are bugs on ufo2ft propagateAnchors filter's side which I'll fix separately).

The main problem when moving the implementation from glyphs-reader to fontir was that there was a circular dependency in the way GDEF categories are computed for Glyphs sources in particular, whereby on the one hand the anchor propagation algorithm needs mark/ligature information, but on the other hand some GDEF glyph categories (Base, Ligature) are inferred based on the presence of anchors and thus need anchors to be propagated first.
In the previous code on `main`, this wasn't a problem because glyphs-reader would use the glyph category/subCategory fields to do its propagate_anchors.rs, which would be performed at the very beginning when the Font was first loaded, so that by the time the GDEF categories were computed all the anchors (including propagated) would already be there.

To resolve the circularity, I resorted to a two-phase approach similar to the one we use for PreliminaryGlyphOrder vs GlyphOrder: i.e. first each frontends populate PreliminaryGdefCategories using only source metadata (such as Glyphs' category/subCategory or the UFO public.openTypeCategories) but without checking the anchors; then after anchor propagation is done inside fontir, the _final_ GdefCategories are computed by also taking the anchors into account.
Note that for DS+UFOs workflow, the `public.openTypeCategories` are presumed to already be "final" and are not extented/pruned based on anchors (at least that's how ufo2ft treats these), so in fontc I follow this (source frontends can set a bool field on the PreliminaryGdefCategory struct to control whether final GdefCategories shoud be inferred from anchors or not).

I kept the current source-level defaults: i.e. for .glyphs source we run anchor propagation by default unless disabled via either "Propagate Anchors" (false) custom parameter (as well as a "propagateAnchors" ufo2ft filter hidden in the masters' userData which is a remnant of ufo2glyphs workflows); whereas for DS + UFO sources anchor propagation is disabled by default but it can be enabled via "propagateAnchors" filter in "com.github.googlei18n.ufo2ft.filters" lib key (like GoogleSans does). And in any case, a tri-state CLI flag --propagate-anchors={true,false} can be used to override these (omitted means follow source).

I ported all the tests from the previous implementation, while adding a few more, and deleted the glyphs-reader/src/propagate_anchors.rs.

I know it's a big change, and I tried to split into logical commits that make review easier as much as possible, but it's inevitable that a lot of modules are affected. I suggest to review single commits in the order the appear.

Last thing, https://github.com/googlefonts/fontc/issues/1661 is not implemented here yet but this should make it easier to implement since we are already interpolating glyph geometry in fontir (for decomposing and flattening) so adding anchor interpolation should be relatively simple.

EDIT: I also included in here a fix for propagating cursive anchors with suffixes from https://github.com/googlefonts/glyphsLib/pull/1130

### DIFF
--- a/fontir/src/propagate_anchors.rs
+++ b/fontir/src/propagate_anchors.rs
@@ -5,14 +5,15 @@
 //! This implementation works at the IR level, processing each NormalizedLocation
 //! independently and then aggregating into variable anchors.
 
-use std::collections::HashMap;
+use std::collections::{BTreeSet, HashMap, HashSet};
 
 use indexmap::IndexMap;
 use kurbo::{Affine, Point, Vec2};
 use smol_str::{SmolStr, format_smolstr};
 use write_fonts::tables::gdef::GlyphClassDef;
+use write_fonts::types::Tag;
 
-use fontdrasil::{coords::NormalizedLocation, types::GlyphName};
+use fontdrasil::{coords::NormalizedLocation, types::GlyphName, variations::VariationModel};
 
 use crate::{
     error::{BadGlyph, Error},
@@ -32,6 +33,8 @@ struct RawAnchor {
 /// This is the main entry point for anchor propagation.
 pub fn propagate_all_anchors(context: &Context) -> Result<(), Error> {
     let gdef_categories = &*context.preliminary_gdef_categories.get();
+    let meta = context.static_metadata.get();
+    let axis_order = meta.axes.axis_order();
 
     // 1. Depth-sort glyphs (process components before composites).
     // The reference implementation (Glyphs.app) does this recursively, but we opt to
@@ -51,6 +54,10 @@ pub fn propagate_all_anchors(context: &Context) -> Result<(), Error> {
     let mut done_anchors: HashMap<(GlyphName, NormalizedLocation), Vec<RawAnchor>> = HashMap::new();
     // Track number of base glyphs for ligature anchor numbering
     let mut base_glyph_counts: HashMap<(GlyphName, NormalizedLocation), usize> = HashMap::new();
+    // Cache of variation models keyed by location set, for interpolating component
+    // anchors at missing locations.
+    let mut variation_models: HashMap<BTreeSet<NormalizedLocation>, VariationModel> =
+        HashMap::new();
 
     for glyph_name_smol in &todo {
         let glyph_name: GlyphName = glyph_name_smol.as_str().into();
@@ -95,6 +102,8 @@ pub fn propagate_all_anchors(context: &Context) -> Result<(), Error> {
                 location,
                 &done_anchors,
                 &mut base_glyph_counts,
+                &axis_order,
+                &mut variation_models,
             );
 
             done_anchors.insert((glyph_name.clone(), location.clone()), anchors);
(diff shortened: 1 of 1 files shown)


fontir/src/glyph.rs
```diff
@@ -844,6 +859,83 @@ impl Work<Context, WorkId, Error> for GlyphOrderWork {
...
+            Some(GlyphClassDef::Ligature) if has_attaching_anchor => Some(GlyphClassDef::Ligature),
+            Some(GlyphClassDef::Ligature) => None, // Ligature without anchors is dropped
+            Some(GlyphClassDef::Base) => Some(GlyphClassDef::Base),
+            Some(GlyphClassDef::Component) => Some(GlyphClassDef::Component),
+            _ if has_attaching_anchor => Some(GlyphClassDef::Base), // Infer Base from anchors
+            _ => None,
+        };
+
+        if let Some(cat) = final_cat {
+            final_categories.insert(glyph_name.clone(), cat);
```
wmedrano: I don't see any uses of `glyph_name` after this point so maybe the `clone` can be removed.

anthrotype: good catch, fixed (and squashed into parent commit [1fe9041](https://github.com/googlefonts/fontc/pull/1832/commits/1fe9041152c83360643250f1f6ba16fd0e398089))

fontbe/src/features/kern.rs
```diff
@@ -14,8 +14,9 @@ use fontdrasil::{
     orchestration::{Access, AccessBuilder, Work},
     types::GlyphName,
 };
+use fontir::ir::GdefCategories;
```
cmyr: lets move this in with the other imports from this module?

fontir/src/ir.rs
```diff
@@ -1952,9 +1978,48 @@ pub struct Component {
...
+    /// the [`transform`](Self::transform) by the font editor.
+    ///
+    /// See: <https://handbook.glyphsapp.com/components/#reusing-shapes/anchors>
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub anchor: Option<SmolStr>,
 }
 
 impl Component {
+    /// Create a regular component (no specific attachment anchor).
+    pub fn new(base: GlyphName, transform: Affine) -> Self {
```
cmyr: to clean up the call-sites further you could make this `base: impl Into<GlyphName>`, and then you wouldn't need to all `clone()` or `.into()` everywhere?

fontc/src/args.rs
```diff
@@ -133,7 +139,9 @@ impl Args {
         if self.erase_open_corners == Some(true) {
             flags.set(Flags::ERASE_OPEN_CORNERS, true);
         }
-        // TODO: add PROPAGATE_ANCHORS flag when we have that implemented
+        if self.propagate_anchors == Some(true) {
```
wmedrano, line 142: Not actionable, just looking to learn.

In practice, how often do these flags change? Will user's want to experiment with `propagate_anchors` for the foreseeable future or is it a mechanism to roll out a feature that basically every user will leave alone.

ufo2fontir/src/toir.rs
```diff
@@ -350,8 +387,53 @@ mod tests {
...
+            Value::Dictionary(
+                [
+                    ("name".to_string(), Value::String("acutecomb".to_string())),
+                    ("index".to_string(), Value::Integer(1.into())),
+                    ("anchor".to_string(), Value::String("top_2".to_string())),
+                ]
+                .into_iter()
+                .collect(),
+            ),
+        ]);
```
cmyr, line 429: might be easier to just inline a PLIST and deserialize it? But not worth rewriting..

fontir/src/orchestration.rs
```diff
@@ -37,6 +37,8 @@ bitflags! {
         const DECOMPOSE_COMPONENTS = 0b100000000;
         // If set, open corners will be erased (Glyphs-native feature)
         const ERASE_OPEN_CORNERS = 0b1000000000;
+        // If set, anchors will be propagated from components to composites
+        const PROPAGATE_ANCHORS = 0b10000000000;
```
wmedrano, line 41: I was confused for a second and thought these were all the same! What do you think of separating bytes with underscore?

`0b100_00000000`

anthrotype:
Thanks for taking a look, I'll fix the regressions tomorrow!

fontc/Cargo.toml
```diff
@@ -61,7 +61,7 @@ chrono.workspace = true
 ordered-float.workspace = true
 rstest.workspace = true
 criterion.workspace = true
-
```
wmedrano: Move closer to top with the other `fontc` repo dependencies?

fontra2fontir/src/toir.rs
```diff
@@ -77,7 +77,6 @@ pub(crate) fn to_ir_static_metadata(font_data: &FontraFontData) -> Result<Static
         Default::default(), // TODO: glyph locations we really do need
         Default::default(),
         Default::default(),
-        Default::default(),
```
wmedrano, line 80: This is used to build the StaticMetadata for `fontra2fontir/src/source.rs`. In that file, `also_completes` has not been updated. Does it get generated somewhere else for fontra or does fontra not need GDef categories?

`StaticMetadataWork` for fontir

```rust
impl Work<Context, WorkId, Error> for StaticMetadataWork {
    fn id(&self) -> WorkId {
        WorkId::StaticMetadata
    }

    fn also_completes(&self) -> Vec<WorkId> {
        vec![WorkId::PreliminaryGlyphOrder] // No PreliminaryGdef categories
    }

    fn exec(&self, context: &Context) -> Result<(), Error> {
        debug!("Static metadata for {:#?}", self.fontdata_file);
        context
            .preliminary_glyph_order
            .set(self.glyph_info.keys().cloned().collect());
        context
            .static_metadata
            .set(create_static_metadata(&self.fontdata_file)?);
        Ok(())
    }
}
```

fontc/src/args.rs
anthrotype: it's mostly to be able to test the effect of anchor propagation (or its absence) without needing to modify the source itself. I'd expect users to prefer enabling/disabling this in the source via appropriate custom parameter (or ufo lib key).

anthrotype:
I think this is ready for merge now. I've reviewed the 5 regressions that Colin reported back in December:
- ElmSans and Bitter are now identical (fixed by non-exporting glyph anchors fix 4132509)
- Murecho still has a diff in GDEF on this branch, but it becomes identical once #1885 is also merged
- RedditMono: the GDEF is now identical, but GPOS markkern.txt still contains a diff for "aeacute" because https://github.com/googlefonts/fontc/issues/1661 is not implemented yet (but I have a branch that fixes this which I will push tomorrow)
- Lora has 1 extra GDEF Base entry + 1 extra MarkToBase rule for "cedilla". This one is the most problematic and I am not sure if/how to fix...

The "cedilla" is a composite glyph referencing "cedillacomb" as a component. In Glyphs.app's GlyphData.xml, "cedilla" has category=Mark, subcategory=Spacing (which is a bit of a misnomer since these are not combining marks, the Unicode category is "Sk" modifier symbol). Glyphs.app skips anchor propagation for category=Mark composite glyphs that already have their own anchors, it keeps them as-is without looking at components. This means cedilla (which happens to have "_bottom" in Lora) keeps only that anchor and doesn't inherit "bottom" from cedillacomb. As a result, fontmake doesn't classify it as GDEF Base, and doesn't add MarkToBase for "cedilla".
fontc only skips propagation for GDEF Mark glyphs, which are defined as Glyphs category=Mark and subCategory=Nonspacing or SpacingCombining. Since Mark/Spacing glyphs like cedilla are not GDEF Mark (they're spacing glyphs that shouldn't act as marks in GPOS), fontc propagates cedillacomb's bottom anchor into cedilla, which causes it to be inferred as GDEF Base and generates one extra MarkToBase rule.

I think fontc's behavior is defensible here:
- cedilla is not a combining mark (Unicode Sk, not Mn/Mc)
- Mark/Spacing glyphs are not marks in the OpenType sense (not GDEF Mark class)
- The _bottom anchor on cedilla is arguably a font bug, a non-combining spacing glyph shouldn't need a mark attachment anchor
- Glyphs.app's blanket category=Mark skip is an implementation shortcut, not a principled rule.

I considered these alternatives to match fontmake for Lora, none of which i'm 100% happy with:

1. Add something like `marks_for_propagation: BTreeSet<GlyphName>` to `PreliminaryGdefCategories` to track all source-level category=Mark glyphs (regardless of subcategory) and use that set alongside the GDEF Mark check when deciding whether to skip propagation. This works (verified Lora identical) but leaks a Glyphs-specific concept into the format-agnostic IR layer (would default to empty for UFO sources, which don't have this distinction).
2. Change glyphsLib to match fontc: only skip propagation for true marks (Mark/Nonspacing and Mark/SpacingCombining), not Mark/Spacing. This would align with Unicode and OpenType definition of a mark, but also deviate from Glyphs.app which is supposed to be the reference implementation for .glyphs sources (I verified that "bottom" is not propagated to "cedilla" in Glyphs.app).
3. Use some heuristic to treat glyphs containing underscore-prefixed anchors as mark-like. Simple change, but conflates a coincidental signal (presence of underscore anchors) with the actual source-level Mark category. It would work with this particular diff in Lora's cedilla, but it may have unwanted side effects.

I'm tempted to merge this PR now and file an issue about Lora to fix later.

PR #1851, opened 2025-12-17
[fea-rs] What if only ligature carets are variable?

cmyr:
I foolishly failed to account for the possibility that a font might not have variable marks or kerning but _might_ have variable ligature caret positions, and so would still need an item variation store.

Whatever was I thinking??

JMM

makes one font covering one obscure script match on crater 🤘

### DIFF
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -207,11 +207,16 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
         let mut ivs = VariationStoreBuilder::new(axis_count);
 
         let (mut gsub, mut gpos) = self.lookups.build(&self.features, &mut ivs, &self.opts);
+        // if ivs hasn't been used, we don't want to create a GDEF table just for it.
         if !ivs.is_empty() {
             self.tables
                 .gdef
                 .get_or_insert_with(Default::default)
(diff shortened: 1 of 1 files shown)


fea-rs/src/compile/compile_ctx.rs
```diff
@@ -207,11 +207,16 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
...
 
         let (mut gsub, mut gpos) = self.lookups.build(&self.features, &mut ivs, &self.opts);
+        // if ivs hasn't been used, we don't want to create a GDEF table just for it.
         if !ivs.is_empty() {
             self.tables
                 .gdef
                 .get_or_insert_with(Default::default)
                 .var_store = Some(ivs);
+        // but if we _do_ have a gdef table, always add the var store,
+        // since we might still add ligature carets to it
```
anthrotype, line 217: we "might" or we "do"? Is there still a possibility that we end up with an empty varstore down the line?

anthrotype: reply to myself: no we won't end up with GDEF with empty varstore because of this guard later on in GdefBuilder::builld(). Only if the var_store has content after building lig carets is it actually added to the table.

https://github.com/googlefonts/fontc/blob/main/fea-rs/src/compile/tables/gdef.rs#L63-L69

Might be worth saying so in the inline comment

PR #1875, opened 2026-01-11
Use resolved values for subscript and superscript defaults where needed

Hoolean:
Happy new year!

---

A few of the superscript and subscript properties are unusual in that they depend on the **resolved** value of other attributes: e.g., superscript size should fallback to the explicit subscript size if provided, not just share its default.

This is hopefully an easy +1 for Noto Serif Oriya, which is now happily building at 100% locally : )

## Worth noting

- The PR adds test coverage for an Oriya-adjacent case, although we could add more to be certain there's no regression
- The second commit does some early rounding to match ufo2ft's precision loss -- where further transforms occur, we need to round as early as ufo2ft does to match the exact final value in all cases -- but this wasn't based on a real-world use-case, and so we could drop it if we prefer simplicity unless compatibility demands it

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -488,13 +488,18 @@ impl GlobalMetricsBuilder {
 
         // https://github.com/googlefonts/ufo2ft/blob/0d2688cd847d003b41104534d16973f72ef26c40/Lib/ufo2ft/outlineCompiler.py#L575-L616
         // NOTE: These defaults depend on other metrics' _resolved_ values.
+        // Where they will be further processed, we need to round them early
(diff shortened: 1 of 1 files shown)


Hoolean:
Oof, just realised this duplicates #1823, sorry @cmyr :speak_no_evil:... there's some parallel universe stuff here like using the closure return to avoid `get(...)`, and the optional rounding commit, so feel free to Frankenstein hack and splice this one if anything ends up useful

cmyr:
I believe this has been resolved, please reopen if I've overlooked anything!

PR #1877, opened 2026-01-12
[fea-rs] Compile format 2 contextual lookups

### LINKED ISSUE #1863: fea-rs does not try to build format 2 contextual subtables
From my initial assessment it seemed like these were never more efficient than format 3, but inevitably there are some cases where they are. This probably explains #1849.

One case where feaLib is choosing format 2 where we aren't, and it showing real binary savings:

```
python3 -m ttx_diff 'https://github.com/notofonts/mongolian?a61ef47935#sources/NotoSansMongolian.designspace'
```

It would be nice if we could figure out a heuristic for deciding if we want to try this format or not, because in the _general_ case it doesn't seem more efficient; it probably starts to become attractive when we are dealing with a larger number of target glyphs?

cmyr:
I had initially struggled to come up with an input where format 2 outperformed format 3, but of course they are out there.

This also updates ttx_diff to normalize these; we were already doing this for chain contexts.

- closes #1863
- should be worth at least a +8 on crater

### DIFF
--- a/fea-rs/src/compile/lookups/contextual.rs
+++ b/fea-rs/src/compile/lookups/contextual.rs
@@ -391,6 +391,60 @@ impl ContextBuilder {
 
         Some(write_layout::SequenceContext::format_1(coverage, rule_sets))
     }
+
+    fn format_2_class_def(&self) -> Option<ClassDefBuilder> {
+        let mut builder = ClassDefBuilder::new();
+        for rule in &self.rules {
+            for (glyphs, _) in &rule.context {
+                let class = glyphs.iter().collect();
+                if !builder.checked_add(class) {
+                    return None;
+                }
+            }
+        }
+        Some(builder)
+    }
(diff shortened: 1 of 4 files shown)


fea-rs/src/compile/lookups/contextual.rs
```diff
@@ -391,6 +391,59 @@ impl ContextBuilder {
...
+            .build();
+
+        let mut rule_sets = vec![Vec::new(); input_map.len() + 1];
+        for rule in &self.rules {
+            let first_item = rule.first_input_sequence_item().to_class().unwrap();
+            let cls_idx = *input_map.get(&first_item.into()).unwrap();
+            let input = rule
+                .context
+                .iter()
+                .skip(1)
```
rsheeter, line 427: The skip is obtuse enough I think maybe a comment is warranted

PR #1880, opened 2026-01-21
Remove (or annotate) unwrap() from fontdrasil

### LINKED ISSUE #1879: [fontdrasil] unwrap in convert() can panic
fontdrasil is a low-level library for coordinate space conversion amongst other things. I don't believe that a library at this level should crash the running application under any circumstances, although I'm aware that there are different opinions on this. Anyway, here's the issue:

https://github.com/googlefonts/fontc/blob/dc1b3d0155f6fe3a1294376e48534cd9e3a31fa8/fontdrasil/src/coords.rs#L373-L382

If the Location contains an axis which is not found in the `axes` parameter, the application panics.

I'm very happy to prepare a PR to remove all `unwrap`s from fontdrasil.

simoncozens:
This PR bans `unwrap`, `expect` and indexing slicing (i.e. panicable operations) from fontdrasil. `unwrap()` is replaced either with a `ok_or` or explained with an annotation.

This *does* change the API somewhat; creating coordinate convertors, and converting coordinates, returns a `Result`.

There is also an unrelated change to the version of `vergen-gitcl` used by fontc. Without this, we were using `vergen-gitcl` 1.0 which pulled in `vergen v9.1.0`, which had a very different API to the one `vergen-gitcl` expected. So I couldn't get things to compile without it.

Fixes #1879

### DIFF
--- a/fontdrasil/src/paths.rs
+++ b/fontdrasil/src/paths.rs
@@ -182,13 +182,11 @@ pub fn safe_filename(name: &str, suffix: &str) -> String {
     }
     filename.extend(suffix.chars());
 
-    #[allow(clippy::indexing_slicing)] // We check that filename is non-empty before indexing
-    if let Some(ch) = filename.first()
+    if let Some(ch) = filename.first_mut()
         && *ch == '.'
     {
-        filename[0] = '_';
+        *ch = '_';
     }
-
     let filename: String = filename.into_iter().collect();
 
     // Windows fears no _


fontdrasil/src/paths.rs
cmyr: maybe simpler to just do,
```suggestion
if let Some(ch) = filename.first_mut()
&& *ch == '.'
{
*ch = '_';
}
```

fontdrasil/src/variations.rs
```diff
@@ -953,13 +972,18 @@ fn delta_weights(
                 })
                 .collect(),
         );
-        trace!("  {} {:?}", loc_idx, weights.last().unwrap());
+        // We *don't* know weights is non-empty
+        if let Some(weight) = weights.last() {
+            trace!("  {} {:?}", loc_idx, weight);
+        }
```
cmyr, line 978: alternatively we could just remove the unwrap, and we will debug-print the Option.

fontdrasil/src/variations.rs
```diff
@@ -953,13 +972,18 @@ fn delta_weights(
...
     }
     weights
 }
 
 #[cfg(test)]
 mod tests {
+    #![allow(clippy::unwrap_used, clippy::indexing_slicing)] // test code, knock yourself out
```
cmyr, line 985: was going to suggest this for tests :)

fontdrasil/src/variations.rs
```diff
@@ -832,13 +834,19 @@ fn regions_for(axis_order: &[Tag], locations: &[NormalizedLocation]) -> Vec<Vari
         .map(|location| {
             let mut region = VariationRegion::new();
             for tag in axis_order {
+                assert!(
+                    minmax.contains_key(tag),
+                    "axis_order contained axes not in locations"
+                );
+                #[allow(clippy::unwrap_used)]
```
cmyr, line 841: totally not necessary to change but: this could also be expressed a bit more concisely by just using `expect` in place of `unwrap` below, with the same text as your assertion here.

simoncozens: True, but I see a philosophical difference. I want to ban `expect`s because they're just `unwrap`s with an error message. I know the actual outcome is the same, but `assert` affirmatively signals that we're testing an internal invariant ("In a mad world this might fail, let's make sure and bug out if it does"), while `.unwrap`/`.expect` is "I am positively sure this won't fail".

PR #1885, opened 2026-02-06
Interpolate anchors when instantiating smart components

### LINKED ISSUE #1884: [glyphs] Smart component anchors not interpolated during instantiation
Smart components can have anchors meant to be used for mark positioning (e.g. top, bottom) whose positions vary across the component's internal axis space. When fontc (in glyphs-reader) instantiates smart components, these anchors are dropped entirely, so downstream anchor propagation has nothing to work with and mark attachment breaks.

In an equivalent fix for glyphsLib and fontmake https://github.com/googlefonts/glyphsLib/pull/1131, @khaledhosny e.g. noted one of his font contains an Arabic smart component that varies in height and has top/bottom anchors that are supposed to be interpolated, otherwise marks would end up at the wrong position or missing entirely.

In glyphsLib, anchor propagation happens before smart component instantiation, so the fix there interpolates already-propagated anchors at decomposition time. In fontc the order is reversed and smart components are instantiated in glyphs-reader before anchor propagation runs (and in #502 we plan to move anchor propagation to fontir, so it must happen afterwards). So here the fix needs to happen earlier, when we interpolate the smart component's contour points we can also interpolate anchor positions and then merge the interpolated anchors into the layer.

anthrotype:
Smart components can have anchors that vary across their internal axis space. Previously, these anchors were dropped during instantiation. Now we interpolate anchor positions using the same variation model as shape interpolation.
This fixes Murecho and other fonts that use smart components with anchors.

Fixes #1884

Equivalent fix to @khaledhosny's https://github.com/googlefonts/glyphsLib/pull/1131

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -3835,16 +3835,19 @@ impl Font {
                             issue,
                         })?;
                         new_shapes.extend(instance.shapes);
-                        // Add anchors from smart component
-                        for anchor in instance.anchors {
-                            // Filter out 'mark' anchors used for internal part-to-part connections
-                            if anchor.name.starts_with('_') {
-                                continue;
-                            }
-                            // When duplicates exist, only keep the last one (matching Glyphs.app).
-                            layer.anchors.retain(|a| a.name != anchor.name);
-                            layer.anchors.push(anchor);
-                        }
+                        // Add anchors from smart component, filtering out
+                        // 'mark' anchors used for internal part-to-part connections.
+                        let new_anchors: Vec<_> = instance
+                            .anchors
+                            .into_iter()
+                            .filter(|a| !a.name.starts_with('_'))
+                            .collect();
+                        // Remove existing anchors that will be replaced by new ones.
+                        // When duplicates exist, keep the last one (matching Glyphs.app).
+                        let new_names: HashSet<&SmolStr> =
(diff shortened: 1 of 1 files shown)


glyphs-reader/src/font.rs
```diff
@@ -3824,20 +3824,27 @@ impl Font {
...
+                        // Add anchors from smart component
+                        for anchor in instance.anchors {
+                            // Filter out 'mark' anchors used for internal part-to-part connections
+                            if anchor.name.starts_with('_') {
+                                continue;
+                            }
+                            // When duplicates exist, only keep the last one (matching Glyphs.app).
+                            layer.anchors.retain(|a| a.name != anchor.name);
+                            layer.anchors.push(anchor);
+                        }
```
cmyr: it's a bit accidentally quadratic (`retain` is iterating the full array, and we're inside another loop) but I  guess the value of N is unlikely to be high?

the alternative would be to first build a set containing the names of the anchors we will add, and then do a single `retain` up front that removes any existing anchors with those names from the layer.

cmyr (approved):
looks good!

glyphs-reader/src/font.rs
anthrotype: >  I guess the value of N is unlikely to be high?

indeed.. but I added the set you asked nonetheless, probably won't make much difference anyway

PR #1890, opened 2026-02-13
Preserve original anchor names through propagation

anthrotype:
After #1832 (propagate anchors in IR), fonts with non-standard caret anchor naming lost caret positions. For example, [ComforterBrush-Pro.glyphs](https://github.com/googlefonts/comforter-brush) defines two caret anchors on "ffi":

- `caret_1` at x=242
- `caret_1_` at x=588

`AnchorKind::new("caret_1_")` parses the suffix "1_", fails `parse::<usize>()`, and falls back to `Caret(1)`, i.e. the same variant as `caret_1`. When `to_name()` method converts both back to strings during anchor propagation, both become "caret_1" and one overwrites the other in the IndexMap.

Before #1832 this didn't matter because `fontbe` consumed `Vec<Anchor>` directly without round-tripping through names. After the merge, propagation serializes anchors via `AnchorKind::to_name()` at 

https://github.com/googlefonts/fontc/blob/ade0a9bc/fontir/src/propagate_anchors.rs#L87

... introducing the lossy conversion.

(fontmake/ufo2ft doesn't have this problem because it never parses caret indices, it just checks `name.startswith("caret_")` and collects positions.. apparently nobody cares about caret indices except me)

The fix is to add `original_name: SmolStr` field to `fontir::ir::Anchor` to preserve the original source name and use that instead of anchor.kind.to_name() when feeding anchors into propagation.

The existing `to_name()` is kept as a public method but documented as lossy.

The first commit adds the regression test that asserts both caret_1 and caret_1_ survive propagation (deliberately fails without the fix). It is followed by the actual fix.

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -1255,7 +1255,10 @@ impl AnchorKind {
         matches!(self, AnchorKind::Base(_) | AnchorKind::Ligature { .. })
     }
 
-    /// Convert AnchorKind back to its original string name representation
+    /// Convert AnchorKind back to its canonical string name representation.
+    ///
+    /// Note: this is lossy for non-standard names (e.g. `caret_1_` becomes `caret_1`).
+    /// Prefer `Anchor::original_name` when the original source name must be preserved.
     pub fn to_name(&self) -> SmolStr {
         match self {
             AnchorKind::Base(name) => name.clone(),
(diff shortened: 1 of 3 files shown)


fontir/src/ir.rs
```diff
@@ -1278,6 +1281,8 @@ impl AnchorKind {
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct Anchor {
     pub kind: AnchorKind,
+    /// The original anchor name from the source, preserved for round-tripping.
+    pub name: SmolStr,
```
rsheeter: Optional: I wonder if original_name would more clearly convey that this is very specifically the original name?

anthrotype: Done

PR #1892, opened 2026-02-16
Preserve explicit anchors during smart component instantiation

anthrotype:
When a glyph uses smart components and also has its own explicit anchors, the smart component's interpolated anchors should not override them. The glyph's explicit anchors represent the designer's intent and take precedence over anchors propagated from smart component parts. This is consistent with similar logic in propagate_anchors.rs.

This PR fixes anchor position regressions in fonts like NotoSansNandinagari where Indic conjuncts (e.g. `KaCa.Nnagari`) had hand-placed anchors that were incorrectly being replaced by interpolated smart component values.

I first push a failing regression test `smart_component_preserves_explicit_anchors` with dedicated test fixture, then follow up with the actual fix.

Verified against NotoSansNandinagari: mark/kern goes from 99.240% to identical with fontmake (glyphsLib 6.12.7)

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -3826,6 +3826,10 @@ impl Font {
         // convert the smart components to normal outlines, per-glyph
         for mut glyph in glyphs_with_smart_components {
             for layer in glyph.layers.iter_mut() {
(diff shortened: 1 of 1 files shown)


rsheeter (approved):
Appreciate the test comments!

Issue #1903, opened 2026-03-04
Support COLR compilation from UFO/.designspace sources


anthrotype:
fontc silently skips COLR/CPAL table generation for UFO/.designspace sources. The plumbing is in place: `ufo2fontir` implements the `Source` trait's `create_color_glyphs_work()` method, but the resulting `PaintGraphWork::exec()` is a stub that writes nothing to the context 

https://github.com/googlefonts/fontc/blob/393b797545f5841899d8bb9cd4b0f0f48257c882/ufo2fontir/src/source.rs#L1877

The backend COLR compiler (fontbe/src/colr.rs) then finds no paint graph data and silently returns.

The Glyphs.app frontend has a working implementation, so the backend compilation logic (COLR layers, CPAL palettes) already exists, the gap is just in extracting color layer data from UFO sources.

We should parse color layer data from UFO sources and populate the IR paint graph. At the very least, until we do that, we could emit a user-visible warning when color layer data is present in the source but cannot be compiled.

PR #1904, opened 2026-03-04
Support COLRv0 from UFO/.designspace sources

anthrotype:
Parse `com.github.googlei18n.ufo2ft.colorPalettes` and `com.github.googlei18n.ufo2ft.colorLayers` from the designspace lib.plist to build COLRv0  and CPAL tables from UFO sources, including support for multi-palette and foreground color (0xFFFF).

Note this only supports the pre-exploded color layer glyphs already in the default UFO layer (not yet also the `colorLayerMapping` variant with multiple UFO layers each mapped to a given color)

Part of fixing #1903

The first commit adds a test COLR ufo adapted from ufo2ft test suite (and a failing integration test). The actual commit with the new feature follows.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -4541,83 +4541,74 @@ mod tests {
         );
     }
 
-    fn assert_colr0(compile: &TestCompile, expected: Vec<(String, Vec<(String, u16)>)>) {
+    fn assert_colr0(compile: &TestCompile, expected: &[(&str, &[(&str, u16)])]) {
         let colr = compile.font().colr().unwrap();
         let layers = colr
             .layer_records()
             .expect("A layer list")
             .expect("A valid layer list");
         assert_eq!(0, colr.version());
-        assert_eq!(
-            expected,
-            colr.base_glyph_records()
-                .expect("v0 records")
-                .expect("Valid v0 records")
-                .iter()
-                .map(|g| (
+        let actual: Vec<(String, Vec<(String, u16)>)> = colr
+            .base_glyph_records()
+            .expect("v0 records")
+            .expect("Valid v0 records")
+            .iter()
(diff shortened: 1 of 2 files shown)


fontc/src/lib.rs
```diff
@@ -4595,6 +4595,54 @@ mod tests {
...
+    fn colr0_from_ufo() {
+        let result = TestCompile::compile_source("COLRv0-var/COLRv0-Regular.ufo");
+        assert_colr0(
+            &result,
+            vec![(
+                "a".to_string(),
+                vec![("a.color0".to_string(), 0), ("a.color1".to_string(), 1)],
+            )],
+        );
+    }
```
rsheeter, line 4605: I realize this mirrors the existing tests but in the light of a new day it's super noisy (all the to-strings, etc) . I wonder how we could reduce noise here? Maybe assert_colr0 could take a more flexible input type or it's second arg could have alternate formation or we could have a builder?

ufo2fontir/src/source.rs
```diff
@@ -1841,6 +1856,192 @@ impl Work<Context, WorkId, Error> for GlyphIrWork {
     }
 }
 
+/// Convert a float color component (0.0–1.0) to u8 (0–255).
+///
+/// Uses banker's rounding (round half to even) to match Python's `round(v * 255)`
+/// used in fontTools' `buildCPAL`.
```
rsheeter: Suggest an actual link

fontc/src/lib.rs
anthrotype: like this? 815b30963c890f3b2aac5cb16edf7fbcbe5e8353

rsheeter: I like it!

PR #1906, opened 2026-03-04
Make it easier to run benchmarks

wmedrano:
Makes it so that `cargo bench` "just works" without having to set up any font directory hierarchy.
- `cargo bench` will automatically download some fonts
- Benchmark can still be customized through environment variable if needed

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -5,8 +5,7 @@
 
 Only relatively large improvements or regressions are effectively detected this
 way: fontc uses [criterion](https://crates.io/crates/criterion) for
-benchmarking. To run the benchmarks, use the following command to run
-`fontc/benches/compile.rs`:
+benchmarking. Use the following command to run `fontc/benches/compile.rs`:
 
 ```shell
 cargo bench -p fontc --bench compile
--- a/fontc/benches/compile.rs
+++ b/fontc/benches/compile.rs
@@ -19,20 +19,18 @@ struct FontSource<'a> {
 }
 
 const DEFAULT_FONT_SOURCES: &[FontSource<'static>] = &[
+    // Compiles quickly.
     FontSource {
         dir: "/tmp/fontc-bench/oswald",
         git_url: Some("https://github.com/googlefonts/OswaldFont.git"),
         source_path: "sources/Oswald.glyphs",
     },
+    // May take a few seconds. Based on threads.svg from --emit-timing, bottlenecked on kern-be,
+    // kern-gather-be, fea, and font.
     FontSource {
-        dir: "/tmp/fontc-bench/googlesans",
-        git_url: Some("git@github.com:googlefonts/googlesans.git"),
-        source_path: "source/GoogleSans/GoogleSans.designspace",
-    },
-    FontSource {
-        dir: "/tmp/fontc-bench/googlesans-flex",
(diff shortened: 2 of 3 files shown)


.github/workflows/rust.yml
```diff
@@ -84,6 +84,12 @@ jobs:
   test-stable-fontc:
     name: fontc tests
     runs-on: ubuntu-latest
+    env:
+      # Disables the built-in benchmarks since:
+      # 1. The benchmark is slow to run on CI
+      # 2. The CI environment doesn't have access to git clone the benchmark's
+      #    default set of font sources.
+      FONTC_BENCH_SRCS: ""
```
cmyr, line 92: do the benchmarks run when we call `cargo test`? Or is it just that we try to build them and that fails?

docs/performance.md
```diff
@@ -3,45 +3,46 @@
 ## Benchmarking with Criterion
 
 
-Only relatively large changes are effectively detected this way: fontc uses
-[criterion](https://crates.io/crates/criterion) for benchmarking. To run the
-benchmarks, use the following command:
+Only relatively large improvements or regressions are effectively detected this
+way: fontc uses [criterion](https://crates.io/crates/criterion) for
+benchmarking. To run the benchmarks, use the following command to run
+`fontc/benches/compile.rs`:
```
cmyr:
```suggestion
benchmarking. Use the following command to run `fontc/benches/compile.rs`:
```

fontc/benches/compile.rs
```diff
@@ -1,23 +1,42 @@
...
+
+const DEFAULT_FONT_SOURCES: &[FontSource<'static>] = &[
+    FontSource {
+        dir: "/tmp/fontc-bench/oswald",
+        git_url: Some("https://github.com/googlefonts/OswaldFont.git"),
+        source_path: "sources/Oswald.glyphs",
+    },
+    FontSource {
+        dir: "/tmp/fontc-bench/googlesans",
+        git_url: Some("git@github.com:googlefonts/googlesans.git"),
```
cmyr: If we're doing this auto-download  business maybe we don't want to use private repos?

wmedrano: Replaced with merriweather4 which is slow and seems to have the same bottlenecks.

docs/performance.md
wmedrano: Done.

.github/workflows/rust.yml
wmedrano: `--all-targets` triggers the benchmarks to run. Seems like each case runs only once without optimizations enabled, which is disastrous for the slow fonts

wmedrano:
> This does raise a general question for me, though, which is whether these benchmarks make a ton of sense? My impression is that critereon is designed more for micro benchmarks, but I haven't dug into that very much.

So having the one liner is nice as I'm reviving some perf oriented changes.

But I agree with criterion not being a good fit. The time scale does not seem to fit (hence the max `10` iterations hardcoded), but its noisy due to being highly multithreaded. I was thinking of de-criterionizing it. Any ideas?

- one-off shell script
- example program, similar to shell script, but I prefer writing Rust
- Part of fontc_crater. Could make sense, but maybe fontc crater wants to focus on diffs

PR #1908, opened 2026-03-06
Optimize coords representation

wmedrano:
- Use `Vec` instead of `BTreeMap`
  - Some time ago I also tried SmallVec, but it didn't outperform `Vec`
- ~7% performance improvement on googlesans-flex with 6 variational axes.
  - Goes to ~8.5% improvement if `from_iter` can assume uniqueness of inputs
  - Smaller improvements on fonts with fewer axes

| Font | Baseline | New | Change |
|------|----------|-----|--------|
| Oswald | 44.101 ms | 42.728 ms | −3.11% |
| Merriweather | 4.9451 s | 4.9399 s | −0.11% |
| GoogleSans | 1.8709 s | 1.8097 s | −3.27% |
| GoogleSansFlex | 3.5438 s | 3.2881 s | −7.22% |

### DIFF
--- a/fontdrasil/src/coords.rs
+++ b/fontdrasil/src/coords.rs
@@ -110,6 +110,8 @@ impl<Space> Coord<Space> {
 /// E.g. a user location is a `Location<UserSpace>`. Hashable so it can do things like be
 /// the key for a map of sources by location.
 #[derive(Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+// Performance: using a Vec instead of BTreeMap is about 7% faster. See:
+// https://github.com/googlefonts/fontc/pull/1908
 pub struct Location<Space>(Vec<(Tag, Coord<Space>)>);
 
 /// A location in [`DesignSpace`].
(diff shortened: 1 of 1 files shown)


fontdrasil/src/coords.rs
```diff
@@ -343,36 +352,49 @@ impl<Space> Location<Space> {
...
-        axes.len() == self.0.len() && axes.iter().all(|tag| self.0.contains_key(tag))
+        axes.len() == self.0.len() && axes.iter().all(|tag| self.contains(*tag))
     }
 
     pub fn insert(&mut self, tag: Tag, pos: Coord<Space>) -> &mut Location<Space> {
-        self.0.insert(tag, pos);
+        match self.0.binary_search_by_key(&tag, |(t, _)| *t) {
+            Ok(i) => {
+                if let Some(entry) = self.0.get_mut(i) {
+                    entry.1 = pos
```
anthrotype: nit: when `binary_search` returns `Ok(i)`, the index is guaranteed valid. `self.0[i].1 = pos` would be sufficient and avoids a dead branch

https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search_by_key

> If the value is found then `Result::Ok` is returned, containing the index of the matching element.

fontdrasil/src/coords.rs
```diff
@@ -343,36 +352,49 @@ impl<Space> Location<Space> {
...
+        self.0.binary_search_by_key(&tag, |(t, _)| *t).is_ok()
     }
 
     pub fn get(&self, tag: Tag) -> Option<Coord<Space>> {
-        self.0.get(&tag).copied()
+        self.0
+            .binary_search_by_key(&tag, |(t, _)| *t)
+            .ok()
+            .and_then(|i| self.0.get(i))
+            .map(|(_, v)| *v)
```
anthrotype: same here, Ok(i) guarantees validity, you can skip the redundant destructure chain

```suggestion
            .map(|i| self.0[i].1)
```

fontdrasil/src/coords.rs
wmedrano: Done, needed to add `#[allow(clippy::indexing_slicing)]` to appease the linter, but it does look cleaner than .get

PR #1910, opened 2026-03-06
Update README

wmedrano:


README.md
```diff
@@ -45,7 +45,14 @@ Install the latest version of Rust, https://www.rust-lang.org/tools/install.
 ### Build a simple test font
 
 ```shell
-$ cargo run -p fontc -- resources/testdata/wght_var.designspace
+# Build a .designspace file
+$ cargo run --release -p fontc -- resources/testdata/wght_var.designspace
```
cmyr: curious why you pass --release in all of these? feels a bit like noise to me but also definitely not a strong opinion

cmyr (approved):
Thanks, this was definitely overdue!

README.md
wmedrano: Ack, reverted release in most places.

For someone without the know, seems safer. Though, it's not as bad as I thought, GSF only takes 5 minutes on this machine without release mode

Issue #542, opened 2023-11-03
fea-rs: support multiple substitution with classes


cmyr:
This is something that feaLib added a little while back, and we should try to match. Without this, we are not matching feaLib's output for the `GSUB_2.fea` test case.

PR #1913, opened 2026-03-07
[fea-rs] Allow classes in inline contextual multiple sub rules

khaledhosny:
This is a follow up to commit 51d33ad7103bf0a5fc1b1865b6d40e86f419a0bd, to allow classes also in inline multiple substitution rules.

This enables the corresponding feaLib test that is now passing, and moves similar test from expected to fail to expected to pass tests.

* Fixes https://github.com/googlefonts/fontc/issues/542

### DIFF
--- a/fea-rs/src/parse/grammar/gsub.rs
+++ b/fea-rs/src/parse/grammar/gsub.rs
@@ -122,11 +122,11 @@ fn finish_chain_rule(parser: &mut Parser, recovery: TokenSet) -> AstKind {
     if parser.eat(Kind::ByKw) {
         if parser.eat(Kind::NullKw) {
             // allowed, continue down to 'expect_semi'
-        } else if glyph::eat_glyph_or_glyph_class(parser, recovery) {
+        } else if glyph::expect_glyph_or_glyph_class(parser, recovery) {
             while glyph::eat_glyph_or_glyph_class(parser, recovery) {
                 continue;
             }
-        } else if !glyph::expect_named_or_unnamed_glyph_class(parser, recovery) {
+        } else {
             // unexpected thing here?
             parser.eat_until(recovery);
             parser.eat(Kind::Semi);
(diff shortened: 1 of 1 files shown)


fea-rs/src/parse/grammar/gsub.rs
```diff
@@ -122,8 +122,8 @@ fn finish_chain_rule(parser: &mut Parser, recovery: TokenSet) -> AstKind {
     if parser.eat(Kind::ByKw) {
         if parser.eat(Kind::NullKw) {
             // allowed, continue down to 'expect_semi'
-        } else if glyph::eat_glyph_name_like(parser) {
-            while glyph::eat_glyph_name_like(parser) {
+        } else if glyph::eat_glyph_or_glyph_class(parser, recovery) {
+            while glyph::eat_glyph_or_glyph_class(parser, recovery) {
                 continue;
             }
         } else if !glyph::expect_named_or_unnamed_glyph_class(parser, recovery) {
```
anthrotype: isn't this branch now unreachable since the previous one handles both glyphs and classes?

khaledhosny: I’m not really sure. The first branch handles `NULL`, the second branch used to handle `glyph` and now also `[glyph]`, and `@class`. What was the the case that was supposed to be handled by the third branch?

anthrotype: i think before your PR, the second and third branches were complementary (2 handled bare glyphs, 3 handled named/unnamed classes); now that the branch 2 you modified handles _both_ glyphs and classes, branch 3 is effectively dead for valid inputs; it can only ever trigger its error recovery (when it returns false, the `eat_until(recovery)` to handle maformed input), because eat_glyph_or_glyph_class already tried everything that expect_named_or_unnamed_glyph_class would try.

khaledhosny: OK, so something like this to handle the error case?
```suggestion
        } else {
```

anthrotype: yeah, plus something like this

```
parser.err("Expected glyph name or glyph class");
```

to mimic the error message that expect_named_or_unnamed_glyph_class would have emitted

khaledhosny: Instead of duplicating the error message, I used `expect_glyph_or_glyph_class()` in the second branch. I think it is clearer now as well.

anthrotype: thanks

Issue #1897, opened 2026-02-22
Exposing insert markers


khaledhosny:
Fontra (not in the main branch yet) uses fontc to compile a so-called shaper font to pass to HarfBuzz for shaping. This is a minimal font with only what is absolutely needed for shaping. The code is in https://github.com/fontra/build-shaper-font.

One thing that Fontra does to avoid having to compile the shaper font for each source font change is that it applies glyph positioning itself. Fontra tries to do this in a way that keeps the output identical to shaping a full font. This is simple if the source font does not have any glyph positioning feature code, but if it does, this complicates things as the source font might contain insertion markers (the `# Automatic code` comment) and it might have code that must be applied after the generated lookups.

So to keep the shaping behavior unchanged between the full font and the minimal font, Fontra now needs to know where the generated lookups would have been inserted, so that it does its glyph positioning at the same point.

fea-rs already maintains insertion markers that tell for each feature the lookup id before which the generated keep will be inserted, but these insertion markers are not exposed outside of fea-rs.

To test this approach, I made changes to fea-rs to expose insertion markers and a few other bits to allow me to get them in build-shaper-font. The changes are here: https://github.com/googlefonts/fontc/compare/main...khaledhosny:fontc:insert-markers

I didn’t open a PR right away because I’m not sure if the maintainers are open to such a change or if they would prefer a different approach. Please let me know what you think and whether I should open a PR with the changes I did instead.

justvanrossum:
There is a new Fontra release that includes this functionality. I wrote a blog post about it:

- https://blog.fontra.xyz/blog/opentype-harfbuzz/

Near the end I try to explain what the insert markers are for, and why we need them.

I am very pleased with how this system overall works (even though it is still imperfect in some darker corners), and I would love to see (the equivalent of) the changes that Khaled made to be integrated back into the fontc upstream.

If the approach can be improved for easier integration, we'd love to hear. This is incredibly useful and powerful stuff, and it would be a shame to keep it hidden as unexposed internals.

PR #1915, opened 2026-03-10
Expose insert markers through Compilation result

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/1897

Surface the `insert_markers` HashMap on the public `Compilation` struct, allowing consumers to know where generated lookups (kern, mark, etc.) would be inserted relative to hand-written feature code.

This is needed by Fontra's [build-shaper-font](https://github.com/fontra/build-shaper-font), which emulates generated GPOS features for interactive editing and needs to know the correct insertion points to maintain shaping fidelity with fully compiled fonts.

Khaled's original approach (cf. https://github.com/googlefonts/fontc/compare/main...khaledhosny:fontc:insert-markers) exposes the entire low-level compilation pipeline (CompilationCtx) so consumers can read ctx.insert_markers between compile() and build().
Whereas in here I kept that pipeline private and instead thread the markers into the existing Compilation result, so consumers use the standard compile::compile() function and read compilation.insert_markers from the output.

The public API surface is narrower and I verified locally that it still works for Fontra's use case (read-only, after compilation), with relatively little changes to `build-shaper-font`.

@khaledhosny @justvanrossum let me know if this would work for you

### DIFF
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -286,7 +286,7 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
                 gpos,
                 opts: self.opts.clone(),
                 gdef_classes,
-                insert_markers: std::mem::take(&mut self.insert_markers),
+                insert_markers: self.insert_markers.clone(),
             },
             self.errors.clone(),
         ))


fea-rs/src/compile/compile_ctx.rs
```diff
@@ -286,6 +286,7 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
                 gpos,
                 opts: self.opts.clone(),
                 gdef_classes,
+                insert_markers: std::mem::take(&mut self.insert_markers),
```
cmyr: let's clone here instead of take... the intention of this design (which might not make sense, but is what it is) is that calling `compile` doesn't change the state of the ctx, so in theory you could make small edits and then compile again.

anthrotype: Done

PR #1922, opened 2026-03-12
[fea-rs] Merge consecutive contextual rules with same context

anthrotype:
This recovers the 33 GSUB-only crater entries that regressed from identical after the fonttools 4.62.0 requirements bump in #1916 (the crater showed a net -17 identicals because 18 unrelated improvements partially offset the 33 GSUB regressions).

It merges consecutive contextual rules in `ContextBuilder::add()` when they share the same backtrack, lookahead, and nested lookup references and differ only in the input glyph at a single position. This merged rules often result in smaller ChainContext{Pos,Subst} Format 3 instead of Format 1 or 2.

This matches fonttools/fonttools#4058 (released in 4.62.0 for inline multi-subst only, fixed https://github.com/fonttools/fonttools/issues/4016), but extends it further to apply the same merging to _all_ contextual rule types: inline multi-subst, inline single-subst, named GSUB/GPOS lookup references, and ignore rules.

I also have a matching fonttools PR almost ready to be pushed which extends the merge to all contextual rule types, so both compilers produce identical output. Which means we will need to coordinate the fonttools release and crater requirements update.

### DIFF
--- a/fea-rs/test-data/compile-tests/mini-latin/good/contextual_format_2.ttx
+++ b/fea-rs/test-data/compile-tests/mini-latin/good/contextual_format_2.ttx
@@ -44,46 +44,78 @@
       <Lookup index="0">
         <LookupType value="6"/>
         <LookupFlag value="16"/><!-- useMarkFilteringSet -->
-        <!-- SubTableCount=2 -->
-        <ChainContextSubst index="0" Format="3">
-          <!-- BacktrackGlyphCount=0 -->
-          <!-- InputGlyphCount=1 -->
-          <InputCoverage index="0">
+        <!-- SubTableCount=1 -->
+        <ChainContextSubst index="0" Format="2">
+          <Coverage>
             <Glyph value="a"/>
             <Glyph value="i"/>
-          </InputCoverage>
-          <!-- LookAheadGlyphCount=1 -->
-          <LookAheadCoverage index="0">
-            <Glyph value="acutecomb"/>
-            <Glyph value="brevecomb"/>
-          </LookAheadCoverage>
-          <!-- SubstCount=1 -->
-          <SubstLookupRecord index="0">
-            <SequenceIndex value="0"/>
-            <LookupListIndex value="1"/>
-          </SubstLookupRecord>
-        </ChainContextSubst>
-        <ChainContextSubst index="1" Format="3">
-          <!-- BacktrackGlyphCount=0 -->
-          <!-- InputGlyphCount=1 -->
-          <InputCoverage index="0">
-            <Glyph value="a"/>
-            <Glyph value="i"/>
-          </InputCoverage>
-          <!-- LookAheadGlyphCount=2 -->
-          <LookAheadCoverage index="0">
-            <Glyph value="ogonekcomb"/>
-            <Glyph value="dotbelowcomb"/>
-          </LookAheadCoverage>
-          <LookAheadCoverage index="1">
-            <Glyph value="acutecomb"/>
-            <Glyph value="brevecomb"/>
-          </LookAheadCoverage>
-          <!-- SubstCount=1 -->
-          <SubstLookupRecord index="0">
-            <SequenceIndex value="0"/>
-            <LookupListIndex value="1"/>
-          </SubstLookupRecord>
+          </Coverage>
+          <BacktrackClassDef>
+          </BacktrackClassDef>
+          <InputClassDef>
+            <ClassDef glyph="a" class="1"/>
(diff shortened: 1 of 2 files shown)


fea-rs/src/common.rs
```diff
@@ -96,6 +96,14 @@ impl GlyphOrClass {
         }
     }
 
+    /// Combine the glyphs from `other` into this value.
+    ///
+    /// After this call, `self` contains the union of glyphs from both operands
+    /// as a `Class` variant.
```
cmyr: this isn't exactly true, because a FEA glyph class is not a set, it's just an ordered sequence; see the docs for `GlyphClass`, above.

fea-rs/test-data/fonttools-tests/bug463.ttx
```diff
@@ -32,7 +32,7 @@
       <Lookup index="0">
         <LookupType value="6"/>
         <LookupFlag value="0"/>
-        <!-- SubTableCount=2 -->
+        <!-- SubTableCount=1 -->
```
cmyr, line 35: Note that these test cases are directly imported from fonttools, and there's a script to update them. I notice the test cases above are also added to fonttools? So one option would be to remove those test cases here, we merge the fonttools PR, and then we can run the update script to bring in the new fonttools tests? I can walk you through that if needed.

fea-rs/src/compile/lookups/contextual.rs
```diff
@@ -304,6 +304,21 @@ impl ContextBuilder {
...
+        // Format 3 with fewer subtables).  Inspired by fonttools PR #4058.
+        if context.len() == 1
+            && let Some(last) = self.rules.last_mut()
+            && last.context.len() == 1
+            && last.backtrack == backtrack
+            && last.lookahead == lookahead
+            && last.context[0].1 == context[0].1
+        {
+            last.context[0].0.extend(&context[0].0);
+            return;
```
cmyr: i could also imagine a method on `ContextRule` that would handle checking for compatibility and merging if possible, in which case this would all look something like,

```rust
let rule = ContextRule { backtrack, context, lookahead };
if self.rules.last_mut().map(|last| last.try_merge(&rule)) != Some(true) {
    self.rules.push(rule);
}
```

fea-rs/test-data/fonttools-tests/bug463.ttx
anthrotype: I know. I cheated and copied the updated tests from the other fonttools PR. They will match when we eventually sync them using that script but you're right it would probably cleaner to skip them and then sync them later

anthrotype: I reverted the manual changes to fonttools-tests/ in https://github.com/googlefonts/fontc/pull/1922/commits/fb428f1bd38019521c214b23e3aa95ab1146d949

we shall update those separately as you requested

anthrotype: fonttools-tests are now in sync with latest fonttools 4.62.1 ([910a2e8](https://github.com/googlefonts/fontc/pull/1922/commits/910a2e89f5b25930f30347b181c6ba4228ce5bd3) force-pushed)

anthrotype: and yes, I did use your shell script :)

anthrotype: and I also had to bump the fonttools in fea_ci_requirements.txt (something to remember next time these fonttools-tests get updated)

fea-rs/test-data/compile-tests/mini-latin/good/contextual_format_2.ttx
```diff
@@ -44,78 +44,46 @@
       <Lookup index="0">
         <LookupType value="6"/>
         <LookupFlag value="16"/><!-- useMarkFilteringSet -->
-        <!-- SubTableCount=1 -->
-        <ChainContextSubst index="0" Format="2">
-          <Coverage>
+        <!-- SubTableCount=2 -->
+        <ChainContextSubst index="0" Format="3">
```
cmyr: this is unfortunate, this test is supposed to be for format 2. Can we open an issue to fix it?

anthrotype: fixed [4e96bed](https://github.com/googlefonts/fontc/pull/1922/commits/4e96bedd90a4ea8d5b810354bbb93b7116a6adb2)

Issue #1886, opened 2026-02-10
[fea-rs] store lookup origin for feature code debugging


simoncozens:
FontTools's feaLib has the ability to generate a `Debg` table which stores source-level information (lookup name, filename, line number) about each lookup. This is helpful for editors and debuggers to pinpoint the effect of lookups back to feature code source. Something similar (ideally, something compatible), would be useful.

cmyr:
huh I've never heard of this before, is it documented anywhere? Is it a feaLib-specific thing?

PR #1933, opened 2026-03-25
[fea-rs] store lookup origin for feature code debugging

khaledhosny:
Store lookup name and location info in `Debg` table following the same structure as `fontTools.feaLib`. But unlike `feaLib`, we don’t store feature info, because:
1) it is redundant when debugging with HarfBuzz trace message as we get the script and feature info from there
2) it is misleading/incomplete as the lookup can be registered under different script/language/feature compos, but `feaLib` only reports one of them.

* Fixes: #1886

### DIFF
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -51,7 +51,7 @@ use super::{
     language_system::{DefaultLanguageSystems, LanguageSystem},
     lookups::{AllLookups, FilterSetId, LookupFlagInfo, LookupId, SomeLookup},
     output::Compilation,
-    tables::{GlyphClassDefExt, LookupDebugInfo, ScriptRecord, Tables},
+    tables::{GlyphClassDefExt, ScriptRecord, Tables},
     tags,
 };
 
@@ -104,8 +104,6 @@ pub struct CompilationCtx<'a, F: FeatureProvider, V: VariationInfo> {
     // and we will use that for the generated lookups.
     // We also store the start pos of the comment, to break ties.
     insert_markers: HashMap<Tag, InsertionPoint>,
-    /// Pending debug info for the next lookup to be started (set in named lookup blocks).
-    pending_lookup_debug_info: Option<LookupDebugInfo>,
 }
 
 impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
@@ -126,7 +124,7 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
             tables: Tables::default(),
             default_lang_systems: Default::default(),
             glyph_class_defs: Default::default(),
-            lookups: Default::default(),
+            lookups: AllLookups::new(opts.compile_debg),
             features: Default::default(),
             mark_classes: Default::default(),
             anchor_defs: Default::default(),
@@ -140,7 +138,6 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
             mark_filter_sets: Default::default(),
             opts,
             insert_markers: Default::default(),
-            pending_lookup_debug_info: Default::default(),
         }
     }
 
@@ -298,11 +295,7 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
                 opts: self.opts.clone(),
                 gdef_classes,
                 insert_markers: self.insert_markers.clone(),
-                debg: self
-                    .tables
-                    .debg
-                    .as_ref()
-                    .map(|d| d.build(&self.tree.sources)),
+                debg: self.tables.debg.as_ref().map(|d| d.build(self.tree)),
             },
             self.errors.clone(),
         ))
@@ -448,17 +441,7 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
         }
 
         self.vertical_feature.begin_lookup_block();
-        if self.opts.compile_debg {
-            // Record location and name now; the LookupId is unknown until the
-            // first rule is added, so we stash it as pending debug info and
-            // consume it in ensure_current_lookup_type.
-            let (file_id, local_range) = self.tree.source_map().resolve_range(name.range());
-            self.pending_lookup_debug_info = Some(LookupDebugInfo {
-                location: (file_id, local_range.start),
-                name: Some(name.text.clone()),
-            });
-        }
-        self.lookups.start_named(name.text.clone());
+        self.lookups.start_named(name.text.clone(), name.range());
     }
 
     fn end_lookup_block(&mut self) {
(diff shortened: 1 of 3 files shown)


fea-rs/src/compile/tables/debg.rs
```diff
@@ -0,0 +1,96 @@
+//! Building the `Debg` debug table.
+//!
+//! The `Debg` table stores source-level information about each lookup,
+//! compatible with fontTools.feaLib format, but unlike feaLib, we always set
+//! feature to null
```
cmyr: if we  could provide a link to the other implementation that could be useful?

fea-rs/src/compile/tables/debg.rs
```diff
@@ -0,0 +1,96 @@
...
+        source_list: Arc<SourceList>,
+    ) -> Self {
+        Self {
+            gsub,
+            gpos,
+            source_list,
+        }
+    }
+
+    pub(crate) fn build(&self) -> Vec<u8> {
```
cmyr: hand rolling the json feels a bit bad, but it also feels a bit bad to make serde_json a hard dependency, so I guess this is fine?

fea-rs/src/compile/tables/debg.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub location: (FileId, usize),
+    pub name: Option<String>,
+}
+
+/// Builder for the `Debg` table.
+#[derive(Clone, Debug)]
+pub(crate) struct DebgBuilder {
+    gsub: Vec<Option<LookupDebugInfo>>,
+    gpos: Vec<Option<LookupDebugInfo>>,
+    source_list: Arc<SourceList>,
```
cmyr: instead of owning a copy of this (which requires us to change API in a bunch of other places) could we just pass it in to the `build` method?

fea-rs/src/compile/compile_ctx.rs
```diff
@@ -432,6 +445,16 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
...
         self.vertical_feature.begin_lookup_block();
+        if self.opts.compile_debg {
+            // Record location and name now; the LookupId is unknown until the
+            // first rule is added, so we stash it as pending debug info and
+            // consume it in ensure_current_lookup_type.
+            let (file_id, local_range) = self.tree.source_map().resolve_range(name.range());
+            self.pending_lookup_debug_info = Some(LookupDebugInfo {
+                location: (file_id, local_range.start),
+                name: Some(name.text.to_string()),
+            });
```
cmyr: so I _think_ I have a possible simplification here, since as far as I understand it what we're doing here is already expressed in existing control flow related to names.

My understanding:

- when we encounter a new name, we are stashing the location of that token.
- later, when we actually start a lookup, we are trying to associate the new lookup ID with the location

If this is true, it sounds very much like how we already handle named lookups: we stash the name when we encounter it, and then anytime we finish a lookup we see if there was a stashed name, and associate the two.

Would that work here? To the best of my understanding that's really all we're trying to do, and it feels like a nice simplification.

basic impl:

- stop storing file/pos, and just store the token range, we can resolve it later
- add a `range` argument to `lookups.start_named`
- in `lookups.finish_current`, update the debug info directly
- resolve ranges to positions in the file only at the end, during compile?

it's possible I've overlooked something, but I _think_ this works (and we could trivially make it also work for anonymous lookups, and could even provide full range information for those lookups) and I think it's simpler.

fea-rs/src/compile/compile_ctx.rs
```diff
@@ -581,14 +604,27 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
         }
     }
 
-    fn ensure_current_lookup_type(&mut self, kind: Kind) -> &mut SomeLookup {
+    fn ensure_current_lookup_type(&mut self, kind: Kind, range: Range<usize>) -> &mut SomeLookup {
```
cmyr, line 593: okay, so this is for handling anonymous lookups? I was hoping this wouldn't be necessary but I think it makes sense.

fea-rs/src/compile/compile_ctx.rs
```diff
@@ -581,14 +604,27 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
...
+                let debug_info = self.pending_lookup_debug_info.take().unwrap_or_else(|| {
+                    let (file_id, local_range) =
+                        self.tree.source_map().resolve_range(range.clone());
+                    LookupDebugInfo {
+                        location: (file_id, local_range.start),
+                        name: None,
+                    }
+                });
+                self.lookups.set_current_debug(debug_info);
+            }
```
cmyr: I think we could simplify all of this by just having `lookups.start_lookup` take a range; this should be cheap enough that we can just always do it, and then we can ignore that range if there was an existing range from a named lookup.

we _would_ want `AllLookups` to be passed the `ops.compile_debg` flag, though, because we would still skip actually stashing the debug info if the flag wasn't set.

fea-rs/src/compile/tables/debg.rs
```diff
@@ -0,0 +1,96 @@
...
+const LOOKUP_DEBUG_INFO_KEY: &str = "com.github.fonttools.feaLib";
+
+/// Debug information for a single lookup.
+///
+/// The source location is stored as a raw `(FileId, local_offset)` pair,
+/// resolved to a `"file:line:col"` string when building the final Debg bytes.
+#[derive(Debug, Clone)]
+pub(crate) struct LookupDebugInfo {
+    pub location: (FileId, usize),
+    pub name: Option<String>,
```
cmyr: This can be `SmolStr`, which saves us an allocation in many cases, and is already the type you get from the token.

fea-rs/src/compile/tables/debg.rs
khaledhosny: currently serde_json is only a test dependency, so I didn’t want to introduce it as a build dependency. I don’t like that we are writing json by hand either, so it is your call.

fea-rs/src/compile/tables/debg.rs
khaledhosny: I think I had it this way at some point, but I had to pass it around a bit, then code went through a few iteration and indeed it looks like the copy is not needed.

fea-rs/src/compile/compile_ctx.rs
```diff
@@ -32,7 +32,7 @@ use write_fonts::{
 use crate::{
     Diagnostic, GlyphIdent, GlyphMap, Kind, NodeOrToken, Opts,
     common::{GlyphClass, GlyphId16, GlyphOrClass, GlyphSet, MarkClass},
-    parse::SourceMap,
+    parse::ParseTree,
```
cmyr, line 35: is this change still necessary?

cmyr (commented):
Okay I think there's one little relic of a previous version of this patch that i've pointed out inline, but otherwise I'm happy to merge. We can keep the inline json for now. 🤷

fea-rs/src/compile/compile_ctx.rs
khaledhosny: `DebgBuilder::build()` needs both the `SourceList` and the `SourceMap` to resolve locations, unless I missed a way to do that with `SourceMap` only.

justvanrossum:
> I considered diverting from the current Debg table design, but I think @justvanrossum prefers keeping things compatible with FontTools version for now.

I have no strong opinion, since Fontra’s main use case is to debug fonts it built itself (with build-shaper-font). I see it as nice-to-have if we can show debug info for fonts built with FontTools. 

But mostly I would like to avoid supporting two incompatible Debg formats, so if you decide to diverge, I may just not support the FontTools version.

fea-rs/src/compile/compile_ctx.rs
cmyr: makes sense, thanks!

PR #1935, opened 2026-03-28
[fontbe] Expose FeatureVariationsProvider and FeaVariationInfo

khaledhosny:
~~Make fontbe’s `FeatureVariationsProvider` public and add a public `make_feature_variations_provider()` that builds a `FeatureVariationsProvider` without needing `StaticMetadata`.~~

Make `FeatureVariationsProvider` and `FeaVariationInfo` public. Also turn `make_gsub_feature_variations` into `FeatureVariationsProvider::new`.

This is to support building feature variations from rules in clients that don’t use the full fontc.

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -9,7 +9,7 @@ use std::{
     sync::Arc,
 };
 
-use feature_variations::FeatureVariationsProvider;
+pub use feature_variations::FeatureVariationsProvider;
 use log::{debug, error, trace, warn};
 use ordered_float::OrderedFloat;
 
@@ -50,7 +50,7 @@ use crate::{
     paths::Paths,
 };
 
-pub mod feature_variations;
+mod feature_variations;
 mod kern;
 mod marks;
 mod ot_tags;
--- a/fontbe/src/features/feature_variations.rs
+++ b/fontbe/src/features/feature_variations.rs
@@ -27,7 +27,7 @@ pub struct FeatureVariationsProvider {
 }
 
 impl FeatureVariationsProvider {
-    // https://github.com/fonttools/fonttools/blob/a1c189bda7c8d970143f92b542f5e26759945a37/Lib/fontTools/varLib/__init__.py#L799
+    // https://github.com/fonttools/fonttools/blob/a1c189bda/Lib/fontTools/varLib/__init__.py#L799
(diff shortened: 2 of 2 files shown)


khaledhosny:
I’m using this in https://github.com/fontra/build-shaper-font/pull/21

fontbe/src/features/feature_variations.rs
```diff
@@ -19,64 +19,67 @@ use crate::error::Error;
...
-            let mut space = NBox::default();
-            for condition in conditions {
-                let axis = static_metadata
-                    .axis(&condition.axis)
-                    .expect("checked already");
-                let min = condition.min.map(|min| min.to_normalized(&axis.converter));
-                let max = condition.max.map(|max| max.to_normalized(&axis.converter));
-                space.insert(condition.axis, min, max);
+impl FeatureVariationsProvider {
+    // https://github.com/fonttools/fonttools/blob/a1c189bda7c8d970143f92b542f5e26759945a37/Lib/fontTools/varLib/__init__.py#L799
```
cmyr: just fyi (and I appreciate this is just copy/paste, and you probably know this!) but you can arbitrarily shorten the SHA in a comment and the link will still resolve on github, e.g. this can be
```suggestion
    // https://github.com/fonttools/fonttools/blob/a1c189bda/Lib/fontTools/varLib/__init__.py#L799
```

fontbe/src/features.rs
```diff
@@ -50,7 +50,7 @@ use crate::{
     paths::Paths,
 };
 
-mod feature_variations;
+pub mod feature_variations;
```
cmyr: instead of making the module public I'd prefer to just reexport the type that you want.

cmyr (approved):
Okay, I mean... this definitely wasn't intended as public API but if you find it useful and understand that it might break at some point in the future than this works?

fontbe/src/features.rs
khaledhosny: Done.

fontbe/src/features/feature_variations.rs
khaledhosny: I indeed just copied the old comment, but it now uses the short version.

cmyr (approved):
thanks!

PR #1941, opened 2026-04-10
[ttx_diff] Faster off-by-one budget picking

cmyr:
In cases where there are a large number of total glyph points (fonts where glyphs have lots of points, or there are just a lot of glyphs) generating the list of all coordinates in order to then call `len` was proving to be the main place we were spending our time.

Instead of counting _all_ points, we now just count the points in the first ten non-empty glyphs and assume that they are representative.

This brings the compare time for rubik pixels from ~8 minutes to ~2 minutes, and shows a similar speedup for a bunch of other fonts, particularly CJK.

### DIFF
--- a/ttx_diff/src/ttx_diff/core.py
+++ b/ttx_diff/src/ttx_diff/core.py
@@ -736,17 +736,19 @@ def allow_fontc_only_variations_postscript_prefix(fontc, fontmake):
 
 
 def allow_some_off_by_ones(fontc, fontmake, container, name_attr, coord_holder):
-    fontmake_num_coords = len(fontmake.xpath(f"//{container}/{coord_holder}"))
+    fontmake_containers = fontmake.xpath(f"//{container}")
+    coord_tag = coord_holder.rpartition("/")[-1]
+    fontmake_num_coords = sum(
+        sum(1 for _ in c.iter(coord_tag)) for c in fontmake_containers
+    )
     off_by_one_budget = int(FLAGS.off_by_one_budget / 100.0 * fontmake_num_coords)
     spent = 0
     if off_by_one_budget == 0:
         return
 
-    coord_tag = coord_holder.rpartition("/")[-1]
     # put all the containers into a dict to make querying more efficient:
-
     fontc_items = {x.attrib[name_attr]: x for x in fontc.xpath(f"//{container}")}
-    for fontmake_container in fontmake.xpath(f"//{container}"):
+    for fontmake_container in fontmake_containers:
         name = fontmake_container.attrib[name_attr]
         fontc_container = fontc_items.get(name)
         if fontc_container is None:


ttx_diff/src/ttx_diff/core.py
```diff
@@ -735,18 +735,36 @@ def allow_fontc_only_variations_postscript_prefix(fontc, fontmake):
...
+        return 0
+    estimated_total = int(
+        sum(sample_counts) / len(sample_counts) * len(fontmake_containers)
+    )
+    return int(FLAGS.off_by_one_budget / 100.0 * estimated_total)
+
+
 def allow_some_off_by_ones(fontc, fontmake, container, name_attr, coord_holder):
-    fontmake_num_coords = len(fontmake.xpath(f"//{container}/{coord_holder}"))
-    off_by_one_budget = int(FLAGS.off_by_one_budget / 100.0 * fontmake_num_coords)
```
rsheeter, line 740: I tend to suspect that if you rewrote this to be string based rather than xpath it would be possible to count num coords fast without needing to estimate

rsheeter:
> ten non-empty glyphs

I suspect this biases us to the same set of ascii glyphs every time, and that this is non-representative for things like a CJK font that also supports latin

ttx_diff/src/ttx_diff/core.py
cmyr: you mean like regex?

cmyr:
would you prefer if we selected our glyphs evenly across the range? or any other ideas?

ttx_diff/src/ttx_diff/core.py
rsheeter: I was originally thinking even more basic, just string scanning. But ... on second glance it feels sus that a parsed xml doc takes so long to count, I wonder if what we're really paying for is xpath. Let me test locally.

rsheeter: Counting by walking the tree is slow but not minutes slow. I ttx'd rubik pixels and counted as follows:

```python
from lxml import etree
import sys
import time

def count(parent, path):
    count = 0
    frontier = [parent]
    path = [p for p in path if p]
    while frontier and path:
        tag = path.pop(0)
        new_frontier = []
        for e in frontier:
            for child in e:
                if child.tag != tag:
                    continue
                # must we go deeper?
                if path:
                    new_frontier.append(child)
                else:
                    count += 1
        frontier = new_frontier
    return count



def allow_some_off_by_ones(ttx_file, container, name_attr, coord_holder):
    times = [("t0", time.perf_counter_ns())]

    root = etree.parse(ttx_file).getroot()
    times.append(("parse", time.perf_counter_ns()))

    # handcount first so it can't possibly benefit from xpath warming
    handcount = count(root, f"//{container}/{coord_holder}".split("/"))
    times.append((f"hand_count {handcount}", time.perf_counter_ns()))

    xpcount = len(root.xpath(f"//{container}/{coord_holder}"))
    times.append((f"xpath_count {xpcount}", time.perf_counter_ns()))

    for (i, (name, t)) in enumerate(times[1:]):
        elapsed = int((t - times[i][1]) / 1e6)
        print(name, f"{elapsed}ms")

def main():
    for ttx_file in sys.argv[1:]:
        allow_some_off_by_ones(ttx_file, "glyf/TTGlyph", "name", "/contour/pt")
        allow_some_off_by_ones(
            ttx_file, "gvar/glyphVariations", "glyph", "/tuple/delta"
        )

if __name__ == "__main__":
    main()
```

The xpath count is so slow I haven't had the patience to wait it out yet. If I comment it out the timing I get is:

```shell
parse 2674ms
hand_count 1997282 975ms  # counting contour pt
parse 2494ms
hand_count 0 0ms  # counting tuple delta
```

grep suggests the count is right:

```
$ grep '<pt' /tmp/rp.ttx | wc -l
1997282

$ grep '<delta' /tmp/rp.ttx | wc -l
0
```

I'm going to try to summon the patience to let the xpath run through but this suggests to me that doing a correct count is fine at 975ms for all the rubik contour pts if I didn't bungle the timing.

rsheeter: I managed to have a cup of tea while waiting for xpath

```
parse 2721ms
hand_count 1997282 1876ms
xpath_count 1997282 948768ms  # is that ... 15 MINUTES?!
parse 2504ms
hand_count 0 0ms
xpath_count 0 253ms
```

My inclination would be to switch to correct non-xpath counting. In the sample below - worst case, it's Rubik Pixels - it's ~500x faster.

cmyr: okay good catch, looks like a more targeted iteration + count will work fine, I've added that.

PR #1942, opened 2026-04-13
[ttx_diff] Speed up name_id_to_name

cmyr:
//glob paths require traversing the entire font, and in some cases the entire font is big and has lots of elements that will not contain name ids (like everything in gvar or glyf, for instance.)

This uses more targetted paths, most importantly by indicating at least the root table in each call.

Tested locally, this changes no output of ttx_diff.


dfrg (approved):
Seems like a nice win for big fonts!

rsheeter:
Is there a measurable speedup?

cmyr:
yes, on a test font (kablamo) this goes from 14s to 0s (where total runtime is ~2m)

PR #1949, opened 2026-04-14
[glyphs] Handle openTypeOS2Type custom param

cmyr:
Another little oversight.

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -3066,6 +3066,17 @@ mod tests {
         assert_eq!(Some(0), context.static_metadata.get().misc.fs_type);
     }
 
+    // Some fonts use the long UFO name "openTypeOS2Type" instead of the Glyphs-native "fsType"
+    #[test]
+    fn reads_fs_type_0x0000_long_name() {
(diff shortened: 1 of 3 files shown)


rsheeter (approved):

rsheeter (approved):

glyphs2fontir/src/source.rs
```diff
@@ -3066,6 +3066,14 @@ mod tests {
         assert_eq!(Some(0), context.static_metadata.get().misc.fs_type);
     }
 
+    // Some fonts use the long UFO name "openTypeOS2Type" instead of the Glyphs-native "fsType"
+    #[test]
+    fn reads_fs_type_0x0000_long_name() {
+        let (_, context) =
+            build_static_metadata(glyphs3_dir().join("fstype_0x0000_long_name.glyphs"));
+        assert_eq!(Some(0), context.static_metadata.get().misc.fs_type);
```
rsheeter, line 3077: as 0 looks like it might be just the  default perhaps a comment exlpaining that if we ignore openTypeOS2Type we'd get some non-0 default would make sense?

PR #1951, opened 2026-04-14
[fontir] Fix anchor propagation for UFO spacing modifier glyphs

cmyr:
This is a claude patch I'm... okay with maybe? but I wanted to post it as a draft for @anthrotype to take a look at, because I find the subtle difference in behaviour between the two source formats a bit awkward and is a surprise to me, so curious to know if you think it makes sense.

I left this patch extremely verbose, but I will remove or condense some of the comments if we're going to actually merge it.

-----
claude says:

In UFO/designspace sources, ufo2ft's PropagateAnchorsFilter never propagates _ (attach-to) anchors from mark components to composite glyphs that are not "ligature marks" by name. fontc was not implementing this rule, causing spacing modifier glyphs (e.g. acute composed of acutecomb) to incorrectly inherit _top and be misclassified as GDEF Mark glyphs.

The fix distinguishes two propagation modes:

- Glyphs.app mode (mark_category_glyphs non-empty): mirrors glyphsLib's propagate_all_anchors, which runs before .glyphs→UFO conversion and always propagates the first component's anchors in full. GDEF classification comes from GlyphData.xml so inheriting a _ anchor does not cause misclassification.

- ufo2ft mode (mark_category_glyphs empty): mirrors ufo2ft's PropagateAnchorsFilter. Components with _ anchors are "mark components"; their _ anchors are never propagated to a composite unless the composite is a ligature mark by name (contains _ but doesn't start with _, per ufo2ft._is_ligature_mark). This caution is necessary because UFO sources typically infer GDEF classes from anchor names.

Fixes the Vollkorn crater diff where acute was appearing in GPOS MarkToBase lookups as a mark rather than a base.

### DIFF
--- a/fontir/src/propagate_anchors.rs
+++ b/fontir/src/propagate_anchors.rs
@@ -98,6 +98,36 @@ pub fn propagate_all_anchors(context: &Context) -> Result<(), Error> {
                 })
                 .unwrap_or_default();
 
+            // There are two anchor propagation pipelines in the Python world, and
+            // we must match whichever one would have run for this source format.
+            //
+            // .glyphs sources: glyphsLib runs `propagate_all_anchors` (in
+            //   glyphsLib/builder/transformations/propagate_anchors.py) via
+            //   `preflight_glyphs()` *before* the font is converted to UFO.
+            //   Anchors are baked into the GSFont object model using Glyphs.app
+            //   semantics: the first component always contributes all its anchors,
+            //   including _ (attach-to) anchors. Because glyphsLib populates
+            //   mark_category_glyphs from GlyphData.xml, GDEF classification is
+            //   explicit and independent of anchor names, so inheriting a _ anchor
+            //   does not accidentally reclassify a spacing modifier as a Mark.
+            //
+            // UFO/designspace sources: ufo2ft's `PropagateAnchorsFilter`
+            //   (ufo2ft/filters/propagateAnchors.py) runs during compilation on
+            //   the UFO objects. It classifies each component as a "mark component"
+            //   (has a _ anchor) or "base component" (no _ anchor). Only base
+            //   components' anchors are propagated. Mark components' _ anchors are
+            //   never propagated to a composite unless the composite is itself a
+            //   "ligature mark" by name (contains _ but doesn't start with _,
+            //   e.g. circumflexcomb_tildecomb). This caution is necessary because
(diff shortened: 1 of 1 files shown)


anthrotype:
TBH I don't feel great about this one (for similar reasons as https://github.com/googlefonts/fontc/pull/1945). I don't like that the font IR has branching behavior based on which frontend produced the data. It's supposed to be the common representation, not a union type with format-specific codepaths.

Plus the logic this is trying to match is a heuristic added [7 years ago](https://github.com/googlefonts/ufo2ft/pull/316) in ufo2ft before we received that obj-c snippet from Georg with the reference implementation of the Glyphs.app's anchor propagation algorithm (which we have already implemented in both fontc and glyphsLib). 

the PropagateAnchors ufo2ft filter is a opt-in and not well advertised so the number of fonts in circulation that make use of this (you mentioned Volkhorn, Cantarell from googlefonts/ufo2ft#316) are probably low.

My inclination would be to deprecate that ufo2ft filter and accept this small diffs and similar ones vs the Glyphs.app implementation already in fontc. If one cares about not marking those spacing modifiers as GDEF marks one should be able to define `openTypeCategories` in the UFO lib -- which is what glyphsLib would do if Volkhorn were re-exported to UFO+DS today (I can definitely tell it was exported from .glyphs with some older version of glyphsLib).

anthrotype:
BTW, regarding the Google CLA bot complaining about the `Co-authored by: Claude`, the instructions we were given from OSPO team is to simply remove that as it's not currently supported. Only humans can sign the CLA and they take full responsibility for the commit, the bot's contribution can be mentioned in the free-form text part of the commit message, but even that is not strictly required and left at one's discretion.

anthrotype:
for Claude specifically, you can turn that off by setting "commit" and "pr" to empty strings in 

https://code.claude.com/docs/en/settings#attribution-settings

cmyr:
I wouldn't commit this with co-authored by, but left it in because I hadn't signed off on the fix, was more posting it for discussion.

I think we do need to acknowledge that there are going to be _some_ differences that leak through into IR; the one  I just rememebered is the [`prefer_gdef_categories_in_fea`](https://github.com/googlefonts/fontc/blob/3518040e6f34782023a48443d16c2865a5c17671/fontir/src/ir/static_metadata.rs#L153) flag.

anthrotype:
this is now fixed after #2021

PR #1956, opened 2026-04-15
[glyphs] maintain start point when reversing closed path

cmyr:
This is a small detail on a very cold code path, but still is the cause of at least one crater diff.

Concretely, when a smart component is flipped (via transform) we reverse the path; however our reversal logic was naive, and didn't keep the same start point.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -789,9 +789,29 @@ impl Shape {
     }
 
     /// If the shape is a path, reverse it
+    ///
+    /// In Glyphs, the last node in the node list is the "start" node of a
+    /// closed contour (see `to_ir_path`). A naive `reverse()` would move the
+    /// start node to position 0, so `to_ir_path` would pick the wrong start.
+    ///
+    /// Python's `GSPath.reverse()` (which works via segments) is equivalent to
+    /// `reverse() + rotate_left(1)`, which keeps the start node at the last
+    /// position. We match that here.
+    ///
+    /// Note: this logic is only correct for closed paths. For open paths the
+    /// start node is `nodes[0]` (see `to_ir_path`), so a plain `reverse()`
+    /// without rotation would be correct. Python's segment-based `reverse()`
+    /// also appears to mishandle open paths (it silently drops the new start
+    /// node via `setSegments`). In practice this method is only called when
+    /// handling smart components, which are always closed?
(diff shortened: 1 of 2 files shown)


glyphs-reader/src/font.rs
```diff
@@ -789,9 +789,29 @@ impl Shape {
...
+    ///
+    /// In Glyphs, the last node in the node list is the "start" node of a
+    /// closed contour (see `to_ir_path`). A naive `reverse()` would move the
+    /// start node to position 0, so `to_ir_path` would pick the wrong start.
+    ///
+    /// Python's `GSPath.reverse()` (which works via segments) is equivalent to
+    /// `reverse() + rotate_left(1)`, which keeps the start node at the last
+    /// position. We match that here.
+    ///
+    /// Note: this logic is only correct for closed paths. For open paths the
```
anthrotype, line 801: so can't this logic only be applied for closed paths?

glyphs-reader/src/font.rs
```diff
@@ -789,9 +789,29 @@ impl Shape {
...
+    /// without rotation would be correct. Python's segment-based `reverse()`
+    /// also appears to mishandle open paths (it silently drops the new start
+    /// node via `setSegments`). In practice this method is only called when
+    /// handling smart components, which are always closed?
+    ///
+    /// <https://github.com/googlefonts/glyphsLib/blob/d42d3b15/Lib/glyphsLib/classes.py#L2369>
     pub(crate) fn reverse(&mut self) {
         if let Shape::Path(path) = self {
             path.reverse();
+            if !path.nodes.is_empty() {
```
```suggestion
            if path.closed && !path.nodes.is_empty() {
```

glyphs-reader/src/font.rs
cmyr: see the note: do we want to fix this in glyphsLib? The codepath seems to only be exercised with closed paths at present,  but it does seem like a bug.

I'll open an issue here and then can open a companion issue in glyphsLib.

anthrotype: yeah i think it's worth fixing in both. and if you say it's currently only exercised for closed paths, you can just do it now in fontc without affecting the crater results

cmyr: https://github.com/googlefonts/glyphsLib/pull/1147

PR #1961, opened 2026-04-16
[glyphs-reader] Treat empty Axis Mappings custom parameter as absent

### LINKED ISSUE #1830: [glyphs] present but empty 'axis values' custom parameter causes miscompilation
This shows up in RobotoSlab. The very simple fix seems to be to treat an empty value as missing, but this font _does_ confuse me:

- it has a weight axis
- there is an empty axis mapping
- there are three masters, which have axis values of 54, 197, and 351.

Those values seem weird. Are these design coordinates? Is there something else going on?

repro:

```
python3 -m ttx_diff 'https://github.com/googlefonts/robotoslab?67af3ce9c4#sources/RobotoSlab.glyphs'
```

anthrotype:
An empty `"Axis Mappings" = {}` custom parameter was being returned as `Some(empty_BTreeMap)`, causing `UserToDesignMapping::new()` to use it as a valid (but empty) mapping and skip the instance-based fallback.

Fix: return `None` from `user_to_design_from_axis_mapping()` when the resulting map is empty, matching glyphsLib's behaviour of treating an empty dict as falsy.

Fixes #1830

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2779,6 +2779,9 @@ fn user_to_design_from_axis_mapping(
                 .add_if_new(*user, *design);
         }
     }
+    if axis_mappings.is_empty() {
+        return None;
+    }
     Some(axis_mappings)
 }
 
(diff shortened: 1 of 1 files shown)


cmyr (approved):
👍

PR #1964, opened 2026-04-16
[glyphs-reader] instances Axis Locations override masters' (#1866)

### LINKED ISSUE #1866: we produce a different axis mapping
Our fvar gives the ExtraBold a wght value of 783.3, compared with 800 from fontmake.

```
python3 -m ttx_diff 'https://github.com/SUSE/suse-font?7159afb255#sources/SUSE-Italic.glyphs'
```

anthrotype:
When a font uses Axis Location custom params on both masters and instances, an instance can legitimately redefine a master's user→design mapping. Previously `add_if_new()` silently dropped the instance entry if the user value was already present (from the master), producing wrong output for fonts like SUSE-Italic (ExtraBold user=800 should map to design=780, not 800).

Switch to last-wins overwrite semantics everywhere, matching glyphsLib's `mapping[userLoc] = designLoc`. 

https://github.com/googlefonts/glyphsLib/blob/71f487f9d24e/Lib/glyphsLib/builder/axes.py#L221

https://github.com/googlefonts/glyphsLib/blob/71f487f9d24e/Lib/glyphsLib/builder/axes.py#L153

Both the master Axis Location path and the instance mapping path now use `add_or_replace` (with a log::warn when overwriting a different value). Removed `add_any_new`, no longer needed.

Rename `add_instance_mappings_if_new` => `add_instance_mappings` since it now always overwrites.

Add an integration test with a minimal .glyphs fixture modeled on SUSE-Italic that compiles end-to-end and checks the avar output.

Fixes #1866

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -2819,12 +2819,18 @@ fn user_to_design_from_axis_location(
             let design = master.axes_values[idx];
             // Last master wins, matching glyphsLib overwrite semantics:
             // https://github.com/googlefonts/glyphsLib/blob/71f487f9d24e/Lib/glyphsLib/builder/axes.py#L221
-            let source_label = format!("master '{}'", master.name.as_deref().unwrap_or(&master.id));
-
-            axis_mappings
+            let old = axis_mappings
                 .entry(axis_location.axis_name.clone())
                 .or_default()
-                .add_or_replace(user, design, &axis.name, &source_label);
+                .add_or_replace(user, design);
+            if let Some(old) = old {
+                let name = master.name.as_deref().unwrap_or(&master.id);
+                warn!(
+                    "Axis {}: master '{}' redefines mapping for user \
+                     location {user} from {old} to {design}",
+                    axis.name, name
+                );
+            }
         }
     }
     Some(axis_mappings)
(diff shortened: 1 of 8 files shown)


anthrotype:
The SUSE-Italic.glyphs `avar` table becomes identical after this change.

anthrotype:
oh looks like I had missed a few other broken test fixtures that don't declare their weightClass #1963 

i'm gonna commit them here if you don't mind

glyphs-reader/src/font.rs
```diff
@@ -2842,6 +2839,39 @@ impl AxisUserToDesignMap {
         self.0.push((user, design));
     }
 
+    fn add_any_or_replace(
+        &mut self,
+        incoming: &AxisUserToDesignMap,
+        axis_name: &str,
+        source_label: &str,
```
cmyr: instead of passing in a string that's only used to debug, I would prefer if the fn returned a Result or a bool, and then the warning was printed at the call-site, if possible?

resources/testdata/glyphs2/WghtVar_Instances_implied_axes.glyphs
```diff
@@ -31,6 +31,7 @@ instanceInterpolations = {
 isBold = 1;
 linkStyle = Regular;
 name = Bold;
+weightClass = Bold;
```
cmyr, line 34: wasn't this fixed in an earlier patch?

cmyr (approved):
looks good, a couple of things inline :)

resources/testdata/glyphs2/WghtVar_Instances_implied_axes.glyphs
anthrotype: it should have but I had missed it https://github.com/googlefonts/fontc/pull/1964#issuecomment-4260774107

glyphs-reader/src/font.rs
anthrotype: done https://github.com/googlefonts/fontc/pull/1964/commits/b88692c6f3906426ce709c92218e010bd6eec5a9

PR #1968, opened 2026-04-22
[propagate_anchors] Don't round deltas when interpolating component anchors

### LINKED ISSUE #1967: [propagate_anchors] interpolate_component_anchors should keep deltas unrounded
in fontir's propagate_anchors.rs, the `interpolate_component_anchors` function is used when a composite has an intermediate/brace layers that a component doesn't. It uses VariationModel.deltas() method to compute deltas and then interpolates from those: 

https://github.com/googlefonts/fontc/blob/7f631b66246051868093f7f07581db08f8d337c2/fontir/src/propagate_anchors.rs#L230-L241

But [VariationModel.deltas()](https://github.com/googlefonts/fontc/blob/7f631b66246051868093f7f07581db08f8d337c2/fontdrasil/src/variations.rs#L261) rounds the deltas to integer (it's used to encode them in the IVS). Whereas here we are using the deltas to interpolate anchors at missing locations, and re-insert them as additional anchor sources. We let these stay floats throughout the pipeline (until the very end when deltas have to be rounded for storage in IVS), so we should use the other `deltas_with_rounding` method which was meant for these use cases exactly:

https://github.com/googlefonts/fontc/blob/7f631b66246051868093f7f07581db08f8d337c2/fontdrasil/src/variations.rs#L272-L280

we do something similar in a couple places, e.g. like this:
https://github.com/googlefonts/fontc/blob/7f631b66246051868093f7f07581db08f8d337c2/fontir/src/glyph.rs#L479-L484

glyphsLib propagate_anchors.py also does _not_ round the deltas when interpolating component anchors at missing brace layers locations (fonttools' VariationModel.interpolateFromMasters defaults to `round=noRound`).

https://github.com/googlefonts/glyphsLib/blob/71f487f9d24e7e9d6ae20415c8a626c25bbf9be1/Lib/glyphsLib/builder/transformations/propagate_anchors.py#L324

There is at least one source on fontc_crater that displays this issue, NotoSerif-Italic.glyphspackage:

````bash
$ python -m ttx_diff --compare default 'https://github.com/notofonts/latin-greek-cyrillic?c4a321e123#sources/NotoSerif-Italic.glyphspackage'
````

```diff
$ diff build/default/fontc.markkern.txt build/default/fontmake.markkern.txt
567286c567286
< uni1DF14 @(x: 283 {185,239,241,268,315,288,288,338,338}, y: 152 {151,151,152,153,153,153,153,153,153})
---
> uni1DF14 @(x: 283 {185,239,241,268,315,288,288,338,338}, y: 152 {151,151,152,152,153,153,153,153,153})
```

I'll send a PR

anthrotype:
Fixes #1967

### DIFF
--- a/fontir/src/propagate_anchors.rs
+++ b/fontir/src/propagate_anchors.rs
@@ -604,6 +604,7 @@ mod tests {
     };
     use fontdrasil::{orchestration::Access, types::Axis};
     use std::collections::BTreeSet;
+    use write_fonts::OtRound;
 
     /// Helper to create a test context with proper setup
     fn test_context() -> Context {
@@ -877,14 +878,14 @@ mod tests {
         );
     }
 
-    /// Helper to assert an anchor exists with expected position at a specific location
-    fn assert_anchor_at_loc(
+    /// Helper to retrieve an anchor position at a specific location, panicking
+    /// if the glyph, anchor, or location is missing.
+    fn anchor_at_loc(
         context: &Context,
         glyph_name: &str,
         anchor_name: &str,
         loc: &NormalizedLocation,
-        expected: Point,
-    ) {
+    ) -> Point {
         let glyph_anchors = context
             .anchors
             .try_get(&WorkId::Anchor(glyph_name.into()))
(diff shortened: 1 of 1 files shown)


anthrotype:
CI fails as expected

```
failures:

---- propagate_anchors::tests::interpolate_component_anchor_uses_unrounded_deltas stdout ----

thread 'propagate_anchors::tests::interpolate_component_anchor_uses_unrounded_deltas' (4300) panicked at fontir/src/propagate_anchors.rs:1895:9:
  left: 153
 right: 152
```

https://github.com/googlefonts/fontc/actions/runs/24779390787/job/72506072869?pr=1968#step:5:249

rsheeter (approved):

PR #1971, opened 2026-04-22
[glyphs-reader] Fix glyph production name lookup order in glyphdata

anthrotype:
To match glyphsLib's resolution order, `query_no_synthesis` must look up a glyph's name by its production name before looking it up by codepoint.

https://github.com/googlefonts/glyphsLib/blob/v6.13.0/Lib/glyphsLib/glyphdata.py#L120-L128
https://github.com/googlefonts/glyphsLib/blob/v6.13.0/Lib/glyphsLib/glyphdata.py#L163-L166

Currently we are doing the opposite of that, doing `find_pos_by_codepoint` before `find_pos_by_prod_name`:

https://github.com/googlefonts/fontc/blob/7f631b66246051868093f7f07581db08f8d337c2/glyphs-reader/src/glyphdata.rs#L534-L545

... which leads to different production glyph names in Inter-Italic and Roman.
In there, a glyph named "triagrt" (with codepoints {0x25BA, 0x25B6} gets incorrectly renamed to "uni25B6" because the codepoint fallback matches U+25B6 before the production name lookup can match "triagrt" (same as original). The correct production name is indeed the latter (AGLFN name for U+25BA).

### DIFF
--- a/glyphs-reader/src/glyphdata.rs
+++ b/glyphs-reader/src/glyphdata.rs
@@ -533,13 +533,13 @@ impl GlyphData {
 
         // No override, perhaps we have a direct answer?
         bundled::find_pos_by_name(name)
+            .or_else(|| find_pos_by_prod_name(name.into()))
             .or_else(|| {
                 codepoints
                     .into_iter()
                     .flat_map(|cps| cps.iter())
                     .find_map(|cp| bundled::find_pos_by_codepoint(*cp))
             })
-            .or_else(|| find_pos_by_prod_name(name.into()))
             .map(|i| {
                 bundled::get(i).unwrap_or_else(|| panic!("We found invalid index {i} somehow"))
             })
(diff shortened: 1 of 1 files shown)


anthrotype:
CI failing as expected:

```
failures:

---- glyphdata::tests::prod_name_lookup_before_codepoint_lookup stdout ----

thread 'glyphdata::tests::prod_name_lookup_before_codepoint_lookup' (5330) panicked at glyphs-reader/src/glyphdata.rs:1358:9:
  left: Some("uni25B6")
 right: Some("triagrt")
```

https://github.com/googlefonts/fontc/actions/runs/24784268438/job/72523713383?pr=1971#step:6:350

rsheeter (approved):
Ideally add a test?

anthrotype:
> Ideally add a test?

I added one, no? https://github.com/googlefonts/fontc/pull/1971/commits/a8ea97f7e63cef53017013763b5d05f6afbb139a

PR #1972, opened 2026-04-23
[fontir] don't leak interpolated sources from non-export flattening into gvar/HVAR

### LINKED ISSUE #1840: gvar and HVAR difference in RedditSansCondensed
in `aringacute.ss02`. My only clue here is that `a.ss02` includes an intermediate layer, and it looks like this might be getting dropped by fontmake, in this particular case? But that doesn't feel like the whole story.

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/1840

When a composite has a non-export component, `flatten_non_export_components_for_glyph` calls `ensure_composite_defined_at_component_locations`, which adds virtual interpolated sources at every location defined by any component — including brace layers on sibling/export components. These extra sources persist on the
composite and produce superfluous intermediate tuples in both gvar and HVAR.

`ensure_composite_defined_at_component_locations` is needed during the flattening loop (for correct nested component resolution), but its interpolated sources should be kept only when they carry data the composite glyph actually needs, that is when a non-export component being flattened has an explicit source at that intermediate location.

As usual, I first push a test that fails without the fix: a composite referencing a non-export component AND a different component with an intermediate layer that leaks into the composite's gvar and HVAR. 
Then a second commit follows with the fix.

I verified that both RedditSansCondensed and NotoSans-Italic become **ttx-diff-identical** after these changes.

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -254,15 +254,20 @@ fn flatten_non_export_components_for_glyph(
     context: &Context,
     glyph: &Glyph,
 ) -> Result<Glyph, BadGlyph> {
-    let glyph = ensure_composite_defined_at_component_locations(context, glyph)?;
-    let mut builder = GlyphBuilder::from(glyph.clone());
-    builder.clear_components();
+    let expanded = ensure_composite_defined_at_component_locations(context, glyph)?;
+    let mut new_sources = HashMap::with_capacity(expanded.sources().len());
 
     log::debug!("flattening non-export components of '{}'", glyph.name);
-    for (loc, instance) in glyph.sources() {
-        let mut new_instance = instance.clone();
-        new_instance.components.clear();
+    for (loc, instance) in expanded.sources() {
+        let mut new_instance = GlyphInstance {
+            width: instance.width,
+            height: instance.height,
+            vertical_origin: instance.vertical_origin,
+            contours: instance.contours.clone(),
+            components: Vec::with_capacity(instance.components.len()),
+        };
 
+        let mut non_export_has_location = false;
         for component in &instance.components {
             let id = WorkId::Glyph(component.base.clone());
             let referenced_glyph = context.glyphs.get(&id);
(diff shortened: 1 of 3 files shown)


rsheeter (approved):

fontir/src/glyph.rs
```diff
@@ -254,15 +254,20 @@ fn flatten_non_export_components_for_glyph(
...
-        let mut new_instance = instance.clone();
-        new_instance.components.clear();
+    for (loc, instance) in expanded.sources() {
+        let mut new_instance = GlyphInstance {
+            width: instance.width,
+            height: instance.height,
+            vertical_origin: instance.vertical_origin,
+            contours: instance.contours.clone(),
+            components: Vec::with_capacity(instance.components.len()),
+        };
```
cmyr, line 269: this is equivalent, yes? DId you just find the previous version confusing?

fontir/src/glyph.rs
```diff
@@ -291,10 +297,22 @@ fn flatten_non_export_components_for_glyph(
...
-        builder.sources.insert(loc.clone(), new_instance);
+        // Keep sources at the glyph's original locations, plus any locations
+        // where a non-export component being flattened has its own source.
+        // Drop the other virtual sources added by
+        // ensure_composite_defined_at_component_locations — they would leak
+        // into gvar as superfluous intermediate tuples.
+        // https://github.com/googlefonts/fontc/issues/1840
+        if glyph.sources().contains_key(loc) || non_export_has_location {
+            new_sources.insert(loc.clone(), new_instance);
+        }
```
cmyr, line 308: i'm curious why we don't keep using the builder pattern, and only add the check? I don't care particularly either way, but it seems more verbose without gaining anything I can identify?

cmyr (approved):
A couple questions inline but seems reasonable!

fontir/src/glyph.rs
anthrotype: Yeah, it's functionally equivalent. I moved away from the builder because with the new filtering logic the builder pattern works against us: `GlyphBuilder::from(expanded.clone())` copies all sources, `clear_components()` iterates them to clear components, and then we either overwrite each source or need an extra `builder.sources.remove(loc)` branch to clean up the ones we're dropping. Building `new_sources` from scratch avoids all that, we only construct what we keep. The explicit `GlyphInstance` construction similarly avoids cloning components just to immediately clear them.

PR #1974, opened 2026-04-24
[smart_components] Fix transform composition and single-layer flattening

### LINKED ISSUE #1973: [smart_components] wrong transform composition and unnecessary component flattening in MonaSans
Two bugs in the same glyphs-reader's `smart_components.rs` module, exemplified by MonaSans.glyphspackage

```
python -m ttx_diff 'https://github.com/github/mona-sans?0d07d5a501#sources/MonaSans.glyphspackage'
```

1) `Shape::apply_affine` composes transforms in the wrong order for `Component` shapes.

When a smart component's inner shape is itself a transformed component (e.g. ninesuperior => nineinferior(smart) => sixinferior(rotate 180°)), `apply_affine` composes the transforms in the wrong order:

```rust
component.transform *= affine;  // component.transform = component.transform * affine  (inner * outer)
                                // should be:  component.transform = affine * component.transform   (outer * inner)
```

The `Path` branch is correct (`affine * node.pt`), but the `Component` branch is inconsistent with it. The result is that the outer translation gets transformed by the inner rotation instead of being applied after it.

In MonaSans, this causes `uni2079` (ninesuperior) to be placed 752 units too low (y=−23 instead of y=729), which also drags the font-wide `head.yMin` from −234 to −421.

The bug only manifests when the inner component has a non-identity 2×2 (rotation, flip, etc.) AND the outer reference has a non-zero translation (otherwise matrix multiplication commutes for the translation part).

2) the single-layer fast path unnecessarily inlines inner component references, flattening the component graph.

When a smart component has only one layer per font master (the smart axes can't actually vary), `instantiate_for_layer` clones the inner shapes and applies the composed transform. If one of those inner shapes is a `Component`, this flattens the component graph — e.g. replacing `ninesuperior => nineinferior` with `ninesuperior => sixinferior`.

glyphsLib handles this case differently: when `len(masters) == 1`, `get_smart_component_variation_model` returns `None` and `instantiate_smart_component` falls back to `pen.addComponent(component.name, component.transform)`, preserving the original component reference.

The flattening produces a semantically valid but structurally different font, causing gvar offset deltas to differ from fontmake's.

anthrotype:
Fix two bugs in smart component instantiation:

1. Shape::apply_affine composed transforms as `inner*outer` instead of `outer*inner` for Component shapes (the Path branch was correct). With a non-identity 2x2 (rotation/flip), this flips the sign of translations: e.g. y=729 became y=-23 in MonaSans ninesuperior.

 2. Single-layer fast path inlined inner component shapes, flattening the component graph (e.g. ninesuperior -> sixinferior instead of ninesuperior -> nineinferior). Now preserves the component reference, matching glyphsLib's model=None fallback.

Fixes #1973

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -1910,6 +1910,44 @@ mod tests {
         );
     }
 
+    /// Verify anchors propagate through a single-layer smart component.
+    ///
+    /// Regression test for <https://github.com/googlefonts/fontc/issues/1973>.
+    /// When a smart component has only one layer per master, we preserve
+    /// the component reference and return no anchors from instantiation.
+    /// Anchors must still reach the outermost composite via propagation
+    /// through the component chain: composite → flipped(smart) → base.
+    #[test]
+    fn propagate_anchors_through_single_layer_smart_component() {
+        let result =
+            TestCompile::compile_source("glyphs3/SmartComponentSingleLayerAnchors.glyphs");
+
+        let anchors = result
+            .fe_context
+            .anchors
+            .try_get(&FeWorkIdentifier::Anchor("composite".into()))
+            .expect("composite should have propagated anchors");
+
+        // base "top" at (200, 400), propagated through:
+        //   flipped [-1, 0, 0, -1, 417, 353] → (217, -47)
+        //   composite translate(0, 376)       → (217, 329)
+        // The 180° rotation renames "top" → "bottom" (see rename_anchor_for_scale).
+        let bottom = anchors
+            .anchors
+            .iter()
+            .find(|a| a.original_name == "bottom")
+            .unwrap_or_else(|| {
+                let names: Vec<_> = anchors.anchors.iter().map(|a| &a.original_name).collect();
+                panic!("composite should have a 'bottom' anchor (renamed from base's 'top'), found: {names:?}")
+            });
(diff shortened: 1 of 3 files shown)


anthrotype:
```
failures:

---- smart_components::tests::single_layer_smart_component_preserves_component_ref stdout ----

thread 'smart_components::tests::single_layer_smart_component_preserves_component_ref' (5462) panicked at glyphs-reader/src/smart_components.rs:970:9:
assertion `left == right` failed: should keep reference to 'inner', not flatten to 'base'
  left: "base"
 right: "inner"

---- smart_components::tests::apply_affine_component_transform_order stdout ----

thread 'smart_components::tests::apply_affine_component_transform_order' (5461) panicked at glyphs-reader/src/smart_components.rs:919:9:
assertion `left == right` failed: should compose as outer * inner; wrong order gives (417, -23)
  left: (417.0, -23.0)
 right: (417.0, 729.0)
```

https://github.com/googlefonts/fontc/actions/runs/24904723190/job/72931048395?pr=1974

cmyr (approved):
Looks good, although the two bugs seem orthogonal?

anthrotype:
they affect the same font! Please don't have me split the PR 🙏

PR #1980, opened 2026-04-25
[ttx_diff] Fix computing source file for gftools mode

cmyr:
My last attempt at this broke a bunch of stuff. But this one has tests! And I tried it locally on a bunch of targets!


anthrotype (approved):
LGTM

PR #1983, opened 2026-04-27
[fea-rs] Fix silent loss of features from .fea files with CR line endings

anthrotype:
fea-rs's lexer `comment()` only terminated `#` comments at `\n` (LF), not `\r` (CR). Files with CR-only line endings (classic Mac) had their entire contents consumed as a single comment token starting from the first `#`.

This silently drops all features from any included `.fea` file that uses CR-only line endings. Discovered via Bungee-Shade, where `sources/1-drawing/features/features.fea` uses CR terminators. fontc lost 13 GSUB features, the entire `latn` script, and 6 `name` UINameIDs.

Repro (on current `main`, before this PR):

```
python -m ttx_diff 'https://github.com/djrrb/Bungee?eb03cf69ad#sources/2-build/Bungee_Basic/Bungee-Shade.ufo'
```

After this PR: output is identical.

Also, `line_offsets()` and `line_containing_offset()` in `parse/source.rs` only recognized `\n`, producing wrong line numbers and garbled context in diagnostics for such files.

### DIFF
--- a/fea-rs/src/parse/source.rs
+++ b/fea-rs/src/parse/source.rs
@@ -237,7 +237,7 @@ impl Source {
 
         (
             offset_idx + 1,
-            self.contents[start_offset..end_offset].trim_end_matches('\n'),
+            self.contents[start_offset..end_offset].trim_end_matches(['\n', '\r']),
         )
     }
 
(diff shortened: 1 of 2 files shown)


anthrotype:
```
failures:

---- parse::lexer::tests::comment_cr_line_endings stdout ----

thread 'parse::lexer::tests::comment_cr_line_endings' (3863) panicked at fea-rs/src/parse/lexer.rs:444:9:
  left: "#(# comment\rlanguagesystem DFLT cool;)"
 right: "#(# comment)"

---- parse::lexer::tests::comment_crlf_line_endings stdout ----

thread 'parse::lexer::tests::comment_crlf_line_endings' (3864) panicked at fea-rs/src/parse/lexer.rs:455:9:
  left: "#(# comment\r)"
 right: "#(# comment)"
```

https://github.com/googlefonts/fontc/actions/runs/25007850874/job/73235418983?pr=1983#step:6:293

rsheeter (approved):

fea-rs/src/parse/source.rs
```diff
@@ -250,13 +250,26 @@ impl Source {
...
+                    result.push(next + 1);
+                    i = next;
+                } else {
+                    result.push(next);
+                }
+            }
+            _ => (),
+        }
+        i += 1;
+    }
```
cmyr: just golf, but I'd write this as something like,
```suggestion
    let mut iter = text.bytes().enumerate().peekable();
    while let Some((i, b)) = iter.next() {
        match b {
            b'\r' if iter.peek().map(|(_, b)| *b) == Some(b'\n') => {
                // treat \r\n as a single line ending
                let _ = iter.next();
                // skip the upcoming \n
                result.push(i + 2);
            }
            b'\n' | b'\r' => result.push(i + 1),
            _ => (),
        }
    }
```

cmyr (approved):
so do we think we have a source that was edited on classic Mac OS?

PR #1986, opened 2026-04-28
Update python deps

cmyr:
This brings in a fix that should give us a meaningful bump on crater.

It does come with a caveat, though: merriweather4 is getting stuck during table packing with this set of python dependencies, so I am temporarily adding it to our manual 'deny list'.


rsheeter:
> brings in a fix

can we link to which fix we're tryign to bring in?

rsheeter (approved):

cmyr:
https://github.com/googlefonts/ufo2ft/pull/977

anthrotype:
> merriweather4 is getting stuck during table packing with this set of python dependencies

strange, I don't see anything in say glyphsLib or ufo2ft that may have potentially caused that. Maybe uharfbuzz (or underlying Harfbuzz) used by fontTools for packing OTL tables?

```diff
- uharfbuzz==0.53.3
+ uharfbuzz==0.54.0
```

PR #1992, opened 2026-04-29
[fea-rs] Strip newlines from string literals

cmyr:
see https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#:~:text=Newlines%20embedded%20within%20the%20string%20are%20removed


rsheeter:
I'm curious, are we observing this to cause problems with sources?

rsheeter (approved):

cmyr:
yes, there is a family that hits this bug. (tajawal)

PR #1995, opened 2026-05-01
[fontdrasil] Fix many-to-one axis map handling

anthrotype:
This is the companion to fonttools/fonttools#4085.

Fixes the remaining `ttx_diff` discrepancy on [SUSE-Italic at 7159afb255](https://github.com/SUSE/suse-font/tree/7159afb255) reported in https://github.com/googlefonts/ufo2ft/issues/978

"output is identical" in ttx-diff after updated fonttools is installed in the fontc `.venv`.

---

`PiecewiseLinearMap::map()` uses `binary_search`, which returns an arbitrary/unspecified index when there are multiple matches:

https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search

When a many-to-one forward map (e.g. user=900 -> design=1000 and user=1000 -> design=1000) is reversed, the value returned from binary_search can differ from fontmake's (which takes the first match).

I changed it to use `partition_point` to always find the first occurrence.

After fixing this I found a second related issue: `to_ir_axis` derived fvar axis min/default/max by round-tripping design coords through `design_to_user`, which is lossy for many-to-one maps.

Changed it to take min/max directly from the user-space values in the mapping, [matching glyphsLib's approach](https://github.com/googlefonts/glyphsLib/blob/75c07d4213e1e7693a220d54c2fed4755cfc4131/Lib/glyphsLib/builder/axes.py#L284-L285).

(I will release a patched fonttools soon and bump fontc requirements after)

### DIFF
--- a/fontdrasil/src/piecewise_linear_map.rs
+++ b/fontdrasil/src/piecewise_linear_map.rs
@@ -136,6 +136,62 @@ mod tests {
     }
 
     #[test]
+    fn many_to_one_reverse() {
+        // https://github.com/googlefonts/ufo2ft/issues/978
+        // user=900 and user=1000 both map to design=1000
+        let user_to_design = PiecewiseLinearMap::new(vec![
+            (OrderedFloat(100_f64), OrderedFloat(100_f64)),
+            (OrderedFloat(800_f64), OrderedFloat(780_f64)),
+            (OrderedFloat(900_f64), OrderedFloat(1000_f64)),
+            (OrderedFloat(1000_f64), OrderedFloat(1000_f64)),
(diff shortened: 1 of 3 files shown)


anthrotype:
https://github.com/googlefonts/fontc/actions/runs/25220991765/job/73952882058?pr=1995#step:4:195

```
thread 'piecewise_linear_map::tests::many_to_one_reverse' (3210) panicked at fontdrasil/src/piecewise_linear_map.rs:151:9:
  left: 1000.0
 right: 900.0
```

cmyr (approved):
nice!

PR #1998, opened 2026-05-01
Fix missing .notdef vertical phantom point deltas

cmyr:
_a small issue affecting one font on crater_:

When os2_typo_ascender/descender vary across masters but the outline ascender/descender are constant, synthesize_notdef was skipping non-default sources entirely, losing the vertical phantom point variation in gvar. Now the variability check also considers the typo metrics, and vertical_origin is set explicitly on all sources.

### DIFF
--- a/fontir/src/glyph.rs
+++ b/fontir/src/glyph.rs
@@ -22,7 +22,7 @@ use write_fonts::{OtRound, types::GlyphId16};
 
 use crate::{
     error::{BadGlyph, Error},
-    ir::{Component, GlobalMetric, Glyph, GlyphBuilder, GlyphInstance, GlyphOrder, StaticMetadata},
+    ir::{Component, Glyph, GlyphBuilder, GlyphInstance, GlyphOrder, StaticMetadata},
     orchestration::{Context, Flags, IrWork, WorkId},
     propagate_anchors::propagate_all_anchors,
 };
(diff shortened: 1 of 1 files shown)


fontir/src/glyph.rs
```diff
@@ -692,17 +693,27 @@ fn synthesize_notdef(context: &Context) -> Result<Glyph, BadGlyph> {
...
         }
-        let outline = make_notdef_outline(upem, ascender.0, descender.0);
+        let outline = make_notdef_outline(upem, metrics.ascender.0, metrics.descender.0);
 
         let instance = GlyphInstance {
             width,
-            height: Some(ascender.0 - descender.0),
+            height: Some(metrics.ascender.0 - metrics.descender.0),
             contours: vec![outline],
+            vertical_origin: Some(metrics.os2_typo_ascender.0),
```
dfrg: I’m guessing this is what fonttools does but using hhea metrics for size and os2 ascender for vertical origin seems weird? Then again, there’s plenty of strange line metrics stuff in skrifa.

fontir/src/glyph.rs
```diff
@@ -1133,6 +1144,38 @@ mod tests {
         assert_eq!(notdef.sources()[&loc0].height, Some(600.));
     }
 
+    /// When os2_typo_ascender varies across masters but the outline ascender/descender
+    /// are constant, the .notdef should still have variable sources so that the
+    /// vertical phantom point deltas are captured in gvar.
+    /// See https://github.com/googlefonts/fontc/issues/XXXX
```
dfrg: Is there an actual issue number for this?

dfrg (approved):

fontir/src/glyph.rs
cmyr: 😅

fontir/src/glyph.rs
cmyr: good call here actually, this isn't exactly necessary.

What is happening is that if vertical_origin is none, we fall back to using os2_typo_ascender; the important thing here is simply that we do create consider that metric when deciding whether or not to create an instance of the glyph at that location.

PR #1999, opened 2026-05-02
[glyphs] Fix parse_int/_float for leading-zero ints

cmyr:
The `numeric_ok` guard in `parse_int`/`parse_float` was rejecting all-digit strings with leading zeros (like "011") to prevent `parse_atom` from misclassifying hex codepoints (like "0041") as integers. But `parse_int`/`parse_float` are called from typed struct parsing where the field is already known to be numeric — the ambiguity that `numeric_ok` solves doesn't exist in that context.

Remove the `numeric_ok` gate entirely and just try parsing directly. Guard the f64 fallback with `is_finite()` to reject "infinity"/"nan".

Fixes compilation of Federant (versionMinor = 011).

### DIFF
--- a/glyphs-reader/src/plist.rs
+++ b/glyphs-reader/src/plist.rs
@@ -655,20 +655,20 @@ impl From<Dictionary> for Plist {
 }
 
 pub(crate) fn parse_int(s: &str) -> Result<i64, Error> {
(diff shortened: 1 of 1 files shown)


dfrg (approved):
👍

PR #2000, opened 2026-05-03
[fea-rs] Allow mark classes in all glyph class contexts

cmyr:
The validation pass rejected mark class references (e.g. @_U) inside glyph class definitions like `@foo = [@bar @_U]`, even though the FEA spec allows mark classes wherever glyph classes are expected. Remove the accept_mark_class parameter and always accept mark classes in validation, matching the behavior already present in the compilation pass.


dfrg (approved):
👍

PR #2011, opened 2026-05-15
stop filtering _-prefixed anchors during smart component instantiation

### LINKED ISSUE #2010: [glyphs-reader] smart component instantiation drops mark-attaching `_`-prefixed anchors
When fontc instantiates smart components in `glyphs-reader/src/font.rs`, it filters out all `_`-prefixed anchors from the interpolated result (PR #1885, commit `7fa98160`). I assumed these were internal part-to-part anchors for aligning the smart components to their base, which serve their purpose once the smart components are turned into contours. It was a precautionary heuristic (inline comment says: "filtering out 'mark' anchors used for internal part-to-part connections"), there was no concrete regression or test case motivating it. The assumption turned out to be wrong, because `_`-prefixed anchors can also be legitimate mark-attaching anchors that are needed for GPOS MarkToBase rules.

E.g. in Padyakke_Libre
```
python -m ttx_diff 'https://github.com/DunwichType/Padyakke_Libre?6b18f96959#source/Padyakke.glyphs'
```

fontmake produces 277 MarkToBase rules in abvm, whereas fontc only 2. Those are legitimate mark attachments.

`knmAU` (U+0CCC, Kannada Vowel Sign AU) is a composite of `_part.knmAU`, a smart component with `partsSettings`. `_part.knmAU` has a `_headline` anchor at (135, 383).

- fontmake (glyphsLib actually) runs anchor propagation before smart component instantiation, so `_headline` is already on `knmAU` when decomposition happens. The mark writer pairs it with `headline` anchors on 277 Kannada base consonants and produces 277 abvm MarkToBase rules.
- in fontc, smart components are instantiated first, and the `_`-prefix filter drops `_headline`, then anchor propagation runs but finds no component to propagate from (already decomposed). This results in only 2 abvm rules instead of 279.

In any case, the filter is redundant because the mark writer's [`used_groups` intersection](https://github.com/googlefonts/fontc/blob/a57c4b6708e7e229272619d53ca17ab557325504/fontbe/src/features/marks.rs#L255-L257) already provides the necessary safety net. Any `_`-prefixed anchor that doesn't have a matching non-prefixed counterpart on an exported glyph in the final font is harmlessly ignored, and no mark rule is generated.

The fix is a trivial one-line change:

```diff
-                            !a.name.starts_with('_') && !explicit_anchor_names.contains(&a.name)
+                            !explicit_anchor_names.contains(&a.name)
```

... which makes ttx-diff report "output is identical" on Padyakke. I'll send a PR.

anthrotype:
Fixes #2010

JMM

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -3994,15 +3994,12 @@ impl Font {
                             issue,
                         })?;
                         new_shapes.extend(instance.shapes);
-                        // Add anchors from smart component, filtering out
-                        // 'mark' anchors used for internal part-to-part connections
-                        // and any that would override the glyph's explicit anchors.
+                        // Add anchors from smart component, skipping any that
+                        // would override the glyph's explicit anchors.
                         let new_anchors: Vec<_> = instance
                             .anchors
                             .into_iter()
-                            .filter(|a| {
-                                !a.name.starts_with('_') && !explicit_anchor_names.contains(&a.name)
-                            })
+                            .filter(|a| !explicit_anchor_names.contains(&a.name))
                             .collect();
                         // When duplicates exist among smart component anchors,
                         // keep the last one (matching Glyphs.app).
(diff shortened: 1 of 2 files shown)


cmyr (approved):
good catch!

PR #2013, opened 2026-05-18
[glyphs2fontir] Include all masters in KerningGroups.locations

### LINKED ISSUE #1857: we don't generate a kern feature for NotoSansHebrew
Possibly a failure to handle the case where _only_ RTL kerning is defined?

anthrotype:
`KerningGroupWork` was building `groups.locations` only from `kerning_ltr` keys, so masters with empty kerning dicts were excluded from the variation model. This caused two symptoms:
 
- RTL-only kerning sources (e.g. NotoSansHebrew): no kern feature emitted at all, since `kerning_ltr` has no entries
 - Fonts where only some masters have kerning (e.g. Geom, 1 of 7 masters): constant kern values instead of interpolating toward zero at non-kerning masters
 
Both font source formats expects that empty kerning dict on a master should still participate in variation. ufo2fontir already does this: it includes all non-glyph-only designspace sources in `kerning_groups.locations` unconditionally.

(font source formats haven't been updated to support truly "sparse" kerning yet, all global font masters are expected to participate in kerning whether or not they contain a non-empty kerning dict, it's absence doesn't make them 'sparse'. An optional lib key has been proposed, I can find the link, but it's out of scope for this bugfix).

After the fix, ttx-diff now reports "output is identical" for NotoSansHebrew and Geom.

Fixes #1857

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -1118,20 +1118,7 @@ impl Work<Context, WorkId, Error> for KerningGroupWork {
             }
         }
 
-        groups.locations = font
-            .kerning_ltr
-            .iter()
-            .filter_map(
-                |(master_id, _)| match font_info.master_positions.get(master_id) {
-                    Some(pos) => Some(pos),
-                    None => {
-                        warn!("Kerning is present for non-existent master {master_id}");
-                        None
-                    }
-                },
-            )
-            .cloned()
-            .collect();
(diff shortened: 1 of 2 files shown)


anthrotype:
the "Build GS" CI job failing is unrelated to my PR (we check out googlesans at HEAD and it seems the font source was updated with broken FEA missing semicolons after `include`s)

cmyr (approved):
looks good!

Issue #2014, opened 2026-05-18
Google Sans CI broken after googlesans#729


anthrotype:
The "Build GS" CI job is broken since googlefonts/googlesans#729 (import-arabic-poc) merged earlier today (2026-05-18). The new Arabic FEA files use two constructs that fea-rs rejects but feaLib accepts:

1. `<anchor NULL>` without `mark @class` on a ligature component
2. `valueRecordDef` inside a feature block instead of at the top-level

Both have been accepted by feaLib since the beginning.
I have two PRs ready that fix each.

anthrotype:
> <anchor NULL> without mark @class on a ligature component

fea-rs required `mark @class` after every anchor, but NULL anchors mean "no attachment here" and don't take a mark class. The FEA spec grammar requires `mark @class` after every anchor, but it is inconsistent with its own [GPOS type 5 example](https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#6.e) wihch uses a final `<anchor NULL>;` without it. feaLib has followed this since its [original implementation](https://github.com/fonttools/fonttools/blob/0532ec85d30b/Lib/fontTools/feaLib/parser.py#L200-L201) (Dec 2015).

> `valueRecordDef` inside a feature block instead of at the top-level

The spec says this is a top-level-only statement, but feaLib has always accepted this in feature and lookup blocks for 11 years, and Google Sans Arabic is a real font that relies on this. It's trivial to support so we should.

anthrotype:
"Build GS" CI job is green again https://github.com/googlefonts/fontc/actions/runs/26046192588/job/76570913022

PR #2015, opened 2026-05-18
[fea-rs] Allow NULL anchor without mark class in pos ligature rules

anthrotype:
Part of #2014.

In GPOS type 5 (mark-to-ligature) rules, `<anchor NULL>` denotes a ligature component with no mark attachment point. fea-rs required `mark @class` after every anchor, rejecting `<anchor NULL>` on non-final components followed by `ligComponent`.

The FEA spec grammar says `<anchor> mark <named mark class>` for each attachment, but the spec's own [GPOS type 5 example](https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#6.e) uses `<anchor NULL>;` without `mark @class`. feaLib has followed the example rather than the grammar since its [original GPOS type 5 implementation](https://github.com/fonttools/fonttools/blob/0532ec85d30b/Lib/fontTools/feaLib/parser.py#L200-L201) (Dec 2015).

The fix peeks at the anchor before consuming it: if NULL and no `mark` follows, skip mark-class parsing. Non-NULL anchors without `mark` still error as before.

### DIFF
--- a/fea-rs/src/parse/grammar/gpos.rs
+++ b/fea-rs/src/parse/grammar/gpos.rs
@@ -217,12 +217,18 @@ fn anchor_mark(parser: &mut Parser, recovery: TokenSet) -> bool {
         return false;
     }
     parser.in_node(AstKind::AnchorMarkNode, |parser| {
+        let is_null = parser.matches(2, Kind::NullKw);
         metrics::anchor(parser, recovery.union(RECOVERY));
-        // we will verify later that the anchor was NULL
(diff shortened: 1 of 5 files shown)


anthrotype:
For reference, this is feaLib's equivalent:
https://github.com/fonttools/fonttools/blob/67ff24a36/Lib/fontTools/feaLib/parser.py#L200-L201

rsheeter (approved):

Issue #1814, opened 2025-12-10
fontmake can include point axes not included by fontc


cmyr:
For instance in the case of piazzolla: fontmake has the following fvar entry, which is missing from fontc:

```
    <Axis>
      <AxisTag>opsz</AxisTag>
      <Flags>0x0</Flags>
      <MinValue>30.0</MinValue>
      <DefaultValue>30.0</DefaultValue>
      <MaxValue>30.0</MaxValue>
      <AxisNameID>Optical size</AxisNameID>
    </Axis>
```

This causes differences to appear in various other tables, like MVAR/HVAR/name.

reproduction:

```
python3 -m ttx_diff 'https://github.com/huertatipografica/piazzolla?fbd775f98a#sources/Piazzolla-Italic.glyphs'
```

anthrotype:
there is a 9 year old TODO in fontTools/varLib.py 

```
# TODO Skip axes that have no variation.
```

https://github.com/fonttools/fonttools/blob/1419daa6680c5c899713598956caa69ad47bcd23/Lib/fontTools/varLib/__init__.py#L104

@behdad should we just prune when min==default==max like fontc does, or do you think there are legitimate use cases for wanting to build an fvar like that?

behdad:
I'm not sure what would break. That said, with avar2 we might want to keep point axes as they might have influence through avar2 mapping, specially for hidden axes.

cmyr:
closing this in favour of #1990, which has more info

PR #2018, opened 2026-05-18
[ttx_diff] Match fontc's logic for detecting variable vs static sources

### LINKED ISSUE #1860: some error in how we decide to build variable tables
It looks like this font defines an axis but only one master, and we're generating fvar and company based on that?

```
python3 -m ttx_diff 'https://github.com/notofonts/makasar?e7902cc5d8#sources/NotoSerifMakasar.designspace'
```

anthrotype:
ttx_diff was using source/master count to decide whether to tell fontmake to build variable or static. fontc uses a different criterion: whether any axis has a non-zero range (min != default or max != default), or in other words it's not a "point" axis. 

This caused ttx_diff to instruct fontmake to build static for single-source designspaces that declare axis ranges, while fontc built them as variable (arguably degenerate having a single master, but GIGO), making the comparison meaningless.

I changed ttx_diff to use fontc's axis-range check for both .designspace and .glyphs sources.

For .glyphs, axis ranges need to be computed from master axis values plus virtual master positions, matching fontc's ir_axes() logic and glyphsLib/builder/axes.py.

I verified that NotoSerifMakasar (1 source, wght 400-700) now produces "output is identical".
Mingzat (1 source, 3 axes, 2 point) now produces a comparable variable build (the remaining fvar/STAT/HVAR/name diffs are #1814: fontc drops point axes that the designspace explicitly declares).

Supersedes #1422.
Fixes #1860

### DIFF
--- a/ttx_diff/src/ttx_diff/core.py
+++ b/ttx_diff/src/ttx_diff/core.py
@@ -438,10 +438,27 @@ def source_is_variable(path: Path) -> bool:
         return False
     if path.suffix == ".designspace":
         dspace = DesignSpaceDocument.fromfile(path)
-        return len(dspace.sources) > 1
+        return any(
+            a.minimum != a.default or a.maximum != a.default for a in dspace.axes
+        )
     if path.suffix == ".glyphs" or path.suffix == ".glyphspackage":
         font = GSFont(path)
-        return len(font.masters) > 1
+        # Virtual masters can extend axis min/max beyond what real masters define
+        # https://github.com/googlefonts/glyphsLib/blob/75c07d42/Lib/glyphsLib/builder/axes.py#L168-L173
+        virtual_masters = [
+            cp.value
+            for cp in font.customParameters
+            if cp.name == "Virtual Master" and not getattr(cp, "disabled", False)
+        ]
+        for i, axis in enumerate(font.axes):
+            values = [m.axes[i] for m in font.masters]
+            for vm in virtual_masters:
+                for entry in vm:
+                    if entry.get("Axis") == axis.name:
+                        values.append(entry["Location"])
+            if min(values) != max(values):
+                return True
+        return False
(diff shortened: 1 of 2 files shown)


ttx_diff/src/ttx_diff/core.py
```diff
@@ -438,10 +438,27 @@ def source_is_variable(path: Path) -> bool:
...
     if path.suffix == ".designspace":
         dspace = DesignSpaceDocument.fromfile(path)
-        return len(dspace.sources) > 1
+        return any(
+            a.minimum != a.default or a.maximum != a.default for a in dspace.axes
+        )
     if path.suffix == ".glyphs" or path.suffix == ".glyphspackage":
         font = GSFont(path)
-        return len(font.masters) > 1
+        # Virtual masters can extend axis min/max beyond what real masters define
```
rsheeter, line 446: I've been working on font compilers for over a year and I don't know what this means, what's a "virtual master"?

rsheeter (approved):

ttx_diff/src/ttx_diff/core.py
anthrotype: https://github.com/googlefonts/fontc/issues/62#issuecomment-1433347088

simoncozens: A single master font may actually be variable; it could vary with a bracket layer, or via feature variation substitutions. Glyphs doesn't define axis ranges up-front, so a virtual master tells the compiler about the range of variation.

anthrotype: > it could vary with a bracket layer

a brace (intermediate) layer you mean. That's exactly what Virtual Masters are for. Extend the axis range to allow intermediate layers in specific glyphs to be defined along axis coordinates that extend beyond the range of the font-wide masters'.

PR #2021, opened 2026-05-27
Read openTypeCategories from DS lib; infer from GlyphData.xml for UFO sources

### LINKED ISSUE #767: GDEF glyph classes in IR
In order to properly generate mark attachment rules, we need to know the GDEF class assignments for glyphs. This information can be specified in various ways:

- in glyphs.app, it is a property on the individual glyphs.
- in UFO, it is an optional field in the root lib, [`public.openTypeCategories`](https://unifiedfontobject.org/versions/ufo3/lib.plist/#publicopentypecategories)

In addition to these locations, the GDEF classes can also be specified via [FEA syntax](http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#9b-gdef-table), and this is the main source of problems. Basically: we would _really_ like to have a single source of truth here. There is also a chicken-egg problem, where for _generated_ features we need to know these class rules.

So: I would like to decide on some standardized behaviour, and I would like to move these classdefs into IR. 

- for glyphs.app sources, the source of truth will always be per-glyph categories.
    - Q: _are these always available, or only if explicitly set by the user? Otherwise I assume they are computed from `GlyphData.xml`?_
- for UFO/designspace sources, the source of truth will be `public.openTypeCategories` dictionary, if present.

Issues to consider (and my current sense of how to address them)

- what if a UFO does not have the `public.openTypeCategories`, but does have explicit classes defined in FEA? _we use the categories from the FEA_.
- what if a UFO has both, and they don't match? _print a warning and use the lib version_.
- what if UFO doesn't have either? _good question_. _Probably I would synthesize one from `GlyphData.xml`? Is there a reason this is bad?_

the FEA spec says that if the GDEF classes are not stated explicitly then they should be generated based on what glyphs appear in mark/base/lig mark attachment rules, but since we are generating most of these lookups automatically, _this really doesn't work for us_. We could infer these based on the lookups we generate, but this runs into possible problems if there are non-base glyphs that include an anchor that is used to attach a component, but which are not expected to participate in mark-to-base? @anthrotype mentioned this as a possible situation, although I'm not sure how common it is.

TL;DR:

- let's add GDEF glyph class defs to IR
- let's get it "not from FEA" by default, and fall back to getting it from FEA
- let's use `GlyphData.xml` to synthesize it if necessary (always for glyph.app except for manual overrides; only for UFO if no lib key _and_ no GDEF block in FEA)

anthrotype:
Fixes #767. Companion to googlefonts/ufo2ft#985 and googlefonts/fontmake#1168.

Two changes to align fontc's UFO/DS category handling with fontmake:

1. Read `public.openTypeCategories` from the designspace lib first, falling back to the default master's UFO lib. The DS lib is the canonical location for designspace sources -- previously fontc only checked the UFO lib. Matches ufo2ft's `OpenTypeCategories.load()` [precedence](https://github.com/googlefonts/ufo2ft/blob/46196892e8/Lib/ufo2ft/util.py#L696-L704).
  
2. Infer preliminary GDEF categories from GlyphData.xml when no explicit `public.openTypeCategories` exist and `propagateAnchors` is enabled. Marks (Nonspacing, Spacing Combining) and ligatures are classified from GlyphData.xml; bases are inferred after anchor propagation via `recompute_gdef_categories` (with `infer_from_anchors: true`). This matches fontmake#1168's build-time category generation, producing identical GDEF output for DS/UFO sources without explicit categories.
  
`GlyphData::new()` made `pub` in `glyphs-reader` so `ufo2fontir` can construct a query instance. The `ufo2fontir` → `glyphs-reader` dependency is a stopgap. Since the `glyphdata` module is self-contained (no dependency on .glyphs parsing), ideally it could move to its own crate so both format-specific frontends can share it without the cross-format coupling. I can do that in a follow-up PR.

Verified locally against several crater fonts with GDEF diffs caused by missing categories (Assistant, MerriweatherSans, MerriweatherSans-Italic, Vollkorn, Vollkorn-Italic, OpenSans-Roman, OpenSans-Italic). The first 5 go from GDEF class diffs to "output is identical". The two OpenSans variants now have identical GDEF/GPOS; residual hmtx/maxp diffs are unrelated.

### DIFF
--- a/ufo2fontir/src/source.rs
+++ b/ufo2fontir/src/source.rs
@@ -607,6 +607,59 @@ fn glyph_categories(
     Ok(categories)
 }
 
+/// Generate preliminary GDEF categories from GlyphData.xml for UFO sources
+/// without explicit `public.openTypeCategories`.
+///
+/// Only marks and ligatures are assigned here; bases are inferred after anchor
+/// propagation by `recompute_gdef_categories` (with `infer_from_anchors: true`).
+fn preliminary_gdef_categories_from_glyphdata(
+    glyph_names: &HashSet<GlyphName>,
+) -> PreliminaryGdefCategories {
+    use glyphs_reader::glyphdata::{Category, GlyphData, Subcategory};
+
+    let glyph_data = GlyphData::new(None);
+    let mut categories = BTreeMap::new();
+    let mut mark_category_glyphs = BTreeSet::new();
+
+    for name in glyph_names {
+        // TODO(anthrotype): pass codepoints for better resolution
+        // https://github.com/googlefonts/fontc/issues/2022
+        let Some(result) = glyph_data.query(name.as_str(), None) else {
+            continue;
+        };
+        if result.category == Category::Mark {
+            mark_category_glyphs.insert(name.clone());
+        }
+        match (result.category, result.subcategory) {
+            (Category::Mark, Some(Subcategory::Nonspacing | Subcategory::SpacingCombining)) => {
+                categories.insert(name.clone(), GlyphClassDef::Mark);
+            }
+            (_, Some(Subcategory::Ligature)) => {
+                categories.insert(name.clone(), GlyphClassDef::Ligature);
+            }
+            _ => {}
+        }
+    }
+
+    log::info!(
+        "Inferred preliminary GDEF categories from GlyphData.xml: {} mark, {} ligature",
+        categories
+            .values()
+            .filter(|c| **c == GlyphClassDef::Mark)
+            .count(),
+        categories
+            .values()
+            .filter(|c| **c == GlyphClassDef::Ligature)
+            .count(),
+    );
+
+    PreliminaryGdefCategories {
+        categories,
+        infer_from_anchors: true,
+        mark_category_glyphs,
+    }
+}
+
 fn postscript_names(lib_plist: &plist::Dictionary) -> Result<Option<PostscriptNames>, BadSource> {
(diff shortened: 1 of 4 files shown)


anthrotype:
> Verified locally against several crater fonts with GDEF diffs...

... only after installing the patched ufo2ft and fontmake in fontc's `.venv`.
We'll need to wait for https://github.com/googlefonts/ufo2ft/pull/985 and https://github.com/googlefonts/fontmake/pull/1168 to be released and then bump fontc's `requirements.txt`.

anthrotype:
I just realized the GlyphData.xml queries don't pass codepoints, so glyphs with non-standard names that aren't in GlyphData by name won't be resolved correctly. Filed #2022 to track this.

anthrotype:
> The ufo2fontir → glyphs-reader dependency is a stopgap

https://github.com/googlefonts/fontc/pull/2023

ufo2fontir/src/source.rs
```diff
@@ -607,6 +607,59 @@ fn glyph_categories(
...
+                categories.insert(name.clone(), GlyphClassDef::Mark);
+            }
+            (_, Some(Subcategory::Ligature)) => {
+                categories.insert(name.clone(), GlyphClassDef::Ligature);
+            }
+            _ => {}
+        }
+    }
+
+    log::info!(
```
cmyr, line 648: this smells a bit bad, if we really want to log this info maybe we should just keep track of the counts as we're adding them?

cmyr (approved):
looks good, one little nit that you can ignore. :)

PR #2023, opened 2026-05-27
move `glyphdata` to fontdrasil crate, feature-gated

anthrotype:
The glyphdata module (GlyphData.xml query API, bundled binary data, generated enum conversions) has no dependency on .glyphs parsing code. Pulling it out of glyphs-reader would let crates like ufo2fontir depend on glyph category/script lookups without dragging in the entire .glyphs parser (see #2021, which I plan to rebase onto this).

Concretely: `glyphdata.rs`, `glyphdata_bundled.rs`, `glyphslib_enums.rs`, the `resources/*.dat` files, and `data/update.py` all move into a new `glyphdata` crate. glyphs-reader drops `quick-xml` and `icu_properties` since only glyphdata used them.

No logic changes, pure structural move.


cmyr:
maybe this should go in fontdrasil, alongside the aglfn? It could be feature gated, if necessary.

anthrotype:
> maybe this should go in fontdrasil, alongside the aglfn? It could be feature gated, if necessary.

makes sense. done!

rsheeter:
I'm not finding it entirely obvious why fontdrasil having this is any better than ufo2fontir having a dependency on glyphs-reader?

Having a separate glyphdata crate seems cleanest, at cost of having Yet Another Crate to worry about. That's not _free_ but it's less of an issue for fontc than fontations.

cmyr:
to my understanding fontdrasil is the place where we put things that are compilation-specific and shared between different crates, and the glyphsdata feels like it is reasonably analogous to the agl/aglfn?

having ufo2fontir depend on glyphs-reader feels like a bigger mixing of concerns, although given that it's going to be in the binary already this isn't a major _practical_ concern...

anthrotype:
i don't have a strong opinion. We're going to enable `glyphdata` feature anyway in fontc. 
Do we know of/forsee any clients of fontdrasil that wouldn't want to also pull in this and the quick-xml and icu_properties deps?
A separate `glyphdata` crate is simpler, and I don't think this is something we need update very frequently.

I can easily git reset and force-push to the previous state, just let me know.

We could also just have ufo2fontir depend on glyphs-reader directly and probably nobody would notice/care. It's just that it feels a bit off..

cmyr:
I mean I don't want to block you on something nitty, go with your best judgement.

rsheeter:
It feels weird to have glyphs specific data in fontdrasil but ... I guess you can argue that data is not glyphs specific anymore?

I would make ufo2fontir depend on glyphs-reader because the feature we're trying to implement is literally a ufo setup that says "do this glyphs specific thing" but I don't feel very strongly about it, any of the proposed options work.

anthrotype:
ok then let's keep as is (ufo2fontir depends on glyphs-reader) and we can revisit this later if need be.

PR #2024, opened 2026-05-28
Warn when default master customParameters differ from font-level values

### LINKED ISSUE #1856: different precedence for customParameters / properties
In at least one font, the same keys occur in the `customParameters` and `properties` dicts of the font itself, as well as for each of the masters, and they do not agree.

- python prefers the values set on the font
- we prefer the value set on the default master

```
python3 -m ttx_diff 'https://github.com/notofonts/bengali?7ab89a0473#sources/NotoSerifBengali.glyphspackage'
```

anthrotype:
Closes #1856

fontc correctly uses font-level custom parameters for font-wide properties like unicodeRanges, fsType, Use Typo Metrics, etc. fontmake (via glyphsLib) accidentally also lets the default master override these, simply because master-level handlers run after font-level ones, overwriting them, and they don't clearly distinguish what belongs to the font as a whole or a particular master.

Glyphs.app doesn't let one add these to a master but some fonts (e.g. NotoSerifBengali.glyphspackage) happen to have some of these in both the font and the default master, and with different values!
It's not worth trying to match this because it's clearly a font source bug that happens to rely on an unintentional quirk in glyphsLib. 

So in this PR, we simply detect when this happen and emit a warning about any such discepancies, so that font developers can fix their sources.

The full list of the font-level custom parameters which are _not_ supported at master level, and for which fontc from now on will emit a helpful message is the following:

- Use Typo Metrics
- Has WWS Names
- fsType / openTypeOS2Type
- isFixedPitch / postscriptIsFixedPitch
- unicodeRanges / openTypeOS2UnicodeRanges
- codePageRanges
- openTypeHeadFlags
- openTypeHeadLowestRecPPEM
- Don't use Production Names
- GASP Table
- panose / openTypeOS2Panose

There are also Glyphs format 2 font-level custom parameters related to name table that have moved to font "properties" in format 3, and which were never supported at the master level. For these we already emit a generic "unknown custom parameter" warning if ever found inside a master, and this is correct because they only belong to the font:

- vendorID / openTypeOS2VendorID
- license / openTypeNameLicense
- licenseURL / openTypeNameLicenseURL
- trademark
- description / openTypeNameDescription
- sampleText / openTypeNameSampleText
- compatibleFullName
- WWSFamilyName / openTypeNameWWSFamilyName
- designers
- designerURL
- manufacturers
- manufacturerURL

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -387,6 +387,8 @@ impl Work<Context, WorkId, Error> for StaticMetadataWork {
             .map(|v| -v)
             .unwrap_or(0.0);
 
+        warn_master_font_custom_param_mismatch(font);
+
         let mut selection_flags = match font.custom_parameters.use_typo_metrics.unwrap_or_default() {
             true => SelectionFlags::USE_TYPO_METRICS,
             false => SelectionFlags::empty(),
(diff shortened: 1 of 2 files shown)


cmyr (approved):
makes sense, now we need to just get in the habit of reading our warnings 🤕

Issue #1829, opened 2025-12-12
the 'non identical feature files' check is naive


cmyr:
In that it only checks file length. This leads to a false negative in one case, `NotoSansThai`.

Longer term we need to agree to an approach for these cases. It would be nice if we could get the sources to use variable FEA, but barring that we could consider a compile + merge approach.

(when I _do_ add code to more aggressively check if these files are identical, I only encounter one failure, so this particular little case isn't very widespread)
```
python3 -m ttx_diff 'https://github.com/notofonts/thai?f8f3f02470#sources/NotoSansThai.designspace'
```

rsheeter:
> It would be nice if we could get the sources to use variable FEA

Can we easily enumerate all the sources that have non-identical feature files? - I would like to poke at moving them to varfea because I really really don't want to merge. That said, if we did have to merge ... would we need to _compile_ or could we merge ASTs?

cmyr:
so there's only one source I know of with non-identical feature files that we aren't detecting, which is `NotoSansThai.designspace`.  From some quick and dirty text munging through the crater failures list, the following fonts are failing the existing check:


```
EkType/Anek/sources/AnekBangla/Masters/AnekBangla.designspace
EkType/Anek/sources/AnekGujarati/Masters/AnekGujarati.designspace
EkType/Anek/sources/AnekGurmukhi/Masters/AnekGurmukhi.designspcae
EkType/Anek/sources/AnekLatin/Masters/AnekLatin.designspace
EkType/Anek/sources/AnekOdia/Masters/AnekOdia.designspace
EkType/Anek/sources/AnekTelugu/Masters/AnekTelugu.designspace
TypeNetwork/Arimo/sources/Arimo-Italic.designspace
TypeNetwork/Arimo/sources/Arimo.designspace
dy/linefont/sources/Linefont.designspace
Roboto-Flex
hyvyys/Tektur/sources/UFO/Tektur.designspace
indestructible-type/Jost/sources/designspace/jostGF-Italic.designspace
indestructible-type/Jost/sources/designspace/jost.designspace
/notofonts/hanifi-rohingya/sources/NotoSansHanifiRohingya.designspace
productiontype/Georama/sources/Georama.designspace
productiontype/Spectral/sources/Spectral-Italic.designspace
productiontype/Spectral/sources/Spectral.designspace
silnrsi/font-andika/source/AndikaItalicdesignspace
silnrsi/font-andika/source/AndikaRoman.designspace
[...]

PR #2027, opened 2026-05-29
Canonicalize include paths when checking features.fea compatibility

anthrotype:
This is an attempt to fix part of #1829, the easier bit (not the merge multiple master features.fea into a variable FEA problem).

Some designspace/UFOs sources (e.g. Roboto-Flex) have masters at different directory depths with features.fea containing relative include paths that point to the _same_ shared file.

Currently we naively check only the lengths of the master UFOs' features.fea files to determine if they are compatible or not. When they only differ because the relative include path differs (while the included file is the same), we reject them as incompatible.

So in this PR, when features.fea files differ across masters, I canonicalize include() paths to absolute paths before comparing them.

Include paths are resolved following the same logic as fea-rs and ufo2ft/feaLib: try the UFO's parent directory first (per the UFO spec, includes resolve relative to where the UFO lives, not inside it), then the directory of the including file.

I decided not to add fea-rs as a dependency to ufo2fontir because it feels a bit like breaking the layer separation between frontend and backend (only fontbe depends on fea-rs right now). Also, parsing the features.fea files just to compare them for equality and then throwing them away feels a bit wasteful, though that's more or less what ufo2ft does when it uses the IncludingLexer (not a full parser) to resolve include statements before comparing them.

This also fixes another latent bug: by only comparing file lengths, we can sometimes assume master features.fea are compatible when in fact they aren't (e.g. they vary in the value records but happen to have the same total file length) and silently take the first one and produce an incorrect font without variations. The correct thing to do in this case is to error (until a proper merge variable feature support is added).

### DIFF
--- a/ufo2fontir/src/source.rs
+++ b/ufo2fontir/src/source.rs
@@ -2256,6 +2256,55 @@ mod tests {
 
     use super::*;
 
+    #[test]
+    fn fea_files_differing_only_by_trailing_newline_are_identical() {
+        // https://github.com/.../Intel-One-Mono: two masters whose features.fea
+        // differ only by a trailing newline must be treated as compatible (as
+        // ufo2ft does, since it compares tokens and ignores whitespace).
+        let tmp = tempfile::tempdir().unwrap();
+        let f1 = tmp.path().join("a.fea");
+        let f2 = tmp.path().join("b.fea");
+        std::fs::write(&f1, "feature kern { pos a b -5; } kern;").unwrap();
+        std::fs::write(&f2, "feature kern { pos a b -5; } kern;\n").unwrap();
+        assert!(fea_files_identical(&f1, &f2).unwrap());
+    }
+
+    #[test]
+    fn fea_files_differing_only_by_comments_are_identical() {
+        // comments don't affect the compiled features, so two masters that
+        // differ only by comments must be treated as compatible (as ufo2ft,
+        // which excludes comments from its token comparison).
+        let tmp = tempfile::tempdir().unwrap();
+        let f1 = tmp.path().join("a.fea");
+        let f2 = tmp.path().join("b.fea");
+        std::fs::write(&f1, "# master A\nfeature kern { pos a b -5; } kern;\n").unwrap();
+        std::fs::write(&f2, "feature kern { pos a b -5; } kern; # note\n").unwrap();
+        assert!(fea_files_identical(&f1, &f2).unwrap());
+    }
+
+    #[test]
+    fn fea_files_include_whitespace_and_relative_paths_normalized() {
+        // `include (path)` with whitespace before the paren is valid FEA (the
+        // old literal "include(" scan missed it), and two masters including the
+        // same file via paths that resolve identically must compare equal.
+        let tmp = tempfile::tempdir().unwrap();
+        std::fs::write(
+            tmp.path().join("shared.fea"),
+            "feature liga { sub a b by c; } liga;\n",
+        )
+        .unwrap();
+        let write_master = |name: &str, include: &str| {
+            let ufo = tmp.path().join(name);
+            std::fs::create_dir(&ufo).unwrap();
+            let fea = ufo.join("features.fea");
+            std::fs::write(&fea, include).unwrap();
+            fea
+        };
+        let a = write_master("A.ufo", "include(../shared.fea)");
+        let b = write_master("B.ufo", "include (../shared.fea)\n");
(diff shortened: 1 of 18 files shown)


anthrotype:
that's a feature, not a bug :)

```
$ fontc googlesans/source/GoogleSans/GoogleSans.designspace -o build/GoogleSans.ttf
[2026-05-29T17:45:44.426745Z ThreadId(1) fontc ERROR] feature files are non-identical: googlesans/source/GoogleSans/GoogleSans-opsz18-wght734-GRAD0.ufo/features.fea, googlesans/source/GoogleSans/GoogleSans-opsz18-wght380-GRAD200.ufo/features.fea
```

ufo2fontir/src/source.rs
```diff
@@ -740,23 +740,81 @@ fn units_per_em<'a>(font_infos: impl Iterator<Item = &'a norad::FontInfo>) -> Re
...
-        .metadata()
-        .map_err(|e| BadSource::new(f1, BadSourceKind::Io(e)))?;
-    let m2 = f2
-        .metadata()
-        .map_err(|e| BadSource::new(f2, BadSourceKind::Io(e)))?;
-    if m1.len() != m2.len() {
-        return Ok(false);
+    let c1 = std::fs::read_to_string(f1).map_err(|e| BadSource::new(f1, BadSourceKind::Io(e)))?;
+    let c2 = std::fs::read_to_string(f2).map_err(|e| BadSource::new(f2, BadSourceKind::Io(e)))?;
+    if c1 == c2 {
```
cmyr: although it probably doesn't matter this in theory could still be incorrect if two files with identical text both have includes but are actually in different directories.

Similarly there are cases at the moment where the current check passes because the length is the same, but one file has like `include!(deva.fea)` and the other has `include!(arab.fea)`.

ufo2fontir/src/source.rs
```diff
@@ -740,23 +740,81 @@ fn units_per_em<'a>(font_infos: impl Iterator<Item = &'a norad::FontInfo>) -> Re
...
+/// directory first, then the directory containing the including file.
+///
+/// Per the UFO spec, includes resolve relative to the directory where the
+/// UFO lives, not inside the UFO package itself.
+fn canonicalize_fea_includes(content: &str, fea_path: &Path) -> String {
+    let fea_dir = fea_path.parent().unwrap_or(Path::new("."));
+    let ufo_parent_dir = fea_dir.parent().unwrap_or(fea_dir);
+    let mut result = String::with_capacity(content.len());
+    let mut remaining = content;
+    while let Some(start) = remaining.find("include(") {
```
cmyr: i really don't love us manually parsing. For instance the current code would match a commented out line, or a string literal.

The more robust option would be to use the fea-rs tokenizer. See https://github.com/googlefonts/fontc/blob/main/fea-rs/src/bin/highlight.rs for a simple example that tokenizes a file, if you're curious.

cmyr (approved):
so I think that this approach is fragile and overly focused to a pretty narrow set of cases. I'm okay with merging this as an explicitly temporary measure, but longer term we want to just handle the merge I think, it shouldn't be that tricky.

ufo2fontir/src/source.rs
anthrotype: Yeah. I explicitly did not do that to avoid having ufo2fontir depend on fea-rs, but if you think that's fine, it certainly would make this feel less hacky.

cmyr: I think it's probably best? Alternatively we could also consider rethinking this so that this check moves into fontir or something, but that would require more careful consideration.

anthrotype: I changed it to use fea-rs' parse_string and FileSystemResolver and compare two iterator of significant tokens (excluding whitespace and comments). Much better

ufo2fontir/src/source.rs
anthrotype: I'm comparing them unconditionally now

anthrotype:
GoogleSans failing is expected.

@cmyr I'd curious to know what is the net loss/gain from this PR over the whole fontc crater, could you run one for me? I don't have enough local storage 😁

rsheeter (approved):
LGTM but would ideally like CI clean, per IM perhaps we could point to your branch of GS or some other font that is interesting as a smoke test until such time as GS is updated to varfea. Is there an issue for that upgrade?

anthrotype:
> Is there an issue for that upgrade?

https://github.com/googlefonts/googlesans/issues/708

PR #2028, opened 2026-06-02
Read zipped UFO (.ufoz) sources

felipesanches:
fontc accepted `.ufo` and `.designspace` but rejected `.ufoz` (a zipped UFO) with UnrecognizedExtension, even though fontmake / fontTools.ufoLib read it. Some upstreams commit only the compact `.ufoz` and unzip it to a build-time `temp/*.ufo`, so google-fonts-sources / fontc_crater report a spurious "missing source" for the fontc side (e.g. aaronbell/LxgwMarkerGothic).

- fontc::Input::new routes `*.ufoz` to the designspace/UFO source; the args.rs help text and ttx_diff's source_is_variable learn the extension.

- DesignSpaceIrSource::new unpacks a `.ufoz` into a tempfile::TempDir and builds from the inner `.ufo`. fontc reads a UFO as a directory tree lazily during compilation, so extraction-to-temp fits; the TempDir is held in the source (Workload owns the source for the whole build) so the unpacked files outlive every work item.

- The UFO payload always lands in a synthetic `<stem>.ufo`, whether the archive wraps it in one top-level directory (`zip -r`) or stores the contents at the root (ufoLib). A single non-junk top-level directory is stripped; archiver junk (__MACOSX/, .DS_Store, Thumbs.db) is ignored.

- The `.ufoz` is untrusted (fontc_crater builds arbitrary repos): entries are written through a size-limited copy (2 GiB ceiling) so a decompression bomb cannot exhaust RAM; zip's enclosed_name guards path traversal.

- Adds the `zip` dependency (default-features off, deflate only; flate2 is already in the tree).

Tests: wrapped/root-contents/junk layouts, an end-to-end build of glyph IR from a `.ufoz` (exercising the lazy temp-dir reads) with parity against the `.ufo`, and error paths (invalid archive, missing metainfo).

Assisted by an AI agent (Claude Opus 4.8)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,8 @@ clap = { version = "4.0.32", features = ["derive"] }
 rayon = "1.6"
 icu_properties = "=2.1"
 smallvec = {version = "1.15", features = ["const_new"]}
+tempfile = "3.3.0"
+zip = { version = "2.1", default-features = false, features = ["deflate"] }
 
 # fontations etc
 write-fonts = { version = "0.44.1", features = ["serde", "read"] }
@@ -29,7 +31,6 @@ norad = { version = "0.18.2", default-features = false }
 # dev dependencies
 criterion = "0.7"
 diff = "0.1.12"
-tempfile = "3.3.0"
 more-asserts = "0.3.1"
 pretty_assertions = "1.3.0"
 temp-env = "0.3.3"
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -15,7 +15,7 @@ use crate::Error;
 #[derive(Serialize, Deserialize, Parser, Debug, Clone, PartialEq)]
 #[command(version)]
 pub struct Args {
-    /// A designspace, ufo, or glyphs file
+    /// A designspace, ufo, ufoz, glyphs, glyphspackage, or fontra source
     #[arg(
         conflicts_with = "source",
         required_unless_present("source"),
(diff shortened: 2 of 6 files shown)


felipesanches:
I reviewed it and it looks good to me.

The security measure (block decompression of data that would result in more than 2 gigabytes) seems a bit paranoid. But still correct, so I think it is fine.

The removal of "junk" folders at first sounded to me as a bit superfluous, but in reality it affects the decision of whether or not a top-level directory is created or if the UFO contents are unzipped directly at the root dir. So removing those junk folders ensures the correct detection is made, leading to proper handling of the extracted files.

This PR will enable us to load `.ufoz` files and, with that, clean up a few of the remaining outliers in [fontc_crater's report](https://googlefonts.github.io/fontc_crater/) (among those "failed to find" targets)

felipesanches:
I've alsoconsidered whether decompressing to filesystem would cause performance issues, but (at least on Linux) these files are save on a temporary directory at /tmp mounted as tmpfs which effectively saves directly to RAM. Other code-paths in the existing fontc code seem to already handle UFOs on the filesystem, so I think that this PR is not the right place for us to try to deal with this matter (if we even think it should be adddressed at all).

cmyr:
so my preference for handling this would be to add ufoz support in norad, and there is a draft PR up that accomplishes that, at https://github.com/linebender/norad/pull/407.

I did consider that it might be easier to add this support directly into fontc, but I think in general it makes sense for norad to support this if anyone does, because it _is_ [part of the UFO spec](https://unifiedfontobject.org/versions/ufo3/#storage-formats).

rsheeter:
+1 to norad, does that mean we close this?

felipesanches:
yes! ;-)

PR #2036, opened 2026-06-11
Allow conditions to use float coordinates too, to match feaLib

### LINKED ISSUE #2034: Coordinates in condition sets should be allowed to be floats
Currently, a feature file like this:

```
conditionset italic_only {
    ital 0.1 1;
} italic_only;

variation rlig italic_only {
    sub a by b;
} rlig;
```

trips up fontc with a:

```
[2026-06-11T11:13:27.303160Z ThreadId(1) fontc ERROR] FEA parsing failed with 1 errors
[2026-06-11T11:13:27.303191Z ThreadId(1) fontc WARN] error: Expected NUM found FLOAT
in a.fea at 2:9
  |
2 |     ital 0.1 1;
  |          ^^^
```

feaLib allows floaty coordinates.

To test, you can make an empty UFO with the above feature code and call fontc. It should parse it correctly and then later crash because there's no ital axis in a single UFO.

Hoolean:
Closes #2034

One day we could also support the coordinate-space suffixes we allow elsewhere, but this PR gets us immediate parity with `feaLib`, and resolves a crash for us on real-world sources.

(See https://github.com/fonttools/fonttools/blob/c5d5359f/Lib/fontTools/feaLib/parser.py#L1978-L1986)


anthrotype (approved):
LGTM!

PR #2040, opened 2026-06-11
[kern] resolve partial-master kern exceptions per-cell to avoid phantom values

### LINKED ISSUE #2035: Variable kern generation emits phantom values for partially-populated conflicting kerning exceptions
This is the fontc counterpart of googlefonts/ufo2ft#988 (see that for the full mechanics). fontbe's kern generation is a documented port of ufo2ft's algorithm: `align_kerning` in fontbe/src/features/kern.rs defines every kerning pair at every location, backfilling pairs missing from an instance via `lookup_kerning_value` (the official [UFO kerning value lookup](https://unifiedfontobject.org/versions/ufo3/kerning.plist/#kerning-value-lookup-algorithm) algorithm).

A glyph-to-class kern exception present in only some masters gets backfilled, at the masters that omit it, with the class-to-class value (or 0 when none exists): resolving the whole key can only generalize the glyph side, so the cascade never reaches a competing class-to-glyph exception that covers one of the key's cells. Since pairs are sorted most-specific-first when building the kern lookups (glyph-glyph, then glyph-to-class, then class-to-glyph, then class-to-class) and the first matching rule wins, the backfilled glyph-to-class pair shadows the class-to-glyph exception on the cells they share, so those cells render a value the master source never resolves to (a "phantom").

Confirmed empirically with a repro extracted from Google Sans, where we originally hit this. fontc produces the identical phantoms as unpatched ufo2ft. The bug is in the kern generation, not in fea-rs, which faithfully compiles whatever rules it is given.

The fix will mirror googlefonts/ufo2ft#990: resolve glyph-to-class exceptions at the cell level instead of as a single key. PR follows.

anthrotype:
Fixes #2035 

A glyph-to-class kern exception that is present in only some masters gets backfilled with the more general class-to-class value, which may shadows competing class-to-glyph exceptions on the cells they share.

Resolve glyph-to-class exceptions cell by cell instead: when every member of the class lands on the same values, keep the compact glyph-to-class pair carrying those cell values; otherwise split it into one glyph-glyph pair per member.

Same fix as googlefonts/ufo2ft#990 (for googlefonts/ufo2ft#988), currently pending review there.

The first commit adds the regression test alone, pushed first so CI demonstrates the failure; the second makes it pass. 

Builds on #2038 and #2039.

(Note: until the ufo2ft fix lands and crater's fontmake picks it up, any affected fonts will show kern diffs against fontmake. I will coordinate the merges so crater doesn't got red)

### DIFF
--- a/fontbe/src/features/kern.rs
+++ b/fontbe/src/features/kern.rs
@@ -265,23 +265,11 @@ fn build_variable_kern_adjustments(
             && let Some(members) = ir_groups.groups.get(second_group)
         {
             // group second-side members by their per-location cell value vector
-            let mut by_value: Vec<(KernAdjustments, Vec<GlyphName>)> = Vec::new();
-            for member in members {
-                if glyph_order.glyph_id(member).is_none() {
-                    continue;
-                }
-                let cell = (
-                    ir::KernSide::Glyph(first.clone()),
-                    ir::KernSide::Glyph(member.clone()),
-                );
+            let mut by_value = BTreeMap::<_, Vec<_>>::new();
+            for member in members.iter().filter(|name| glyph_order.contains(*name)) {
+                let cell = (first.clone().into(), member.clone().into());
                 let vector = resolve(&cell);
-                match by_value
-                    .iter_mut()
-                    .find(|(existing, _)| *existing == vector)
-                {
-                    Some((_, members)) => members.push(member.clone()),
-                    None => by_value.push((vector, vec![member.clone()])),
-                }
+                by_value.entry(vector).or_default().push(member.clone());
             }
             if by_value.len() == 1 {
                 // no cell diverges: keep the compact glyph-to-class pair, but
--- a/fontbe/src/features/kern.rs
+++ b/fontbe/src/features/kern.rs
@@ -151,7 +151,7 @@ impl Work<Context, AnyWorkId, Error> for GatherIrKerningWork {
             .map(|(_, ki)| (ki.location.clone(), ki.as_ref().to_owned()))
             .collect();
 
-        let adjustments = build_variable_kern_adjustments(&ir_groups, &kern_by_pos);
+        let adjustments = build_variable_kern_adjustments(&ir_groups, &glyph_order, &kern_by_pos);
 
         let adjustments: Vec<_> = adjustments
             .into_iter()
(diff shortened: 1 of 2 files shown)


anthrotype:
fails as expected (actual = left, expected = right, see comments in the test)

https://github.com/googlefonts/fontc/actions/runs/27365717077/job/80864563806?pr=2040#step:4:567

```diff
thread 'tests::partial_master_kern_exception_is_resolved_per_cell' (5527) panicked at fontc/src/lib.rs:2337:9:
assertion failed: `(left == right)`

Diff < left / right > :
 [
<    "iotadasiaoxia @side2.GRK_iotaLeft: [30, 50]",
<    "qhwee-ethiopic @side2.ethi_ca: [30, 30]",
<    "uoA @side2.uoRight: [50, 70]",
>    "iotadasiaoxia iotadasiaoxia: [30, 50]",
>    "iotadasiaoxia iotapsilioxia: [30, 30]",
>    "qhwee-ethiopic ca-ethiopic: [40, 30]",
>    "qhwee-ethiopic cee-ethiopic: [30, 30]",
>    "uoA @side2.uoRight: [30, 70]",
     "@side1.GRK_iotaRight iotapsilioxia: [20, 30]",
     "@side1.GRK_iotaRight @side2.GRK_iotaLeft: [60, 50]",
     "@side1.ethi_qhee ca-ethiopic: [40, 50]",
     "@side1.ethi_qhee @side2.ethi_ca: [30, 50]",
     "@side1.uoLeft uoX: [30, 50]",
     "@side1.uoLeft uoY: [30, 50]",
     "@side1.uoLeft @side2.uoRight: [50, 50]",
 ]
```

fontbe/src/features/kern.rs
```diff
@@ -244,13 +258,83 @@ fn build_variable_kern_adjustments(
...
+                    ir::KernSide::Glyph(member.clone()),
+                );
+                let vector = resolve(&cell);
+                match by_value
+                    .iter_mut()
+                    .find(|(existing, _)| *existing == vector)
+                {
+                    Some((_, members)) => members.push(member.clone()),
+                    None => by_value.push((vector, vec![member.clone()])),
+                }
```
cmyr: total code golf, but I think you can write this as,


```suggestion
            // group second-side members by their per-location cell value vector
            let mut by_value = BTreeMap::<_, Vec<_>>::new();
            for member in members.iter().filter(|name| glyph_order.contains(*name)) {
                let cell = (first.clone().into(), member.clone().into());
                let vector = resolve(&cell);
                by_value.entry(vector).or_default().push(member.clone());
```

cmyr (approved):
make sense!

anthrotype:
I will merge this after the other ufo2ft PR also lands

anthrotype:
ufo2ft https://github.com/googlefonts/ufo2ft/pull/990 was meged and released in https://github.com/googlefonts/ufo2ft/releases/tag/v3.8.2

i'll bump fontc's requirements.txt in a separate PR and merge the two PRs in short order so the fontc crater will be happy

anthrotype:
ufo2t bumped in #2047 ... merging

PR #2041, opened 2026-06-12
[fontir] skip missing components with a warning instead of decomposing the whole glyph

### LINKED ISSUE #1858: we (accidentally?) convert composites to simple glyphs if a component is missing
This appears to be an unintended side effect:

https://github.com/googlefonts/fontc/blob/0267a52700707e70efe98d762484e867eee7a8a8/fontir/src/glyph.rs#L770-L772

The intention here appears to be to inline non-export components, but it has the effect of also causing us to inline components in any glyph with a component that is missing, period.

A simple fix here would be to remove missing components from the component list before we hit this check?

anthrotype:
Fixes #1858

A component pointing at a glyph that doesn't exist used to survive until the glyph-order resolution pass, where the dangling reference tripped `convert_components_to_contours` and decomposed the entire composite into a simple glyph, including its other, valid components. fontmake instead keeps the glyph a composite and skips only the missing component (fontTools' ttGlyphPen logs a "skipped non-existing component" warning).

Add a `prune_missing_components preflight` in fontir's GlyphOrderWork, before anchor propagation and non-export flattening, that strips component references whose base glyph doesn't exist. It warns once per dropped reference,
matching fontmake, which warns but builds the font anyway.

Test commit first to confirm CI catches the bug, the fix lands in a follow-up.

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -851,6 +851,22 @@ mod tests {
     }
 
     #[test]
+    fn missing_component_does_not_decompose_siblings() {
+        // 'A' references 'B' (present) and 'F' (missing). A missing component
+        // must not cause the whole glyph to be decomposed into a simple glyph;
+        // 'A' should stay a composite referencing only 'B', matching fontmake.
+        // https://github.com/googlefonts/fontc/issues/1858
+        let result = TestCompile::compile_source("glyphs3/MissingComponent.glyphs");
+        let b_gid = result.get_gid("B");
+
+        let RawGlyph::Composite(glyph) = result.read_be_glyph("A") else {
+            panic!("'A' should remain a composite glyph, not be decomposed");
+        };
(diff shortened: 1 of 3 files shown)


anthrotype:
Fails as expected:
https://github.com/googlefonts/fontc/actions/runs/27407481798/job/81000116228?pr=2041#step:4:567

```
thread 'tests::missing_component_does_not_decompose_siblings' (5489) panicked at fontc/src/lib.rs:863:13:
'A' should remain a composite glyph, not be decomposed
```

(Rustfmt and Clippy CI jobs are failing for unrelated reasons, looks like the latest `time` 0.3.48 broke fea-rs, as we don't pin our Cargo.lock, will file a separate issue)

anthrotype:
fontc tests went from RED to GREEN (Clippy/Rustfmt are still red purely from the time 0.3.48 bug)

cmyr (approved):
👍

PR #2044, opened 2026-06-12
Make WARN the default log level

### LINKED ISSUE #2032: Missing components are silently ignored
Possibly related to #1858, this font compiles with no warnings or errors:

```
{
.appVersion = "3414";
.formatVersion = 3;
date = "2026-06-09 16:49:12 +0000";
familyName = "New Font";
fontMaster = (
{
id = m01;
name = Regular;
}
);
glyphs = (
{
glyphname = A;
lastChange = "2026-06-09 16:49:36 +0000";
layers = (
{
layerId = m01;
shapes = (
{
ref = F;
}
);
width = 0;
}
);
unicode = 65;
}
);
unitsPerEm = 1000;
versionMajor = 1;
versionMinor = 0;
}

```

anthrotype:
fontc inherited env_logger's default of ERROR, so genuine problems that are logged at warn (e.g. skipped missing components, glyphs absent from the default master) were invisible unless the user passed --log warn or set RUST_LOG.

fontmake shows warnings by default, and users rely on that to learn their source has issues.

Let's default the filter to warn instead. RUST_LOG and --log can still override in either direction (I verified locally that --log error and RUST_LOG=error both suppress the warnings).

Fixes #2032.

### DIFF
--- a/fontc/src/args.rs
+++ b/fontc/src/args.rs
@@ -115,7 +115,7 @@ pub struct Args {
     #[arg(long = "vv", default_value = "false")]
     pub verbose_version: bool,
 
-    /// Set the log level, either globally or per module.
+    /// Set the log level, either globally or per module. Defaults to warn.
     ///
     /// See <https://docs.rs/env_logger/latest/env_logger/#enabling-logging> for format.
     #[arg(long)]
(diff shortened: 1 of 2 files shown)


cmyr (approved):
lgtm

PR #2045, opened 2026-06-12
Update GitHub Actions to latest stable versions

anthrotype:
GitHub deprecates Node 20 actions on June 16th 2026 (runners will force Node 24), so I took the oppurtunity to bump everything to the current stable major.

JMM


cmyr (commented):
👍

PR #2049, opened 2026-06-17
Stamp fontc version in name

### LINKED ISSUE #2048: Smile while I write my name in you
Make sure fontc writes it's name and version, presumably in name, so we can do things like count how many Google Fonts are built with fontc based on the shipping binaries.

anthrotype:
Fixes #2048

Append fontc's version at the end of the font version string (name ID 5), makeotf-style, so a shipping binary identifies the compiler that built it e.g. to count how many Google Fonts are built with fontc or which version they used etc.

Example: "Version 1.000;fontc 0.6.1-dev.394+gd62ba016.dirty".

The stamp is basically a [SemVer](http://semver.org/) rendering of [`git describe`](https://git-scm.com/docs/git-describe): just the tag on a release ("fontc 0.6.0"), a next-patch `dev` pre-release suffix with the commit count plus a short git SHA between releases; or the crate version when fontc is built without/outside of git (e.g. a `cargo install` build from tarball).

The ";" suffix is blessed by the OpenType [`name` table spec](https://learn.microsoft.com/en-us/typography/opentype/spec/name) as "helpful to separate different pieces of version information".
So it should pass QA tools' checks, and the font version itself before the semicolon is unchanged.

Also: `fontc --version` / `--vv` now report that same string instead of the bare crate version, so a development build no longer looks like a clean release as it is at the moment.

And ttx_diff strips the ";fontc ..." suffix before diffing against fontmake, so it stays out of crater.

### DIFF
--- a/fontbe/src/name.rs
+++ b/fontbe/src/name.rs
@@ -61,12 +61,11 @@ impl Work<Context, AnyWorkId, Error> for NameWork {
             name_records.sort();
         }
 
-        // Stamp the caller's compiler identity (e.g. "fontc 0.6.1-dev...") into
-        // the version string (name ID 5), so the tool that built a shipping font
-        // binary can be identified. The whole stamp is supplied by the caller;
-        // fontbe doesn't assume it was fontc.
-        if let Some(stamp) = context.compiler_version.as_deref() {
-            stamp_compiler_version(&mut name_records, stamp);
+        // Stamp the fontc version into the version string (name ID 5) so a
+        // shipping font binary records which fontc built it (see #2048). fontc
+        // threads the version in; None means don't stamp.
+        if let Some(version) = context.compiler_version.as_deref() {
+            stamp_compiler_version(&mut name_records, version);
         }
 
         context.name.set(Name::new(name_records));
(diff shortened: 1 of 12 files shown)


fontbe/src/name.rs
```diff
@@ -96,3 +143,120 @@ fn merge_name_records(records: Vec<NameRecord>, fea_names: &Name) -> Vec<NameRec
...
+    #[test]
+    fn leaves_non_stamp_fontc_note() {
+        // A human note that merely starts a segment with "fontc " (no version)
+        // is not our stamp (a digit always follows ours) so it survives.
+        let mut records = vec![version_record("Version 1.000;fontc is great")];
+        stamp_compiler_version(&mut records, "fontc 0.6.0");
+        assert_eq!(
+            records[0].string.as_str(),
+            "Version 1.000;fontc is great;fontc 0.6.0"
+        );
```
cmyr: is this something we really care about? it seems like complexity for something that is extremely hypothetical

fontbe/src/name.rs
```diff
@@ -61,11 +61,58 @@ impl Work<Context, AnyWorkId, Error> for NameWork {
...
+/// prior stamp from the same tool.
+///
+/// `stamp` is the caller's `<tool> <version>` (e.g. "fontc 0.6.1-dev..."); the
+/// replace marker `;<tool> ` is derived from its leading token. Replacing rather
+/// than keeping matters when a source version string carries a *stale* stamp
+/// from an earlier build (e.g. a round-tripped `openTypeNameVersion`).
+///
+/// A stamp not of that `<tool> <digit-led version>` shape is still appended, but
+/// re-runs won't detect/replace it (idempotence relies on the shape).
+fn stamp_compiler_version(records: &mut [NameRecord], stamp: &str) {
```
cmyr: this handles the idea of a tool in a very general way, but does that make sense? Is the tool ever _not_ `fontc`? Is there any time we'd want to stamp the compiler version and have it be some arbitrary string?


Basically: if we had clearer invariants here, we could simplify the code quite a bit; as written it's actually quite hard to figure out what's going on.

fontbe/src/name.rs
```diff
@@ -61,11 +61,58 @@ impl Work<Context, AnyWorkId, Error> for NameWork {
...
+    let marker = format!(";{tool} ");
+    for record in records.iter_mut() {
+        if record.name_id != NameId::VERSION_STRING {
+            continue;
+        }
+        let value = &mut *record.string;
+        // A prior stamp is always appended last, so it runs to the end; drop it
+        // before re-stamping. Only treat the marker as a stamp when a digit
+        // follows (the version is SemVer, digit-led), so a human-authored note
+        // like "...;fontc rules" is left untouched.
```
cmyr: except that a human authored note like 'fontc 4ules' _would_ be stripped. Does this matter to anyone?

fontc/src/lib.rs
```diff
@@ -1540,13 +1551,22 @@ mod tests {
         assert_eq!(vec![Path::new("font.ttf")], outputs);
     }
 
+    /// Strip the build-dependent ";fontc <version>" tag that fontc stamps onto
+    /// the version string (name ID 5), so tests can assert stable values.
+    /// See https://github.com/googlefonts/fontc/issues/2048
+    fn strip_fontc_version_tag(s: &str) -> &str {
+        s.split_once(";fontc ").map_or(s, |(head, _)| head)
+    }
```
cmyr, line 1559: this is the code I'd expect us to be actually using to strip an existing record in the stamp routine, above.? If the code in our testing is this simple, probably the business logic can be, too..

cmyr (commented):
broadly looks good but I do wonder why this patch isn't a bit simpler?

anthrotype:
Yeah, you're right, this got too general. I’ll keep stale-version replacement because otherwise the provenance can lie, but collapse the marker logic to `;fontc ` and simplify. Thanks for reviewing

fontbe/src/orchestration.rs
```diff
@@ -864,6 +864,10 @@ type BeContextMap<T> = ContextMap<AnyWorkId, T, BePersistentStorage>;
 pub struct Context {
     pub flags: Flags,
 
+    /// The fontc version to stamp into the name table version string (name ID 5),
+    /// e.g. "0.6.1-dev.394+gd62ba016", or None to not stamp. See name.rs.
+    pub compiler_version: Option<Arc<str>>,
```
cmyr, line 869: okay one last question: is it actually useful for us to store this, instead of jus computing it directly when we want to do the 'stamp'? Do we ever use it for anything else?

anthrotype: Well yeah.. you don't want fontc to depend on git at runtime

EDIT: I probably misunderstood your question

anthrotype: fontbe owns building the name table, fontbe can't see fontc's version, so fontc passes it in via Context, same as any other build input. Or do you mean fontc should post process the name table after fontbe builds one?

anthrotype: I considered a fontbe build.rs and discarded it because fontbe's CARGO_PKG_VERSION is not the same as fontc's, so compiling from outside a git repo would stamp the wrong version

anthrotype:
@cmyr are you ok if I merge this or you want to take another look?

cmyr:
go for it!

PR #2054, opened 2026-07-06
[ci] Build GoogleSans smoke test from its default branch

### LINKED ISSUE #2033: [ci] Point GoogleSans' smoke test to upstream `main` once it has variable FEA
In #2027 I temporarily pinned our CI smoke test to build GoogleSans off of this PR branch https://github.com/googlefonts/ufo2ft/pull/990 where I merged GS's per-master features.fea into a single variable FEA (which fontc supports).

We should revert to building off their `main` branch once that PR or an equivalent one land.

https://github.com/googlefonts/googlesans/issues/708

anthrotype:
googlefonts/googlesans#740 landed the variable FEA on main, so the temporary pin to the merge-fea-ast branch is no longer needed.

Fixes #2033.

JMM

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -224,16 +224,10 @@ jobs:
           which ttx
           ttx --version
 
-      # Temporarily build the variable-FEA branch of GoogleSans (googlefonts/googlesans#740)
-      # as a smoke test. The static per-master features.fea on the GS default branch are
-      # intentionally rejected as incompatible by this check (see #1829), so building main
-      # here would fail CI. Drop this `ref` once variable FEA lands on the GS default branch
-      # (tracked in googlefonts/googlesans#708).
       - name: Check out GS source repository
         uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


anthrotype:
@cmyr actually CI failed 😅 
I was looking into why...

cmyr:
@anthrotype it's the issue with the expired credential that we had discussed earlier.

anthrotype:
no, I had fixed that (refreshed the token). Only the GS Italic build failed, it's failing to load the font info data from one of the UFO files. I'll continue debugging

anthrotype:
looks like the Italic has some malformed trailing whitespace in fontinfo.plist's openTypeOS2WinAscent that's tripping up norad (i'll send a PR to googlesans to fix)

anthrotype:
https://github.com/googlefonts/googlesans/pull/750

cmyr:
oops, sorry. should fix that in norad, in any case..

PR #2056, opened 2026-07-08
[glyphs-reader] Don't let the name-based ligature guess override an explicit set but uknown category

### LINKED ISSUE #2055: Category guessing leads to bizarre results (fontc edition)
This is the fontc equivalent of https://github.com/googlefonts/glyphsLib/issues/756 

If a glyph name is made up of other glyph names concatenated with an underscore (`ttadeva_viramadeva_ttadeva`), Glyphs (/glyphsLib/fontc) assumes it is a ligature and assigns it GDEF category 2. However, Indic *conjuncts* are not ligatures, in the sense that they only have one anchor which applies to the whole glyph (`nukta`) as opposed to per-component anchors (`nukta_1`, `nukta_2` etc.). Assuming that it's a ligature when it isn't means that anchoring rules are not generated for the whole-glyph anchors, and so marks don't attach.

To persuade Glyphs and fontmake that this is not a component-based ligature, the usual trick is to explicitly set `category=Letter; subCategory=Other;` (or equivalently `subCategory=Conjunct;`).

This trick works for fontmake:

```
$ fontmake -o ttf -g Glegoo.glyphs --keep-overlaps --no-production-names  2>/dev/null ; ttx -o - -t GDEF master_ttf/Glegoo-Regular.ttf| grep ttadeva_viramadeva_ttadeva
Dumping "master_ttf/Glegoo-Regular.ttf" to "<stdout>"...
Dumping 'GDEF' table...
      <ClassDef glyph="ttadeva_viramadeva_ttadeva" class="1"/>
```

But fontc ignores the explicit category assignment and reverts to guessing:

```
$ fontc Glegoo.glyphs ; ttx -o - -t GDEF build/font.ttf | grep ttadeva_viramadeva_ttadeva
Dumping "build/font.ttf" to "<stdout>"...
Dumping 'GDEF' table...
      <ClassDef glyph="ttadeva_viramadeva_ttadeva" class="2"/>
```

[Glegoo.glyphs.gz](https://github.com/user-attachments/files/29734287/Glegoo.glyphs.gz)

(nb @rsheeter: This was discovered while converting SFD to Glyphs)

anthrotype:
`parse_category` returns `None` both when a category/subCategory is absent and when it is set but unrecognized. The GlyphData lookup fallback then filled the empty slot from a name-based heuristic, so an underscore-named glyph with
an explicit subCategory that fontc doesn't recognize could silently revert to a Ligature (breaking mark attachment).

Track instead whether the source sets an explicit value and only fall back when it doesn't. This also matches glyphsLib, which uses the subCategory override string verbatim and and only falls back to guessing when it it's unset/falsy.

This should prevent future footguns if/when Glyphs.app adds new subCategories that fontc doesn't recognize yet.

Closes https://github.com/googlefonts/fontc/issues/2055

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -3168,6 +3168,17 @@ impl RawGlyph {
         let mut sub_category = parse_category(self.sub_category.as_deref(), &self.glyphname);
         let mut production_name = self.production_name;
 
+        // Whether the source explicitly set a non-empty category/subCategory. A value that
+        // was set but not recognized parses to None (parse_category logs and drops it), so
+        // we track "was set" separately to distinguish it from "absent". When the author
+        // set a value, we don't let the name-based GlyphData lookup override it -- even if
+        // we failed to parse it -- otherwise an underscore-named glyph explicitly marked as
+        // e.g. `subCategory = <something fontc doesn't recognize>` silently reverts to being
+        // guessed as a Ligature (breaking mark attachment).
+        // https://github.com/googlefonts/fontc/issues/2055
+        let category_was_set = self.category.as_deref().is_some_and(|s| !s.is_empty());
+        let sub_category_was_set = self.sub_category.as_deref().is_some_and(|s| !s.is_empty());
+
         let codepoints = self
             .unicode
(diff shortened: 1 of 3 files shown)


cmyr (approved):
makes sense!

Issue #2052, opened 2026-06-23
Glyphs FEA predicate tokens: glyph-metadata attributes and nested boolean expressions


anthrotype:
Follow-up to #92. Phase 1 there scopes 1.0 to name-only predicates which need nothing beyond the GlyphMap fea-rs already has. This issue is the deferred remainder: predicates that query glyph metadata, and nested boolean expressions.

As far as I can tell, only two crater sources need this:

ZalandoSans (zalando/sans): category/subCategory:

```
$[category like 'Number']
$[subCategory like 'Space']
```

Borel (RosaWagner/Borel): the full NSPredicate subset plus nested/parenthesised boolean, e.g. one `@short` class meaning "every exporting glyph that is an uppercase Letter with a Unicode value whose name starts with AE/C/E/T/X":

```
@short = [$[category like "Letter" && case==upper && unicode != nil && (name beginswith "AE" OR name beginswith "C" OR name beginswith "E" OR name beginswith "T" OR name beginswith "X")]];
```

What this needs on top of phase 1:

- Threading glyph metadata (category, subCategory, case, unicode) through to wherever predicates are expanded. fea-rs currently only receives glyph names. `case` in particular isn't retained by glyphs-reader today (it's dropped in glyphdata.rs), so that needs adding to the reader first.
- The `like` operator (`fnmatch`-style) and the `nil` literal (for `unicode != nil`).
- A real boolean parser: mixing `&&`/`OR`, parenthesised grouping, and `NOT` with proper precedence (phase 1 only needs a flat single-operator chain).

These are two fairly independent chunks: ZalandoSans needs only the metadata threading + `like`; Borel additionally needs `case` (including the glyphs-reader change), `unicode`/`nil`, and the nested-boolean parser.

PR #2057, opened 2026-07-09
[fea-rs] Support name-only Glyphs predicate tokens (phase 1)

### LINKED ISSUE #92: Handle predicate tokens and number values in Glyphs FEA extensions
Continuing from #91, implement some or all of https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/Lib/glyphsLib/builder/tokens.py#L12 and fail on whatever parts are not implemented. Used by https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/Lib/glyphsLib/builder/features.py#L85. Consider looking at our actual sources to determine what is needed.

anthrotype:
Fixes #92 by implementing Phase 1 of Glyphs.app predicate tokens in fea-rs: name-only `$[...]` predicates inside glyph classes.

The supported subset is deliberately narrow: `name beginswith`, `name endswith`, `name contains`, `name ==`, `name !=`, and flat `and`/`or` chains. That covers the crater cases that only need glyph names, without threading GlyphData into fea-rs. Predicates that need glyph metadata or a richer boolean grammar (`category`, `subCategory`, `case`, `unicode`, `like`, `matches`, `not`, parentheses, mixed/nested expressions) fail during validation and are left to #2052.

The predicate body is lexed as one token and parsed separately because Glyphs documents `$[...]` as [Apple NSPredicate code](https://glyphsapp.com/learn/tokens), not FEA syntax. Ordinary predicate cases do not tokenize into useful FEA nodes: single quotes are contextual-substitution markers in FEA, and `!=` is not a FEA operator. This mirrors glyphsLib's TokenExpander approach: capture the `$[...]` body, subparse it, and expand it to a glyph class. For `or`, the evaluator preserves glyphsLib's observable ordering: it evaluates the left clause over the glyph list, then appends new matches from the right clause, so class order can differ from evaluating the whole expression once per glyph.

DynaPuff is now identical to fontmake through ttx_diff, and Nastaliq's classes that use predicates match glyphsLib expansion in the default build; remaining Nastaliq table diffs are pre-existing contextual GSUB/GPOS differences, not predicate expansion.


fea-rs/src/compile/glyphsapp_syntax_ext/mod.rs
```diff
@@ -0,0 +1,15 @@
+//! Handling glyphsapp syntax extensions.
```
cmyr, line 1: just a personal preference I suppose but I don't use `mod.rs` anymore; you can just have `glyphsapp_syntax_ext.rs` and then `glyphsapp_syntax_ext/other_files.rs` and it treats it as a module and I find it easier to navigate? 🤷

cmyr (commented):
okay, going to just leave some preliminary thoughts that we can discuss. I haven't gotten to all the meat of this PR yet, but I have some concerns about the parsing layer.

- _lexing_ should be entirely about breaking the input stream into meaningful units. There is no reason to have predicates be a special case here, although you will likely need to add support for a few additional tokens (`<`, `>`, & co). You definitely don't need to have a special case in the lexer for "unterminated predicate"; any number of constructs in FEA can be unterminated, and that's something that we handle during parsing, not lexing. (the only errors we have to handle in lexing is when certain tokens _themselves_ are malformed, like if the user never closes a string literal).
- _parsing_ is where you make sense of the tokens produced by the lexer, and turn them into AST nodes. Given that we have an incompletely supported grammar, the parsing layer will potentially be a bit funny, but the basic idea should be pretty straightforward: you're going to consume tokens, and if you can parse the predicate you will produce something like a `GlyphsAppPredicate` ast node, and if you can't you will produce an error during parsing, and we will stop there.
- Parsing is all rigorously tested; how that works is documented in [test-data/parse-tests/README.md](https://github.com/googlefonts/fontc/blob/353b5d9b3379d9c23c19b56db5bd26825f5cd826/fea-rs/test-data/parse-tests/README.md#L1). A nice thing about this system is that we also test that invalid inputs produce meaningful errors. 
- on the AST types: this is where a bunch of the stuff that you've defined in the compile module should live. At the very simplest you would end up with something like a `GlyphsPredicate` node that contains an object, an operation, and a value; and then you would have nodes for each of those types that encapsulate the possible constructions that are valid.

In my dream world, we would have one PR that just introduced the lexing and parsing, and then we would have a separate PR that handled compilation, but no worries if that's burdensome.

For what it's worth there's also some good prior art here: this overlaps a lot with the numbers syntax, which landed in https://github.com/googlefonts/fontc/pull/892. 


In any case it's nice to see progress here, but unless there is some major complication I'm overlooking I think we should try and have this functionality work as much like the rest of fea-rs as is reasonably possible.

anthrotype:
Fair points. I should have looked more closely at the #892 precedent. For the record, the separate parser wasn't an accident: Glyphs documents `$[...]` as Apple NSPredicate code, which is an open-ended foreign grammar (like/matches regexes, keypaths, function calls), and glyphsLib itself captures the token opaquely (everything up to the first `]`, never lexed), and sub-parses it on its own. So keeping the FEA grammar ignorant of it, with predicate parsing living next to the evaluator that compilation needs regardless, looked like a clean boundary.

But with the predicate fully modeled as typed AST nodes there's no duplicate parser at all. Compile just walks nodes, same as the number values, and the subset we actually need for #92/#2052 is bounded, no open-endedness concern in practice. So, agreed: I'll restructure along the #892 lines, split into two PRs (lexing/parsing + typed AST first, then compilation), and I'll keep `glyphsapp_syntax_ext.rs` as the module root rather than introducing `mod.rs`.

One note on where I'm drawing the phase-1 line: the grammar accepts the structural surface (clause chains, keyword and symbolic operators, quoted/bare values) and validation enforces the supported subset, in the spirit of the number values grammar's comment "just eat values and operators; we will validate later" ([metrics.rs:174](https://github.com/googlefonts/fontc/blob/353b5d9b3379d9c23c19b56db5bd26825f5cd826/fea-rs/src/parse/grammar/metrics.rs#L174)). So `$[name like "a*"]` parses but fails validation with a pointer to #2052, while structurally foreign constructs (NOT, parentheses, nesting) are parse errors.

cmyr:
yea I get that the predicate syntax is a bit extensive, but ultimately we're going to need to parse it, and if we're doing that we should do it in the parser. FWIW I don't think the grammar is _that_ bad, and is described here:

https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Predicates/Articles/pBNF.html#//apple_ref/doc/uid/TP40001796-CJBDBHCB

to my mind, our parsing is going to look like:

`GlyphsCompoundPredicate = GlyphsPredicate [AND/OR GlyphsCompoundPredicate]`
`GlyphsPredicate = GlyphsPredicateObject GlyphsPredicateOperation GlyphsPredicateValue`
`GlyphsPredicateObject = Ident` (in theory this can be a 'keypath', e.g. multiple idents separated by `.`? but not necessary at first)
`GlyphsPredicateOperation = Ident | == | > | < | etc`
`GlyphsPredicateValue = String | Number | Ident`

this might be missing out a few things (like NOT and parens?) but it should get us most of the way there.



> One note on where I'm drawing the phase-1 line: the grammar accepts the structural surface (clause chains, keyword and symbolic operators, quoted/bare values) and validation enforces the supported subset, in the spirit of the number values grammar's comment "just eat values and operators; we will validate later" ([metrics.rs:174](https://github.com/googlefonts/fontc/blob/353b5d9b3379d9c23c19b56db5bd26825f5cd826/fea-rs/src/parse/grammar/metrics.rs#L174)). So `$[name like "a*"]` parses but fails validation with a pointer to #2052, while structurally foreign constructs (NOT, parentheses, nesting) are parse errors.

I think this is exactly the right idea: the grammar doesn't know if a given predicate _makes sense_, it just sees if it is structurally well formed. The idea with the validate/compilation split is that once we're in the compilation phase we generally don't need to worry about error handling, since we've checked our invariants in validation, and this is enforced by tests.

PR #2059, opened 2026-07-13
[kern] skip non-default kernless sources in the variable kern path

anthrotype:
Match googlefonts/ufo2ft#995 (merged as googlefonts/ufo2ft#997): a non-default full source with no kerning must not add a variation-model location, or the kern hollows toward 0 there instead of interpolating across it. The default is always kept, since it anchors the model.

Applies to both frontends:

* glyphs2fontir filters `master_positions` (checking LTR+RTL, so an RTL-only master like NotoSansHebrew/fontc#1857 stays)
* ufo2fontir filters the `KerningGroupWork` location loop via a kerning-only UFO read.

This refines my #2013's blunt include-all-masters, which had matched the pre- googlefonts/ufo2ft#995. The KerningEmptyMaster test flips to the new constant-kern expectation.

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -1131,14 +1131,7 @@ impl Work<Context, WorkId, Error> for KerningGroupWork {
             .iter()
             .filter_map(|(master_id, pos)| {
                 let keep = master_id.as_str() == default_master_id.as_str()
-                    || font
-                        .kerning_ltr
-                        .get(master_id)
-                        .is_some_and(|k| !k.is_empty())
-                    || font
-                        .kerning_rtl
-                        .get(master_id)
-                        .is_some_and(|k| !k.is_empty());
+                    || font.has_kerns_for_master(master_id);
                 keep.then(|| pos.clone())
             })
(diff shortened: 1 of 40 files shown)


anthrotype:
(note ufo2ft with has not been released yet, so we should hold merging this)

glyphs2fontir/src/source.rs
```diff
@@ -1120,7 +1120,28 @@ impl Work<Context, WorkId, Error> for KerningGroupWork {
...
+            .filter_map(|(master_id, pos)| {
+                let keep = master_id.as_str() == default_master_id.as_str()
+                    || font
+                        .kerning_ltr
+                        .get(master_id)
+                        .is_some_and(|k| !k.is_empty())
+                    || font
+                        .kerning_rtl
+                        .get(master_id)
+                        .is_some_and(|k| !k.is_empty());
```
cmyr: total golf but i'd be tempted to move this to a method like `has_kerns_for_master(&self, master_id: &str)`

cmyr (approved):
looks good!

Issue #339, opened 2023-06-13
What should happen if ufo's define inconsistent kerning groups?


rsheeter:
https://github.com/googlefonts/fontc/pull/338 assumes kerning groups _should_ be identical across ufos in a designspace. But ... what if that's a bad assumption? - prior art in python tries hard to handle this case by fiddling with the groups:

1. https://github.com/fonttools/fonttools/blob/4923a353bb75c0cb1f771c6f90edb63e35b3a7a2/Lib/fontTools/varLib/merger.py#L484-L508
1. https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/misc/classifyTools.py#L5
1. https://github.com/robotools/fontMath/issues/22
1. https://github.com/robotools/fontMath/pull/245

rsheeter:
Per IM, having the groups be inconsistent reflects user or application error in handling sources. The reclassification process attempts to mitigate, but having correct (consistent groups) sources to begin with remains preferable.

anthrotype:
The ufo2ft varialble kern feature writer (which fontc kern writer derives from) has the same issue: googlefonts/ufo2ft#992

`ufo2fontir` reads only the default master's `groups.plist` as authoritative kerning groups. Non-default masters' `groups.plist` files are never loaded.
`KerningInstanceWork::resolve()` then validates each source's kerning references against the default group map. Any kern pair that references a group unique to a non-default master is silently dropped ("not a valid group name; ignored"). The per-master group partition never reaches the backend.

This means a designspace whose masters define different kerning group memberships (e.g. glyph Y in `@kern2.right` at Regular but `@kern2.other` at Bold) compiles variable kerning that matches neither master's source kerning.

kerning groups (`groups.plist`) are independent of the feature text compatibility check, so a designspace with identical per-master `features.fea` but divergent groups would silently hit this.

Glyphs sources are immune since groups are per-glyph, shared across masters (as it should be, one could say).

The `fontTools.varLib.merger` module can handle GPOS PairPos2 with different classes.
In `_PairPosFormat2_align_matrices`, it loads each source's own groups, computes the common refinement of the per-master class partitions (glyphs that are in the same group across all masters stay classed; divergent glyphs are split out),
and resolves each refined cell against its own source's kerning cascade.

fontc and ufo2ft should do the same, so that when font sources are switched to use the new variable font compiler they produce the same kerning.

[...]

PR #2063, opened 2026-07-14
[kern] Remove unused kern group rename machinery 

anthrotype:
#636 broke the kern group rename map from #338. The `kerning_groups_for` calls it deduplicated were not duplicates: one loaded the default master's groups, the loop loaded each non-default source's own groups, so reusing the first call's result left the loop comparing the default's groups against themselves.

Since then, `old_to_new_group_names` only ever holds identity entries and kerning referencing a non-default master's own name for a group is dropped rather than renamed.

No test noticed: `groups_renamed_to_match_master` only asserts that the merged map equals the default's groups, which holds whether renaming works or not.

The first commit adds the missing assertions against today's code. The WghtVar-Bold.ufo test fixture has carried a deliberately-renamed group since #338 precisely to exercise the rename path, and the map never maps it. A kern pair referencing that name is dropped (when my PR for #339 lands, it will assert the pair survives).

The second commit removes the machinery rather than attempting to repair it. Proper handling of cross-master group differences (renaming included, with end-to-end coverage) will come with the port of googlefonts/ufo2ft#996 (#339), whose kern group reconciliation subsumes what this loop was meant to do.

JMM

### DIFF
--- a/fontir/src/ir.rs
+++ b/fontir/src/ir.rs
@@ -166,11 +166,6 @@ pub struct KerningGroups {
     ///
     /// Must be a subset of the master locations.
     pub locations: BTreeSet<NormalizedLocation>,
-
-    /// Optional group renaming map, meant for [KerningInstance] to consume
-    ///
-    /// The rhs should be the name used in the groups map.
-    pub old_to_new_group_names: BTreeMap<KernGroup, KernGroup>,
 }
 
 /// IR representation of kerning for a location.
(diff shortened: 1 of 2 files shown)


anthrotype:
The first commit was pushed alone, its green CI check shows the new assertions passing with the rename machinery still in place: https://github.com/googlefonts/fontc/actions/runs/29348074906.

The following commit then deletes the unused machinery and the dropped-kerning test continues to pass (backing up the "no output change" claim).

PR #2064, opened 2026-07-15
Update and complete Fontra backend

### LINKED ISSUE #2050: fontra's data structure has changed
fontc expects the "axes" field in font_info to be a list of axis definitions:

https://github.com/googlefonts/fontc/blob/d62ba0166eb9af8116be5f55673244ecdb012026/fontra2fontir/src/fontra.rs#L33-L37

but it isn't any more. It's a dictionary of axes and avar2 cross-axes mappings:

https://github.com/fontra/fontra/blob/3581940106e3bb6728de553bf76f308849a5b152/src/fontra/core/classes.py#L35-L41

```python
@dataclass(kw_only=True)
class Axes:
    axes: list[Union[FontAxis, DiscreteFontAxis]] = field(default_factory=list)
    mappings: list[CrossAxisMapping] = field(default_factory=list)
    elidedFallBackname: Optional[str] = None
    customData: CustomData = field(default_factory=dict)
```

khaledhosny:
Updates the Fontra font parsing to the current state of Fontra upstream code.

* Fixes #2050


rsheeter:
Awesome! I wonder if we can split up for review, perhaps pulling the earliest commits into standalone non-draft PRs?

khaledhosny:
Except for the last two commits, all the commits are about parsing Fontra files, I can drop the last two from this PR.

khaledhosny:
This should be ready for review.

fontra2fontir/src/fontra.rs
```diff
@@ -1,230 +1,723 @@
...
 
 use fontdrasil::{paths::string_to_filename, types::GlyphName};
-use fontir::error::{BadSource, PathConversionError};
+use fontir::error::{BadSource, BadSourceKind, PathConversionError};
 use serde::Deserialize;
 use write_fonts::types::Tag;
 
 pub(crate) type AxisName = String;
 pub(crate) type LayerName = String;
+pub(crate) type GlyphMap = BTreeMap<GlyphName, Vec<u32>>;
```
cmyr, line 23: I assume this is a 'name -> [codepont]' map?

fontra2fontir/src/fontra.rs
```diff
@@ -1,230 +1,723 @@
...
     serde_json::from_str(&raw).map_err(|e| BadSource::custom(p, e))
 }
 
-/// serde type used to load font-data.json
+/// An entry of glyph-info.csv
+#[derive(Debug, Clone, Default, PartialEq, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct GlyphInfo {
+    pub(crate) category: Option<String>,
+    pub(crate) sub_category: Option<String>,
```
cmyr: In general I prefer to use `SmolStr` over `String` for anything that is an identifier; this is the type that we use most often in fontir/fontbe, and so it saves having to convert later on.

fontra2fontir/src/fontra.rs
```diff
@@ -1,230 +1,723 @@
...
 }
 
-impl FontraFontData {
-    pub(crate) fn from_file(p: &Path) -> Result<Self, BadSource> {
-        from_file(p)
+/// Corresponds to a Fontra OpenTypeFeatures
+/// <https://github.com/fontra/fontra/blob/469a001f8/src/fontra/core/classes.py#L59>
+#[derive(Debug, Clone, Deserialize)]
+#[serde(default, rename_all = "camelCase")]
+pub(crate) struct OpenTypeFeatures {
```
cmyr, line 101: could `language` be a `Tag`? This only works if we expect the sources to always contain well-formed tags, and are willing for it to be an error if they are not...

fontra2fontir/src/fontra.rs
```diff
@@ -1,230 +1,723 @@
...
+///
+/// The file is split into blank line-delimited sections (TYPE, GROUPS1, GROUPS2, VALUES), repeated
+/// once per kerning type. Each section contains semicolon-delimited entries.
+///
+/// The legacy single-GROUPS format is not supported.
+///
+/// <https://github.com/fontra/fontra/blob/469a001f8/src/fontra/backends/fontra.py#L538>
+fn parse_kerning(fontra_dir: &path::Path) -> Result<BTreeMap<String, Kerning>, BadSource> {
+    let kerning_file = fontra_dir.join("kerning.csv");
+    if !kerning_file.is_file() {
```
cmyr, line 330: I'd find `.exists()` maybe a bit clearer in terms of the intent here? but not a blocker..

fontra2fontir/src/fontra.rs
```diff
@@ -1,230 +1,723 @@
...
+    if row.is_empty() || row[0].is_empty() {
+        return Err(BadSource::custom(
+            kerning_file,
+            format!("expected TYPE value string (line {line_number:?})"),
+        ));
+    }
+    Ok(Some(row[0].to_string()))
+}
+
+fn kerning_read_groups<'a>(
```
cmyr, line 403: I find this (free fns pass a &mut impl Iterator) to be a slightly funny way of writing a parser, but 🤷. I would probably have some helper struct with methods? Again, not a blocker.

cmyr (approved):
this all looks reasonable to me, happy to see this kept up to date!

fontra2fontir/src/fontra.rs
khaledhosny: Yes.

fontra2fontir/src/fontra.rs
khaledhosny: It was ported from the corresponding Python code and I tried to follow the same structure to avoid unintentional changes, I also have limited Rust experience to get a sense of what feels Rusty and what looks funny. If you can suggest a different way to approach this, I can try it.

fontra2fontir/src/fontra.rs
khaledhosny: The only features language currently supported by Fontra is `fea`, and I don’t see fontc supporting any other language anytime soon. So I think this is mostly a placeholder for future extension.

fontra2fontir/src/fontra.rs
khaledhosny: `is_file` was already use for font info and glyph files, and I copied the same pattern. But now I see the other places return `BadSourceKind::ExpectedFile` error, while this one just skips it.

fontra2fontir/src/fontra.rs
khaledhosny: I switched these and few others to `SmolStr`

khaledhosny:
Can we get another review/merge this PR?

rsheeter:
I see @cmyr approved so merging. Ty!

PR #2062, opened 2026-07-14
requirements.txt: bump ufo2ft to 3.9.0

anthrotype:
Picks up the variable-features kern path fixes for divergent per-master kern groups and kernless masters (googlefonts/ufo2ft#992, #995), plus refreshes all transitive deps.

PR #2065, opened 2026-07-15
[kern] Reconcile divergent per-master kerning groups

### LINKED ISSUE #339: What should happen if ufo's define inconsistent kerning groups?

anthrotype:
Fixes #339 and ports googlefonts/ufo2ft#996 to fontc.

When masters partition glyphs into kerning groups differently, resolve each master's kerning against its own groups and refine the output classes into the coarsest common refinement of the per-master partitions (the same partition varLib.merger computes from the compiled per-master ClassDefs), so class kerning compiles the way fontmake would when building a VF from DS/UFOs sources with divergent kern groups.

Previously fontc read only the default master's groups.plist: a divergent glyph took the default's grouping everywhere (wrong kerning value), and a group existing only in a non-default master was dropped along with its kerning.

The first commit adds the (initially failing) tests; the fixtures are generated by the tests themselves, masters sharing identical box glyphs and differing only in groups/kerning.

The second commit carries each source's own groups into IR (`KerningInstance.groups`), and validates each source's kerning against its own groups.

The third rewrites `build_variable_kern_adjustments` with per-source resolution through the UFO value-lookup cascade, and refined classes grouping members by which group holds them in each source. Divergence is detected on the source-declared groups, but the refinement counts only groups that are referenced by each source's own kerning, matching what reaches compiled ClassDefs in varLib.merger. Glyph-to-class pairs continue to be resolved per cell (googlefonts/ufo2ft#988, googlefonts/ufo2ft#994 semantics).

ttx_diff vs fontmake with ufo2ft 3.9.0 is identical (normalized GPOS) on the test fixtures and on JosefinSlab, whose [shipped VF](https://github.com/google/fonts/blob/47831f08ec6d6d7ad6b465f23dc9f9a890a2a04b/ofl/josefinslab/JosefinSlab%5Bwght%5D.ttf) carries this bug live.

Also note that the merged `KerningGroups.groups` map is now **unread** in the BE. The `KerningGroups` artifact itself keeps an orchestration role: its locations record which sources participate in kerning (the ufo2ft#995 skip), the BE iterates exactly those, and the `KerningGroupWork`'s completion spawns the per-location KernInstance works. Retiring the merged map and renaming the now misleadingly-named work is a planned follow-up to keep this diff reviewable.

We should merge after #2059 and the ufo2ft 3.9.0 bump (#2062), in the same crater window. Against pinned ufo2ft 3.8.2 it would spray kern diffs across crater.

### DIFF
--- a/fontbe/src/features/kern.rs
+++ b/fontbe/src/features/kern.rs
@@ -117,10 +117,20 @@ impl Work<Context, AnyWorkId, Error> for GatherIrKerningWork {
         let ir_groups = context.ir.kerning_groups.get();
         let ir_kerns = context.ir.kerning_at.all();
 
-        // convert the groups stored in the Kerning object into the glyph classes
-        // expected by fea-rs:
-        let groups = ir_groups
-            .groups
+        // Add IR kerns to builder. IR kerns are split by location so put them back together again.
+        let kern_by_pos: HashMap<_, _> = ir_kerns
+            .iter()
+            .map(|(_, ki)| (ki.location.clone(), ki.as_ref().to_owned()))
+            .collect();
+
+        // Resolve every pair at every location, reconciling sources whose
+        // kerning groups disagree. `refined_groups` is the reconciled output
+        // class partition (see build_variable_kern_adjustments).
+        let (refined_groups, adjustments) =
+            build_variable_kern_adjustments(&ir_groups, &kern_by_pos);
+
+        // convert the refined groups into the glyph classes expected by fea-rs:
+        let groups = refined_groups
             .iter()
             .filter_map(|(class_name, glyph_set)| {
                 let glyph_class: GlyphSet = glyph_set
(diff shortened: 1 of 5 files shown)


rsheeter (approved):
I don't have time to fully dig into all the nuances but the code and outcomes seem reasonable enough.

Issue #551, opened 2023-11-08
Support Glyphs.app-style contextual anchor positioning


anthrotype:
As implemented by @simoncozens in glyphsLib as optional ufo2ft featureWriter:

https://github.com/googlefonts/glyphsLib/pull/921

simoncozens:
More generally it would be super useful to have a summary (probably in README.md) of what does and what doesn't work yet. Last I looked mark feature writing wasn't there at all, but now it sounds like something is happening (#492)?

PR #2067, opened 2026-07-15
 Support the ufo2ft featureWriters lib key to disable or force-append automatic featues

### LINKED ISSUE #1045: ufo2ft.featureWriters lib key may be used to disable kern and/or mark feature generation
In https://github.com/docrepair-fonts/lunasima-fonts - sources/masters/Lunasima-Regular.ufo/lib.plist contains

```xml
<?xml version='1.0' encoding='UTF-8'?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <dict>
    <key>com.github.googlei18n.ufo2ft.featureWriters</key>
    <array>
      <dict>
        <key>class</key>
        <string>KernFeatureWriter</string>
        <key>options</key>
        <dict>
          <key>mode</key>
          <string>skip</string>
        </dict>
      </dict>
    </array>
```

when this private lib key "com.github.googlei18n.ufo2ft.featureWriters" is present, ufo2ft uses this list _instead of_ the built-in list of feature writers to generate things like kern or mark/mkmk features.

In this particular case, this has the effect of _not_ building mark/mkmk features for Lunasima font, and as a result, no GPOS table is generated when this is built using fontmake (there isn't any kerning either...).

fontc instead ignores this and goes on to build mark/mkmk from the glyph anchors and a non-empty GPOS is built.

ttx_diff.py understandably reports a very bad score for Lunasima

Image

Ideally ttx_diff could check for this lib key and tell fontc to only build mark or kern or neither accordingly.

### LINKED ISSUE #1842: Extra GDEF marks in NotoMusic due to ignored featureWriters key in userData
exhibited in NotoMusic.

```sh
python3 -m ttx_diff 'https://github.com/notofonts/music?2c97b9ea69#sources/NotoMusic.glyphs'
```

### LINKED ISSUE #1843: difference in ordering of curs & kern lookups in NotoMusic
There are a few things going on in the GPOS here, and it's hard to tease it apart; but one obvious thing is that we are ordering the generated cursive lookup ahead of user-provided kern lookups, which doesn't match fontmake.

It might be significant that the kern block includes an `# Automatic Code` marker, but that we don't actually generate any code to insert there?

### LINKED ISSUE #1868: support `com.github.googlei18n.ufo2ft.featureWriters` key (UFO+DS)
This can be used to suppress mark generation, as in

```
python3 -m ttx_diff 'https://github.com/docrepair-fonts/agdasima-fonts?c971400d77#sources/Agdasima-Regular.designspace'
```

anthrotype:
Adds support for the `com.github.googlei18n.ufo2ft.featureWriters` lib key (UFO/designspace lib and Glyphs font-level userData), so sources can disable individual automatic feature writers or request "append" placement instead of the default "skip" (or insert at marker) behavior.

fontc already implemented "skip" mode; this adds "append" (fea-rs learns forced append placement for generated features) and wires the key through fontir/ufo2fontir/glyphs2fontir to fontbe. Key absent means the default writer list, key present is a full replacement, an empty list disables all automatic features.

The plumbing: each frontend parses the key (the designspace lib, already merged with the default master's lib.plist, or the Glyphs font-level userData) into validated `FeatureWriterSpec` entries on `StaticMetadata.misc`, so validation fires once at IR-build time and the config rides the cached IR. Backends resolve that into a `FeatureWriterPlan`: one settings slot (`mode` + `features` subset) per writer, `None` meaning disabled, plus a bool for Gdef.
The plan is consulted at the generation sites: `feature_writer_todo_list `decides per tag whether to generate and whether to append; append-forced tags reach fea-rs through the new `FeatureBuilder::force_append` method, and the Gdef flag gates both the GDEF class-def overwrite and ligature caret generation.

Validation mirrors ufo2ft's current default `loadFeatureWriters(ignoreErrors=True)`: entries ufo2ft would fail to import (e.g. unknown writers, wrong module/class pairs) are warned and dropped, while recognized options that fontc currently doesn't implement (`ignoreMarks=false`, `quantization != 1`, `groupMarkClasses=true`, a valid GdefFeatureWriter `features=` subset, or the old `kernFeatureWriter2`) are hard errors. So we deliberately refuse rather than silently build something different, although some of these (quantization, kern2's lookup organization) could be worth implementing or decide to relaxe later.

The glyphsLib's legacy `ContextualMarkFeatureWriter` alias is accepted as a plain Mark writer, since glyphsLib >= 6.10 defines it as a shim subclass of ufo2ft's MarkFeatureWriter (when the latter gained contextual anchors support; btw, contextual anchor generation itself is tracked separately #551).

You'll note that in a few places fontc is deliberately stricter than ufo2ft: a non-list key value, a non-string module, and a writer listed more than once are hard errors where ufo2ft silently builds without the affected writers (or, for duplicates, runs every entry in order). AFAIK none of these appear in any crater source.

Also one known limitation: custom writer order is not honored for "append" placement (generated features keep the default curs -> kern -> marks merge order).

I verified ttx-diff identical to fontmake on Agdasima, NotoMusic, KumbhSans, Lunasima, Gulzar, NotoSansTakri, AkayaKanadaka and OpenSans' GPOS/GDEF tables.

I verified this with ttx_diff against fontmake: Agdasima, NotoMusic, KumbhSans, Gulzar, NotoSansTakri and AkayaKanadaka are identical; OpenSans Roman and Italic have identical GPOS/GDEF (pre-existing unrelated diffs untouched); Lunasima is identical except a pre-existing unrelated head flags diff (bit 8).

Fixes #1868, fixes #1045, fixes #1842, fixes #1843 (the latter three are older reports/manifestations of the ignored key).

PS: The stutter is mandatory: GitHub requires "full syntax for each issue" (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)... a parser that can close an issue but not cross a comma.

EDIT: I changed some names after review feedback in be9cf0e7: FeatureWriterPlan -> FeatureGenerationPlan, FeatureWriterSettings -> FeatureGenerationSettings, resolve_feature_writers -> resolve_feature_generation, and the misc.feature_writers field -> feature_generation.

### DIFF
--- a/fea-rs/src/compile/feature_writer.rs
+++ b/fea-rs/src/compile/feature_writer.rs
@@ -59,6 +59,7 @@ pub struct FeatureBuilder<'a> {
     pub(crate) lig_carets: BTreeMap<GlyphId16, Vec<CaretValue>>,
     mark_filter_sets: &'a mut HashMap<GlyphSet, FilterSetId>,
     feature_variations: Option<RawFeatureVariations>,
+    append_features: HashSet<Tag>,
 }
 
 pub trait LookupSubtableBuilder: Sized {
@@ -145,6 +146,7 @@ impl<'a> FeatureBuilder<'a> {
             mark_filter_sets,
             feature_variations: Default::default(),
             lig_carets: Default::default(),
+            append_features: Default::default(),
         }
     }
 
@@ -229,6 +231,15 @@ impl<'a> FeatureBuilder<'a> {
         })
     }
 
+    /// Force a generated feature to be appended, ignoring any insertion marker.
+    ///
+    /// Generated lookups for `tag` will be placed after all user-defined
+    /// lookups even if the FEA contains a `# Automatic Code` marker for that
+    /// feature. This models a feature writer running in `append` mode.
+    pub fn force_append(&mut self, tag: Tag) {
+        self.append_features.insert(tag);
+    }
+
     fn get_filter_set_id(&mut self, cls: GlyphSet) -> FilterSetId {
         let next_id = self.mark_filter_sets.len();
         *self.mark_filter_sets.entry(cls).or_insert_with(|| {
@@ -246,6 +257,7 @@ impl<'a> FeatureBuilder<'a> {
             features,
             lig_carets,
             feature_variations,
+            append_features,
             ..
         } = self;
         ExternalFeatures {
@@ -254,6 +266,7 @@ impl<'a> FeatureBuilder<'a> {
             sub_lookups,
             feature_variations,
             lig_carets,
+            append_features,
         }
     }
 }
@@ -297,6 +310,9 @@ pub(crate) struct ExternalFeatures {
     pub(crate) features: BTreeMap<FeatureKey, FeatureLookups>,
     pub(crate) lig_carets: BTreeMap<GlyphId16, Vec<CaretValue>>,
     pub(crate) feature_variations: Option<RawFeatureVariations>,
+    /// Features whose generated lookups must ignore insertion markers and be
+    /// appended after all user-defined lookups (feature writer `append` mode).
+    pub(crate) append_features: HashSet<Tag>,
 }
 
 /// A position in a feature where generated code should be inserted.
@@ -323,6 +339,8 @@ struct MergeCtx<'a> {
     all_feats: &'a mut AllFeatures,
     // that were explicit
     insert_markers: &'a HashMap<Tag, InsertionPoint>,
+    // features forced to append, ignoring any insertion marker
+    append_features: HashSet<Tag>,
     ext_pos_lookups: BTreeMap<LookupId, PositionLookup>,
     ext_sub_lookups: BTreeMap<LookupId, SubstitutionLookup>,
     ext_features: BTreeMap<FeatureKey, FeatureLookups>,
@@ -510,11 +528,14 @@ impl MergeCtx<'_> {
     }
 
     fn do_curs(&mut self) {
-        let curs_pos = self
-            .insert_markers
-            .get(&CURS)
-            .copied()
-            .unwrap_or_else(|| self.insertion_point_for_append());
+        let curs_pos = if self.append_features.contains(&CURS) {
+            self.insertion_point_for_append()
+        } else {
+            self.insert_markers
+                .get(&CURS)
+                .copied()
+                .unwrap_or_else(|| self.insertion_point_for_append())
+        };
         self.finalize_lookups_for_feature(CURS, curs_pos);
     }
 
@@ -533,11 +554,18 @@ impl MergeCtx<'_> {
                     DIST => [Some(DIST), kern],
                     _ => [kern, dist],
                 });
-        let marker = features_we_are_writing
-            .into_iter()
-            .flatten()
-            .find_map(|k| self.insert_markers.get(&k).copied())
-            .unwrap_or_else(|| self.insertion_point_for_append());
+        // kern and dist are set together by the caller; if either is forced to
+        // append, ignore both markers.
+        let marker = if self.append_features.contains(&KERN) || self.append_features.contains(&DIST)
+        {
+            self.insertion_point_for_append()
+        } else {
+            features_we_are_writing
+                .into_iter()
+                .flatten()
+                .find_map(|k| self.insert_markers.get(&k).copied())
+                .unwrap_or_else(|| self.insertion_point_for_append())
+        };
 
         let lookups = self.take_lookups_for_features(&[KERN, DIST]);
         if !lookups.is_empty() {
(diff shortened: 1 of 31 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -486,6 +571,7 @@ impl Work<Context, WorkId, Error> for StaticMetadataWork {
         )
         .map_err(Error::VariationModelError)?;
         static_metadata.misc.selection_flags = selection_flags;
+        static_metadata.misc.feature_writers = feature_writers_from_user_data(&font.user_data)?;
```
rsheeter: IIUC this is more like flags or configuration for feature production. It's *not* feature writers in the python sense. That would lead me to want to name it differently. For example, feature_flags (like selection_flags).

fontir/src/ir/feature_writers.rs
rsheeter, line 1: I wonder if we could broadly rename our feature writer stuff to focus more on what it _does_ which seems to be mostly to configure fea-rs? - the feature writer naming everywhere feels like maybe it over-carries python concepts across.

Could be as simple as s/FeatureWriter/Feature/ in a bunch of names?

anthrotype: Yeah, the writer vocabulary was leaking out of the parsing layer. 

I renamed  the resolved side: FeatureWriterPlan -> FeatureGenerationPlan, FeatureWriterSettings -> FeatureGenerationSettings, resolve_feature_writers -> resolve_feature_generation, and the misc field is now feature_generation. I kept "writer" only where we parse the literal ufo2ft schema: FeatureWriterSpec, KnownFeatureWriter and validate_feature_writer match actual class/module strings like KernFeatureWriter. So the Python concept appears exactly where it's being decoded and nowhere downstream.

For the renamed types I went with FeatureGeneration* rather than bare Feature* because fea-rs already owns that namespace (FeatureBuilder, FeatureKey, FeaturesSource); FeaturePlan or FeatureSettings would read as FEA-layer types.

And feature_flags undersells the content a bit: entries carry a mode and optional feature subsets, not just on/off bits.

fea-rs/src/compile/feature_writer.rs
```diff
@@ -533,11 +554,18 @@ impl MergeCtx<'_> {
...
+        let marker = if self.append_features.contains(&KERN) || self.append_features.contains(&DIST)
+        {
+            self.insertion_point_for_append()
+        } else {
+            features_we_are_writing
+                .into_iter()
+                .flatten()
+                .find_map(|k| self.insert_markers.get(&k).copied())
+                .unwrap_or_else(|| self.insertion_point_for_append())
+        };
```
cmyr: just a thought but there could maybe be a reuseable fn for this, like `insert_pos_for_features(&self, features: &[Tag]) -> InsertionPoint`?

fontbe/src/features/marks.rs
```diff
@@ -35,7 +35,10 @@ use crate::{
     },
 };
 use fontir::{
-    ir::{self, Anchor, AnchorKind, GdefCategories, GlyphAnchors, GlyphOrder, StaticMetadata},
+    ir::{
+        self, Anchor, AnchorKind, FeatureGenerationPlan, GdefCategories, GlyphAnchors, GlyphOrder,
+        StaticMetadata, resolve_feature_generation,
```
cmyr: a total nit but rust style discourages importing bare fns, and instead calling them through the containing module (so `ir::resolve_feature_generation(args)`, instead of bare `resolve_feature_generation(args)`)

fontbe/src/features.rs
```diff
@@ -717,24 +744,70 @@ fn get_script_language_systems(ast: &ParseTree) -> HashMap<UnicodeShortName, Vec
...
+/// Tags whose generated lookups must be appended (insertion markers ignored).
+///
+/// A tag is force-appended exactly when its writer is enabled in `append` mode;
+/// this is independent of the FEA, so it needs no AST.
+fn append_forced_tags(plan: &FeatureGenerationPlan) -> Vec<Tag> {
+    let mut tags = Vec::new();
+    for (settings, all) in [
+        (&plan.curs, &[CURS][..]),
+        (&plan.kern, &[KERN, DIST][..]),
+        (&plan.mark, &[MARK, MKMK, ABVM, BLWM][..]),
```
cmyr: wth is going on here with the `[..]`? Is this some fancy trick I don't understand? Like coercing to a slice from references to different fixed sized arrays I guess??

i'd probably do,

```rust
    [
        (&plan.curs, [CURS].as_slice()),
        (&plan.kern, [KERN, DIST].as_slice()),
        (&plan.mark, [MARK, MKMK, ABVM, BLWM].as_slice()),
    ]
    .into_iter()
    .filter_map(|(settings, tags)| {
        (settings.as_ref().map(|settings| settings.mode) == Some(FeatureWriterMode::Append))
            .then_some(tags.iter().copied())
    })
    .flatten()
    .collect()
```

but I won't argue this is much better..

cmyr (approved):
well this is certainly nasty, but the existing code was already nasty so I don't think this is much worse. Ultimately it does feel bad to be jumping through so many hoops in order to match python, but maybe that's a philosophical question for future debate...

fontbe/src/features.rs
rsheeter: Gemini feels the current code is "a concise, idiomatic way to create a temporary, borrowable slice"

anthrotype:
since this was LGTM'd and the only follow ups were addressing the original (approved) review, I'm going to merge this now before it goes stale

Issue #2061, opened 2026-07-14
Automatically set the overlap 


MariannaPaszkowska:
Hi all, while working on the Google Sans/Flex projects we noticed a need for automatically setting the overlap bit flag 6 in the glyf table.

This flag is used to prevent rendering issues during the rasterization overlapping outlines. 

We need a tool that would allow us to detect glyphs that need this flag automatically. It should consider all of the variable design space of that glyph (for example, in some cases some shapes might intentionally overlap in bolder or condensed weights)
@RickyDaMa and @Hoolean are working on integrating one such tool but there are a couple of functions missing: one to simplify beziers and another to remove overlaps. 
@simoncozens already has implemented those in Python, and we could easily port those to Rust. 

I thought it would be a good idea to add those to the compiler itself. 

I also think that rendering issues resulting from those bits unset are common so perhaps we'd even want to use that as default on all compiles. Perhaps it won't be so costly on the run time. 

Please let us know what you think. 
@RickyDaMa and @Hoolean will follow with more technical details and are ready to start working on it. 
Cc @aaronbell @davelab6

RickyDaMa:
Related: #717

Currently, we only have external tools that do this, such as the proof-of-concept https://github.com/rsheeter/glyph_overlaps and a hand-rolled script that relies on manual classification. #717 points towards the use of [`linesweeper`](https://radicle.network/nodes/rosa.radicle.network/rad%3Az2jbcL8tV3Fqqqk1yPyGkDTjgaTfH) as the library for doing overlap detection, and this seems like a good home for a robust implementation of this.

There are a couple of primitives that overlap detection relies on that aren't part of `linesweeper`, namely overlap detection & removal. However, both are implemented in Simon's Python wrapper of `linesweeper` [here](https://github.com/simoncozens/linesweeper-py/blob/1014ed5f185a941a04a4c948d9d5eda2849ab366/src/linesweeper/__init__.py#L33-L46). These functions are both easily back-portable to Rust, to either be kept here, or potentially even contributed back to `linesweeper` itself, given their relative simplicity and utility beyond fonts (depending on the maintainer's appetite).

From there, we'd suggest implementing the remainder of the work to detect overlaps and set the bits in fontc itself. This is much more convenient for designers, who no longer need to think about adding an extra tool/operation to their build process. The compiler has access to raw outlines at every source position to push through `linesweeper` as it's building `glyf`, where the overlap bits live too.

[...]

Issue #717, opened 2024-02-12
Handle overlaps when compiling static fonts


rsheeter:
Overlaps produce cutouts when drawn with evenodd fill. Apparently older versions of macos and _maybe_ some printers do this.

When compiling a static (non-variable) font do at least one of:

* At least set OVERLAP_SIMPLE (only if there are overlaps?), https://learn.microsoft.com/en-us/typography/opentype/spec/glyf
* Actually remove the overlaps.
   * No Rust implementation equivalent ot Skia pathops appears to exist, would likely have to call into pathops
   * @simoncozens https://github.com/simoncozens/rust-font-tools/blob/main/fonttools-cli/src/bin/ttf-remove-overlap.rs does this via https://docs.rs/crate/skia-safe/latest

The importance of actually removing the overlaps is a bit unclear to me. It used to be done for macos but now we have the overlap simple flag. Rumor has it that old printers might need actual removal but I don't have specific examples.

rsheeter:
Courtesy of @simoncozens , "There are many experienced designers out there sharing their perspective on this topic why it’s not needed to outline your text anymore but it’s still the go-to method for most designers whenever they’re preparing their files for print. " (https://www.jukeboxprint.com/blog/is-it-still-necessary-to-outline-your-fonts)

PR #2070, opened 2026-07-28
Overlap detection & removal primitives

RickyDaMa:
This PR adds functions to concatenate `BezPath`s, as well as functions that use [`linesweeper`](https://radicle.network/nodes/rosa.radicle.network/rad%3Az2jbcL8tV3Fqqqk1yPyGkDTjgaTfH) to perform overlap removal & detection. In future PRs these can then be used to do overlap removal for statics (#717) and automatically set the overlap bit (#2061).

Outstanding questions:
1. Do we want to expose `linesweeper` types internally within the fontc crates? Currently `linesweeper::Error` is exposed.
2. Do we need to specify an epsilon value ourselves? `linesweeper` uses `1e-6` by default.
3. Is this code living in the right place? I can foresee `fontir` needing the overlap removal method and `fontbe` needing overlap detection, so my rationale was that the logic should go in the "general utilities & types" crate.
4. What sort of QA do we want around this code to be satisfied that it's doing the right thing? Tests are pretty minimal right now while this is a proof of concept, but when it comes to actually setting the overlap bit and such we probably want to be a lot more rigorous

For the TODOs in code that talk about potential linesweeper features, I have opened [an issue](https://radicle.network/nodes/rosa.radicle.network/rad%3Az2jbcL8tV3Fqqqk1yPyGkDTjgaTfH/issues/53ddb00297b3ca82056f3c588d66f1bac3efba52) and [a PR](https://radicle.network/nodes/rosa.radicle.network/rad%3Az2jbcL8tV3Fqqqk1yPyGkDTjgaTfH/patches/d4a09be08f0cd74d5f5c30f7c62a333d1135e122).

### DIFF
--- a/fontdrasil/src/overlaps.rs
+++ b/fontdrasil/src/overlaps.rs
@@ -5,7 +5,7 @@ use thiserror::Error;
 /// An error from an overlap operation.
 #[derive(Debug, Error)]
 #[error(transparent)]
-pub struct OverlapError(#[from] linesweeper::Error);
+pub struct OverlapError(linesweeper::Error);
 
 /// Fill rule for interpreting which regions of a path are inside.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
@@ -16,11 +16,11 @@ pub enum FillRule {
     EvenOdd,
 }
 
-impl From<FillRule> for LsFillRule {
-    fn from(fill_rule: FillRule) -> Self {
-        match fill_rule {
-            FillRule::NonZero => Self::NonZero,
-            FillRule::EvenOdd => Self::EvenOdd,
+impl FillRule {
+    fn to_linesweeper(self) -> LsFillRule {
+        match self {
+            FillRule::NonZero => LsFillRule::NonZero,
+            FillRule::EvenOdd => LsFillRule::EvenOdd,
         }
     }
 }
@@ -40,12 +40,17 @@ pub fn remove_overlaps(
     // TODO: if linesweeper gave us an owned iterator, we could just have this method return a
     //       Contour iterator instead of a Vec<BezPath>, and then re-use it in has_overlaps without
     //       incurring allocation overhead
-    let beziers =
-        linesweeper::binary_op(bez_path, &BezPath::new(), fill_rule.into(), BinaryOp::Union)?
-            .contours()
-            .cloned()
-            .map(|contour| contour.path)
-            .collect::<Vec<_>>();
+    let beziers = linesweeper::binary_op(
+        bez_path,
+        &BezPath::new(),
+        fill_rule.to_linesweeper(),
+        BinaryOp::Union,
+    )
+    .map_err(OverlapError)?
+    .contours()
+    .cloned()
+    .map(|contour| contour.path)
+    .collect::<Vec<_>>();
     Ok(beziers)
 }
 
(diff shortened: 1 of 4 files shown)


fontdrasil/src/overlaps.rs
```diff
@@ -0,0 +1,155 @@
...
+    // TODO: Contours could implement .len?
+    if non_zero_beziers.contours().count() != even_odd_beziers.contours().count() {
+        return Ok(true);
+    }
+    // TODO: Could Contours implement PartialEq?
+    //       Ok(non_zero_beziers == even_odd_beziers)
+    let has_overlaps = non_zero_beziers
+        .contours()
+        .zip(even_odd_beziers.contours())
+        .all(|(non_zero_contour, even_odd_contour)| non_zero_contour.path == even_odd_contour.path);
```
anthrotype: Wait, shouldn't the logic be flipped here?

```diff
-        .all(|(non_zero_contour, even_odd_contour)| non_zero_contour.path == even_odd_contour.path);
+        .any(|(non_zero_contour, even_odd_contour)| non_zero_contour.path != even_odd_contour.path);
```

RickyDaMa: Shoot sorry, forgot to push my latest before opening the PR 🙈 This expression ended up getting refactored, but should be semantically correct now

anthrotype:
Sorry for the delay. Overall this looks good, and with the follow-up PRs you've merged I'm happy for it to land. Regarding your outstanding questions, in order:

1. We should not expose `linesweeper` types. fontdrasil sits at the base of the crate graph, so a leaked type forces fontir/fontbe to depend on linesweeper just to name it. See the now merged https://github.com/daltonmaag/fontc/pull/2 

2. linesweeper's 1e-6 default is ok. The linesweeper's `binary_op` scales eps to the input bbox with a floor of 1e-6 (https://docs.rs/linesweeper/0.4.0/src/linesweeper/lib.rs.html#123-125), and with the usual font-unit magnitudes the floor always wins, so it's effectively exact for our purposes. A `binary_op_with_eps` also exists if we ever want to tune it. In general, I think we should err toward reporting overlaps, because a false positive causes one unnecessary union, whereas a false negative might cause rendering artifacts.

3. fontdrasil is the right place for the reason you gave

4. I added some negative tests plus one that asserts curves are preserved, and moved linesweeper to `[workspace.dependencies]` to match the other deps (thanks for the quick merge https://github.com/daltonmaag/fontc/pull/1). 

Beyond that, before wiring this into overlap removal for statics (#717), I guess one could perform a corpus pass over real glyphs using ttx_diff. The latter already normalizes contour order (and all-offcurve start points) when the point data matches, but two different boolean engines (linesweeper vs Skia pathops in the fontmake pipeline) won't generally produce identical outlines. I expect different segmentation and/or starting points even where the geometry is right. Maybe for overlap-removed glyphs the comparison will need to be more geometric than textual.

As for #2061, fontmake currently doesn't do this automatically for VFs. I guess it could be an opt-in flag.. I haven't made up my mind yet. Worth maybe also profiling how expensive it is to perform this double-union detection for every glyph of each master (let alone those overlaps that only occur in between masters).

anthrotype (approved):

anthrotype:
my LGTM stands but i'm holding off on merging this because adding linesweeper dep to fontdrasil means everything downstream picks it up while nothing calls it yet. You can keep building #717 or #2061 on top of this and can then land them together (or land this immediately before)

RickyDaMa:
Thanks Cosimo! We've got higher priorities at the moment, but will do exactly that when we get back to this

PR #2071, opened 2026-07-29
Support Glyphs 3 component slant transforms

### LINKED ISSUE #1127: components in Glyphs3 sources can have a 'slant' attribute, in addition to 'pos', 'scale' and 'angle'
As seen in https://github.com/googlefonts/glyphsLib/issues/1047

apparently this wasn't supported in glyphsLib either (and thus fontmake)

fontc should match Glyphs.app and support slanted components (and glyphsLib will need fixing at the same time).

The above linked issue documents the expected order of the partial transformations comprising the full affine:

translate * slant * rotate * scale

Currently fontc (and glyphLib) only does:

translate * rotate * scale

here: 
https://github.com/googlefonts/fontc/blob/645cdff5588d1930dab3d777c9099501130bc93c/glyphs-reader/src/font.rs#L1821-L1843

anthrotype:
Parse the Glyphs 3 component `slant` field and compose it in Glyphs.app's `translate * slant * rotate * scale` order. Add equivalent Glyphs 2 and Glyphs 3 fixtures written by Glyphs.app to cover the conversion.

Companion glyphsLib fix: [googlefonts/glyphsLib#1160](https://github.com/googlefonts/glyphsLib/pull/1160).

Fixes #1127.

### DIFF
--- a/glyphs-reader/src/font.rs
+++ b/glyphs-reader/src/font.rs
@@ -1646,6 +1646,8 @@ struct RawShape {
     pos: Vec<f64>,             // v3
     angle: Option<f64>,        // v3
     scale: Vec<f64>,           // v3
+    /// Horizontal and vertical slant, in degrees (v3)
+    slant: Vec<f64>,
 
     #[fromplist(alt_name = "attr")]
     attributes: ShapeAttributes,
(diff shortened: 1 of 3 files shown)


rsheeter (approved):

anthrotype:
thanks for the review, I will hold this until https://github.com/googlefonts/glyphsLib/pull/1160 is also approved/merged/released so we can synchronize the fontc deps bump and avoid fontc crater regressions.

PR #2072, opened 2026-08-01
[glyphs] Apply *origin to layer geometry, not just propagated anchors

anthrotype:
Glyphs.app treats `*origin` as a layer-wide coordinate shift: everything else on the layer is drawn as if `*origin` were at (0, 0) and then translated back by -origin. fontc only applied this to anchors, during propagation, and never to contours or components. Even that anchor adjustment was silently skipped for non-composite glyphs, because propagate_anchors only writes propagated anchors back for composite glyphs, so a simple glyph with `*origin` came out unshifted everywhere.

This applies the shift during Glyphs source conversion instead, to contours, components and the raw anchor collection, so simple glyphs are covered too.

propagate_anchors keeps the current `origin_adjusted_anchors` for other potential IR callers, but on the Glyphs path it becomes a no-op.

(related to #769 which handled the anchor half of this in 2024; this is the rest of it)

glyphsLib does the same thing as of googlefonts/glyphsLib#1155, which merged yesterday (July 31), so the two compilers will continue to agree.

We should hold this until a glyphsLib release containing googlefonts/glyphsLib#1155 is out and fontc requirements are bumped.

### DIFF
--- a/glyphs2fontir/src/source.rs
+++ b/glyphs2fontir/src/source.rs
@@ -4291,6 +4291,31 @@ unitsPerEm = 1000;
         );
     }
 
+    trait ExecForTest {
+        fn try_create_static_metadata(
+            &self,
+            source: &impl Source,
+        ) -> Result<Arc<StaticMetadata>, Error>;
+        fn create_static_metadata(&self, source: &impl Source) -> Arc<StaticMetadata>;
+    }
+
+    impl ExecForTest for Context {
+        fn try_create_static_metadata(
+            &self,
+            source: &impl Source,
+        ) -> Result<Arc<StaticMetadata>, Error> {
+            let work = source
+                .create_static_metadata_work()
+                .expect("To create static metadata work");
+            let task_context = self.copy_for_work(work.read_access(), work.write_access());
+            work.exec(&task_context).map(|_| self.static_metadata.get())
+        }
+
+        fn create_static_metadata(&self, source: &impl Source) -> Arc<StaticMetadata> {
+            self.try_create_static_metadata(source).unwrap()
+        }
+    }
+
     #[test]
     fn reads_feature_writers_from_font_user_data() {
         let _ = env_logger::builder().is_test(true).try_init();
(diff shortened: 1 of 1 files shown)


glyphs2fontir/src/source.rs
```diff
@@ -1695,6 +1705,30 @@ impl Work<Context, WorkId, Error> for GlyphIrWork {
...
+/// The layer is drawn as if the origin anchor were at (0, 0), so contours,
+/// components and every other anchor are translated by -origin. Advance
+/// width and metrics are unaffected.
+///
+/// See <https://github.com/googlefonts/glyphsLib/pull/1155> for a reference
+/// implementation (glyphsLib applies this before anchor propagation).
+const GLYPHS_ORIGIN_ANCHOR: &str = "*origin";
+
+/// The `*origin` anchor's position, if the layer has one.
+fn layer_origin(layer: &Layer) -> Option<kurbo::Point> {
```
rsheeter, line 1719: Nit: layer.origin() or layer.find_origin() would be nicer, might it make sense to use a trait or move this into glyphs-reader?

glyphs2fontir/src/source.rs
```diff
@@ -4398,4 +4446,192 @@ unitsPerEm = 1000;
...
+);
+width = 200;
+}
+);
+}
+);
+unitsPerEm = 1000;
+}"#,
+        )
+        .unwrap();
```
rsheeter, line 4581: IMHO this would be nicer as an actual .glyphs file we load

glyphs2fontir/src/source.rs
```diff
@@ -4398,4 +4446,192 @@ unitsPerEm = 1000;
...
+                .variant(WorkId::StaticMetadata)
+                .variant(WorkId::PreliminaryGlyphOrder)
+                .variant(WorkId::PreliminaryGdefCategories)
+                .build(),
+        );
+        source
+            .create_static_metadata_work()
+            .unwrap()
+            .exec(&task_context)
+            .unwrap();
```
rsheeter: This is noisy and seems to repeat quite a bit. A helper, perhaps via a wee trait, could at least spare the unwraps?

`source.create_static_metadata_work().exec(task_context)`? Bonus points for automating the setup of the copy_for_work based on read/write access of work.

rsheeter (approved):

glyphs2fontir/src/source.rs
rsheeter: I felt nerd-sniped and tried it in #2073. Feel free to use or not, or to leave standalone if we'd prefer to keep the PRs split.

PR #2073, opened 2026-08-05
Try to denoise tests a little

rsheeter:
Exploratory based on https://github.com/googlefonts/fontc/pull/2072#discussion_r3717213660


cmyr (approved):
was wondering why this needed a trait, but I see it's a foreign type. Agree this is an improvement.

Issue #903, opened 2024-08-17
[fontir] Port VariationModel changes from fonttools


behdad:
Specifically the change in: https://github.com/fonttools/fonttools/pull/3601
Which applies to https://github.com/googlefonts/fontc/blob/main/fontir/src/variations.rs

anthrotype:
Closed by #2075

PR #2075, opened 2026-08-05
[variations] Regions reach axis extremes

### LINKED ISSUE #2012: Interpolation issue in Geist Pixel
Just noticed that when compiling https://github.com/vercel/geist-font/tree/Pixel-for-GF/sources/GeistPixel.glyphspackage in fontmake, glyphs and fontc, fontc is the odd one out:

Image

The default "pixel" component is a block at ELSP=100 and a line at ELSP=80, so at 80-and-a-bit it should still be a line. fontc makes it jump straight to a block.

For more fun: the fontc build is bigger than the Glyphs build but smaller than the fontmake build:

```
2.9M	build/font.ttf
2.0M	build/GeistPixelVF.ttf
3.4M	fonts/GeistPixel/variable/GeistPixel[ELSH].ttf
```

cmyr:
This reflects changes made to varlib in
https://github.com/fonttools/fonttools/pull/3601.

In short, we used to determine the reach of each variation purely from the min/max of the master locations, so an axis whose outermost master sits short of normalized ±1 had every region clipped there and stopped varying past that master.

With this patch we match the behaviour of current varLib; regions span the full normalized range.

Fixes https://github.com/googlefonts/fontc/issues/2012

### DIFF
--- a/fontc/src/lib.rs
+++ b/fontc/src/lib.rs
@@ -4394,6 +4394,51 @@ mod tests {
         );
     }
 
+    /// Variation regions reach the end of the axis, not the outermost master.
+    ///
+    /// The test font's wght axis runs 400..900 but its heaviest master is at 700, so
+    /// nothing sits at normalized 1.0; masters land at 0, 0.4 (a sparse brace layer)
+    /// and 0.6. We used to stop each region at the outermost master, which cut variation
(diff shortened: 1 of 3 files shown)


anthrotype:
this also closes https://github.com/googlefonts/fontc/issues/903

cmyr:
@anthrotype does that constitute an approval? 😇

anthrotype (approved):
LGTM

PR #2076, opened 2026-08-05
Migrate from env_logger to tracing

wmedrano:
Use `tracing` instead of `env_logger`. The functionality is mostly the same, but the directives available for the `--log` flag are slightly different
- `--log debug|info|warn|error` still works
- Filtering and more advanced stuff has different syntax. See docs for [env_logger](https://docs.rs/env_logger/latest/env_logger/#filtering-results) and [tracing](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#example-syntax)
- Motivation: Enables #2077 for better performance traces.

# Before

image

# After

image

### DIFF
--- a/fontc/src/main.rs
+++ b/fontc/src/main.rs
@@ -1,13 +1,60 @@
 use std::io::Write;
 
 use clap::Parser;
+use tracing::{error, warn};
+use tracing_log::NormalizeEvent;
+use tracing_subscriber::{
+    EnvFilter,
+    fmt::{FmtContext, FormatEvent, FormatFields, format::Writer},
+    registry::LookupSpan,
+};
 
 mod args;
 
 use args::Args;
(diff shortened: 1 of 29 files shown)


fontc/src/main.rs
```diff
@@ -1,13 +1,60 @@
...
+};
 
 mod args;
 
 use args::Args;
 use fontbe::orchestration::AnyWorkId;
 use fontc::{Error, JobTimer};
-use log::{error, warn};
+
+struct LogFormatter;
```
wmedrano, line 18: `LogFormatter` matches the old log output format. If we want to simplify and don't need the exact same format, this is the default:

Image

rsheeter (approved):
Are there any instructions that need to change, e.g. examples that talk about the svg output?

PR #2077, opened 2026-08-05
Export timings to "Trace Event Format"

wmedrano:
- This is a json file that can be viewed at ui.perfetto.dev
- Same format as used in Chrome
- Let's us delete our custom tracing and ui in fontc/src/timing.rs

# Demo

https://ui.perfetto.dev/#!/?s=12fd0a66454858159e461e60abb91d905d239d57

image

# Previous Format

image


cmyr (approved):
always feels good to delete code :)

PR #2079, opened 2026-08-11
Fixup how we init tracing_subscriber

cmyr:
Funny gotcha in this crate, but ::fmt()::init() creates a builder and then inits it without looking at the environment; and ::fmt::init() calls a free fn in the fmt module that _does_ look at the environment. (I was seeing an issue where doing `RUST_LOG=debug cargo run fontc_crater --etc --etc` wasn't showing me the expected log messages)

JMM


rsheeter (approved):
Boo environment vars, but makes sense.

PR #2082, opened 2026-08-18
[fea-rs] Parse Glyphs.app glyph predicates (phase 1)

anthrotype:
First of two PRs for #92 (supersedes the #2057 draft): this one teaches the fea-rs lexer and parser the Glyphs.app `$[...]` predicate surface; a stacked follow-up PR will expand name-only predicates at compile time.

The grammar follows the BNF sketched in https://github.com/googlefonts/fontc/pull/2057#issuecomment-4936914752, one node per production:

- `GlyphsCompoundPredicate = GlyphsPredicate [AND/OR GlyphsCompoundPredicate]` -> `GlyphsPredicateNode`, with the and/or connectives as distinct `GlyphsPredicateAnd`/`GlyphsPredicateOr` tokens
- `GlyphsPredicate = GlyphsPredicateObject GlyphsPredicateOperation GlyphsPredicateValue` -> `GlyphsPredicateClauseNode`
- `GlyphsPredicateObject = Ident` -> `GlyphsPredicateAttr` token
- `GlyphsPredicateOperation = Ident | == | > | < | ...` -> `GlyphsPredicateOpNode`
- `GlyphsPredicateValue = String | Number | Ident` -> `GlyphsPredicateValueNode`

The grammar deliberately accepts more than the compiler will initially support: anything predicate-shaped parses (chains of attribute/operator/value clauses joined by and/or connectives, keyword and symbolic operators, quoted and bare values), one whole token at a time, in the spirit of the number-values grammar's "just eat values and operators; we will validate later". Only constructs that don't fit that shape at all (NOT, parentheses, nesting) are parse errors; they point at #2052 (our phase 2 which tracks the parts of the predicate language we don't support yet). Narrowing the well-formed predicates down to the supported subset is validation's job, and in this PR validation is a stub rejecting every predicate ("not yet supported at compile time"), so compiled output is unchanged until the follow-up lands.

There are some minor known divergences from glyphsLib's opaque regex capture (`\$\[([^\]]+)\]`), summarized below, but none of these lets a predicate that both toolchains accept select different glyphs; where glyphsLib silently evaluates something other than what was written, fontc errors instead:

- a `"` or `#` inside a single-quoted value trips FEA's own string/comment lexing and breaks the parse (impossible in a real glyph-name value);
- operators and connectives must be their own tokens: glued spellings (`name contains"x"`, `x&&name`) are parse errors, though glyphsLib's boundary-free regexes accept some of them (every operator example in the Glyphs docs is spaced);
- trailing input after a complete clause is a parse error; glyphsLib silently drops whatever its capture leaves unconsumed (`name != 123abc` becomes `name != 123`, which selects every glyph).

NOTE: Most of the diff (especially the second commit) is test fixtures: a parse-tree fixture pinning the accepted surface, and error fixtures for each rejected family.


fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -109,6 +115,255 @@ fn glyph_class_list_member(parser: &mut Parser, recovery: TokenSet) -> bool {
...
+// written (dropped trailing input, boolean/integer typing of bare words),
+// fontc reports an error instead. They are documented, not fixed.
+fn eat_glyphs_predicate(parser: &mut Parser, recovery: TokenSet) -> bool {
+    let recovery = recovery.add(Kind::RSquare);
+    parser.in_node(AstKind::GlyphsPredicateNode, |parser| {
+        // The caller only enters on an adjacent `$[`; pin that invariant in
+        // debug builds but eat gracefully so a drifted guard cannot panic.
+        debug_assert!(parser.matches(0, Kind::Dollar) && parser.matches(1, Kind::LSquare));
+        parser.eat(Kind::Dollar);
+        parser.eat(Kind::LSquare);
```
cmyr: if this is ever not true something is majorly wrong. I'd just do,

```rust
assert!(parser.eat(Kind::Dollar));
assert!(parser.eat(Kind::LSquare));
```

fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -109,6 +115,255 @@ fn glyph_class_list_member(parser: &mut Parser, recovery: TokenSet) -> bool {
...
+            recovery,
+        );
+        return false;
+    }
+    // A `not` word or a `!` (Bang) at the head of a clause is negation, deferred
+    // to #2052. Both `! name` and `!name` begin with a Bang (because `!` delimits
+    // idents), so the single Bang check covers both spellings.
+    if parser.current_token_text().eq_ignore_ascii_case("not") || parser.matches(0, Kind::Bang) {
+        parser.err_recover(
+            "not predicates are not yet supported (see fontc#2052)",
```
cmyr:
```suggestion
            "negation (not/!) are not yet supported in predicates (see fontc#2052)",
```

fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -109,6 +115,255 @@ fn glyph_class_list_member(parser: &mut Parser, recovery: TokenSet) -> bool {
...
+            recovery,
+        );
+        return false;
+    }
+
+    parser.in_node(AstKind::GlyphsPredicateClauseNode, |parser| {
+        if !eat_glyphs_predicate_attr(parser) {
```
cmyr: I would rename to `expect_glyphs_predicate_attr` (and ditto for the other three fns here) and then move the error reporting into those fns; then this just becomes,

```rust
expect_glyphs_predicate_attr(parser)
&& expect_glyphs_predicate_op(parser, recovery)
&& expect_glyphs_predicate_value(parser, recovery)
```

fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -109,6 +115,255 @@ fn glyph_class_list_member(parser: &mut Parser, recovery: TokenSet) -> bool {
...
+        return false;
+    }
+
+    // Composite symbolic operators are two adjacent tokens; check adjacency of
+    // the pair (as the `!=` guard above does) before eating, so all sites share
+    // `glyphs_predicate_tokens_are_adjacent` rather than re-deriving offsets.
+    parser.in_node(AstKind::GlyphsPredicateOpNode, |parser| {
+        if is_word {
+            parser.eat_raw();
+            true
```
cmyr: I would `eat_remap` here, to something like `GlyphsPredicateOpName`?  It's very defensive, but it guards against a situation where there is an op name that overlaps with a keyword or something else that doesn't just parse as an ident, which could fall through the cracks later?

fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -109,6 +115,255 @@ fn glyph_class_list_member(parser: &mut Parser, recovery: TokenSet) -> bool {
...
+}
+
+fn eat_glyphs_predicate_attr(parser: &mut Parser) -> bool {
+    if !is_glyphs_predicate_word(parser.current_token_text()) {
+        return false;
+    }
+    parser.eat_remap(parser.nth(0).kind, AstKind::GlyphsPredicateAttr)
+}
+
+fn eat_glyphs_predicate_op(parser: &mut Parser, _recovery: TokenSet) -> bool {
```
cmyr: it would be nice to document the actual list of ops? According to [the apple docs](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Predicates/Articles/pBNF.html#:~:text=%22=%22%20|%20%22!=%22%20|%20%22%3C%22%20|%20%22%3E%22%20|%20%22%3C=%22%20|%20%22%3E=%22), the supported (non-string) operators are just, `"=" | "!=" | "<" | ">" | "<=" | ">="`; and `==` isn't supported at all, although it's used heavily in the glyphs.app documentation. I must be missing something...

fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -109,6 +115,255 @@ fn glyph_class_list_member(parser: &mut Parser, recovery: TokenSet) -> bool {
...
+    );
+    if !is_word && !is_symbol {
+        return false;
+    }
+    // A `!` is an operator only as the start of an adjacent `!=`.
+    if parser.matches(0, Kind::Bang)
+        && !(parser.matches(1, Kind::Eq) && glyphs_predicate_tokens_are_adjacent(parser, 0, 1))
+    {
+        return false;
+    }
```
cmyr, line 260: okay I think I have a simpler way to do this whole fn. Basically:

- we give the parser the ability to combine multiple lexemes into a single token.
- we use this to remap `!=`, `<=` etc to `Kind::GlyphsPredicateOpNotEq`, Kind::GlyphsPredicateOpLeq` etc.
- we don't need `GlyphsPredicateOpNode`, since all ops are now single tokens
- we have a `GlyphsPredicateOp` enum in the `typed` module.

this should be a significant simplification? here's a sketch of the new fn..


```rust
/// either the name of an operation or one of [==, !=, <=, >=, =>, =<, <>, <, >,]
fn expect_glyphs_predicate_op(parser: &mut Parser, recovery: TokenSet) -> bool {
    if is_glyphs_predicate_word(parser.current_token_text()) {
        return expect_glyphs_predicate_op_name(parser);
    }

    if parser.eat_adjacent_remap(Kind::Eq, Kind::Eq, AstKind::GlyphsPredicateOpEq)
        || parser.eat_adjacent_remap(Kind::Bang, Kind::Eq, AstKind::GlyphsPredicateOpNe)
        || parser.eat_adjacent_remap(Kind::LAngle, Kind::RAngle, AstKind::GlyphsPredicateOpNe)
        || parser.eat_adjacent_remap(Kind::LAngle, Kind::Eq, AstKind::GlyphsPredicateOpLe)
        || parser.eat_adjacent_remap(Kind::Eq, Kind::LAngle, AstKind::GlyphsPredicateOpLe)
        || parser.eat_adjacent_remap(Kind::RAngle, Kind::Eq, AstKind::GlyphsPredicateOpGe)
        || parser.eat_adjacent_remap(Kind::Eq, Kind::RAngle, AstKind::GlyphsPredicateOpGe)
        || parser.eat_remap(Kind::Eq, AstKind::GlyphsPredicateOpEq)
        || parser.eat(Kind::LAngle)
        || parser.eat(Kind::RAngle)
    {
        return true;
    }
    parser.err_recover("expected glyphs predicate", recovery);
    false
}
```

where `eat_adjacent_remap` looks like,


```rust
    /// Eat the next two tokens as a *single* token of kind `remap`.
    ///
    /// The two tokens must not be separated by whitespace.
    pub(crate) fn eat_adjacent_remap(
        &mut self,
        first: impl TokenComparable,
        second: impl TokenComparable,
        remap: Kind,
    ) -> bool {
        if self.matches(0, first)
            && self.matches(1, second)
            && self.nth_range(0).end == self.nth_range(1).start
        {
            self.do_bump::<2>(remap);
            return true;
        }
        false
    }
```


Doing this and then digging into the logic in typed.rs also made me think that (despite my earlier advice, I acknowledge) it's _probably_ cleaner to also just parse all of the named operators here as well; for me this looked like,


```rust
fn expect_glyphs_predicate_op_name(parser: &mut Parser) -> bool {
    let kind = match parser.current_token_text() {
        text if text.eq_ignore_ascii_case("beginswith") => AstKind::GlyphsPredicateOpBeginsWith,
        text if text.eq_ignore_ascii_case("endswith") => AstKind::GlyphsPredicateOpEndsWith,
        text if text.eq_ignore_ascii_case("contains") => AstKind::GlyphsPredicateOpContains,
        text if text.eq_ignore_ascii_case("like") => AstKind::GlyphsPredicateOpLike,
        text if text.eq_ignore_ascii_case("matches") => AstKind::GlyphsPredicateOpMatches,
        _ => {
            parser.err_and_bump("unknown glyphs predicate operator");
            return false;
        }
    };
    parser.eat_remap(TokenSet::IDENT_LIKE, kind)
}
```

This is all self-contained, and it saves us from having a bunch of custom logic and hand-written types in typed.rs, which feels cleaner.

fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -109,6 +115,255 @@ fn glyph_class_list_member(parser: &mut Parser, recovery: TokenSet) -> bool {
...
+        .is_some_and(|c| c.is_ascii_digit())
+    {
+        AstKind::Number
+    } else {
+        AstKind::Ident
+    };
+
+    parser.in_node(AstKind::GlyphsPredicateValueNode, |parser| {
+        parser.eat_remap(parser.nth(0).kind, target)
+    })
```
cmyr: if values are always single tokens than we probably don't need a node type here, either; we can use `ast_enum!` in typed.rs to define a `GlyphsAppPredicateValue` enum that will have 'number' and 'ident' variants.

fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -245,6 +503,165 @@ mod tests {
...
+        let mut sink = AstSink::new(fea, FileId::CURRENT_FILE, None);
+        let mut parser = Parser::new(fea, &mut sink);
+        assert!(eat_glyph_name_like(&mut parser));
+        let stray_bang = parser.matches(0, Kind::Bang);
+        assert!(sink.errors().is_empty(), "'hi' should parse cleanly");
+        assert!(stray_bang, "the stray '!' remains as Bang");
+    }
+
+    #[test]
+    fn glyphs_predicate_has_structured_children() {
```
cmyr: this should be a parse test; that system is designed for exactly this (checking that some input has the expected parsed structure)

fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -245,6 +503,165 @@ mod tests {
...
+        ));
+        assert_eq!(clauses[1].value().unwrap().text(), "latn");
+        assert!(matches!(
+            predicate.connectives().next(),
+            Some(typed::GlyphsAppPredicateConnective::Or(_))
+        ));
+    }
+
+    #[test]
+    fn glyphs_predicate_accepts_keyword_words_and_connectives() {
```
cmyr: ditto, this is much cleaner expressed as a parse test.

fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -245,6 +503,165 @@ mod tests {
...
+        assert_eq!(clauses[1].attr().unwrap().text(), "mark");
+        assert_eq!(clauses[1].value().unwrap().text(), "mark");
+        assert!(matches!(
+            predicate.connectives().next(),
+            Some(typed::GlyphsAppPredicateConnective::And(_))
+        ));
+    }
+
+    #[test]
+    fn glyphs_predicate_types_keyword_operators() {
```
cmyr: if we go ahead and parse all the named operators we can make this a parse test too.

fea-rs/src/parse/lexer.rs
```diff
@@ -101,6 +101,13 @@ impl<'a> Lexer<'a> {
             b'\\' => Kind::Backslash,
             b'-' => self.hyphen_or_minus(),
             b'=' => Kind::Eq,
+            // `!` is `Bang` (and an ident delimiter, see `is_special`) so that
+            // `name!="x"` lexes as `name` `!` `=` `"x"`. This changes the error
+            // shape of already-invalid `!`-adjacent input: `hi!` used to lex as
+            // one bad Ident (rejected as an invalid glyph name), and now lexes
+            // as `hi` (a valid name) plus a stray `!` that a caller rejects
+            // downstream. No legal FEA contains `!`, so nothing valid is affected.
```
cmyr: I don't think this comment is really adding much?

fea-rs/src/tests/compile.rs
```diff
@@ -147,6 +147,32 @@ fn compile_fea_variable(fea: &str, test_name: &str) -> Compilation {
         .expect("compilation should succeed")
 }
 
+#[test]
+fn glyphs_predicate_fails_validation_until_compile_support_is_added() {
```
cmyr: if this is still relevant is it well expressed as a normal compile test (text files).

fea-rs/src/token_tree/typed.rs
```diff
@@ -431,6 +431,68 @@ ast_enum!(GlyphsAppNumberValue {
...
+// a connective between predicate clauses; the grammar sorts the `and`/`&&` and
+// `or`/`||` spellings into two kinds, so the classification is in the tree
+ast_enum!(GlyphsAppPredicateConnective {
+    And(GlyphsAppPredicateAnd),
+    Or(GlyphsAppPredicateOr),
+});
+
+/// An operator written in a Glyphs.app glyph predicate.
+#[derive(Clone, Debug)]
+pub enum GlyphsAppPredicateOp {
```
cmyr: yea it's annoying that our deferred parsing pushes us in this direction, and it's an argument for just doing the parsing ahead of time.

fea-rs/test-data/parse-tests/bad/glyphsapp_predicate_bad.fea
```diff
@@ -0,0 +1,24 @@
+@missing_value = [$[name == ]];
```
cmyr: i'd generally prefer to split these different failures into more cases, if there's a reasonable way to group them?

cmyr (approved):
this looks good! There are a few details around the parsing logic that I have some thoughts on (and have left detailed comments about inline) but big picture this makes sense to me :)

fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -189,114 +193,96 @@ fn eat_glyphs_predicate_clause(parser: &mut Parser, recovery: TokenSet) -> bool
...
-            //
-            // The quoted content is otherwise opaque: scan to the closing quote
-            // with predicate-local stop points only, not the caller's
-            // statement-level recovery set, whose keywords (e.g. `by` in a GSUB
-            // rule) may legitimately appear inside the quotes. `]` still bounds
-            // the scan; glyphsLib's `$[([^\]]+)]` capture cannot contain one
-            // either.
             parser.eat_until(TokenSet::new(&[Kind::SingleQuote, Kind::RSquare]));
             parser.expect_recover(Kind::SingleQuote, recovery);
             true
```
cmyr: we should return the result of `expect_recover`, and not blinding `true` here I think?

fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -311,25 +297,24 @@ fn eat_glyphs_predicate_value(parser: &mut Parser, recovery: TokenSet) -> bool {
...
+            .current_token_text()
+            .chars()
+            .next()
+            .is_some_and(|c| c.is_ascii_digit())
+        {
+            AstKind::Number
+        } else {
+            AstKind::Ident
+        };
+        return parser.eat_remap(parser.nth(0).kind, target);
```
cmyr, line 328: hmm, could this just be,

```rust
parser.eat(Kind::Number) || parser.eat(Kind::Ident)
```
?

fea-rs/src/token_tree/typed.rs
```diff
@@ -1918,118 +1925,47 @@ impl GlyphsAppPredicateClause {
...
-            "=>" => Self::GreaterThanOrEqual(node),
-            "=<" => Self::LessThanOrEqual(node),
-            _ => Self::UnknownKeyword(node),
-        }
-    }
-
-    /// The exact parsed operator node, including its source range.
-    pub(crate) fn node(&self) -> &GlyphsAppPredicateOpNode {
+    /// The operator spelling, as written in the source.
+    pub(crate) fn text(&self) -> &str {
```
cmyr: is this still going to be necessary?

fea-rs/src/token_tree/typed.rs
```diff
@@ -1918,118 +1925,47 @@ impl GlyphsAppPredicateClause {
...
-                // strip the surrounding quotes; `get` avoids panicking on a
-                // malformed short token (e.g. a lone `"` recovered from an
-                // unterminated string) or a non-char-boundary slice.
-                .and_then(|text| text.get(1..text.len().saturating_sub(1)))
+            Self::Bare(token) => token.text().to_string(),
+            Self::Number(token) => token.text().to_string(),
+            // strip the surrounding quotes; `get` avoids panicking on a
+            // malformed short token (e.g. a lone `"` recovered from an
+            // unterminated string) or a non-char-boundary slice.
+            Self::DoubleQuoted(token) => token
```
cmyr: why not just `token.text().trim_matches(")`?

fea-rs/src/token_tree/typed.rs
```diff
@@ -1918,118 +1925,47 @@ impl GlyphsAppPredicateClause {
...
+                .text()
+                .get(1..token.text().len().saturating_sub(1))
                 .unwrap_or_default()
                 .to_owned(),
             Self::SingleQuoted(node) => {
```
cmyr, line 1946: i do find this singlequoted business the trickiest bit of parsing here, and i'm not sure how best to think about it? I do think there's probably a better approach but I'm not sure what it is. :/

fea-rs/test-data/parse-tests/bad/glyphsapp_predicate_malformed_bad.ERR
```diff
@@ -0,0 +1,89 @@
...
+error: Unexpected token, expected glyph or glyph class
+in ./test-data/parse-tests/bad/glyphsapp_predicate_malformed_bad.fea at 9:18
+  | 
+9 | @spaced_dollar = [$ [name == "x"]];
+  |                   ^
+
+error: Expected ';'
+in ./test-data/parse-tests/bad/glyphsapp_predicate_malformed_bad.fea at 9:33
+  | 
+9 | @spaced_dollar = [$ [name == "x"]];
```
cmyr, line 46: this is a funny error message... it's stopping at the first `]` and is then confused by the second one. Maybe `]` doesn't actually need to be in our recovery set? 🤷

fea-rs/test-data/parse-tests/bad/glyphsapp_predicate_malformed_bad.ERR
```diff
@@ -0,0 +1,89 @@
...
+error: Expected glyph or glyph class
+in ./test-data/parse-tests/bad/glyphsapp_predicate_malformed_bad.fea at 11:8
+   | 
+11 |     sub $[name == "not-a-class-member"] by B;
+   |         ^
+
+warning: ';' should only follow a statement
+in ./test-data/parse-tests/bad/glyphsapp_predicate_malformed_bad.fea at 11:44
+   | 
+11 |     sub $[name == "not-a-class-member"] by B;
```
cmyr: this is a funny one too...

fea-rs/test-data/parse-tests/bad/glyphsapp_predicate_value_bad.fea
```diff
@@ -0,0 +1,4 @@
+# numeric value forms that glyphsLib's `\d+` typing would silently truncate
```
cmyr, line 1: so we're matching glyphsLib.. what does glyphs.app do?

cmyr (approved):
cool I think this is looking good. A few more notes inline, but I hope that overall the architecture of fea-rs is starting to make more sense :)

anthrotype:
Thanks for both review rounds! Each is addressed in its own follow-up commit so the incremental diffs are easy to review. The first commit adopted your `eat_adjacent_remap` sketch wholesale (per-op token kinds, eager named-operator parsing), and the one I just pushed covers the second round. I'll squash them into the stack once you're
happy. (You'll probably do a third round before I can even finish replying to all the inline comments 😅)

fea-rs/src/parse/grammar/glyph.rs
anthrotype: that pBNF page is just a narrower layer than the one we have to match. It just lists one spelling per operator and doesn't mention the alternates.

Apple's [*Predicate Format String Syntax*](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Predicates/Articles/pSyntax.html) page (the prose companion to the one you linked) documents the basic comparison operators in pairs, one description per pair:

- `=`, `==` -- "The left-hand expression is equal to the right-hand expression."
- `>=`, `=>` -- greater than or equal
- `<=`, `=<` -- less than or equal
- `!=`, `<>` -- not equal

More to the point, our parity target isn't `NSPredicate` but glyphsLib's `TokenExpander` (`glyphsLib/builder/tokens.py`), which does its own regex parse and spells out all four pairs.

(`=>` and `=<` are listed in that regex but unreachable in practice, but that a glyphsLib's bug)

fea-rs/test-data/parse-tests/bad/glyphsapp_predicate_value_bad.fea
anthrotype: So Glyphs.app's engine silently selects nothing for name predicates with numeric values (glyphsLib is either
silent or raises ValueError). The quoted form is a plain string comparison in all three implementations (unlikely with `==` since a glyph name can't start with a digit anyway). That's why these are rejected with "must be quoted".

fea-rs/src/parse/grammar/glyph.rs
anthrotype: Almost, but the value position also accepts spellings that don't lex as Ident or Number: e.g. keyword-shaped words (`$[language == NULL && mark == mark]` in the good fixture, where `NULL` and `mark` lex as their keyword kinds) and leading-zero digit runs (`077` lexes as Octal). The remap folds all of them onto the two value kinds so the typed enum stays two-variant. 

Matching on the token text rather than the kind is also what keeps parity with glyphsLib's bare-value pattern
(Python `\w+`, Unicode-aware).

fea-rs/test-data/parse-tests/bad/glyphsapp_predicate_malformed_bad.ERR
anthrotype: `$ [` fails the adjacency check so the class parser treats `$` as a bad member ("Unexpected token, expected glyph or glyph class"), and its recovery (`eat_glyph_class_list`'s `recovery.add(RSquare)`, which predates this PR) stops at the first `]` and closes the class early. The leftover `]` then trips the statement parser twice.

Dropping RSquare from that set would change recovery for every malformed glyph class in the language, not just predicates, so that feels like its own change rather than part of this PR.

fea-rs/src/token_tree/typed.rs
anthrotype: The constraint is that the lexer is context-free and `'` is the glyph-marker token for contextuals, so a single-quoted value can never be a single lexeme without a lexer mode. And the mode switch would need parser feedback (`'` means "string" only in this one position), which seemed worse than the scan.

Maybe If a later phase grows more quoted contexts, a lexer-level approach may be worth revisiting then...

fea-rs/src/token_tree/typed.rs
anthrotype: Yes. The yet-to-be-filed compile PR's validation diagnostics do quote the spelling through `GlyphsAppPredicateOp::text()`. The `dead_code` will be removed there.

fea-rs/src/parse/grammar/glyph.rs
cmyr: hmm, okay. I'll register that I don't love it, but if it's the expedient answer that's fine. I'd make sure we capture why we do it this way in a comment (this might already exist, i can't see the context here)

fea-rs/src/token_tree/typed.rs
cmyr: hmm, that might not be necessary? errors from fea-rs are attached to spans, so the error will be showing exactly what word is the issue, which might be enough? but whatever you think you need..

fea-rs/src/token_tree/typed.rs
cmyr: the other way to do this would be to reparse into a new token after lexing, when we're in the predicate parsing code (sort of like how we handle == and <=)

fea-rs/src/parse/grammar/glyph.rs
```diff
@@ -40,7 +40,29 @@ pub(crate) fn named_glyph_class_decl(parser: &mut Parser, recovery: TokenSet) {
...
+        || eat_named_or_unnamed_glyph_class(parser, recovery)
+        || eat_misplaced_glyphs_predicate(parser, recovery)
+}
+
+/// A `$[...]` glyphs predicate outside a glyph class.
+///
+/// Parse it in full so recovery resumes cleanly after it, but report that it
+/// is not valid in this position.
+fn eat_misplaced_glyphs_predicate(parser: &mut Parser, recovery: TokenSet) -> bool {
+    if parser.matches(0, Kind::Dollar)
```
cmyr: this is just for better error reporting? in practice I'm not sure it's worth doing this. The point of the `eat_` functions is that they advance the cursor and return `true` if the given construct exists. `expect_` functions try to eat the given construct, and report an error if it isn't found.

if we wanted this logic to check for a dangling glyphs predicate, I would _probably_ choose to do that in `eat_glyph_class_list`; I would have a case right at the start of that fn that looked like,

```rust
if parser.matches(0, Kind::Dollar) && parser.matches(1, Kind::LSquare) {
    parser.error("glyph class predicate must be declared inside class, like `[ $[predicate] ]`");
    parser.eat_until(recovery);
    return true;
```

anthrotype: I believe a `$[...]` at a rule position seems a likely enough mistake to deserve a real message so I kept it but moved it into `eat_glyph_class_list` as you suggested; `eat_misplaced_glyphs_predicate` is gone.

It still parses the predicate in full rather than `eat_until`, so the `sub $[...] by X;` case recovers at `by` with a single error

fea-rs/src/parse/grammar/glyph.rs
anthrotype: > make sure we capture why we do it this way in a comment 

Done in 4b95f6ce

fea-rs/src/token_tree/typed.rs
anthrotype: That could work, though I'd rather not rework it in this PR, the scan is contained and tested. Could be done as a follow-up if you feel strongly

fea-rs/src/token_tree/typed.rs
anthrotype: Yeah the span was enough (dropped in 4b95f6ce). My compile branch's operator diagnostic will rely on that instead of quoting the spelling.

anthrotype:
I promised to squash before merge but then I forgot. Oh well, I doesn't really matter. I'll try to remember next time

PR #2083, opened 2026-08-18
Reject ambiguous mappings in PiecewiseLinearMap::new

### LINKED ISSUE #937: let PiecewiseLinearMap::new return error if input mappings contain duplicate keys
slice::binary_search isn't well defined with multiple values https://github.com/googlefonts/fontc/issues/933

In https://github.com/googlefonts/fontc/pull/935#issuecomment-2331996519 I commented

> I now also wonder if PiecewiseLinearMap::new (and as a result CoordConverter::new) should be fallible, in case we ever construct them with duplicate from inputs which would lead to a similar ambiguous situation.
>
> in fonttools this situation never occurs because mappings are a dict where keys are unique by definition.
When reading a designspace axis' maps, the list of (user, design) tuples is turned into a dict (with a dict comprehension, where last key wins in case of duplicates):
https://github.com/fonttools/fonttools/blob/c2119229cfb02cdb7c5a63374ef29d3d514259e8/Lib/fontTools/designspaceLib/__init__.py#L1031-L1047
>
> not sure if fontmake warns about this ambiguity (maybe glyphsLib does when constructing the DS axes)

hdimer:
Fixes #937.

## The bug

`PiecewiseLinearMap::new` sorted the `(from, to)` pairs and unzipped them into parallel `Vec`s with no duplicate detection and no error path. Two mappings sharing a `from` but disagreeing on `to` were both stored, and `map()` then resolved an exact hit via `partition_point` and returned whichever sorted first, silently dropping the other mapping.

On `main` (ee249ef):

```rust
let plm = PiecewiseLinearMap::new(vec![(0.0, 0.0), (10.0, 100.0), (10.0, 999.0)]);
// len() == 3, map(10.0) == 100.0 -- (10.0, 999.0) is silently discarded
```

## The fix

`new` now returns `Result`, with a `DuplicateMapInput` variant on the existing `fontdrasil::error::Error`. `CoordConverter::new` was already fallible, so both call sites just propagate with `?` and nothing outside `fontdrasil` changes. (`PiecewiseLinearMap` is crate-private, so this isn't a public API break; `Error` does gain a variant.)

Two judgement calls worth your attention:

**Duplicate `from` is rejected, duplicate `to` is not.** Per @anthrotype in #1528, "duplicate keys are bad, but duplicate values are fine". `reverse()` therefore builds its map without the check, since reversing a many-to-one map is exactly how duplicate `from` values legitimately arise, and `map()` resolves those to the first match on purpose (the ufo2ft#978 behaviour).

**Exact duplicate mappings are collapsed rather than rejected.** Repeating a mapping identically is redundant, not ambiguous, and nothing gets discarded. This is not hypothetical: `fea-rs`'s `simple_axis` builds `[(min, min), (default, default), (max, max)]`, so any axis with `min == default` repeats a mapping, and rejecting those broke the `fea-rs` suite. `CoordConverter::unmapped` already handles the same case with `mappings.dedup()`, so this follows the idiom that was already in the file. Happy to make it strict instead if you'd rather callers pre-dedup.

## Things to weigh

- **This makes fontc stricter than fontmake.** fonttools builds its mapping from a `dict`, so a conflicting duplicate `<map>` collapses last-wins and never errors; fontc previously took first-after-sort. A source in the wild with a genuinely conflicting duplicate will now fail a build that used to succeed (quietly disagreeing with fontmake). That's what #937 asks for, but a crater run is the real gate here, not the unit tests. Reachable from `.designspace` and fontra; the glyphs path keys `axis_mappings` by user value, so duplicates can't survive parsing.
- **`avar` gets slightly more spec-correct.** `to_segment_map` is the only production consumer of `CoordConverter::iter()`. For a source with an exactly-repeated `<map>`, it previously emitted a `SegmentMaps` with a duplicate `fromCoordinate`, which the spec forbids. `fvar` is unaffected.
- **Out of scope, but noted:** `CoordConverter::default_idx` indexes the caller's mappings, not `user_to_design`, which has always been sorted and is now also deduped. Nothing in production reads the field, so I only documented it rather than recomputing it. Say the word if you'd like it fixed properly, here or separately.

## Testing

`cargo fmt`, `cargo clippy --all-features --all-targets -D warnings`, and `cargo doc -D warnings` are clean, and the full suite passes: fontdrasil 63, fontir 180, glyphs-reader 278, glyphs2fontir 89, ufo2fontir 80, fontra2fontir 20, fontbe 132, fontc 202. I couldn't run the `fea-rs` ttx tests or `fontc_crater` locally (no `ttx` on PATH, no `cmake` for `tidy-sys`); both fail identically on a clean checkout here, and `fontc_crater` has no `fontdrasil` dependency.

I mutation-tested the new tests rather than trusting them: dropping the `dedup()`, deduping by `from` alone (which would reintroduce the original silent drop), and comparing `to` instead of `from` are each caught by at least two tests.

I use AI tooling to help with my open source work. This patch was written and tested with that assistance, and I've reviewed and verified all of it myself.

### DIFF
--- a/fontdrasil/src/coords.rs
+++ b/fontdrasil/src/coords.rs
@@ -176,6 +176,7 @@ impl Coord<NormalizedSpace> {
 // suggest <= 10 mappings is typical, we can afford the bytes.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 pub struct CoordConverter {
+    /// Indexes the caller's mappings, not `user_to_design`, which is sorted and deduped.
     pub(crate) default_idx: usize,
     pub(crate) user_to_design: PiecewiseLinearMap,
     design_to_user: PiecewiseLinearMap,
@@ -197,7 +198,7 @@ impl CoordConverter {
                 .iter()
                 .map(|(u, d)| (u.into_inner(), d.into_inner()))
                 .collect(),
-        );
+        )?;
 
         let design_coords: Vec<_> = mappings.iter().map(|(_, d)| d).collect();
         #[allow(clippy::unwrap_used)] // We checked above that mappings is not empty
@@ -216,7 +217,7 @@ impl CoordConverter {
         if *design_max > design_default {
             examples.push((design_max.into_inner(), 1.0.into())); // right of default *must* be +1
         }
-        let design_to_normalized = PiecewiseLinearMap::new(examples);
+        let design_to_normalized = PiecewiseLinearMap::new(examples)?;
 
         let design_to_user = user_to_design.reverse();
         let normalized_to_design = design_to_normalized.reverse();
@@ -587,6 +588,7 @@ mod tests {
     #![allow(clippy::unwrap_used)] // test code
 
     use super::{CoordConverter, DesignCoord, NormalizedCoord, NormalizedLocation, UserCoord};
+    use crate::error::Error;
     use write_fonts::types::Tag;
 
     // From <https://github.com/googlefonts/fontmake-rs/blob/main/resources/text/units.md>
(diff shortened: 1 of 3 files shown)


fontdrasil/src/coords.rs
```diff
@@ -176,6 +176,7 @@ impl Coord<NormalizedSpace> {
 // suggest <= 10 mappings is typical, we can afford the bytes.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 pub struct CoordConverter {
+    /// Indexes the caller's mappings, not `user_to_design`, which is sorted and deduped.
     pub(crate) default_idx: usize,
```
anthrotype: Could we drop the stored `default_idx` instead of documenting it? It is only used locally by `new` to select `design_default`; no production code reads `converter.default_idx`, and the current tests only assert that internal state. Keeping it after `user_to_design` is sorted/deduped leaves an index into discarded input, which can be misleading. I'd just keep the constructor parameter and remove those direct test assertions. The derived serialization is not a compatibility commitment.

anthrotype (commented):
The core approach looks right: exact duplicates are harmlessly collapsed, conflicting input coordinates become a well-scoped error, and reverse() preserves the intentional many-to-one semantics from #1528. Apart from the default_idx cleanup, I have no code concerns.

fontdrasil/src/coords.rs
hdimer: Done, dropped it. Kept the parameter, `design_default` is picked from it locally in `new` and the `DefaultOutOfBounds` check is unchanged.

`unmapped_coords_get_deduped` had nothing in it but `default_idx` assertions, so rather than delete the test I kept the two things the index was standing in for: the dedup count, and that the default still normalizes to 0. Folded the four cases into a table to match the loop style in `agl.rs`. The other three tests already assert `len()` plus the value at every vertex, so those lines just came out.

One note since you mentioned serialization: `CoordConverter` rides into the persisted IR yaml via `StaticMetadata`. Nothing in the workspace uses `deny_unknown_fields`, so a stale IR dir with `default_idx` in it still deserializes fine, and no fixtures contain the key.

anthrotype (approved):
Thanks. mind also signing the CLA?

hdimer:
> Thanks. mind also signing the CLA?

All done. Let me know if you need anything else.

PR #2084, opened 2026-08-19
fea-rs: accept bare defaults in variable metrics

anthrotype:
Match fontTools' variable-scalar shorthand by allowing one bare value to supply the default location. Reject duplicate bare defaults and test that the shorthand compiles identically to spelling out the default location.

https://github.com/fonttools/fonttools/pull/4024

For example, with a `wght` default of 400, `(-12 wght=700:22)` is equivalent to `(wght=400:-12 wght=700:22)`.

/cc @khaledhosny

### DIFF
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -1338,20 +1338,22 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
         let mut locations = HashMap::new();
         for metric_loc in metric.location_values() {
             let mut pos = NormalizedLocation::new();
-            for axis_value in metric_loc.location().items() {
-                let tag = axis_value.axis_tag().to_raw();
-                // All the tags are valid if we made it here, safe to unwrap
-                let (_, axis) = var_info.axis(tag).unwrap();
-                let coord = match axis_value.value().parse() {
-                    super::AxisLocation::Normalized(value) => NormalizedCoord::new(value),
-                    super::AxisLocation::User(value) => {
-                        UserCoord::new(value).to_normalized(&axis.converter)
-                    }
-                    super::AxisLocation::Design(value) => {
-                        DesignCoord::new(value).to_normalized(&axis.converter)
-                    }
-                };
-                pos.insert(tag, coord);
+            if let Some(location) = metric_loc.location() {
+                for axis_value in location.items() {
+                    let tag = axis_value.axis_tag().to_raw();
+                    // All the tags are valid if we made it here, safe to unwrap
+                    let (_, axis) = var_info.axis(tag).unwrap();
+                    let coord = match axis_value.value().parse() {
+                        super::AxisLocation::Normalized(value) => NormalizedCoord::new(value),
+                        super::AxisLocation::User(value) => {
+                            UserCoord::new(value).to_normalized(&axis.converter)
+                        }
+                        super::AxisLocation::Design(value) => {
+                            DesignCoord::new(value).to_normalized(&axis.converter)
+                        }
+                    };
+                    pos.insert(tag, coord);
+                }
             }
             locations.insert(pos, metric_loc.value().parse_signed());
         }
--- a/fea-rs/src/compile/validate.rs
+++ b/fea-rs/src/compile/validate.rs
@@ -1377,8 +1377,19 @@ impl<'a, V: VariationInfo> ValidationCtx<'a, V> {
             return;
         };
 
+        let mut bare_default_seen = false;
         for location_val in metric.location_values() {
-            for item in location_val.location().items() {
+            let Some(location) = location_val.location() else {
+                if bare_default_seen {
+                    self.error(
+                        location_val.range(),
+                        "duplicate value for the default location",
+                    );
+                }
+                bare_default_seen = true;
+                continue;
+            };
+            for item in location.items() {
                 let Some((_, axis)) = var_info.axis(item.axis_tag().to_raw()) else {
                     self.error(item.axis_tag().range(), "unknown axis");
                     continue;
--- a/fea-rs/src/parse/grammar/metrics.rs
+++ b/fea-rs/src/parse/grammar/metrics.rs
@@ -114,7 +114,7 @@ pub(crate) fn expect_metric(parser: &mut Parser, recovery: TokenSet) -> bool {
 /// (<https://glyphsapp.com/learn/tokens#g-number-values>)
 ///
 /// A variable metric has the syntax `(<location_value>+)`
-/// where `<location_value>` is `<location_spec>:<number>`
+/// where `<location_value>` is `<location_spec>:<number>` or a bare default `<number>`
 /// and a `<location_spec>` is `<axis_tag>=<number>,+`
 ///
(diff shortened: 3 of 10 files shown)


anthrotype:
After filing this, I found that a scalar like e.g. `(10 wght=400:20 wght=700:30)` with `wght` defaulting to 400 supplied two values for the same default location, which became identical when VariationModel expanded each location to the model’s axes (adding absent axes at normalized zero); then the subsequent HashMap rebuild would nondeterministically keep either 10 or 20..

The follow-up commit rejects these collisions matching fontTools and fixing the pre-existing sparse explicit form too.

fea-rs/src/compile/compile_ctx.rs
```diff
@@ -1338,20 +1338,22 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
...
-                    super::AxisLocation::Normalized(value) => NormalizedCoord::new(value),
-                    super::AxisLocation::User(value) => {
-                        UserCoord::new(value).to_normalized(&axis.converter)
-                    }
-                    super::AxisLocation::Design(value) => {
-                        DesignCoord::new(value).to_normalized(&axis.converter)
-                    }
-                };
-                pos.insert(tag, coord);
+            if let Some(location) = metric_loc.location() {
```
cmyr, line 1341: could we do `let Some(location) = metric_loc.location() else { continue };` and have a bit less nesting?

I guess we could also do a filter_map on the iterator.

fea-rs/src/compile/validate.rs
```diff
@@ -1378,7 +1378,10 @@ impl<'a, V: VariationInfo> ValidationCtx<'a, V> {
         };
 
         for location_val in metric.location_values() {
-            for item in location_val.location().items() {
+            let Some(location) = location_val.location() else {
```
cmyr, line 1382: okay yea this is what i was imagining...

fea-rs/src/parse/grammar/metrics.rs
```diff
@@ -131,12 +131,21 @@ fn eat_metric(parser: &mut Parser, recovery: TokenSet) -> bool {
...
     parser.in_node(AstKind::VariableMetricNode, |parser| {
         parser.eat_raw();
+        let mut bare_default_seen = false;
         if parser.in_node(AstKind::LocationValueNode, |parser| {
-            expect_variation_location_and_value(parser, recovery.add(Kind::RParen))
+            expect_variation_location_and_value(
+                parser,
+                recovery.add(Kind::RParen),
+                &mut bare_default_seen,
+            )
```
cmyr: okay, i really don't love threading through this &mut bool.

if I understand correctly, the bare default would always be first? In that case I think a simpler approach is to just have something like where we optionally accept a single number, first, and then stick with the existing `expect_variation_location_and_value`. something like,


```suggestion
           parser.eat(Kind::Number) || eat_variation_location_and_value(parser, recovery.add(Kind::RParen))
```

and then we don't need to change much else, since we'll error on any subsequent bare number?

the only modest downside being that we don't get as useful an error message, although that's easy enough to address.

fea-rs/test-data/compile-tests/mini-latin/bad/variable_metrics_duplicate_default.ERR
```diff
@@ -0,0 +1,5 @@
+error: duplicate value for the default location
```
cmyr, line 1: excellent, I find these tests so much easier to follow.

fontdrasil/src/variations.rs
```diff
@@ -379,14 +379,17 @@ impl VariationModel {
...
+            for (location, value) in seqs {
+                let mut location = location.to_owned();
+                location.fit_to_axes(self.axis_order());
+                match normalized.entry(location) {
+                    Entry::Occupied(entry) => {
+                        return Err(DeltaError::DuplicateLocation(entry.key().clone()));
+                    }
+                    Entry::Vacant(entry) => entry.insert(value.clone()),
+                };
+            }
```
cmyr: you could also just do,

```rust
if let Some(existing) = normalized.insert(location.to_owned(), value.clone()) {
// return error
};

**Review by cmyr — APPROVED** [review]
makes sense, one little idea for a possible simplification.

File: fea-rs/src/compile/compile_ctx.rs (hunk 1 above)
Comment (anthrotype) [reply fea-rs/src/compile/compile_ctx.rs:1341]: `None` is the bare-default form here, so we still need to insert the empty `pos`; `continue` would drop that value. The nesting is preserving that distinction.

File: fea-rs/src/parse/grammar/metrics.rs (hunk 3 above)
Comment (anthrotype) [reply fea-rs/src/parse/grammar/metrics.rs]: unfortunately the bare value is not constrained to first position: feaLib accepts it anywhere, and its builder test includes one in the middle. We therefore need state across entries to accept one bare value in any position and diagnose a second...

Comment (khaledhosny) [reply fea-rs/src/parse/grammar/metrics.rs]: We can change feaLib if it makes things simpler here, the real question is what does Adobe version do?

File: fontdrasil/src/variations.rs (hunk 5 above)
Comment (anthrotype) [reply fontdrasil/src/variations.rs]: thanks, done

File: fea-rs/src/parse/grammar/metrics.rs (hunk 3 above)
Comment (anthrotype) [reply fea-rs/src/parse/grammar/metrics.rs]: I just checked. AFDKO 5.0.1 accepts the bare value in any position. I tested both middle and last forms with `addfeatures`, and they both compiled:

```fea
(wght=200u:-30 -40 wght=900u:-60)
(wght=200u:-30 wght=900u:-60 -40)
```

The grammar likewise defines every repeated entry as `(locationSpecifier COLON)? NUM`, rather than giving the first position special treatment. 

see https://github.com/adobe-type-tools/afdko/blob/a843a0a87d9db0ea62d5ce719900acf5749c143e/c/addfeatures/hotconv/FeatParser.g4#L206-L212

So I think we should keep the stateful parse rather than restrict feaLib/fea-rs.

khaledhosny: Makes sense.

cmyr: okay makes sense. That said, my instinct is that the parser isn't the best place to enforce the "only seen once" invariant. What I would personally try:

- update `eat_variation_location_and_value` (as done here) so that it will accept either `number` or `location_spec : number`
- during validation, iterate through all the `LocationValueNode`s and make sure that there is no duplicate definition at the default location.

The only way I can see having the parser enforce this condition is if doing it elsewhere incurs some unacceptable overhead. Is there anything else I'm missing?

anthrotype: Ah... No, you’re not missing anything. I simply mirrored feaLib's parser-level check, but for fea-rs this is actually cleaner as a validation invariant. Validation already walks every `LocationValue`, so it can track the first bare default with no meaningful extra work, while the parser helpers stay stateless. Thanks!

anthrotype: much better now, I moved the “seen once” invariant into validation’s existing `LocationValue` loop and left the parser stateless. The duplicate fixture now exercises validation rather than parser failure.

cmyr (approved):
looks good!

PR #2085, opened 2026-08-20
[fea-rs] Compile name-only Glyphs predicate tokens

### LINKED ISSUE #92: Handle predicate tokens and number values in Glyphs FEA extensions

anthrotype:
Follow-up to #2082.

Turns the typed `$[...]` predicate AST into a small evaluator that expands the token into the matching glyphs, reproducing glyphsLib's matching semantics and member order.

Validation walks the typed nodes and reports what we defined in #92 as phase-1 boundary errors (non-name attribute, like/matches operators, unquoted values, mixed and/or) on the offending child, each with a pointer to phase 2 #2052. 

Compilation then converts the already-validated subset without further error handling, matching fea-rs' validate/compile philosopy.

Values must be quoted in phase 1. [The Glyphs docs](https://glyphsapp.com/learn/tokens) require quoting for string values and every predicate in the crater corpus quotes them; glyphsLib is laxer, but types some bare values as booleans or integers   rather than strings (using unanchored prefix match, so even `noon` reads as a `no` boolean 😅). Note our grammar still parses bare values. That's what makes the targeted "must be quoted" diagnostic possible, and phase 2's documented attributes can take bare booleans, numbers and constants (see #2052).

Two minor documented limitations remain:

1) Matches are emitted in GID order while glyphsLib uses source order; however, the disagreement between the order of glyphs stored in the source and its `glyphOrder` custom parameter is rare in practice, mostly limited to hand- or script-edited files, since Glyphs.app normally keeps the former in sync with the latter upon saving. I might file an issue in glyphsLib or fix it outright later.

2) The evaluated set is fontc's _compiled_ glyph map, which can include synthesized glyphs (bracket glyphs, a synthesized .notdef) that are absent from glyphsLib's. I think that's arguably a feature :)

Fixes #92. That's the last open issue on the 1.0 milestone 🎉


anthrotype:
For the record, this makes at least two fonts in the crater compile glyph predicates identically to fontmake: DynaPuff ("output is identical"), and Noto Nastaliq Urdu (matches glyphsLib's expansion exactly, the remaining diffs are pre-existing mark-attachments unrelated to predicates).

fea-rs/src/compile/glyphsapp_syntax_ext/predicate.rs
```diff
@@ -0,0 +1,432 @@
...
+//! also accepts bare values, but types some of them as booleans or integers
+//! rather than strings, so validation rejects every unquoted value with a
+//! quoting diagnostic instead of mirroring that behavior.
+
+use std::collections::HashSet;
+use std::hash::Hash;
+
+use crate::typed;
+
+/// A comparison operator over a glyph's name.
```
cmyr: although currently we only handle the name attribute, I wonder if it's actually helpful to document that here?

fea-rs/src/compile/glyphsapp_syntax_ext/predicate.rs
```diff
@@ -0,0 +1,432 @@
...
+enum Connective {
+    And,
+    Or,
+}
+
+/// A single `name <op> "value"` comparison.
+#[derive(Clone, Debug)]
+struct Clause {
+    op: Op,
+    value: String,
```
cmyr: I'd use `SmolStr` here.

fea-rs/src/compile/glyphsapp_syntax_ext/predicate.rs
```diff
@@ -0,0 +1,432 @@
...
+                .filter(|(_, name)| self.clauses.iter().all(|clause| clause.matches(name)))
+                .map(|(id, _)| *id)
+                .collect(),
+        }
+    }
+}
+
+fn clause_from_typed(clause: &typed::GlyphsAppPredicateClause) -> Clause {
+    // a clause missing any of its parts is a parse error and never compiles
+    let attr = clause.attr().expect("checked by the grammar");
```
cmyr: the general pattern is to unwrap in `typed`, with it being implicit that the structure is enforced by the grammar.

fea-rs/src/compile/glyphsapp_syntax_ext/predicate.rs
```diff
@@ -0,0 +1,432 @@
...
+//! quoting diagnostic instead of mirroring that behavior.
+
+use std::collections::HashSet;
+use std::hash::Hash;
+
+use crate::typed;
+
+/// A comparison operator over a glyph's name.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum Op {
```
cmyr: something to note is that this basically duplicates the `GlyphsAppPredicateOp` in the AST. maybe we could just use it directly? It looks like this might save us some code...

fea-rs/src/compile/glyphsapp_syntax_ext/predicate.rs
```diff
@@ -0,0 +1,432 @@
...
+        self.op.matches(name, &self.value)
+    }
+}
+
+/// A parsed, name-only Glyphs.app glyph predicate.
+#[derive(Clone, Debug)]
+pub(crate) struct Predicate {
+    clauses: Vec<Clause>,
+    // `None` when there is a single clause.
+    connective: Option<Connective>,
```
cmyr: we could also just default to `And`?

fea-rs/src/compile/glyphsapp_syntax_ext/predicate.rs
```diff
@@ -0,0 +1,432 @@
...
+            let this = match conn {
+                typed::GlyphsAppPredicateConnective::And(_) => Connective::And,
+                typed::GlyphsAppPredicateConnective::Or(_) => Connective::Or,
+            };
+            assert!(
+                connective.is_none() || connective == Some(this),
+                "mixed connectives are rejected by validation"
+            );
+            connective = Some(this);
+        }
```
cmyr: could also just be map/reduce?

```suggestion
        let connective = node
            .connectives()
            .map(|conn| match conn {
                typed::GlyphsAppPredicateConnective::And(_) => Connective::And,
                typed::GlyphsAppPredicateConnective::Or(_) => Connective::Or,
            })
            .reduce(|prev, this| {
                assert_eq!(prev, this, "mixed connectives are rejected by validation");
                this
            });
```

fea-rs/src/compile/glyphsapp_syntax_ext/predicate.rs
```diff
@@ -0,0 +1,432 @@
...
+    match value {
+        T::Bare(_) | T::Number(_) => {
+            unreachable!("unquoted values are rejected by validation")
+        }
+        _ => {
+            assert!(!text.is_empty(), "empty values are rejected by validation");
+            text
+        }
+    }
+}
```
cmyr: I think this is overkill? I would just use value.text() and trust your validation + tests.

fea-rs/src/compile/glyphsapp_syntax_ext/predicate.rs
```diff
@@ -0,0 +1,432 @@
...
+                return Some(pred);
+            }
+            if let Some(inner) = child.as_node()
+                && let Some(found) = find_predicate(inner)
+            {
+                return Some(found);
+            }
+        }
+        None
+    }
```
cmyr: there's a slightly easier way to do this, via a [fn in the `parse` module](https://github.com/googlefonts/fontc/blob/313f2c63aeaa115add9df0e7eef4abc03ea10def/fea-rs/src/parse/grammar/mod.rs#L352).

Again this makes me wonder if the types here are duplicating the AST types without really adding much? I'm not sure. The nice thing about holding onto the AST types is that it lets us provide contextual error messages or warnings wherever it might be useful.

fea-rs/src/compile/glyphsapp_syntax_ext.rs
```diff
@@ -1,5 +1,9 @@
 //! Handling glyphsapp syntax extensions
 
+pub(crate) mod predicate;
```
cmyr: it doesn't seem like this needs to be `pub`.

fea-rs/src/compile/validate.rs
```diff
@@ -1270,6 +1267,77 @@ impl<'a, V: VariationInfo> ValidationCtx<'a, V> {
...
+    /// Enforce the Phase 1 subset of Glyphs.app glyph predicates.
+    ///
+    /// The grammar accepts the structural predicate surface; here we reject the
+    /// constructs deferred to fontc#2052, attaching each diagnostic to the
+    /// offending attribute/operator/value/connective child rather than the whole
+    /// predicate. Conversion and evaluation live in
+    /// [`crate::compile::glyphsapp_syntax_ext::predicate`].
+    fn validate_glyphs_predicate(&mut self, predicate: &typed::GlyphsAppPredicate) {
+        for clause in predicate.clauses() {
+            if let Some(attr) = clause.attr() {
```
cmyr: what does it mean if we _don't_ have an attr? Is that an allowed state? Given that we `expect` this later on in compilation, having this return `Option<_>` here is mostly confusing..

fea-rs/src/tests/compile.rs
```diff
@@ -807,3 +799,143 @@ markClass a <anchor 150 -10> @top;
...
+    let fea_path = write_temp_fea(fea, test_name);
+    let glyph_map = GlyphMap::new(glyph_names.iter().copied()).unwrap();
+    Compiler::<'_, NopFeatureProvider, MockVariationInfo>::new(fea_path, &glyph_map)
+        .compile_binary()
+}
+
+// A Glyphs.app '$[...]' predicate inside a class compiles to the same thing as
+// the equivalent explicit class. This exercises the full lexer -> grammar ->
+// compile path (the predicate unit tests cover the evaluator in isolation).
+#[test]
```
cmyr: I think this can be expressed as a text-based compile test?

Some tests have accumulated in this module through various PRs but in general I would prefer that anything that can be expressed via the normal testing harness uses it (among other things this makes it easy for us to port tests to python, if we wish)

cmyr (approved):
okay looks good, some notes inline!

fea-rs/src/compile/glyphsapp_syntax_ext/predicate.rs
anthrotype: Good call. It did save code (net -53 lines). In the second commit (https://github.com/googlefonts/fontc/pull/2085/commits/a32ece2ac9538ebcd9a89c5165d4135b03807f2e) I deleted Op/Clause/Predicate and the conversion step entirely: `evaluate_predicate()` now takes the typed node and matches on GlyphsAppPredicateOp directly, following the approach of the neighboring `resolve_glyphs_app_expr`. I still collect each clause's `(op, value text)` up front once before the glyph loops, since value.text() allocates and the loop runs per glyph.

fea-rs/src/compile/glyphsapp_syntax_ext/predicate.rs
anthrotype: Done. The tests now go through `debug_parse_output` + `eat_glyphs_predicate`. I also agree on the broader question. The duplicate types are gone and evaluation works off the AST types directly.

anthrotype:
Thanks for the review! I addressed the notes in two commits: the first for the mechanical ones, the second takes up your question about the duplicated op types (the mirror types are gone entirely, see the inline replies).

PR #2068, opened 2026-07-22
Compile Fontra fonts

khaledhosny:
(Should be merged after ~#2064~ #2080, #2134, #2135)

PR #2087, opened 2026-08-25
feature-gate fontra support behind an off-by-default 'fontra' feature

anthrotype:
.fontra input is experimental and still hits a `todo!()` in fontra2fontir, so I think it's best not to build it by default, especially in light of the upcoming fontc 1.0

Without the feature a .fontra path will now report a clean error instead of panicking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,8 +51,8 @@ $ cargo run -p fontc -- resources/testdata/wght_var.designspace
 # Build a .glyphs file
 $ cargo run -p fontc -- resources/testdata/glyphs3/WghtVar.glyphs
 
-# Build a .fontra file
-$ cargo run -p fontc -- resources/testdata/fontra/minimal.fontra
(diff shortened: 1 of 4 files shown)


khaledhosny:
Is fontc 1.0 coming soon? #2068 should fix all the `todo!()` (it is currently blocked on my finishing #2080 which I’m working on right now).

anthrotype:
> Is fontc 1.0 coming soon?

yeah, but to be clear on scope: fontra support is not part of what we promised for 1.0. The intent is matching fontmake's output, and fontmake has no fontra support to match. Your fontra PRs can land in 1.x releases as they mature, independent of the 1.0 timeline, there's no need to rush them before the release.

Also, apparently the team consensus is this feature gate isn't needed, shipping with incomplete fontra support is harmless, and subsequent fontra improvements won't be treated as breaking changes.

Closing.

PR #2090, opened 2026-08-26
Fix release binary archiving and bump to 1.0.0-rc.2

anthrotype:
The `fontc-v1.0.0-rc.1` release run failed in the macOS build job (that's why I did a pre-release 😅 )

maturin 1.15 now builds the wheel from the extracted sdist in a temp directory when --sdist is passed, so the archive steps no longer found the binaries under the workspace `target/`. 

We can pin `CARGO_TARGET_DIR` to that to restore the old layout.

Nothing was published from the failed run (PyPI and GitHub release jobs were skipped), but the rc.1 tag was pushed, so this bumps to rc.2 rather than reusing the tag name.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -99,6 +99,10 @@ jobs:
     runs-on: ${{ matrix.platform.os }}
     env:
       TOOL: ${{ needs.detect.outputs.tool }}
+      # When --sdist is passed, maturin (observed with 1.15) builds the wheel
+      # from the extracted sdist in a temp dir; pin cargo's target dir to the
+      # workspace so the archive steps below can find the binaries.
+      CARGO_TARGET_DIR: ${{ github.workspace }}/target
     steps:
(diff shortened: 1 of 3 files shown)


anthrotype:
the second time it worked all the way through:

https://github.com/googlefonts/fontc/actions/runs/32975393226

That validates the release pipeline end to end ahead of the final 1.0: all 7 targets built, wheels + sdist published to PyPI as https://pypi.org/project/fontc/1.0.0rc2/ (maturin converts the SemVer version to PEP 440 automatically), the GitHub release was created at https://github.com/googlefonts/fontc/releases/tag/fontc-v1.0.0-rc.2 with all binary assets and correctly auto-marked as pre-release (#2088), auto-generated release notes picked the right base (fontc-v0.6.0), and fontc --version installed from the wheel reports `fontc 1.0.0-rc.2` from the same git tag.

pip won't pick the rc by default (pre-releases need `--pre`),and the plugin-notification workflow correctly refused to dispatch for a pre-release version (as expected).

PR #2093, opened 2026-08-27
skip pre-releases in notify-plugin CI job instead of failing

anthrotype:
The `X.Y.Z` check rejected pre-release versions with exit 1, so rc tags (e.g. `fontc-v1.0.0-rc.2`) left the notify-plugin CI red.

We can simply skip the dispatch when it's not a full release. 

(A malformed fontc-v tag still fails the run)

JMM

### DIFF
--- a/.github/workflows/notify-plugin.yml
+++ b/.github/workflows/notify-plugin.yml
@@ -37,7 +37,10 @@ jobs:
         env:
           GITHUB_EVENT_WORKFLOW_RUN_HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
 
-      - name: Validate version format (X.Y.Z)
+      # Only notify the plugin of full releases; pre-releases (e.g. 1.0.0-rc.1)
+      # skip the dispatch instead of failing the run.
+      - name: Check version format (X.Y.Z)
+        id: check
         run: |
           VERSION="${STEPS_VERSION_OUTPUTS_VERSION}"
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-plugin.yml
```diff
@@ -39,18 +42,20 @@ jobs:
...
-          if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-            echo "ERROR: Invalid version format: $VERSION"
-            echo "Expected format: X.Y.Z (e.g., 0.5.0)"
-            exit 1
+          if [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "is_release=true" >> $GITHUB_OUTPUT
+            echo "Version validated: $VERSION"
+          else
+            echo "is_release=false" >> $GITHUB_OUTPUT
+            echo "Pre-release version: $VERSION; skipping plugin notification"
```
cmyr, line 57: okay so we will allow any version, but only publish ones that follow X.Y.Z?

cmyr (approved):
so let's land a PR that makes zizmor happy, and then we can land these changes?

.github/workflows/notify-plugin.yml
anthrotype: No this isn't about what we publish, it's just to gate the workflow dispatch for the Glyphs.app fontc plugin on actual full releases (skipping it otherwise). Currently it is red for pre-releases, it should just be grey.

Issue #2042, opened 2026-06-12
CI broken with "conflicting implementation in crate `time`"


anthrotype:
CI is currently failing on every PR with [E0119](https://doc.rust-lang.org/error_codes/E0119.html).

The `time` crate published [0.3.48](https://crates.io/crates/time/0.3.48) which adds a trait impl that collides, via [coherence](https://rustc-dev-guide.rust-lang.org/coherence.html) overlap checker, with fea-rs's blanket impl at

https://github.com/googlefonts/fontc/blob/d201554c1cd63eef4f1f3c3bffbbf88b57010d03/fea-rs/src/common.rs#L44-L48

Upstream bug https://github.com/time-rs/time/issues/783

Since we don't commit Cargo.lock, CI resolves the latest.

The error only occurs when passing --all-features to cargo:

```
    error[E0119]: conflicting implementations of trait
      `From<...HourBase>` for type `<...HourBase as ...ModifierValue>::Type`
      --> fea-rs/src/common.rs:44:1
       = note: conflicting implementation in crate `time`:
               - impl From<HourBase> for <HourBase as ModifierValue>::Type;
       = note: upstream crates may add a new impl of trait
               `Into<fontdrasil::types::GlyphName>` for type `HourBase` in
               future versions
```

We could pin time to =0.3..47 or maybe narrow the blanket `impl<T: Into<GlyphName>> From<T> for GlyphIdent`...

PR #2094, opened 2026-08-27
Commit Cargo.lock for reproducible builds

anthrotype:
Resurrects #1723

I think the "reproducible fontc builds" argument still holds: without a lockfile every fresh checkout resolves whatever crates.io serves that day. That's how crater ended up silently updating dependencies between runs (#1723), and how `time` crate v0.3.48 broke CI on every open PR overnight (#2042) until we pinned it (#2043) and unpinned it again (#2051).
Incidents like these are invisible to git bisect since the tree never changed. The [Cargo team's 2023 guidance](https://blog.rust-lang.org/2023/08/29/committing-lockfiles/) argues that .gitignoring lockfiles removes 

> an important piece of history from code bases, making bisecting to find the root cause of a bug harder for maintainers.

The same applies to what we release. `cargo publish` already embeds a Cargo.lock in binary crates (the current fontc-0.6.0.crate on crates.io carries one) but it's whatever the release environment happened to resolve on a local developer's machine on publish day: a snapshot recorded nowhere in git. Today `cargo install --locked fontc` reproduces that accident; whereas with a committed lockfile it reproduces the reviewed graph that the CI tested and the tag tagged (note: plain `cargo install fontc` keeps resolving unlocked whether we the lockfile is committed).

A second argument we didn't consider back then is "reproducible font builds", which is the other side of the coin. Since #2049 we stamp ";fontc <version>" into name ID 5 as a provenance claim, and we even have a `repeatable-builds.sh` script that tests that one fontc binary builds byte-identical fonts deterministically. Committing the lockfile would add the missing link between git rev -> binary. Currently, two builds of the same commit/tag could stamp the same version string yet behave differently if the graph underneath silently changed in between.

The objections raised back then were: a committed lockfile needs an update process; and automatically picking up the latest compatible versions is a feature.

Yes, with the committed lockfile we lose free semver-compatible updates on rebuild, which only arrive when someone runs cargo update. But note we can't keep the free updates without the surprise breakage. The refresh is one mechanical cargo update command followed by a PR. We can do that say at the start of a release cycle to validate the refreshed graph.

For the freshness objection, my second commit adds a weekly scheduled job that runs `cargo update` before the usual check/test so the latest compatible versions still get exercised. Breakage surfaces there instead of every open PR, and an update stops being something that happens to us and becomes something we do.

Plus, another scheduled weekly job runs `cargo audit --deny yanked` over the committed lockfile. It earned its keep before it ever ran on CI: the first local run surfaced two high-severity DoS advisories against quick-xml 0.38 published in June, present in every build since, floating or pinned. This PR bumps our direct requirement (glyphs-reader), the remaining 0.38 path is norad (already fixed on `main` and awaiting a release).

If/after the lockfile lands, Dependabot security alerts will also start covering the Rust graph. Today they only flag the Python stuff from the requirements.txt (see the [Security tab](https://github.com/googlefonts/fontc/security) above). This audit job is not redundant with Dependabot, because the latter doesn't see yanks or the RustSec's "unmaintained" class (ansi_term and bincode apparently are listed as such). Also, a failing workflow run is a louder than the Security tab (did anybody ever pay attention to that one?).

On merge conflicts, the resolution is mechanical (take either side, run cargo check) and documented in the README's Contributing section.

I think it's worth landing this before the final 1.0 tag so the 1.0 release artifacts ship the lockfile.


anthrotype:
(zizimor hates me)

anthrotype:
The org zizmor policy scans workflows on first touch, and rust.yml's pre-existing unpinned action refs failed its mandatory check after this PR touched the file.

Since the action pins are security-relevant and there are a lot of them: every `uses:` carries the intended version as a trailing comment, so the pins can be re-derived instead of trusted. To verify the whole set, this is the command my agent used:

```sh
grep -rhoE '[A-Za-z0-9_./-]+@[0-9a-f]{40} # [^ ]+' .github/workflows/*.yml | sort -u |
while read -r pin _ tag; do
  repo=$(cut -d/ -f1,2 <<< "${pin%%@*}"); sha=${pin##*@}
  got=$(git ls-remote "https://github.com/$repo" "refs/tags/$tag*" | tail -n1 | cut -f1)
  [ "$got" = "$sha" ] && echo "ok $pin" || echo "CHECK $pin (-> ${got:-no tag})"
done
```

It prints:

```
ok actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803
ok actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c
ok actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1
ok actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
CHECK dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c (-> no tag)
ok getsentry/action-setup-venv@5a80476d175edf56cb205b08bc58986fa99d1725
ok PyO3/maturin-action@e83996d129638aa358a18fbd1dfb82f0b0fb5d3b
ok pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33
ok softprops/action-gh-release@3d0d9888cb7fd7b750713d6e236d1fcb99157228
```

The `sort -u` collapses repeated sites to unique pins (9 across both workflows). The `tail -n1` matters: for annotated tags, ls-remote lists both the tag object and the peeled `^{}` commit, and the commit is what must be pinned. All tag pins in this PR (and the release.yml ones from #2088) verify exact.

The one `CHECK` is dtolnay/rust-toolchain@stable, a moving branch rather than a tag (the repo's only tag, a floating v1, drifts the same way). It cannot be re-derived after the fact, only confirmed to exist in that repo; its pinned SHA is currently identical to the stable branch tip.

anthrotype:
(note: the dtolnay/rust-toolchain pin freezes the installer, not the compiler. I double-checked their actions.yml and`rustup toolchain install stable` still resolves at run time: https://github.com/dtolnay/rust-toolchain/blob/4360b52568e2003a75bf9bc1d59f33a8e3fc893c/action.yml#L93)

cmyr:
take a look at https://github.com/googlefonts/fontations/pull/2002 for how I handled this in fontations; I suggest the same approach here, and it can be a separate PR.

anthrotype:
Thanks! Ive done that in #2095

I can rebase this and the other one on top once that's merged

anthrotype:
@cmyr @rsheeter is this good to go now?

cmyr (approved):
looks good!

PR #2095, opened 2026-08-27
Manage github actions deps

anthrotype:
Same treatment as googlefonts/fontations#2002: pin all remaining unpinned actions to commit SHAs, add the missing permissions blocks, and let dependabot bump the pins monthly.

The one local wrinkle: zizmor flags notify-plugin.yml's workflow_run trigger as a "fundamentally insecure workflow trigger", so it gets an inline ignore -- the justification is in a comment next to it.

Unblocks #2093; #2094 drops its own zizmor commit on rebase.


cmyr (approved):
looks good!

PR #2099, opened 2026-08-27
[fea-rs] Move some predicate tests to be file-based

cmyr:
These tests were jumping through hoops to check error conditions, and we have a very nice pattern for exactly that.

JMM


anthrotype (approved):
LGTM

Nit: the old test also asserted that these predicates fail during validation. The shared bad-test harness accepts CompilationFail too, so that specific invariant is no longer covered. Not sure if we should bother retaining a small assertion for that? Up to you

cmyr:
To my mind the distinction between the validation and compilation passes are opaque to the consumer, as long as we're showing a meaningful error I think it's okay?

anthrotype:
ok yeah

PR #2101, opened 2026-08-27
[fea-rs] Fix crasher when contextual input is empty

cmyr:
Found while looking at the empty mark classes stuff.

JMM

### DIFF
--- a/fea-rs/src/compile/compile_ctx.rs
+++ b/fea-rs/src/compile/compile_ctx.rs
@@ -952,10 +952,13 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
         let target = input.target();
         let replacement = node.inline_rule().and_then(|r| r.replacements().next());
         //FIXME: warn if there are actual lookups here, we don't support that
-        if let Some((target, replacement)) =
+        if let Some((target_ids, replacement)) =
             self.validate_single_sub_inputs(&target, replacement.as_ref())
         {
-            let context = target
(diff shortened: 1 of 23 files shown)


fea-rs/src/compile/compile_ctx.rs
```diff
@@ -1238,11 +1245,37 @@ impl<'a, F: FeatureProvider, V: VariationInfo> CompilationCtx<'a, F, V> {
...
+        node: &impl ContextualRuleNode,
+        context: &[(GlyphOrClass, Vec<LookupId>)],
+    ) -> bool {
+        let input = node.input();
+        let mut found_empty = false;
+        for (item, (glyphs, _)) in input.items().zip(context) {
+            if glyphs.is_empty() {
+                self.error(
+                    item.target().range(),
+                    "Empty glyph class in contextual rule",
```
anthrotype: report_empty_contextual_input emits “Empty glyph class in contextual rule”, while the new fixtures expect the GSUB- and GPOS-specific diagnostics. That makes all seven new cases fail in CI. Maybe pass the diagnostic text into the helper, or align the expected output?

anthrotype:
`add_reverse_contextual_sub` does not go through `report_empty_contextual_input` so `rsub []' by b;` still compiles successfully, whereas feaLib rejects it as an empty glyph class in substitution. You should probably add a similar guard and regression fixture?

cmyr:
good catch on rsub, added a commit to cover that.

anthrotype (approved):

PR #2107, opened 2026-09-01
Hoist resolve_variable_metric out of fontbe

cmyr:
This will let me reuse this same delta computation logic in tests in fea-rs, which will be very useful for the fea-merge work.

No function change.

JMM


dfrg (approved):
LGTM

PR #2108, opened 2026-09-01
Fea test variable syntax

cmyr:
based on #2107, this makes our MockVariationStore compute real variable metrics, based on the same logic we use in fontc.

This lets us run some fonttools tests we'd been ignoring previously, and _that_ uncovered a bug, fixed in the second commit.


dfrg (approved):
👍

PR #2114, opened 2026-09-03
[crater] Skip comparison when fontc output unchanged

cmyr:
On average compiling a target with fontc accounts for 8% of the the runtime of crater (with a _cached_ fontmake output), with the majority of the rest of the time spent in the ttx_diff python code itself: dumping the ttx and doing various bits of xml normalization and cleanup.

This points us to an easy and significant win: avoid doing any of that ttx/xml work if fontc's output is unchanged.

ttx_diff now writes the sha256 of the font it built to fontc.sha256, and takes an --expected_fontc_ttf_hash; on a match it exits 3 without comparing anything. Crater records that hash alongside the result and passes it back on the next run. A hit costs a fontc build and a hash instead of the whole comparison.

Only results that describe a specific pair of fonts are cached: a runtime failure could have any cause, and if fontc did not produce a font there is no hash to key on. The cached result is only trusted when fontmake's output came from the cache in the same run, and the existing invalidation still applies (the whole cache is dropped when ttx_diff or the pinned python deps change.)

### DIFF
--- a/fontc_crater/src/ci/results_cache.rs
+++ b/fontc_crater/src/ci/results_cache.rs
@@ -2,7 +2,10 @@
 
 use std::path::{Path, PathBuf};
 
-use crate::Target;
+use crate::{
+    RunResult, Target,
+    ttx_diff_runner::{DiffError, DiffOutput},
+};
 
 static CACHE_DIR_NAME: &str = "crater_cached_results";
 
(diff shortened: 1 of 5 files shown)


dfrg (approved):
Seems like a nice win for crater runs!

PR #2115, opened 2026-09-03
fea-merge part 1: PendingCompilation

cmyr:
This refactors fea-rs to introduce the idea of a PendingCompilation, which stores all of the state accumulated during the compilation pass of a single FEA source tree, without finalizing it into OpenType tables.

This is intended to be the basis of support for merging per-master non-variable FEA into a variable font.

This is not a functional change, it just moves things around: instead of a `build()` method on `CompilationCtx`, you first get a `PendingCompilation`, and then you call `build` on that to get the final output. (we do actually keep the `build()` on `CompilationCtx`, it just uses this new API under the covers).

This also moves a few tasks out of the `compile()` method, so that all we're doing there is accumulating the state, not doing any finalization.

The end result of this is that we can stash all of the state accumulated from a single FEA source (graph), and this will be the state we work with when we do the merge logic.


dfrg (approved):
LGTM!

fea-rs/src/compile/compiler.rs
```diff
@@ -146,13 +146,16 @@ impl<'a, F: FeatureProvider, V: VariationInfo> Compiler<'a, F, V> {
...
 
-        // we 'take' the errors here because it's easier for us to handle the
-        // warnings using our helper method.
-        let messages = std::mem::take(&mut ctx.errors);
+        // diagnostics are handled in one place, after build: the feature
+        // provider and the finalizers run during build, so they can contribute
+        // messages of their own.
+        let (compilation, messages) = ctx.build().map_err(|errors| {
+            CompilerError::CompilationFail(DiagnosticSet::new(errors, &tree, self.max_n_errors))
+        })?;
```
anthrotype, line 154: I just noticed after merging this: on the Err branch the whole diagnostic vector now goes straight into `CompilationFail`, so when compilation fails warnings are no longer printed (self.print_warnings is ignored), they count towards "failed with N errors" (and max_n_errors caps a mixed list). Before this PR the diagnostics were taken out of the context and run through `print_warnings_return_errors` before `build()`, so both outcomes went through it. 

not urgent since fontc is unaffected (fontbe calls `compile::compile` which never split), this only changes the fea-rs CLI and the ttx test harness with VERBOSE.

Routing both outcomes through the helper could restore it:

```rust
    let (compilation, messages) = match ctx.build() {
        Ok((compilation, messages)) => (Some(compilation), messages),
        Err(messages) => (None, messages),
    };
    let diagnostics = DiagnosticSet::new(messages, &tree, self.max_n_errors);
    print_warnings_return_errors(diagnostics, self.print_warnings, self.max_n_errors)
        .map_err(CompilerError::CompilationFail)?;
    Ok(compilation.expect("no errors, so build succeeded"))
```

PR #2116, opened 2026-09-03
fea-merge part 2: basic API and structure

cmyr:
This adds the basic API/shape of the merging logic, without implementing merging any for of the relevant tables or their values.

This also includes a temporary commit pinning to fontations:HEAD; I'll do a release there and get an update merged here as a separate PR, and then drop that commit.

### DIFF
--- a/fea-rs/src/compile.rs
+++ b/fea-rs/src/compile.rs
@@ -18,6 +18,7 @@ pub use feature_writer::{
 };
 pub use language_system::LanguageSystem;
 pub use lookups::{FeatureKey, LookupId};
+pub use merge::{MergeError, merge};
 pub use opts::Opts;
 pub use output::{Compilation, PendingCompilation};
 pub use tables::Os2Builder;
@@ -35,6 +36,7 @@ mod glyph_range;
 pub(crate) mod glyphsapp_syntax_ext;
 mod language_system;
 mod lookups;
+mod merge;
 mod opts;
 mod output;
 mod tables;
@@ -80,7 +82,7 @@ pub fn compile<V: VariationInfo, T: FeatureProvider>(
 /// Compile one master's (non-variable) FEA into the intermediate state used for merging.
 ///
 /// A designspace may have a different `features.fea` per master. To combine
-/// them, compile each one with this function, [`merge`] the results into a
+/// them, compile each one with this function, [`merge()`] the results into a
 /// single variable [`PendingCompilation`], and call [`PendingCompilation::finish`]
 /// once; the feature writer therefore runs once, on the merged result.
 ///
--- a/fea-rs/src/compile.rs
+++ b/fea-rs/src/compile.rs
@@ -19,7 +19,7 @@ pub use feature_writer::{
 pub use language_system::LanguageSystem;
 pub use lookups::{FeatureKey, LookupId};
 pub use opts::Opts;
-pub use output::Compilation;
+pub use output::{Compilation, PendingCompilation};
 pub use tables::Os2Builder;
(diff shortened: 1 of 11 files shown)


fea-rs/src/compile/language_system.rs
```diff
@@ -17,7 +17,7 @@ pub struct LanguageSystem {
 /// Track languagesystem statements
 ///
 /// Seeing no statements is the same as seeing 'DFLT dflt'.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
```
anthrotype, line 20: This derive makes `check_structure` reject a difference that isn't one: a master with no `languagesystem` statement and one with `languagesystem DFLT dflt;` compile identically, but `has_explicit_entry` differs, so they fail with `MergeError::LanguageSystems`. Since only the resulting set matters for the merge, I'd compare `items` alone:

```rust
    #[derive(Clone, Debug)]
    pub(crate) struct DefaultLanguageSystems { ... }

    impl PartialEq for DefaultLanguageSystems {
        fn eq(&self, other: &Self) -> bool {
            self.items == other.items
        }
    }
```

and a test alongside `language_systems_must_match`:

```rust
#[test]
fn implicit_and_explicit_dflt_are_the_same_language_system() {
    let a = "feature kern { pos a b -20; } kern;";
    let b = "languagesystem DFLT dflt; feature kern { pos a b -20; } kern;";
    assert_eq!(merged_binary(&[a, b]), one_shot_binary(a));
}
```

anthrotype (approved):
LGTM

fea-rs/src/compile.rs
```diff
@@ -77,6 +79,32 @@ pub fn compile<V: VariationInfo, T: FeatureProvider>(
...
+/// once; the feature writer therefore runs once, on the merged result.
+///
+/// Unlike [`compile`] this takes no [`VariationInfo`]: a master's FEA is by
+/// definition not variable, so variable metrics, conditionsets and glyphsapp
+/// number values are rejected, as they would be when compiling a static font.
+pub fn compile_for_merge(
+    tree: &ParseTree,
+    glyph_map: &GlyphMap,
+    opts: Opts,
+) -> Result<PendingCompilation, DiagnosticSet> {
```
anthrotype, line 96: nit: on success this drops the master's warnings (they stay in `pending.errors` and only the default's survive to `finish`). The doc comment on `merge()` says

> the caller should report each master's own diagnostics before merging

but the API gives it no way to. Mirroring `compile`'s `Result<(PendingCompilation, DiagnosticSet), DiagnosticSet>` would fix that.

cmyr: in practice I'm not sure this is an issue? if merge succeeds it seems unlikely that there would be a warning in only a non-default master's FEA, and the flip side of reporting a bunch of warnings for each master seems strictly worse from a UI standpoint?

anthrotype: fair enough. Looking at where warnings actually come from I agree: the validation ones are already logged per master in fontbe's first pass, and the compile-stage ones are structural, which merge requires to match anyway. 

Still that merge() doc I quoted earlier says something the API doesn't let it do on success

https://github.com/googlefonts/fontc/blob/ba7e87fea13dfb74a82a2bba3dcf3cc9a76b4e45/fea-rs/src/compile/merge.rs#L39-L40

could be reworted to say the other masters' warnings are dropped and why. Not blocking.

cmyr: actually I've come back around, I think it makes sense for the API on these methods to be the same, so i'll return the warnings 😆

PR #2118, opened 2026-09-03
fea-merge part 3: GPOS scaffolding

cmyr:
This is the basic structure for merging GPOS, without support for merging the individual lookup types.

----
[fea-rs] Align GPOS lookups across masters when merging

Per lookup index the masters must agree on count, kind, lookupflags, mark filtering set and extension-ness; an anonymous-lookup count difference gets an error message that says where it usually comes from, since a value-only edit to an inline contextual rule can cause it.

Lookups that are identical across masters are passed through untouched, so identical masters still round-trip byte-for-byte. The subtable policy by type follows varLib: cursive and mark lookups merge subtable-by-subtable and require equal counts, because matching there is per (mark, base) pair within one subtable; contextual lookups must be identical. Single and pair lookups are flattened before merging, which lands with their value merging. Until then any lookup whose values differ is reported as unsupported.

### DIFF
--- a/fea-rs/src/compile/lookups.rs
+++ b/fea-rs/src/compile/lookups.rs
@@ -259,7 +259,7 @@ impl PositionLookup {
         }
     }
 
-    fn kind(&self) -> Kind {
+    pub(crate) fn kind(&self) -> Kind {
         match self {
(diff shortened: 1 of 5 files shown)


anthrotype (approved):
LGTM

PR #2119, opened 2026-09-03
fea-merge part 4: singlepos

cmyr:
[fea-rs] Merge values, starting with SinglePos

This adds merge_metric, the core logic for merging values across masters. merge_value_record applies that per field, so the value format of the result is the union of the masters' formats.

Uses this new logic to merge SinglePos lookups. A lookup applies at most one subtable to a glyph, the first that covers it, so each master's subtables flatten to one map (first wins) without changing behaviour. The merged map is rebuilt as one subtable and the builder re-splits it.

### DIFF
--- a/fea-rs/src/compile/merge/single.rs
+++ b/fea-rs/src/compile/merge/single.rs
@@ -1,0 +1,106 @@
+//! Merging single positioning lookups.
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use write_fonts::tables::{
+    gpos::builders::{SinglePosBuilder, ValueRecordBuilder},
+    layout::builders::LookupBuilder,
+};
+
+use super::{MergeCtx, MergeError, lookups::AlignedLookup};
+use crate::{common::GlyphId16, compile::VariationInfo};
+
+impl<V: VariationInfo> MergeCtx<'_, V> {
+    /// Merge a single positioning lookup across masters.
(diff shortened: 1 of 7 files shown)


fea-rs/src/compile/merge/single.rs
```diff
@@ -0,0 +1,93 @@
...
+            merged_binary(&[
+                "feature kern { pos a 10; } kern;",
+                "feature kern { pos a 10; pos b 20; } kern;",
+            ]),
+            one_shot_binary("feature kern { pos a 10; pos b (wght=400:0 wght=900:20); } kern;")
+        );
+    }
+
+    #[test]
+    fn subtables_flatten_first_wins() {
```
anthrotype, line 84: nit: `merge_value_record`'s field-wise union isn't covered directly; one test for it is worth having:

```rust
    #[test]
    fn fields_set_in_different_masters_are_unioned() {
        assert_eq!(
            merged_binary(&[
                "feature kern { pos a 10; } kern;",
                "feature kern { pos a <10 0 0 0>; } kern;",
            ]),
            one_shot_binary(
                "feature kern { pos a <(wght=400:0 wght=900:10) 0 (wght=400:10 wght=900:0) 0>; } kern;"
            )
        );
    }
```

anthrotype (approved):

PR #2120, opened 2026-09-03
fea-merge part 5 (pair pos)

cmyr:
Two commits, adding support for glyph- and class-based pairpos lookups.

### DIFF
--- a/fea-rs/src/compile/merge/pair.rs
+++ b/fea-rs/src/compile/merge/pair.rs
@@ -34,6 +34,49 @@ struct PairSubtable<'a> {
     classes: Vec<Vec<ClassPair<'a>>>,
 }
 
+/// The record a class subtable applies to a pair whose first glyph it covers
+/// but whose second glyph is in none of its classes.
+///
+/// A static so that outcome can be returned by reference, like a real cell,
+/// and stop the search through later subtables.
+static EMPTY: ValueRecordBuilder = ValueRecordBuilder {
+    x_advance: None,
+    y_advance: None,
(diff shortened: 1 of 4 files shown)


fea-rs/src/compile/merge/pair.rs
```diff
@@ -0,0 +1,405 @@
...
+
+        let mut builder = PairPosBuilder::default();
+        for ((first, second), (r1, r2)) in merged_pairs {
+            builder.insert_pair(first, r1, second, r2);
+        }
+
+        let all_pairs = || per_master.iter().flat_map(PairLookup::class_pairs);
+        let classes1 = refine(all_pairs().map(|(class1, ..)| *class1));
+        let classes2 = refine(all_pairs().map(|(_, class2, ..)| *class2));
+        for class1 in &classes1 {
```
anthrotype, line 232: this loop is masters x refined cells x class rules, since `effective_class_value` scans every rule of every class subtable to find the one covering `first`; on a real multi-master kern.fea it could get noticeably heavy. The classes within a class subtable are disjoint on each side ([ClassDefBuilder::can_add](https://github.com/googlefonts/fontations/blob/e62be6fa7fd3ec88221500d373408df45ae73992/write-fonts/src/tables/layout/builders.rs#L157) makes sure of that), so a glyph -> class index per subtable built once in `PairLookup::new` would turn the scan into a couple of map lookups. Not blocking, fine to leave for a follow-up.

anthrotype (approved):

PR #2121, opened 2026-09-03
fea-merge part 6: anchors + cursive pos

cmyr:
[fea-rs] Merge anchors, starting with cursive

Anchors are sparse rather than zero-filled: a master that lacks an anchor is left out of the variation model, as varLib does via getSubModel, so merge_metric grows a Missing::{Zero, Sparse} choice. Since fontdrasil's model needs a value at the default location, an anchor present only in non-default masters is an error, matching varLib's "Base master not found". varLib merges only format 1 anchors; a contour point has no home in the format 3 anchor a varying position needs, so an anchor with one may not vary.

Cursive attachment is the first consumer: subtable by subtable, coverage unioned, entry and exit anchors merged over the masters that have them.

### DIFF
--- a/fea-rs/src/compile/merge/metric.rs
+++ b/fea-rs/src/compile/merge/metric.rs
@@ -4,25 +4,38 @@ use std::collections::HashMap;
 
 use fontdrasil::coords::NormalizedLocation;
 use write_fonts::tables::{
-    gpos::builders::ValueRecordBuilder,
+    gpos::builders::{AnchorBuilder, ValueRecordBuilder},
     layout::builders::{DeviceOrDeltas, Metric},
 };
 
 use super::{MergeCtx, MergeError};
 use crate::compile::{VariationInfo, compile_ctx::metric_from_deltas};
 
+/// What a master without a value contributes.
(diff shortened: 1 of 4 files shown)


fea-rs/src/compile/merge/metric.rs
anthrotype, line 54: now that `None` can mean "left out" rather than zero, the device-table check above is wrong for the `Missing::Sparse` case: it compares every slot against `Some(first)`, so a master that just doesn't have the anchor makes the merge report `DeviceDiffers` even though the anchors that do exist agree. When missing is Sparse it should only compare the present values, still checking first that the default master has one (otherwise `[None, Some(device)]` would get through).

This fails on the current branch:

```rust
    #[test]
    fn device_tables_can_be_sparse() {
        let var_info = var_info();
        let ctx = ctx(&var_info, &[0.0, 1.0]);
        let device = Metric {
            default: 10,
            device_or_deltas: Device::new(11, 12, &[1, 1]).into(),
        };
        let merged = ctx
            .merge_metric(&[Some(&device), None], Missing::Sparse, 0)
            .unwrap();
        assert_eq!(merged, device);
    }
```

anthrotype (approved):

PR #2123, opened 2026-09-03
fea-merge part 8: GDEF ligature carets

cmyr:
Ligature glyphs are unioned; the masters that have one must agree on its number of carets and on which are contour points, and positions are merged over the masters that have the glyph, so they must be present at the default master. varLib's LigCaretList goes through the generic merger, so it requires the caret lists themselves to be identical and interpolates only the positions; this is a superset of that. This is what Anek and Public Sans need, whose masters differ only in LigatureCaretByPos.

### DIFF
--- a/fea-rs/src/compile/merge/carets.rs
+++ b/fea-rs/src/compile/merge/carets.rs
@@ -1,0 +1,200 @@
+//! Merging GDEF ligature carets.
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use write_fonts::tables::layout::builders::{CaretValueBuilder, DeviceOrDeltas, Metric};
+
+use super::{MergeCtx, MergeError};
+use crate::{common::GlyphId16, compile::VariationInfo};
+
+impl<V: VariationInfo> MergeCtx<'_, V> {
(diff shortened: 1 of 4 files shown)


fea-rs/src/compile/merge/carets.rs
```diff
@@ -0,0 +1,200 @@
...
+                return Err(MergeError::LigatureCarets { master, glyph });
+            }
+
+            let mut values = Vec::with_capacity(default.len());
+            for (i, caret) in default.iter().enumerate() {
+                let CaretValueBuilder::Coordinate { .. } = caret else {
+                    // a point index is not a position; the masters that have
+                    // the glyph must agree on it
+                    if carets.iter().flatten().any(|other| other[i] != *caret) {
+                        return Err(MergeError::LigatureCarets { master: 0, glyph });
```
anthrotype, line 71: is master 0 intended here? the loop is over the default's carets, so if this fires it's some other master whose point index differs from the default's, and that's the one worth naming; the check above already uses position() for that.

Also, this reuses LigatureCarets, whose message says "different number or kind of carets", so a differing contour point gets reported as a count/kind mismatch. Maybe widen the message to cover contour points, or give it its own variant.

anthrotype (approved):

PR #2124, opened 2026-09-03
[fea-rs] Match mark classes by their marks, not their names

cmyr:
A crater run revealed an issue in our initial handling of mark classes, which this commit resolves.

The first pass at the merge logic compared mark classes by name in two places, and both were wrong for real fonts:

- The per-glyph membership check assumed a glyph is in one mark class, but a mark is routinely in several, one per lookup (Tektur's ogonek is in @ogonek and @bottom in every master). Which class the check saw first came out of a HashMap, so the "conflict" it reported was random.

- Within a subtable, marks were matched by class name. ufo2ft names its generated classes after the lookup index, so Arimo's masters call the same class @_MARKCLASS_0_lookup_6 and @_MARKCLASS_0_lookup_7. The binary has no class names, only a partition of the marks into classes, and that is what varLib effectively compares.

The global map now just unions membership; it only feeds GDEF class inference, where names are irrelevant. Within a subtable, each master's class names are aligned to shared ones by the marks they contain: two marks that share a class in one master must share it in every master that has both, and that is the only error left (MarkClassMismatch, replacing MarkClassConflict). Bases and ligature components are keyed by the shared names so their anchors merge per class as before.

### DIFF
--- a/fea-rs/src/compile/merge.rs
+++ b/fea-rs/src/compile/merge.rs
@@ -8,8 +8,6 @@
 use std::collections::HashMap;
 
 use fontdrasil::coords::NormalizedLocation;
-use smol_str::SmolStr;
-use write_fonts::types::GlyphId16;
 
 use super::{PendingCompilation, VariationInfo};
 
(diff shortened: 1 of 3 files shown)


anthrotype:
> Arimo's masters call the same class @_MARKCLASS_0_lookup_6 and @_MARKCLASS_0_lookup_7. The binary has no class names, only a partition of the marks into classes, and that is what varLib effectively compares

amazing :)

anthrotype (approved):
LGTM

varLib's own TODO in `_MarkBasePosFormat1_merge` describes exactly that: https://github.com/fonttools/fonttools/blob/34be2443a7539b0b36467e9cca63ef53d4506fd6/Lib/fontTools/varLib/merger.py#L697-L704

PR #2125, opened 2026-09-03
fea-merge finale: fontbe integration

cmyr:
This is the final bit of the fea-merge work, wiring the merge logic into fontbe.

When FEA is identical, we will still take the fast path and just build it once; when it differs across masters we will compile each master's FEA in parallel and then attempt to merge them.

This gets us compiling a bunch of new fonts, and has us matching fontmake on many of them.

### DIFF
--- a/fontbe/src/features.rs
+++ b/fontbe/src/features.rs
@@ -48,7 +48,7 @@ use crate::{
     error::Error,
     orchestration::{
         AnyWorkId, BeWork, Context, ExtraFeaTables, FeaFirstPassOutput, FeaRsKerns, FeaRsMarks,
-        WorkId,
+        FeaSourceIdx, WorkId,
     },
 };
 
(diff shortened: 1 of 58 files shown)


fontbe/src/features.rs
```diff
@@ -74,8 +74,16 @@ const MKMK: Tag = Tag::new(b"mkmk");
 const ABVM: Tag = Tag::new(b"abvm");
 const BLWM: Tag = Tag::new(b"blwm");
 
+/// Parse, validate, and first-pass compile one of the font's FEA sources.
+///
+/// A designspace can have a features.fea per master; each distinct source is
+/// compiled by its own instance of this work. Only the default master's output
+/// ([`WorkId::DEFAULT_FEATURES_AST`]) is consumed today, see
+/// [`FeatureCompilationWork`].
```
anthrotype, line 82: stale now, every source's output is consumed

fontbe/src/features.rs
```diff
@@ -403,6 +400,128 @@ impl FeatureCompilationWork {
...
+    let default_location = static_metadata
+        .default_location()
+        .subset_axes(&static_metadata.axes);
+    let Some(default) = masters
+        .iter()
+        .position(|(location, _)| *location == default_location)
+    else {
+        return Err(Error::VariableFeaNoDefaultMaster(features.n_sources()));
+    };
+    masters.swap(0, default);
```
anthrotype, line 523: the master index in MergeError is the position after this swap, so when the merge fails the user gets "master 3: ..." and no way to tell which features.fea that is. fontbe knows the path and location of each master at this point; can we map the index back to the source, or attach them here, before this reaches crater?

anthrotype (approved):
LGTM, thanks for this monumental effort! 🎉 


(PS: the first commit's body still says "rewrite me")

fontbe/src/features.rs
cmyr: good call, but i'm going to punt on this right now, opened https://github.com/googlefonts/fontc/issues/2142 to track.

PR #2127, opened 2026-09-04
[fea-rs] Preserve lookupflag across a redundant script statement

cmyr:
See https://github.com/fonttools/fonttools/pull/4169 for more detail on this issue.

In fea-rs, the issue was straightforward: we hardcoded self.script_ to "DFLT" instead of seeding it from the declared languagesystem statements.

This caused our guard in set_script to be incorrect, which could cause lookups to be registered under DFLT instead of the declared script.

Alongside the fonttools patch above, this produces at least a +20 identical on crater.

### DIFF
--- a/fea-rs/src/compile/features.rs
+++ b/fea-rs/src/compile/features.rs
@@ -44,7 +44,9 @@ pub(crate) struct ActiveFeature {
     pub(crate) tag: Tag,
     condition_set: Option<ConditionSet>,
     default_systems: DefaultLanguageSystems,
-    current_lang_sys: Option<LanguageSystem>,
+    current_lang_sys: LanguageSystem,
+    /// Whether a script or language statement has been seen
+    seen_script_lang: bool,
     lookups: HashMap<LanguageSystem, Vec<LookupId>>,
     script_default_lookups: HashMap<Tag, Vec<LookupId>>,
 }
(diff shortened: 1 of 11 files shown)


fea-rs/src/compile/features.rs
```diff
@@ -349,12 +351,13 @@ impl ActiveFeature {
             condition_set,
             script_default_lookups: Default::default(),
             lookups: Default::default(),
-            current_lang_sys: Default::default(),
+            current_lang_sys: default_systems.first(),
```
anthrotype, line 354: Should we not keep the current script used by `set_language()` synchronized with this initial system?
`ActiveFeature::current_lang_sys` now has `default_systems.first()` as its initial state, but `CompilationCtx::script` remains None.

For example: 

```fea
languagesystem latn dflt;

feature test {
    language dflt;
    sub a by b;
} test;
```

When `language dflt;` is processed, `set_language()` reads `self.script`, gets `None` and falls back to DFLT. This PR therefore emits the substitution under DFLT/dflt, whereas https://github.com/fonttools/fonttools/pull/4169 emits it under the initial latn/dflt system.

FontTools initializes both `script_` and `language_` in `start_feature()` to `min(self.language_systems)`. Here, we could either initialize `CompilationCtx::script` in `start_feature()` alongside the construction of `ActiveFeature`, or have `set_language()` derive the script from `current_lang_sys()`, avoiding the duplicated state.

Could we also add the example above as a fixture?

cmyr: good catch, will clean this up.

anthrotype (approved):

cmyr:
going to merge this; it will cause a small regression on crater until a fonttools release lands.

PR #2130, opened 2026-09-08
[ci] Update GoogleSans designspace paths after upstream repo reorg

anthrotype:
googlefonts/googlesans#768 moved the sources from source/GoogleSans/ to sources/, so the repeatable-builds job now fails on every PR with a missing file error before fontc runs (see e.g. the Build GS job on #2127).

JMM


cmyr (approved):
thank you, was confused by this!

PR #2133, opened 2026-09-09
[fontbe] treat a table that fails to pack as an error

### LINKED ISSUE #2128: fontc silently drops GPOS when write-fonts fails to pack it
When write-fonts can't resolve the offset overflows for a table, fontc writes the font without that table and exits 0. The only trace is a debug-level log line, so with the default log level there is no indication at all that anything went wrong.

The cause is that the final assembly serializes every table with `dump_table(table).ok()`, which turns a `PackingFailed` error into None, and font.rs then treats None as "this table has no content" and skips it:

https://github.com/googlefonts/fontc/blob/beb0b3506c3265f17622b735a59a2935925c748a/fontbe/src/orchestration.rs#L808-L813
https://github.com/googlefonts/fontc/blob/beb0b3506c3265f17622b735a59a2935925c748a/fontbe/src/font.rs#L223-L230

I ran into it with a Myanmar font that has a couple of very large chaining contextual lookups (several thousand rules each, one of them only representable as format 3): fea-rs builds the GPOS, the repacker promotes all lookups to extension and still ends with unresolved overflows, and the resulting font has no GPOS at all, so all marks and kerning are gone.

GSUB happens to be luckier because `FeaFirstPassOutput::new` dumps it during the first FEA pass and propagates the error properly (that's the "Generating bytes for GSUB failed: Table packing failed with N overflows" you get):

https://github.com/googlefonts/fontc/blob/beb0b3506c3265f17622b735a59a2935925c748a/fontbe/src/orchestration.rs#L370-L379

GPOS is only dumped at the end, through the `.ok()` path.

I think a packing failure should be a hard error like it is for gvar or hmtx (and crater would presumably have caught this earlier if it had been).

anthrotype:
Final font assembly serializes every table with `dump_table(table).ok()`, so when write-fonts can't resolve the overflows for a table (GPOS in my case) the Err becomes None, font.rs takes that as "no content for this table" and fontc writes the font without it and exits 0:

https://github.com/googlefonts/fontc/blob/e70d542ad7b29c6ae7e64f695bbcab5d7c2382f1/fontbe/src/orchestration.rs#L808-L813

Now to_bytes returns the write-fonts error wrapped in a DumpTableError tagged with the table's own tag, and every arm of bytes_for propagates it, which is what the first-pass GSUB dump (and gvar, hmtx) already did. The "No content" branch stays for a font that has no avar.

The first commit adds the test on its own so CI fails on it: a GPOS with a single SinglePosFormat2 subtable of 33000 value records, which is itself over 64k so the Offset16 to its own coverage can never fit (promoting the lookup to extension only widens the lookup->subtable link, and SinglePos isn't split). The second commit is the fix. Also fixed the BASE dump context string that rendered as "Generating bytes for dump BASE failed failed".

Fixes #2128

### DIFF
--- a/fontbe/src/font.rs
+++ b/fontbe/src/font.rs
@@ -237,3 +237,61 @@ impl Work<Context, AnyWorkId, Error> for FontWork {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use fontdrasil::orchestration::Access;
+    use fontir::orchestration::Context as IrContext;
+    use write_fonts::{
+        tables::{
+            gpos::{PositionLookup, SinglePos, ValueRecord},
+            layout::{Lookup, LookupFlag, LookupList},
+        },
+        types::GlyphId16,
+    };
+
+    use super::*;
+
+    /// A GPOS with a single SinglePosFormat2 subtable that is itself larger
+    /// than 64k, so the Offset16 to its own coverage table can never fit.
+    /// Promoting the lookup to extension only widens the link from the
+    /// lookup to the subtable, and write-fonts does not split SinglePos.
+    /// If it ever learns to, this will fail with "GPOS should not pack" and
+    /// needs a different unpackable shape.
+    fn unpackable_gpos() -> Gpos {
+        const NUM_GLYPHS: u16 = 33000;
+        let coverage = (0..NUM_GLYPHS).map(GlyphId16::new).collect();
+        let values = (0..NUM_GLYPHS)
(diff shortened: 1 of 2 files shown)


anthrotype:
repro: https://github.com/googlefonts/fontc/actions/runs/34363908023/job/102507606763?pr=2133#step:4:394

```
thread 'font::tests::failure_to_pack_a_table_is_an_error' (4302) panicked at fontbe/src/font.rs:285:53:
GPOS should not pack: None
```

dfrg (approved):

Issue #2008, opened 2026-05-13
avar2


simoncozens:
fontc currently has no avar2 support. Here is an example of how to build avar2 tables in fontations given a set of input/output location mappings: https://github.com/simoncozens/babelfont-rs/blob/main/babelfont/src/convertors/fontir/avar2.rs

I'd provide a PR but I'm not sure how you're storing cross-axis mappings in your IR format, but hopefully this will be a useful reference point for implementation.

rsheeter:
https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#example-of-all-mappings-elements-together appears to describe a source representation

PR #2134, opened 2026-09-09
[fontir/fontbe] Support building avar version 2 from axis mappings

khaledhosny:
Helps with https://github.com/googlefonts/fontc/issues/2008, and used by https://github.com/googlefonts/fontc/pull/2068.

### DIFF
--- a/fontbe/src/avar.rs
+++ b/fontbe/src/avar.rs
@@ -1,14 +1,21 @@
 //! Generates a [avar](https://learn.microsoft.com/en-us/typography/opentype/spec/avar) table.
 
+use std::collections::{HashMap, HashSet};
+
 use fontdrasil::{
-    coords::NormalizedCoord,
+    coords::{NormalizedCoord, NormalizedLocation},
     orchestration::{Access, Work},
     types::Axis,
+    variations::{VariationModel, resolve_variable_metric},
 };
-use fontir::orchestration::WorkId as FeWorkId;
+use fontir::{ir::StaticMetadata, orchestration::WorkId as FeWorkId};
 use log::debug;
 use write_fonts::{
-    tables::avar::{Avar, AxisValueMap, SegmentMaps},
+    read::tables::variations::NO_VARIATION_INDEX,
+    tables::{
+        avar::{Avar, AxisValueMap, SegmentMaps},
+        variations::{DeltaSetIndexMap, ItemVariationStore, ivs_builder::VariationStoreBuilder},
+    },
     types::F2Dot14,
 };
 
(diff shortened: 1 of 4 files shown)


fontbe/src/avar.rs
```diff
@@ -82,6 +89,82 @@ fn to_segment_map(axis: &Axis) -> SegmentMaps {
...
+    let axis_tags = axes.axis_order();
+
+    let mut input_locations: Vec<NormalizedLocation> = mappings
+        .iter()
+        .map(|mapping| mapping.input.clone())
+        .collect();
+    let mut output_locations: Vec<NormalizedLocation> = mappings
+        .iter()
+        .map(|mapping| mapping.output.clone())
+        .collect();
```
cmyr: just code golf, but this could be:
```rust
    let (mut input_locations, mut output_locations): (Vec<_>, Vec<_>) = mappings
        .iter()
        .map(|item| (item.input.clone(), item.output.clone()))
        .unzip();
```

fontbe/src/avar.rs
```diff
@@ -82,6 +89,82 @@ fn to_segment_map(axis: &Axis) -> SegmentMaps {
...
+
+    // If base-master is missing, insert it at zero location.
+    if !input_locations.iter().any(|loc| !loc.has_any_non_zero()) {
```
cmyr: instead of the double negative you could write,

```rust
if !input_locations.iter().any(Location::is_default)
```

fontbe/src/avar.rs
```diff
@@ -82,6 +89,82 @@ fn to_segment_map(axis: &Axis) -> SegmentMaps {
...
+        input_locations.insert(0, NormalizedLocation::new());
+        output_locations.insert(0, NormalizedLocation::new());
+    }
+
+    let mut locations = HashSet::new();
+    for loc in input_locations.iter_mut() {
+        loc.fit_to_axes(&axis_tags);
```
cmyr: just to confirm the intention, this is mutating the location in place, which means the value in input_locations will also be changed when it is used later.

That being the case, would it make more sense to do `fit_to_axes` when we construct input_locations originally?

fontbe/src/avar.rs
```diff
@@ -82,6 +89,82 @@ fn to_segment_map(axis: &Axis) -> SegmentMaps {
...
+        );
+    }
+
+    let (store, optimized) = store_builder.build();
+    let var_idxes: Vec<u32> = var_idxes
+        .into_iter()
+        .map(|value| match value {
+            Some(value) => optimized.get(value).unwrap().into(),
+            None => NO_VARIATION_INDEX,
+        })
```
cmyr: I'd just do `map` + `unwrap_or(NO_VARIATION_INDEX)`? not blocking, but a few places here where you match on option where unwrap_or would probably be more idiomatic.

cmyr (approved):
some non-blocking nits inline. Will caveat that I'm not very familiar with avar2, but happy to get this in and we can iterate as needed.

fontbe/src/avar.rs
khaledhosny: Yes. I moved it up where `input_locations` is constructed.

cmyr:
@khaledhosny looks like this collided with your earlier PR? will merge when the conflicts are resolved

khaledhosny:
> @khaledhosny looks like this collided with your earlier PR? will merge when the conflicts are resolved

Should be fixed now.

PR #2138, opened 2026-09-09
[crater] suppress git SHA in name table version / clear cache when otl-normalizer changes

cmyr:
Two commits.

I added logic in #2114 to check when a font generated by fontc was identical to a previous build, so we could skip comparison in  that case; however this wasn't working in CI, because fontc stamps its version in the name table, and includes the current git sha. By definition, this sha is different between two crater runs, so the generated fonts were never identical.

This is fixed by suppressing the git bit of the version when building for crater.

@anthrotype an interesting bit of this is that it didn't reproduce locally, since build.rs does not rebuild just because HEAD moves. In CI, we're always cloning the repo from scratch and doing a clean build, so we always run build.rs, but locally it's possible for the sha in the version to be stale. Do we care? I think not, but wanted to at least bring it up.

(The second commit is boring, it just ensures we clear the cache if otl-normalizer has changes)


wmedrano (approved):

anthrotype:
Thanks for the heads up

> do we care

Yes I do https://github.com/googlefonts/fontc/pull/2139

PR #2143, opened 2026-09-14
[fea-rs] allow include statements inside table and lookup blocks

anthrotype:
The spec allows `include()` anywhere a statement ending in semicolon is allowed. Both feaLib (which resolves includes at the token level) and makeotf (which parses each included file with the entry point of the enclosing block) accept it inside table and lookup blocks. fea-rs only handled includes at the top level and inside feature blocks: a table block stopped at the include keyword and errored at the closing brace, and a file included from a lookup block was parsed as a top-level file, so the first pos rule in it was rejected.

AFDKO-style sources do both routinely, e.g. Source Serif 4 has `table OS/2 { include(...); }`, `table name { include(...); }` and `lookup KERN useExtension { include(../kern.fea); }`, and doesn't build with fontc because of it.

This PR parses an included file in the scope of the block that included it, like makeotf does: the nine known table kinds use their own table item parser, lookup blocks use lookup statements. An include statement at the top level of a file that was itself included from inside a block now inherits that block's scope (previously it fell back to top-level parsing). A token that can't start an item in such a file (a stray `}`, say) is now reported instead of silently dropping the rest of the file.

Since the same file can now legitimately mean different things depending on where it's included from, a file is parsed once per (file, scope) pair instead of once per file; before, the first scope encountered won, which with table scopes could splice e.g. hhea entries into an OS/2 table and panic in the compiler.

Includes inside cvParameters, featureNames and the STAT names are still not supported (never seen one in the wild).

With this, stock Source Serif 4 gets past parsing; it still needs googlefonts/fontations#2128 and googlefonts/fontations#2136 released in write-fonts to build.


cmyr:
So I think there's a bit of room for some simplification here, and I've opened #2147 with some suggested changes. Take a look and see if that makes sense to you?

anthrotype:
@cmyr looks good, thanks! Can we merge this now?

cmyr (approved):
looks good!

Issue #2078, opened 2026-08-10
[glyphs] Generate STAT axis values from instances, matching glyphsLib >= 6.14


anthrotype:
glyphsLib now auto-generates STAT data from a Glyphs source's instances (googlefonts/glyphsLib#1158, in v6.14.0): one AxisValue per distinct instance coordinate on each axis, the default value elided, and a STAT-only `ital` axis pinned at 0, or 1 for an italic family.
It also honours the `Export STAT Table`, `Elidable STAT Axis Value Name` and `Style Name as STAT entry` custom parameters.

fontc writes only a stub with design axis records but no axis values (see `fontbe/src/stat.rs` comment "we support only a very simple STAT at time of writing"), unless a STAT table block arrives via FEA.

So as soon as we bump past the pinned glyphsLib 6.13.1, every Glyphs source with Variable Font Settings instances will get a STAT that differs from fontmake's.

PR #2148, opened 2026-09-16
[glyphs-reader] accept named weight and width classes in Glyphs 3 files

anthrotype:
The Glyphs 3 format defines an instance's weightClass and widthClass as integers and Glyphs.app writes them that way, but glyphsLib writes the names (Bold, Condensed, ...) in v3 files too and Glyphs.app reads either. Glyphs 2 files already get their named classes converted on load, but in a Glyphs 3 file the reader panics with a ParseFloatError on a named class when it infers the instance's user location for the axis mapping. Separately, "Medium (normal)" could never match the width lookup, because the key is compared after lowercasing and stripping spaces.

This PR runs the name-to-number conversion we already do for Glyphs 2 files on both format versions, so the rest of the reader only ever sees numbers; an unknown name is an error either way, as it was for v2.

I stumbled on this while working on #2078, where I used glyphsLib to make the test sources.


rsheeter (approved):
Ew. Good catch.

PR #2149, opened 2026-09-16
[fea-rs] Canonicalize glyph classes in contextual rules

cmyr:
In the general case, order matters for GlyphClass, but in contextual
rules every position becomes a coverage table or class def, with set
semantics. This sorts and dedups each position when a rule is added
(and turns a one-glyph class into a plain glyph), so that the derived
PartialEq compares positions as sets.

When merging adjacent rules we concatenate their inputs and canonicalize
the result, replacing GlyphOrClass::extend, which had no other caller.

This was motivated by a specific fea merge failure, where two masters
had the same glyphs in a different order in a glyph class, which failed
our 'gsub must be equal' check. With this patch, that check passes.

----

- this also fixes a logic error related to `extend`, which was the cause of diff for ClimateChange.
- I've snuck a second commit in here that fixes a bug latent in the difference between two similar methods, `to_glyph` and `single_glyph`, which hid a potential crasher.


dfrg (approved):
LGTM!

PR #2152, opened 2026-09-16
[fontir] Round sub/sup Y offsets before deriving X offsets

cmyr:
Opening this for discussion: it's a situation where we have a crater diff related to slightly different rounding behaviour. I do not think the particular behaviour is meaningful. Do we wan this patch?

**pros**:
- the new code is not meaningfully _worse_ than the old code
- it is genuinely useful to get unimportant diffs off of crater, to better surface the important ones

**cons**
- we're adding unit tests for behaviour we don't actually care about?

My personal preference is probably to take this patch, but to have a key factor in that decision being the fact that it isn't introducing significant complexity.

-----

ufo2ft's OS/2 builder stores each sub/superscript value as otRound(value) and then derives the X offsets from the *stored* (already rounded) Y offsets.

fontc fed the raw fallback float (upem * 0.075, upem * 0.35) into adjust_offset instead. For a upem where those products are not integral, and a non-zero italic angle, the derived X offset can round to a different integer than fontmake's.

This... doesn't really matter, but the new code isn't any worse than the previous code, and it makes some numbers green.


anthrotype (approved):
> This... doesn't really matter, but the new code isn't any worse than the previous code, and it makes some numbers green.

I don't disagree :)

PR #2155, opened 2026-09-16
[glyphs] Propagate anchors regardless of ufo2ft filters

cmyr:
If the default master's `userData` had a ufo2ft filters list, we only set PROPAGATE_ANCHORS when propagateAnchors appeared in it. But glyphsLib propagates anchors before it ever looks at filters, gated only by the "Propagate Anchors" custom parameter (default true), so any source with that userData key (e.g. anything produced by ufo2glyphs) silently lost propagation in fontc while fontmake still propagated.

With this patch, the custom parameter alone decides; a propagateAnchors filter entry is a no-op, as in glyphsLib.

This fixes major layout diffs in NotoSans.


anthrotype (approved):
LGTM

(I had the exact same change sitting on my machine, so I can offer you this #2160 if you like)

PR #2157, opened 2026-09-17
[glyphs] Don't assign GDEF classes to color layer glyphs

### LINKED ISSUE #1870: colr glyph GDEF categories
we have GDEF entries for various generated `.color` glyphs, and fontmake does not.

cmyr:
The .colorN glyphs we split from Glyphs color layers keep the parent's anchors, so we inferred Base for all of them and then emitted mark attachment rules for them too. glyphsLib computes openTypeCategories before it creates these glyphs, so fontmake gives them no class at all.

With this patch, PreliminaryGdefCategories carries a set of glyphs that never get a class, and glyphs2fontir puts the split color glyphs in it.

glyphsLib does regenerate categories after adding bracket glyphs, so a font with both bracket and color layers gets classes for its color glyphs in fontmake. We don't emulate that.

- closes https://github.com/googlefonts/fontc/issues/1870


anthrotype (approved):

anthrotype:
> glyphsLib computes openTypeCategories before it creates these glyphs, so fontmake gives them no class at all

this is correct to do regardless of the order in which glyphsLib computes the categories since .colorN glyphs are paint implementation details and don't participate in shaping, giving them GDEF base classes would be misleading.

> a font with both bracket and color layers gets classes for its color glyphs in fontmake

ugh. that's a bug which we should not chase 👍

